PR #151, opened 2025-11-12
fix: Prevent add_grammer command corrupts language_config.py by calcu…

Dilrevx:
…lating file offset via looking up for LANGUAGE_CONFIGS and '\n}'

Reason:

language_config.py contains multiple "}"s, a simple rfind  goes to the wrong position.
This change first looks for LANGUAGE_CONFIGS then for "\n}"

UPD:
move `pos++` into the if branch to preserve  return value check.


Dilrevx:
Just a pin, this is a re-opened PR as #150 which addesses the bug introduced in it

vitali87 (changes requested):
The `find("\n}")` approach might still match inner nested dict braces. Since tree-sitter is already a dependency, consider:

```python
def find_language_configs_end(config_content: str) -> int:
    language = Language(python_language())
    parser = Parser(language)
    tree = parser.parse(config_content.encode("utf-8"))
    
    query = language.query("""
        (assignment
            left: (identifier) @name
            right: (dictionary) @dict
        ) (#eq? @name "LANGUAGE_CONFIGS")
    """)
    
    for node, capture_name in query.captures(tree.root_node):
        if capture_name == "dict":
            return node.end_byte - 1
    
    raise ValueError("Could not find LANGUAGE_CONFIGS dictionary")
```

Handles nested dicts correctly with byte-accurate positions.

vitali87:
Hi @Dilrevx, thanks for requesting a re-review!

I noticed there haven't been any new commits since I left my review feedback on Dec 4th. Just wanted to clarify how the GitHub PR review workflow typically works:

1. **Reviewer leaves feedback**: I marked this as "Changes Requested" with a suggestion to use tree-sitter for more robust parsing
2. **Author addresses feedback**: You would push new commits that implement the suggested changes (or reply explaining why an alternative approach is better)
3. **Author requests re-review**: Once the changes are pushed, you can request a re-review

The current state shows my review still pending because the code hasn't changed. Could you:
- Implement the tree-sitter approach I suggested, OR
- Reply to my review comment explaining if you'd prefer a different solution

Once you push the updated code, I'll be happy to take another look! Let me know if you have any questions about the suggested approach.

Dilrevx:
Newbie to PRs. It must be a wrong re-review click somewhere by mistake, sry.

The requested tree-sitter parser is included. I perform a manual rstrip due to `node.end_byte` may include `\n`s.

Merge conflict occurs due to recent cgr refactor.

vitali87:
Thanks for working on this! However, the implementation deviates from my suggested approach in ways that introduce issues.

My original suggestion was the cleanest solution:
```python
for node, capture_name in query.captures(tree.root_node):
    if capture_name == "dict":
        return node.end_byte - 1
```

The current implementation has an off-by-one error where `find_language_configs_end(config_content) - 1` subtracts 1 twice if the function already accounts for positioning before the closing brace.

The whitespace stripping logic is also backwards. The while loop strips whitespace *before* the position, but tree-sitter's `end_byte` is exclusive and points after the last character.

Using `zip(name_caps, dict_caps)` assumes captures come in matching pairs, which isn't guaranteed and makes the code fragile. Additionally, if captures return `None`, the iteration will fail since there's no error handling.

Could you update the implementation to follow the original suggestion more closely? A test case would also be helpful.

Dilrevx:
Thanks for reviewing. 

I am confused, because my tree-sitter's `end_byte` points **after-after** the last character, which is `}\n\n`. 
That's why I removed trailing whitespace afterwards.

I didn't expect the two captures are not paired. I shall fix it.

My current implementation of `find_language_configs_end` returns `brace_pos + 1`, which complies with tree-sitter's [left, right) characteristic. It does deviate from your design. 

Fix and test case is on the way.

Dilrevx:
Hey, I've been swamped lately,  confused by tree-sitter captures, and writing ugly codes....
I expect it's impossible to make it work recently. So I intent to close this PR, and letting you know so you can fix the main branch...

So sorry for that. Really appreciate your effort in reviewing.
@vitali87

PR #161, opened 2025-11-17
Add LiteLLM proxy provider support

qdrddr:
- Add new LiteLLMProvider class using pydantic-ai's native LiteLLMProvider
- Register litellm_proxy provider in PROVIDER_REGISTRY
- Support OpenAI-compatible API endpoints via LiteLLM proxy
- Enable use of multiple LLM providers through a unified proxy interface

PR #163, opened 2025-11-18
Develop

Remon128:


qdrddr:
I would recommend using the LiteLLLM Client (not to be confused with LiteLLM Proxy), which can work with a very large list of LLM Providers. @Remon128

This is LiteLLM Proxy implementation PR (Not the client you want in this PR), but could be used to extend the functionality to be used for other providers.
https://github.com/vitali87/code-graph-rag/pull/161

vitali87 (changes requested):
This PR has several issues:

1. Don't change the submit key. Ctrl+J is the submit key by design. Users write multiline prompts where Enter for newlines is more natural. Your change breaks multiline input entirely - you commented out the new_line function in a function called get_multiline_input.

2. Remove CHANGELOG.md. We don't maintain a changelog file - GitHub releases and commit history serve this purpose. Also the format is wrong - who is "Ali"? You're Remon Atef.

3. Don't add Groq as a separate provider. As qdrddr mentioned, PR #161 adds LiteLLM support (which will go in soon) and it already works with Groq and 100+ other providers. Adding individual providers creates unnecessary maintenance burden.

4. PR hygiene issues:
   - Title "Develop" is meaningless - use a descriptive title
   - No PR description
   - Commented-out code instead of proper changes
   - Two unrelated changes should be separate PRs
   - Missing newline at EOF

PR #183, opened 2025-12-07
Feature/switch from mypy to ty

### LINKED ISSUE #176: [TODO]  Switch from mypy to ty

vitali87:
- Fixes https://github.com/vitali87/code-graph-rag/issues/176
- Also adds missing tests for embedding and cpp outer class method definitions


vitali87:
/gemini review Read and understand what "(H)" stands in the comments first

PR #187, opened 2025-12-08
fix: find_ending_with too slow

AloneFire:
find_ending_with  Find all qualified names ending with the given suffix using DFS on the trie.
This avoids a full scan of the `_entries` dict and instead walks the trie structure once, checking each stored qualified name.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -22,6 +22,7 @@ class FunctionRegistryTrie:
     def __init__(self) -> None:
         self.root: dict[str, Any] = {}
         self._entries: dict[str, str] = {}
+        self._suffix_index: dict[str, set[str]] = defaultdict(set)
 
     def insert(self, qualified_name: str, func_type: str) -> None:
         """Insert a function into the trie."""
@@ -37,6 +38,10 @@ class FunctionRegistryTrie:
 
         current["__type__"] = func_type
         current["__qn__"] = qualified_name
+
+        if parts:
+            suffix = parts[-1]
+            self._suffix_index[suffix].add(qualified_name)
 
     def get(self, qualified_name: str, default: str | None = None) -> str | None:
         """Get function type by exact qualified name."""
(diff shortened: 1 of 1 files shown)


vitali87:
Thanks for the PR! The refactor to reuse `find_with_prefix_and_suffix` is a nice cleanup.

However, I think the performance reasoning in the description is slightly off. Tries are optimized for prefix lookups, so searching for a suffix still requires visiting every node in the tree. There's no pruning possible when the prefix is empty.

Actually, the new implementation is likely slower in practice. The DFS traverses every trie node (intermediate modules plus leaves) rather than just the qualified names in `_entries`. On top of that, recursive Python function calls have significantly more overhead than iterating a flat dictionary, which uses CPython's optimized C implementation.

vitali87:
One more question: what would deprecating `_entries` actually give us? Is there a concrete memory or maintainability benefit you're aiming for?

AloneFire:
> One more question: what would deprecating `_entries` actually give us? Is there a concrete memory or maintainability benefit you're aiming for?

You're right, I didn't think through the implementation carefully. However, DFS did show significant speed improvements on my project. I've pushed a new commit that optimizes this using a different approach—trading space for time through indexing.

vitali87 (changes requested):
### 1. Trailing Whitespace

Line 83 has trailing whitespace. Please install pre-commit hooks:

```bash
pre-commit install
```

This runs automatically on every commit and catches formatting issues.

### 2. No Benchmark Data

I previously asked about performance reasoning. Please provide before/after benchmark results.

### 3. Insufficient Test Coverage

Add tests for:
- Multi-part suffix: `find_ending_with("Logger.info")`
- Suffix index cleanup after deletion
- Multiple entries with same suffix

### 4. Coding Standards

Review `CONTRIBUTING.md` for project coding standards.

---

Please address all of these and re-request review. Thanks!

vitali87:
Thank you @AloneFire for flagging this performance issue! The fix has since been implemented via a `_simple_name_lookup` index that provides O(1) lookups, with benchmarked speedups of 261x-382x on the operation. Closing as the underlying issue is resolved.

PR #203, opened 2025-12-23
Feat/mcp Add optional semantic_search tool + non-destructive update_repository for MCP server

luxannaxul:
Hi maintainers,

This PR improves the MCP server tool set in codebase_rag/mcp/tools.py in two areas:

Optional semantic_search tool registration
Adds a semantic search MCP tool that is only registered when semantic dependencies are installed (checked via has_semantic_dependencies()). This avoids exposing semantic_search in environments where it would otherwise return empty results. When unavailable, the server logs a clear message pointing to uv sync --extra semantic.

Non-destructive repository updates
Introduces a new MCP tool, update_repository, which runs GraphUpdater without clearing the database first. This complements the existing index_repository tool (which still performs a full wipe + reindex) and makes incremental updates available via MCP as well.

Thanks for reviewing!

### DIFF
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -533,7 +533,7 @@ class MCPToolsRegistry:
         `get_code_snippet` or `get_function_source_by_id` to inspect implementations.
 
         Args:
-            query: Natural language description of the desired functionality.
+            natural_language_query: Natural language description of the desired functionality.
             top_k: Maximum number of results to return (default: 5).
 
         Returns:
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -10,7 +10,10 @@ from codebase_rag.graph_updater import GraphUpdater
 from codebase_rag.parser_loader import load_parsers
 from codebase_rag.services.graph_service import MemgraphIngestor
 from codebase_rag.services.llm import CypherGenerator
-from codebase_rag.tools.code_retrieval import CodeRetriever, create_code_retrieval_tool
+from codebase_rag.tools.code_retrieval import (
+    CodeRetriever,
+    create_code_retrieval_tool,
+)
 from codebase_rag.tools.codebase_query import create_query_tool
 from codebase_rag.tools.directory_lister import (
     DirectoryLister,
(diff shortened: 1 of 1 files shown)


vitali87:
Hey, thanks for the contribution! The feature itself is really useful - having a non-destructive `update_repository` option and conditional `semantic_search` registration is exactly what we need. Nice thinking on the incremental updates.

A few things to address before we can merge:

**1. Coding Standards - String Centralization**

The project follows a strict "no hardcoded strings" policy. Tool names, descriptions, and log messages need to live in their respective modules:

- Tool names → `constants.py` (add to `MCPToolName` enum)
- Descriptions → `tools/tool_descriptions.py` (add to `MCP_TOOLS` dict)
- Log messages → `logs.py`

This ensures the new tools will also show up in the auto-generated README (we have a dynamic generator that pulls from these sources).

**2. Input Schema Structure**

The existing tools use `MCPInputSchema` and `MCPInputSchemaProperty` TypedDicts from `types_defs.py`. Would be great to keep that pattern consistent rather than inline dicts.

**3. No Docstrings**

Per our coding standards, we don't use docstrings - code should be self-documenting. Please remove all the docstrings from the new methods (`update_repository`, `semantic_search`, etc.). The tool descriptions in `tool_descriptions.py` serve as the documentation for MCP tools.

**4. Tool Descriptions Tone**

The warning about `index_repository` clearing data is important, but the current phrasing is a bit intense:
```
"NEVER USE THIS EXCEPT YOU ARE DIRECTLY ASKED TO, IT CLEANS ALL EXISTING EMBEDDINGS!!!"
```

Could we tone this down to something like:
```
"WARNING: Clears the entire database including embeddings. Use update_repository for incremental updates. Only use when explicitly requested."
```

**5. Minor Grammar**

A few spots have "except you" which should be "unless you" (e.g., "except you know exact names" → "unless you know exact names").

---

Would you mind making these adjustments? Happy to help if you have questions about where things should go. The feature is solid, just needs to align with how the rest of the codebase is structured.

vitali87 (changes requested):
See my comment above for details on what needs to be addressed.

luxannaxul:
I create a new PR with the new policy for doc strings and constant string definitions, becouse updating this one would be a mess with my amount of git knowledge. I hope that is fine

PR #215, opened 2026-01-04
feat: add --exclude flag and interactive prompt for directory exclusion

vitali87:


vitali87:
@greptile

vitali87:
@greptile

xxxajk:
Still processes the directory contents, it should be totally skipping them, unless it is tagging them to ignore in memgraph??
Seems like a waste of time... exclude should mean exclude, and not even look within it or do anything whatsoever.

PR #232, opened 2026-01-08
Fix/path utils exclude before include

### LINKED ISSUE #230: path_utils ignores excluded paths
The fix is very simple.
Eliminate the code block that checks include_paths.

Rationale:
If we check what to exclude first, then it's obvious to include it.

vitali87:
## Summary

Fixes #230

The bug was in `codebase_rag/utils/path_utils.py:15-20` where `include_paths` was checked before `exclude_paths`, causing an early return that bypassed exclusion checks.

**Before (buggy):**
```python
if include_paths:
    if path in include_paths:
        return False  # Early return, exclude check never reached
# exclude check here
```

**After (fixed):**
```python
if exclude_paths and any(part in exclude_paths for part in dir_parts):
    return True  # Check exclusions FIRST
if include_paths:
    if path in include_paths:
        return False
```

This ensures user-provided exclusions are always respected, even within included directories.

## Changes

- Check `exclude_paths` before `include_paths` in `should_skip_path()`
- Add 18 comprehensive tests across 5 new test classes:
  - `TestIncludeExcludeInteraction` (5 tests) - core bug reproduction
  - `TestIncludePathsEdgeCases` (4 tests)
  - `TestExcludePathsEdgeCases` (4 tests)
  - `TestIgnoreSuffixesInteraction` (2 tests)
  - `TestDirectoryVsFileBehavior` (3 tests)

## Test plan

- [x] All 2725 tests pass (`make test-parallel-all`)
- [x] New `TestIncludeExcludeInteraction` tests verify the fix

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -7,7 +7,7 @@ from loguru import logger
 from . import cli_help as ch
 from . import constants as cs
 from . import logs as ls
-from .config import settings
+from .config import load_cgrignore_patterns, settings
 from .graph_updater import GraphUpdater
 from .main import (
     app_context,
@@ -104,12 +104,15 @@ def start(
             style(cs.CLI_MSG_UPDATING_GRAPH.format(path=repo_to_update), cs.Color.GREEN)
         )
 
-        exclude_paths = frozenset(exclude) if exclude else None
+        cgrignore = load_cgrignore_patterns(repo_to_update)
+        cli_excludes = frozenset(exclude) if exclude else frozenset()
+        exclude_paths = cli_excludes | cgrignore.exclude or None
         unignore_paths: frozenset[str] | None = None
         if interactive_setup:
             unignore_paths = prompt_for_unignored_directories(repo_to_update, exclude)
         else:
             app_context.console.print(style(cs.CLI_MSG_AUTO_EXCLUDE, cs.Color.YELLOW))
+            unignore_paths = cgrignore.unignore or None
 
         with connect_memgraph(effective_batch_size) as ingestor:
             if clean:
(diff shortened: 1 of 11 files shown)


vitali87:
@gemini-code-assist This is **not a regression** - it's the **original intended behavior**.

The existing test `test_include_paths_overrides_default_skip` (written before this PR) explicitly tests that `include_paths` overrides `IGNORE_PATTERNS`:

```python
def test_include_paths_overrides_default_skip(self, tmp_path: Path) -> None:
    include_paths = frozenset({"submodule1/.git"})
    
    assert should_skip_path(file_in_root_git, tmp_path)  # Root .git skipped
    assert not should_skip_path(
        file_in_sub1_git, tmp_path, include_paths=include_paths
    )  # submodule1/.git NOT skipped - intentionally overrides IGNORE_PATTERNS
```

This test **passes** with both the old and new code. The behavior is unchanged.

The original design is:
1. `include_paths` = "I want to include these paths even if they'd normally be ignored"
2. `exclude_paths` = "I want to exclude these paths on top of the defaults"

This PR ensures `exclude_paths` takes precedence over `include_paths` (fixing #230), while preserving the original behavior that `include_paths` can override `IGNORE_PATTERNS`.

If we wanted `IGNORE_PATTERNS` to take precedence over `include_paths`, that would be a **different feature request**, not part of this bug fix.

vitali87:
@gemini-code-assist This suggestion proposes a **design change**, not a bug fix.

**Current behavior** (intentional):
- `include_paths` = "un-ignore these specific paths that would normally be excluded"
- Other paths not in `include_paths` are still processed normally

**Your suggestion**:
- `include_paths` = "ONLY process these paths, skip everything else" (whitelist)

This would break the existing test `test_include_paths_overrides_default_skip` and fundamentally change the feature's semantics. The current behavior is intentional - it allows users to selectively include paths like `submodule1/.git` that would normally be ignored, without requiring them to explicitly list every single path they want to process.

If whitelist behavior is desired, that would be a separate feature request with its own discussion.

vitali87:
@greptile

vitali87:
**@greptile-apps The operator precedence analysis is incorrect.**

The claim is that `cli_excludes | cgrignore.exclude or None` evaluates as `cli_excludes | (cgrignore.exclude or None)`. This is **wrong**.

`|` having **higher precedence** than `or` means `|` binds **tighter** (is evaluated **first**), so the expression parses as `(cli_excludes | cgrignore.exclude) or None`.

**Proof via AST:**
```python
>>> import ast
>>> ast.dump(ast.parse('a | b or c', mode='eval'), indent=2)
Expression(
  body=BoolOp(
    op=Or(),
    values=[
      BinOp(left=Name(id='a'), op=BitOr(), right=Name(id='b')),
      Name(id='c')]))
```

The `BinOp(BitOr)` is nested inside `BoolOp(Or)`, confirming `(a | b) or c`.

**Proof via runtime:**
```python
>>> frozenset() | frozenset() or None
None
>>> frozenset({'foo'}) | frozenset() or None
frozenset({'foo'})
```

No `TypeError` occurs. The code works as intended:
1. Compute `cli_excludes | cgrignore.exclude` (set union)
2. If result is empty (falsy), return `None`

The confidence score of 1/5 is based on a misunderstanding of Python operator precedence.

PR #283, opened 2026-01-27
fix: Update Memgraph index 

antejavor:
This PR introduces the label property index since the constraints and indexes are separated in Memgraph storage.

Context on the topic: https://memgraph.com/docs/fundamentals/constraints#uniqueness-constraint

### DIFF
--- a/codebase_rag/cypher_queries.py
+++ b/codebase_rag/cypher_queries.py
@@ -99,8 +99,7 @@ def build_constraint_query(label: str, prop: str) -> str:
 
 
 def build_index_query(label: str, prop: str) -> str:
-    """Build a label-property index query for efficient MERGE operations.
-    """
+    """Build a label-property index query for efficient MERGE operations."""
     return f"CREATE INDEX ON :{label}({prop});"
 
 
--- a/codebase_rag/cypher_queries.py
+++ b/codebase_rag/cypher_queries.py
@@ -99,7 +99,6 @@ def build_constraint_query(label: str, prop: str) -> str:
 
 
 def build_index_query(label: str, prop: str) -> str:
-    """Build a label-property index query for efficient MERGE operations."""
     return f"CREATE INDEX ON :{label}({prop});"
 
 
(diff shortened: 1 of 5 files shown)


codebase_rag/cypher_queries.py
```diff
@@ -98,6 +98,11 @@ def build_constraint_query(label: str, prop: str) -> str:
     return f"CREATE CONSTRAINT ON (n:{label}) ASSERT n.{prop} IS UNIQUE;"
 
 
+def build_index_query(label: str, prop: str) -> str:
+    """Build a label-property index query for efficient MERGE operations."""
```
vitali87: ditto

codebase_rag/tests/test_node_relationship_coverage.py
```diff
@@ -245,6 +245,25 @@ def capture_query(query: str) -> None:
                 f"Missing constraint for {label.value}. Expected query: {expected}"
             )
 
+    def test_ensure_constraints_creates_all_indexes(self) -> None:
+        """Verify that label-property indexes are created for MERGE performance."""
```
vitali87: Please drop the docstring - they are not allowed encouraged in this repo.

codebase_rag/services/graph_service.py
```diff
@@ -185,6 +186,22 @@ def ensure_constraints(self) -> None:
             except Exception:
                 pass
         logger.info(ls.MG_CONSTRAINTS_DONE)
+        self._ensure_indexes()
+
+    def _ensure_indexes(self) -> None:
+        """Create label-property indexes for efficient MERGE operations.
```
vitali87: ditto

vitali87: If you really want to keep this explanation, consider adding human-labelled comments with (H)
``` # (H) In Memgraph, uniqueness constraints do NOT automatically create indexes.                                                                                                          
  # (H) Without explicit indexes, MERGE operations perform full scan operations.                                                                                                          
  def _ensure_indexes(self) -> None:                                                                                                                                                      
      ...```

vitali87 (changes requested):

PR #287, opened 2026-01-30
fix: filter embedding generation by project name

evander-wang:
Fixes cross-repository embedding generation by filtering functions from the current project only. Previously, indexing a repository would attempt to generate embeddings for all functions in the database, including those from other projects, leading to file not found errors.


vitali87:
@greptile

vitali87 (changes requested):
Please fix the greptile's suggestion

Issue #307, opened 2026-02-19
[Bug]: Contradicting information about env setup


Alameyo:
### Bug Description

We can't run the project on Vertex AI with the example environment configuration. I may be doing something wrong, but then it would require some clarification in the documentation

### Steps to Reproduce

If you try to run cgr using this config
```
# Example 4: Google Vertex AI
# ORCHESTRATOR_PROVIDER=google
# ORCHESTRATOR_MODEL=gemini-2.5-pro
# ORCHESTRATOR_PROJECT_ID=your-gcp-project-id
# ORCHESTRATOR_REGION=us-central1
# ORCHESTRATOR_PROVIDER_TYPE=vertex
# ORCHESTRATOR_SERVICE_ACCOUNT_FILE=/path/to/service-account.json

# CYPHER_PROVIDER=google
# CYPHER_MODEL=gemini-2.5-flash
# CYPHER_PROJECT_ID=your-gcp-project-id
# CYPHER_REGION=us-central1
# CYPHER_PROVIDER_TYPE=vertex
# CYPHER_SERVICE_ACCOUNT_FILE=/path/to/service-account.json
```
Cgr will tell me I am missing GOOGLE_API_KEY env and even give me link where to get this API key
```
cgr start --repo-path /home/ubuntu/workspace/some-repo-i-am-testing

─── API Key Missing ───────────────────────────────────────────────

  Error: GOOGLE_API_KEY environment variable is not set.
         This is required to use Google AI for cypher.

  To fix this:

  1. Get your API key from:
     https://console.cloud.google.com/apis/credentials

  2. Set it in your environment:
     export GOOGLE_API_KEY='your-key-here'

     Or add it to your .env file in the project root:
     GOOGLE_API_KEY=your-key-here

  3. Alternatively, you can use a local model with Ollama:
     (No API key required)


```

however if I set up GOOGLE_API_KEY I will get error saying extra input is not allowed in envs.
```
Traceback (most recent call last):
  File "/home/ubuntu//code-graph-rag/.venv/bin/cgr", line 4, in <module>
    from codebase_rag.cli import app
  File "/home/ubuntu/code-graph-rag/codebase_rag/cli.py", line 12, in <module>
    from .config import load_cgrignore_patterns, settings
  File "/home/ubuntu/code-graph-rag/codebase_rag/config.py", line 325, in <module>
    settings = AppConfig()
               ^^^^^^^^^^^
[...]

PR #308, opened 2026-02-19
fix: use service account file for vertexai client initialization

Alameyo:
Fixes issue https://github.com/vitali87/code-graph-rag/issues/307.

There are 2 parts of this PR.

-  First, if VertexIAI is used, don't check for GOOGLE_API_KEY because the service account file is used instead.
-  Second, add vertexai variable and service account to the client initialization

Without these I was unable to run cgr on VertexAI

### DIFF
--- a/codebase_rag/tools/document_analyzer.py
+++ b/codebase_rag/tools/document_analyzer.py
@@ -35,6 +35,8 @@ class DocumentAnalyzer:
         if orchestrator_provider == cs.Provider.GOOGLE:
             if orchestrator_config.provider_type == cs.GoogleProviderType.VERTEX:
                 self.client = genai.Client(
+                    vertexai=True,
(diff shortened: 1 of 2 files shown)


vitali87 (changes requested):
@Alameyo The validate_api_key fix (part 1) has already been merged in #349 with additional type safety and test coverage, so that part is now redundant.                                                                                                                                                                                                                                                   But, the document_analyzer.py change (passing vertexai=True and credentials= to genai.Client) is still needed and not covered by #349. Could you rebase this branch on main, drop the config.py change, and keep just the document_analyzer.py fix? That way we can get the remaining piece merged cleanly.

Alameyo:
@vitali87 I have removed changes to config.py and rebased it onto the current main branch.

Alameyo:
@vitali87, anything still preventing the merge?

PR #339, opened 2026-02-20
fix(graph): guard flush_all with try/except on exception path in __exit__

### LINKED ISSUE #338: [Bug]: bug(graph_service): flush_all in __exit__ can mask original exception on connection failure
### Bug Description

File
`codebase_rag/services/graph_service.py:78-89`

Code
def __exit__(
    self,
    exc_type: type | None,
    exc_val: Exception | None,
    exc_tb: types.TracebackType | None,
) -> None:
    if exc_type:
        logger.exception(ls.MG_EXCEPTION.format(error=exc_val))
    self.flush_all()
    if self.conn:
        self.conn.close()
        logger.info(ls.MG_DISCONNECTED)
Bug
When an exception occurs inside the `with MemgraphIngestor(...) as ingestor:` block (e.g., a database connection error), `exit` is called with `exc_type` set. It logs the exception but then unconditionally calls `self.flush_all()` on line 86. If the original exception was caused by a connection failure, `flush_all()` will also fail (since `self.conn` is still set but the connection is broken), raising a new exception. This new exception replaces the original exception, making it impossible to diagnose the root cause. The original exception is lost because Python's context manager protocol propagates only the most recent exception from `exit`.

### Steps to Reproduce

NA

### Expected Behavior

`flush_all()` should be skipped when there is an active exception, or it should be wrapped in its own `try/except` block to prevent masking the original error:

def __exit__(self, exc_type, exc_val, exc_tb) -> None:
    if exc_type:
        logger.exception(ls.MG_EXCEPTION.format(error=exc_val))
    else:
        self.flush_all()
    if self.conn:
        self.conn.close()
        logger.info(ls.MG_DISCONNECTED)
Or alternatively, wrap the flush in a try/except:

try:
    self.flush_all()
except Exception as flush_error:
    logger.error(f"Failed to flush during cleanup: {flush_error}")

### Actual Behavior

NA

### Error Logs

```shell

```

### Programming Language

Other/Not Applicable

### Version

0.0.69

### Operating System

macOS

### Python Version

Python 3.12.2

### LLM Provider

Not Applicable

### Additional Context

_No response_

### Checklist

- [x] I have searched existing issues to ensure this is not a duplicate
- [x] I have included all relevant error logs and steps to reproduce
- [x] I am using a supported version of Python (3.12+)

vitali87:
## Summary
- Wrap `flush_all()` in `MemgraphIngestor.__exit__` with `try/except` on the exception path so a secondary flush failure never masks the original error
- On the success path, `flush_all()` raises normally so callers are notified of data loss

## Test plan
- [ ] Verify that normal `with` block usage still flushes and raises on flush failure
- [ ] Verify that when an exception occurs inside the `with` block, `flush_all()` is attempted but its failure is logged instead of masking the original exception

Closes #338


vitali87:
@greptile

vitali87:
@greptile

PR #352, opened 2026-02-21
ci(pypi): add trusted publisher workflow for PyPI releases

vitali87:
## Summary
- Add GitHub Actions workflow that publishes to PyPI on release creation
- Uses OIDC trusted publishers (no API tokens needed)
- Requires the `pypi` GitHub environment (already created) and the pending publisher registered on PyPI (already configured)

## Usage
1. Create a GitHub Release with a tag (e.g., `v0.0.76`)
2. The workflow builds with `uv build` and publishes via `pypa/gh-action-pypi-publish`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,37 @@
+name: Publish to PyPI
+
+on:
+  release:
+    types: [published]
+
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile

PR #359, opened 2026-02-25
feat(memgraph): add optional authentication support

vitali87:
## Summary

- Add `MEMGRAPH_USERNAME` and `MEMGRAPH_PASSWORD` environment variables and config fields
- Update `MemgraphIngestor` to accept optional `username`/`password` and pass them to `mgclient.connect()` conditionally
- Validate that both credentials are provided together (or neither)
- Thread credentials through all connection sites (`main.py`, `mcp/server.py`, `realtime_updater.py`)

Based on #160 by @qdrddr, rebased onto current `main` with adaptations for codebase conventions.

## Test plan

- [x] Unit tests for auth validation (username only, password only, both, neither)
- [x] Unit tests for conditional `mgclient.connect()` call (with and without auth)
- [x] Full test suite passes (3169 passed)


vitali87:
@greptile

PR #366, opened 2026-02-26
style(docs): use graph icon logo and default to dark mode

vitali87:
## Summary

- Replace full logo with just the graph icon (three colored circles with edges) for a cleaner header
- Regenerate favicon from the cropped graph icon
- Default to dark mode regardless of OS preference by removing the `media` query from the dark palette entry

## Test plan

- [ ] Verify the graph icon appears in the docs header instead of the full logo with text
- [ ] Verify the favicon in the browser tab shows the graph icon
- [ ] Confirm dark mode loads by default on first visit (even with OS set to light mode)
- [ ] Confirm the light mode toggle still works


vitali87:
@greptile

PR #405, opened 2026-03-01
fix(realtime): handle non-code files and filter spurious events

bhargavchippada:
## Summary

This PR fixes three issues in the real-time file watcher (`realtime_updater.py`):

1. **Filter spurious file system events**: Only process `MODIFIED`, `CREATED`, and `deleted` events
   - Previously, read-only events like `opened` and `closed_no_write` (triggered by IDEs accessing files) would cause files to be deleted from the graph but not recreated
   - This happened because Step 1 (deletion) ran for all events, but Step 3 (recreation) only ran for `MODIFIED`/`CREATED` events

2. **Delete File nodes for non-code files**: Added query to delete `File` nodes
   - The existing `CYPHER_DELETE_MODULE` query only deletes `Module` nodes (for code files)
   - Non-code files like `.md`, `.json`, etc. were never removed from the graph when deleted from the filesystem

3. **Create File nodes for ALL file types**: Added `process_generic_file()` call for all files
   - Previously, only code files with recognized language configs were indexed in real-time
   - Non-code files were only indexed during the initial full scan, not during real-time updates

## Test plan

- [x] Created a `.md` file → verified it appears in the graph
- [x] Deleted the `.md` file → verified it's removed from the graph
- [x] Opened a file in IDE without modifying → verified no false deletions occur
- [x] Modified a Python file → verified code structure is updated correctly

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -848,9 +848,11 @@ PAYLOAD_QUALIFIED_NAME = "qualified_name"
 class EventType(StrEnum):
     MODIFIED = "modified"
     CREATED = "created"
+    DELETED = "deleted"
 
 
 CYPHER_DELETE_MODULE = "MATCH (m:Module {path: $path})-[*0..]->(c) DETACH DELETE m, c"
+CYPHER_DELETE_FILE = "MATCH (f:File {path: $path}) DETACH DELETE f"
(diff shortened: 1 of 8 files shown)


vitali87:
Merged manually after resolving conflicts. Thanks @bhargavchippada!

PR #452, opened 2026-03-17
fix(cpp): attach CALLS edges to Method nodes instead of Module

### LINKED ISSUE #450: [Question]: C++ indexer attaches CALLS edges to Module node instead of individual Method nodes
### Question

When indexing a C++ project, CALLS relationship edges are attributed to the Module (file) node rather than to the specific Method node whose body contains the call. This means call-graph queries like "which function calls handleArtifact?" return zero results, even though the method nodes themselves are correctly indexed.

### What I'm Trying to Achieve

Nodes Connected with relationship of method - > method when exsists.
Example: get_cpu_usage() calls -> read_cpu()

### What I've Tried

Steps to reproduce

Index a C++ project that has a class with multiple methods, where one method calls another (e.g. handleArtifactWatcherCb calls handleArtifact in Player.cpp).
Query for callers of a specific method:
Cypher
MATCH (c:Function|Method)-[:CALLS]->(h:Method)
WHERE h.name = 'handleArtifact'
RETURN c.name, c.qualified_name
Observe that the result is empty.

Actual behavior
Zero results are returned for method-to-method queries.
However, when querying with the Module as source, the CALLS edges are found:

`MATCH ()-[r:CALLS]->(m:Method)
WHERE m.name = 'handleArtifact'
RETURN startNode(r), m.qualified_name`

Returns:

startNode: (:Module { name: 'Player.cpp', path: 'src/Player.cpp' })
target: Player.src.Player.Player.handleArtifact
So the graph correctly records "file Player.cpp calls handleArtifact" but loses attribution to which specific method inside that file made the call.

### Operating System

macOS

### Python Version

3.12

### Code-Graph-RAG Version

0.1.0

### Installation Method

uv

### Context/Code

```shell

```

### Additional Information

_No response_

### Checklist

- [x] I have searched existing issues and discussions
- [x] This is a question, not a bug report (use Bug Report template for bugs)

vitali87:
## Summary
- Fix C++ method name extraction in `_process_methods_in_class` to use `cpp_utils.extract_function_name()` instead of generic `_get_node_name()`, which returns `None` for C++ `function_definition` nodes (they nest the name inside `declarator`, not a direct `name` field)
- This caused all C++ class methods to be skipped in Pass 2 (class method call processing), so no `Method --CALLS--> Method` edges were created
- Pass 3 (module-level) still captured the calls, incorrectly attributing them to the Module node

## Test plan
- [ ] Index a C++ project with class methods that call other methods
- [ ] Verify `MATCH (c:Method)-[:CALLS]->(h:Method) WHERE h.name = 'handleArtifact' RETURN c.name` returns the calling method
- [ ] Verify no regressions: 3345 passed, 10 skipped, 3 xfailed (identical to main)

Closes #450


vitali87:
@greptile

vitali87:
@greptile

PR #453, opened 2026-03-18
fix(cli): make --clean work independently of --update-graph and delete hash cache

### LINKED ISSUE #434: fix: --clean should also delete the repo hash cache
## Problem

When running \`cgr start --clean\`, the Memgraph database is wiped but the per-repo hash cache file (\`.cgr-hash-cache.json\`) is left intact. On the next \`--update-graph\` run, the hash cache marks every file as unchanged, so 0 functions are re-indexed into the now-empty database.

This means \`--clean\` followed by \`--update-graph\` silently produces an empty graph unless the user manually deletes the cache file first.

## Steps to reproduce

\`\`\`sh
cgr start --clean --repo-path /some/repo --update-graph
# Graph is wiped, but .cgr-hash-cache.json still marks all files as unchanged
# Result: 0 nodes indexed
\`\`\`

## Multi-repo scenario (makes this worse)

When indexing multiple repos into the same database, \`--clean\` only deletes the cache for the repo passed via \`--repo-path\`. Other repos' caches remain stale. Example:

\`\`\`sh
# Wipe DB and reindex repo A (deletes A's cache correctly)
cgr start --clean --update-graph --repo-path /repos/A

# Reindex repo B — its cache still reflects the pre-wipe state,
# so all files appear unchanged and 0 nodes are added
cgr start --update-graph --repo-path /repos/B
\`\`\`

The user must manually \`rm /repos/B/.cgr-hash-cache.json\` before re-indexing B.

## Expected behaviour

\`--clean\` should delete (or invalidate) the hash cache at \`<repo-path>/.cgr-hash-cache.json\` so the subsequent update processes all files from scratch. For multi-repo workflows, users need to either delete all caches manually or run \`--clean --update-graph\` once per repo.

Longer term, a \`--clean-all\` flag or a standalone \`cgr clean\` command that wipes both the database and all known repo caches would address the multi-repo case cleanly.

## Proposed fix (single-repo)

In \`cli.py\`, after \`ingestor.clean_database()\`:

\`\`\`python
cache_path = repo_to_update / cs.HASH_CACHE_FILENAME
if cache_path.exists():
    cache_path.unlink()
\`\`\`

This is a one-liner fix with no architectural impact.

### LINKED ISSUE #436: bug: --clean has no effect without --update-graph
## Problem

\`--clean\` is documented as wiping the database, but it silently does nothing unless \`--update-graph\` is also passed.

The implementation gates \`clean_database()\` inside \`if update_graph:\`:

\`\`\`python
if update_graph:
    ...
    with connect_memgraph(...) as ingestor:
        if clean:
            ingestor.clean_database()
\`\`\`

So \`cgr start --clean --repo-path /some/repo\` connects to nothing and exits successfully, giving the user no indication that the database was not wiped.

## Expected behaviour

Either:
1. \`--clean\` works independently of \`--update-graph\` (wipe the database without re-indexing), or
2. A clear error/warning is emitted when \`--clean\` is passed without \`--update-graph\`

Option 1 is more useful and allows separating the "wipe" step from the "index" step in scripts.

## Workaround

Always combine the flags: \`cgr start --clean --update-graph --repo-path ...\`

vitali87:
## Summary
- `--clean` now works without `--update-graph`, connecting to Memgraph and wiping the database independently (fixes #436)
- `--clean` now deletes the `.cgr-hash-cache.json` file so subsequent `--update-graph` runs re-index all files from scratch (fixes #434)
- Both behaviors work for standalone `--clean` and combined `--clean --update-graph`

## Test plan
- [x] `--clean` alone wipes the database
- [x] `--clean` alone deletes the hash cache
- [x] `--clean` alone succeeds when no cache file exists
- [x] `--clean` alone does not invoke GraphUpdater
- [x] `--clean` alone shows completion message
- [x] `--clean --update-graph` deletes the hash cache
- [x] `--clean --update-graph` calls clean_database
- [x] `--update-graph` without `--clean` preserves the hash cache

Closes #436
Closes #434

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -86,7 +86,7 @@ def _delete_hash_cache(repo_path: Path) -> None:
                 cs.Color.YELLOW,
             )
         )
-        cache_path.unlink()
+        cache_path.unlink(missing_ok=True)
 
 
 @app.command(help=ch.CMD_START)
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -77,6 +77,18 @@ def _info(msg: str) -> None:
         app_context.console.print(msg)
 
 
+def _delete_hash_cache(repo_path: Path) -> None:
+    cache_path = repo_path / cs.HASH_CACHE_FILENAME
+    if cache_path.exists():
+        _info(
+            style(
+                cs.CLI_MSG_CLEANING_HASH_CACHE.format(path=cache_path),
+                cs.Color.YELLOW,
+            )
+        )
+        cache_path.unlink()
+
+
 @app.command(help=ch.CMD_START)
 def start(
     repo_path: str | None = typer.Option(
@@ -140,8 +152,6 @@ def start(
         )
         raise typer.Exit(1)
 
-    _update_and_validate_models(orchestrator, cypher)
-
     effective_batch_size = settings.resolve_batch_size(batch_size)
 
     if clean and not update_graph:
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile

PR #456, opened 2026-03-20
feat: add --project-name flag to override qualified-name prefix

### LINKED ISSUE #435: feat: add --project-name flag to override qualified-name prefix
## Problem

The project name used as the qualified-name prefix for all indexed nodes is derived from the last component of the repo path:

\`\`\`python
self.project_name = repo_path.resolve().name
\`\`\`

This creates confusing results when:

1. The repo is nested inside a same-named parent directory. For example, indexing \`~/code/BRender/BlazingRenderer\` produces correct names, but if someone previously indexed \`~/code/BRender\` directly, all nodes get the prefix \`BRender\` instead. Stale \`BRender.*\` nodes then coexist with fresh \`BlazingRenderer.*\` nodes in the same database with no way to distinguish or clean them.

2. The directory name is not a good identifier (e.g. \`v1.3.2\`, \`src\`, \`main\`).

## Proposed solution

Add a \`--project-name\` CLI flag that overrides the derived name:

\`\`\`sh
cgr start --update-graph \
  --repo-path ~/code/BRender/BlazingRenderer \
  --project-name BlazingRenderer
\`\`\`

Implementation is minimal: accept the flag in \`cli.py\`, pass it to \`GraphUpdater.__init__\`, and use it with a fallback:

\`\`\`python
self.project_name = project_name or repo_path.resolve().name
\`\`\`

vitali87:
## Summary
- Adds `--project-name` CLI flag to the `start` command to override the auto-derived project name (which defaults to `repo_path.resolve().name`)
- Fixes edge cases where directory name is a poor identifier (e.g. `src`, `v1.3.2`, nested same-name parent dirs)
- Closes #435

## Changes
- `cli_help.py` — added `HELP_PROJECT_NAME` constant
- `cli.py` — added `--project-name` option, passed through to `GraphUpdater`
- `graph_updater.py` — accepts optional `project_name` param with fallback to directory name
- `tests/test_project_name_flag.py` — 23 tests covering defaults, overrides, edge cases, factory propagation, and qualified name integration

## Test plan
- [x] 23 new tests all pass
- [x] Full suite: 3,424 passed, 85 skipped, 3 xfailed
- [x] 3 pre-existing failures (LLM API key + root chmod) — unrelated
- [x] Zero regressions

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -131,6 +131,11 @@ def start(
         min=1,
         help=ch.HELP_BATCH_SIZE,
     ),
+    project_name: str | None = typer.Option(
+        None,
+        "--project-name",
+        help=ch.HELP_PROJECT_NAME,
+    ),
     exclude: list[str] | None = typer.Option(
         None,
         "--exclude",
@@ -199,6 +204,7 @@ def start(
                 queries=queries,
                 unignore_paths=unignore_paths,
                 exclude_paths=exclude_paths,
+                project_name=project_name,
             )
             updater.run()
 
--- a/codebase_rag/cli_help.py
+++ b/codebase_rag/cli_help.py
@@ -73,6 +73,10 @@ HELP_GRAMMAR_URL = (
 )
 HELP_KEEP_SUBMODULE = "Keep the git submodule (default: remove it)"
 
+HELP_PROJECT_NAME = (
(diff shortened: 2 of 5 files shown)


vitali87:
@greptile

vitali87:
@greptile

PR #458, opened 2026-03-20
fix: catch PermissionError in directory lister for out-of-sandbox paths

### LINKED ISSUE #274: _get_safe_path does not catch raised value error
Tools that use paths should tell the LLM that the path requested is out of the allowed sandbox, and reply back with the suggestion to either use the whole path to the project, or to hint that the CWD is the project root.

vitali87:
## Summary
- Catches `PermissionError` from `_get_safe_path` in `list_directory_contents` and returns a helpful error message instead of an unhandled exception
- Adds `DIRECTORY_PATH_OUTSIDE_ROOT` error constant that tells the LLM the path is outside the project root and suggests using a relative path
- Closes #274

## Changes
- `tools/directory_lister.py` — wrap `_get_safe_path` call in try/except PermissionError
- `tool_errors.py` — add `DIRECTORY_PATH_OUTSIDE_ROOT` message
- `tests/test_directory_lister.py` — 2 new tests for relative and absolute out-of-bounds paths

## Test plan
- [x] 20 tests pass (including 2 new)
- [x] Lint and format clean

### DIFF
--- a/codebase_rag/tests/test_directory_lister.py
+++ b/codebase_rag/tests/test_directory_lister.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 from pydantic_ai import Tool
 
+from codebase_rag import tool_errors as te
 from codebase_rag.tools.directory_lister import (
     DirectoryLister,
     create_directory_lister_tool,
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile

PR #460, opened 2026-03-20
fix: type check error in stats command and add test coverage

vitali87:
## Summary
- Fixes `ty check` failure: `int(row.get("count", 0))` fails because `ResultValue` includes non-int types. Added proper isinstance narrowing.
- Adds 7 tests for the stats command covering: node table, relationship table, totals, empty graph, connection error, multi-label nodes, empty labels fallback.

## Changes
- `cli.py` — fix type narrowing for count value
- `tests/test_stats_command.py` — 7 new tests using mocked Memgraph connection

## Test plan
- [x] 7 tests pass
- [x] Lint and format clean


vitali87:
@greptile

PR #465, opened 2026-03-21
fix: resolve Anthropic/Azure API keys from env when default 'ollama' key is set

vitali87:
## Summary
- Fixes a bug where `ANTHROPIC_API_KEY` and `AZURE_API_KEY` env vars were ignored because the config passed the default `"ollama"` api_key, which is truthy and prevents the env var fallback
- The `or` in `self.api_key = api_key or os.environ.get(...)` never triggers when `api_key="ollama"`

## Root cause
When a user runs `cgr start --orchestrator anthropic:claude-opus-4-6` with `ANTHROPIC_API_KEY` set, the config's `api_key` field defaults to `"ollama"` (from the Ollama default). Since `"ollama"` is truthy, `AnthropicProvider.__init__` uses it as the API key instead of reading `ANTHROPIC_API_KEY` from the environment. The Anthropic API then rejects it with a 401.

## Fix
Check if `api_key` equals `DEFAULT_API_KEY` ("ollama") and treat it as unset, allowing the env var fallback to work. Applied to both `AnthropicProvider` and `AzureOpenAIProvider`.

## Test plan
- [x] 30 provider tests pass
- [x] Verified manually: `ANTHROPIC_API_KEY=test-key` correctly reaches the provider

### DIFF
--- a/codebase_rag/providers/base.py
+++ b/codebase_rag/providers/base.py
@@ -44,6 +44,11 @@ class ModelProvider(ABC):
         pass
 
 
+def _resolve_api_key(api_key: str | None, env_var: str) -> str | None:
+    effective = api_key if api_key and api_key != cs.DEFAULT_API_KEY else None
+    return effective or os.environ.get(env_var)
+
+
 class GoogleProvider(ModelProvider):
     __slots__ = (
         "api_key",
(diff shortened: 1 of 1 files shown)


vitali87:
@greptile

PR #511, opened 2026-03-27
fix: skip CALLS relationship when callee is a Class node

### LINKED ISSUE #497: fix(graph): Module CALLS Class relationship is semantically incorrect
## Problem

The knowledge graph can contain `Module -[CALLS]-> Class` edges, which are semantically incorrect. A Module does not "call" a Class. This was identified in the [gdotv blog analysis (Part 2)](https://gdotv.com/blog/codebase-rag-knowledge-graph-analysis-part-2/) as a "modeling flaw" where the CALLS relationship is being overloaded to represent structural containment rather than invocation.

The actual code relationship is that a class is declared/defined locally within a function scope inside a module (e.g., a class declared inside a function body in C++). The current code incorrectly creates a CALLS edge for this.

## Root Cause

In `codebase_rag/parsers/call_processor.py` line 194-203, `_process_module_level_calls` processes top-level calls in a file using `cs.NodeLabel.MODULE` as the caller type. When the call resolver resolves a reference to a Class node, the result is a `Module -[CALLS]-> Class` edge, which misrepresents the relationship.

```python
def _process_module_level_calls(self, root_node, module_qn, language, queries):
    self._ingest_function_calls(
        root_node, module_qn, cs.NodeLabel.MODULE, module_qn, language, queries
    )
```

## Expected Behavior

Module-level references to classes should either:
1. Not produce CALLS edges (classes are not "called" at module level, they are defined/used)
2. Or produce a more appropriate relationship type (e.g., DEFINES for locally-scoped class definitions)

## Impact

Misleading CALLS edges confuse both graph queries and LLM reasoning. When an LLM is asked "what does this module call?", classes should not appear in the results. This also pollutes call graph analysis.

## Affected Components

- `codebase_rag/parsers/call_processor.py` (lines 194-203, `_process_module_level_calls`)
- `codebase_rag/parsers/call_resolver.py` (resolution logic that returns Class as callee)

vitali87:
## Summary
- Filter out Class nodes from CALLS relationship creation in call_processor
- Module CALLS Class edges are semantically incorrect — modules don't call classes, they may instantiate or reference them
- Identified by gdotv blog analysis of knowledge graph structural integrity

Closes #497

### DIFF
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -376,6 +376,7 @@ FILE_OUTSIDE_ROOT = "Security risk: Attempted to {action} file outside of projec
 CALL_PROCESSING_FILE = "Processing calls in cached AST for: {path}"
 CALL_PROCESSING_FAILED = "Failed to process calls in {path}: {error}"
 CALL_FOUND_NODES = "Found {count} call nodes in {language} for {caller}"
+CALL_SKIP_CLASS = "Skipping CALLS edge from {caller} to {call_name} (callee is Class node: {callee_qn})"
 CALL_FOUND = (
     "Found call from {caller} to {call_name} (resolved as {callee_type}:{callee_qn})"
 )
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile

vitali87:
@codecov-commenter

PR #517, opened 2026-03-27
fix: add lexicographic tie-breaking for deterministic call resolution

### LINKED ISSUE #515: fix: non-deterministic graph output across runs for JS/TS, Rust, and Java repos
## Summary

The graph builder produces non-deterministic output across identical runs on the same repo. Node and relationship **counts** are stable, but the **content** (which specific calls resolve to which targets) varies between runs.

## Affected languages

JavaScript, TypeScript, Rust, and Java repos are most affected. Pure C, Python, PHP, Go, and Lua repos are deterministic.

## Reproduction

Run `export_goldens.py` twice consecutively on the same commit and diff the output. Repos like axios (JS), zod (TS), bat (Rust), and gson (Java) produce different graph digests despite identical inputs.

## Root cause hypothesis

Call resolution in `_try_resolve_via_trie` uses `find_ending_with` which returns sorted results, but the **trie itself** is populated in file-processing order. When multiple candidates have equal import distance scores, `min()` picks one arbitrarily based on iteration order of the candidates list. Different runs may populate the trie in subtly different order due to:

1. `os.walk` returning files in filesystem-dependent order (partially fixed by sorting dirnames/filenames)
2. Set iteration order for `simple_name_lookup` (Python sets are hash-ordered, which varies with PYTHONHASHSEED)
3. Dict insertion order for `_entries` affecting the linear scan fallback

## Suggested fix

Ensure `find_ending_with` returns results in a fully canonical order, and ensure tie-breaking in `_import_distance_fast` is deterministic (e.g., break ties by lexicographic QN order, which `min()` with a tuple key `(distance, qn)` would achieve).

## Impact

Prevents strict golden-based correctness testing for ~50% of benchmark repos. Currently, only 7 of 15 candidate repos produce deterministic output.

vitali87:
## Summary
- Add lexicographic qualified-name tie-breaking when multiple call resolution candidates have equal import-distance scores
- Ensures deterministic graph output across runs for JS/TS, Rust, and Java repos
- Fixes non-determinism that prevented golden-based correctness testing for ~50% of benchmark repos

Closes #515

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -395,6 +395,7 @@ class GraphUpdater:
                 )
             ):
                 eligible.append(filepath)
+        eligible.sort()
         return eligible
 
     def _process_files(self, force: bool = False) -> None:
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -221,7 +221,7 @@ class CallResolver:
             return None
 
         possible_matches.sort(
-            key=lambda qn: self._calculate_import_distance(qn, module_qn)
+            key=lambda qn: (self._calculate_import_distance(qn, module_qn), qn)
         )
         best_candidate_qn = possible_matches[0]
         logger.debug(ls.CALL_TRIE_FALLBACK, call_name=call_name, qn=best_candidate_qn)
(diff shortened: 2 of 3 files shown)


vitali87:
@greptile

PR #526, opened 2026-03-28
fix: replace Java stdlib subprocess with static prefix check and sorted os.walk

vitali87:
## Summary

- Replace the `javac`/`java` subprocess in `_extract_java_stdlib_path` with a static prefix check against known Java stdlib package prefixes
- Replace `rglob("*")` in `_collect_eligible_files` with sorted `os.walk` for deterministic file processing order

## Motivation

The Java stdlib extractor spawned `javac` and `java` subprocesses to determine if a fully qualified name belongs to the standard library. This caused:

1. **Non-deterministic output**: JVM timing and temp file handling produced different results across runs, causing relationship counts to vary by 4-8 across identical inputs
2. **Slow performance**: ~200ms per subprocess call
3. **Fragile**: Required JDK installation; failed silently without one

The `rglob("*")` call returned files in filesystem-dependent order, which combined with the existing `eligible.sort()` was mostly deterministic but still varied across different temp directory copies.

## Approach

**Static prefix check**: The Java standard library uses a known, finite set of top-level package prefixes. This is the same approach used by IntelliJ IDEA and Eclipse JDT for stdlib detection.

```python
JAVA_STDLIB_PREFIXES = (
    "java.", "javax.", "jdk.", "com.sun.", "sun.",
    "org.w3c.", "org.xml.", "org.ietf.", "org.omg.", "netscape.",
)
```

For class entities (uppercase first letter), the last component is stripped to get the package path. For package entities (lowercase), the full path is preserved. This correctly handles both `java.util.ArrayList` (resolves to `java.util`) and `java.nio.file.*` (preserves `java.nio.file`).

**Sorted os.walk**: Replaces `rglob("*")` with `os.walk` that sorts both `dirnames` and `filenames` at each level, ensuring identical traversal order regardless of filesystem layout.

## Test plan

- [x] All 3672 existing tests pass (0 failures)
- [x] `test_java_stdlib_introspection` passes (validates correct stdlib package resolution)
- [x] Java stdlib extraction is now 5000x faster (0.04μs vs 200ms per lookup)
- [x] No subprocess calls for Java stdlib detection

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1932,7 +1932,7 @@ CPP_STDLIB_ENTITIES = frozenset(
     }
 )
 
-# (H) Java common class names for heuristic detection
+# (H) Java stdlib package prefixes for static stdlib detection
 JAVA_STDLIB_PREFIXES = (
     "java.",
     "javax.",
@@ -1946,6 +1946,7 @@ JAVA_STDLIB_PREFIXES = (
     "netscape.",
 )
 
+# (H) Java common class names for heuristic detection
 JAVA_STDLIB_CLASSES = frozenset(
     {
         "String",
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1933,6 +1933,19 @@ CPP_STDLIB_ENTITIES = frozenset(
 )
 
 # (H) Java common class names for heuristic detection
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile

vitali87:
@greptile

vitali87:
@greptile

PR #533, opened 2026-06-27
fix(java): resolve field-access chains as method receivers and nested field access

### LINKED ISSUE #200: Java type inference chain method calls and field access
@vitali87 
I had a  question regarding the java type inference that it currently does not handle java chain method calls and field access such as :

```
class Address {
    public String zipCode = "90210";
}

class User {
    public Address address = new Address();
}

public class Main {
    public static void main(String[] args) {
        User obj = new User();
        
        // Chained Field Access
        String zip = obj.address.zipCode;
        System.out.println("Zip: " + zip);
    }
}
```

```
class QueryBuilder {
    private String query = "SELECT * ";

    public QueryBuilder from(String table) {
        this.query += "FROM " + table;
        return this; // Returning 'this' enables the chain
    }

    public QueryBuilder where(String condition) {
        this.query += " WHERE " + condition;
        return this;
    }

    public void execute() {
        System.out.println("Executing: " + query);
    }
}

public class Main {
    public static void main(String[] args) {
        QueryBuilder obj = new QueryBuilder();

        // Chained Method Calls
        obj.from("users").where("id = 1").execute();
    }
}
```

```
class Engine {
    public void start() {
        System.out.println("Engine started!");
    }
}

class Car {
    // The field
    public Engine engine = new Engine();
}

public class Main {
    public static void main(String[] args) {
        Car obj = new Car();

        // Mixed: obj (Car) -> field (engine) -> method (start)
        obj.engine.start();
    }
}
```

Also in java type inference the sequence of inferencing is :

```
            # 2. Analyze local variable declarations in the scope
            self._analyze_java_local_variables(scope_node, local_var_types, module_qn)
            # 3. Analyze field declarations from the containing class
            self._analyze_java_class_fields(scope_node, local_var_types, module_qn)
```

Which I think should be:

```
            # 2. Analyze field declarations from the containing class
            self._analyze_java_class_fields(scope_node, local_var_types, module_qn)

            # 3. Analyze local variable declarations in the scope
            self._analyze_java_local_variables(scope_node, local_var_types, module_qn)
```

Parameters first (already correct)
Class fields second - Fields are declared at class level and their types are fully known without dependencies on local variables
Local variables third - Local variables can reference parameters and fields in their initializers

_Originally posted by @HRV1527 in https://github.com/vitali87/code-graph-rag/discussions/199_

vitali87:
Fixes #200.

Java type inference did not resolve method calls whose receiver is a field-access chain, nor multi-level field-access type inference. Verified against the issue's exact snippets:

- `obj.engine.start()` (field access then method) produced **no** CALLS edge.
- `obj.address.zipCode` (nested field access) did not infer a type.
- `obj.from("users").where("id = 1").execute()` (method chains) already worked and still does.

## Root causes

- `_resolve_java_object_type` (`method_resolver.py`) only resolved a receiver that was a single name (local var, `this`, import, or class). A dotted field-access receiver like `obj.engine` fell through to `None`, so the method never resolved.
- `_infer_java_field_access_type` (`variable_analyzer.py`) decoded the receiver as a plain variable name; for a nested access the object is itself a `field_access` (`obj.address`), which is not a variable, so the lookup returned `None`.

## Fixes

- Added `_resolve_field_access_chain_type`: resolve the base, then walk each field's declared type across classes via the existing `_lookup_java_field_type`, then resolve the method on the resulting type. Handles arbitrary depth (`obj.address.city.ping()`).
- Made `_infer_java_field_access_type` recurse when the object node is itself a field access.

Both changes are additive fallbacks that run only after the existing resolution paths fail, so `this.engine.start()`, static calls, and single-level access are unchanged.

## Tests

New `test_java_field_access_chains.py` (RED before, GREEN after):

- mixed field then method (`obj.engine.start()`)
- multi-level field then method (`obj.address.city.ping()`)
- nested field-access type inference via `var` (`var c = obj.address.city; c.ping()`)

Full non-integration suite: 3951 passed, 17 skipped. ty and ruff clean.

## Note

The issue also suggested reordering class-field analysis before local-variable analysis. That was not changed: the analyzers infer types from declarations rather than by evaluating initializers against a partial map, and the field merge already guards against clobbering locals, so the order is not the cause of these bugs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,6 +791,7 @@ my_build_output
 - **protobuf**
 - **defusedxml**: XML bomb protection for Python stdlib modules
 - **huggingface-hub**: Client library to download and publish models, datasets and other repos on the huggingface.co hub
+- **griffe**: Signatures for entire Python programs. Extract the structure, the frame, the skeleton of your project, to generate API documentation or find breaking changes in your API.
 <!-- /SECTION:dependencies -->
 
 ## 🤖 Agentic Workflow & Tools
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2250,6 +2250,7 @@ TS_FALSE = "false"
 # (H) Tree-sitter field names for child_by_field_name
 TS_FIELD_NAME = "name"
 TS_FIELD_TYPE = "type"
+TS_SCOPED_TYPE_IDENTIFIER = "scoped_type_identifier"
 TS_FIELD_SUPERCLASS = "superclass"
 TS_FIELD_INTERFACES = "interfaces"
 TS_FIELD_TYPE_PARAMETERS = "type_parameters"
--- a/codebase_rag/parsers/java/method_resolver.py
+++ b/codebase_rag/parsers/java/method_resolver.py
@@ -11,7 +11,11 @@ from ... import logs as ls
 from ...decorators import recursion_guard
 from ...types_defs import ASTNode, NodeType
 from ..utils import safe_decode_text
-from .utils import extract_method_call_info, get_class_context_from_qn
+from .utils import (
+    extract_class_info,
+    extract_method_call_info,
+    get_class_context_from_qn,
+)
 
 if TYPE_CHECKING:
     from pathlib import Path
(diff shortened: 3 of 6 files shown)


vitali87:
@greptile-apps

vitali87:
@greptile-apps

vitali87:
@greptile-apps

vitali87:
@greptile-apps

vitali87:
@greptile-apps

vitali87:
@greptile-apps

vitali87:
@greptile-apps

PR #535, opened 2026-06-27
fix(parser): correct module-level call attribution and close Rust/C call-recall gaps

vitali87:
## What

Corrects how `CALLS` edges are attributed and closes call-recall gaps in three languages. Foundation for trustworthy reachability analysis (e.g. dead-code detection).

## Changes

**1. Module-level call attribution (all languages)**
Previously every call nested inside a function was *also* emitted with the enclosing module as the caller, so `(Module)->callee` edges existed for calls the module never makes. Now the module is the caller only for genuine top-level calls (module-load time, including `if __name__ == "__main__"` blocks). `call_processor._filter_top_level_calls`.

**2. Rust turbofish calls**
`generic_function::<T>()` was captured by neither the call query nor name extraction, so the call was dropped. Added a `generic_function` query pattern and unwrap it to the underlying callee identifier.

**3. Rust macro-internal calls**
Calls inside macro token-trees (`println!("{}", f(x))`) were never `call_expression` nodes, so `f` was missed. Added a query pattern matching an identifier immediately followed by a token-tree, with a precision guard so bare interpolated identifiers (`println!("{}", value)`) do not become calls.

**4. C function-body calls**
C `function_definition` nests its name in a declarator (no `name` field), so the call-attribution pass skipped every C function and emitted zero C `CALLS`. C now uses the same declarator-aware name extractor as C++ (`_C_FAMILY_LANGUAGES`).

## Tests (red → green)

- New: `test_module_call_attribution` (top-level vs nested attribution), `test_rust_call_recall` (turbofish, macro calls, and the bare-identifier precision guard).
- Corrected two tests that had encoded the old over-attribution: `test_cpp_cross_file_singleton` (calls attributed to the enclosing method, not the file) and `test_lua_modern_features` (precise edge assertions instead of a duplicate-inflated count).
- The previously-skipped `test_c_language` call suite (24 tests) now passes.
- Full non-infra suite: 4088 passed, 0 failed.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1987,7 +1987,11 @@ TS_SCALA_COMPILATION_UNIT = "compilation_unit"
 TS_SCALA_FUNCTION_DEFINITION = "function_definition"
 TS_SCALA_FUNCTION_DECLARATION = "function_declaration"
 TS_SCALA_CALL_EXPRESSION = "call_expression"
-TS_SCALA_GENERIC_FUNCTION = "generic_function"
+# (H) Shared tree-sitter node type: a call with explicit type args, e.g. Rust
+# (H) turbofish `f::<T>()` and Scala `f[T]()`. Its `function` field holds the
+# (H) actual callee (identifier or scoped_identifier).
+TS_GENERIC_FUNCTION = "generic_function"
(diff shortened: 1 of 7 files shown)


vitali87:
@greptile-apps

PR #537, opened 2026-06-27
feat(evals): add L2 module-call attribution metric with ast oracle

vitali87:
## What

Adds **L2 module-call attribution** to the eval harness: does cgr attribute the right calls to the *module* (caller side)?

A call runs at module-load time -- and so belongs to the module -- iff it is a top-level statement, a decorator, or a default-argument expression, i.e. NOT inside a function body. The L3 execution trace cannot measure this: it records the innermost *function* frame as the caller and drops `<module>` frames, so module-level attribution is its structural blind spot. An `ast` oracle fills it.

## How

- **`evals/module_calls.py`**
  - `oracle_module_calls`: AST walk that visits decorators and argument defaults in the enclosing scope but function bodies as function scope; collects first-party, non-dunder module-level callee names.
  - `cgr_module_calls`: every `CALLS` edge whose caller is a `Module` node, keyed by `(module_file, callee_simple_name)`; a constructor resolved to `Class.__init__` is credited to `Class`.
  - `score_module_calls`: precision/recall over those edges, plus a `main()` for ad-hoc reporting (`uv run python -m evals.module_calls --target codebase_rag`).
- **README** gains an L2 section.

## Tests

`test_eval_module_calls.py` — fixture guard asserting precision == recall == 1.0 on a controlled case (top-level call, default-argument call, `__main__` call, and a nested call that must NOT be module-attributed).

## What it surfaced

On the whole `codebase_rag` target the metric is a sound lower bound (P=0.85, R=0.74) that exposes two real, separate gaps (the fixture proves attribution itself is correct):

- **Recall**: constructing first-party classes with no explicit `__init__` (NamedTuple/dataclass/pydantic) emits no edge — cgr has no method node to target. Tracked with the dead-code Class work.
- **Precision**: the trie suffix-match fallback occasionally resolves a module-level call to an unrelated first-party name.

### DIFF
--- a/codebase_rag/tests/test_eval_module_calls.py
+++ b/codebase_rag/tests/test_eval_module_calls.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from evals.module_calls import (
+    cgr_module_calls,
+    oracle_module_calls,
+    score_module_calls,
+)
+
+_FIXTURE = """def make_default():
+    return 1
+
+
+def helper():
+    return 2
+
+
+def main():
+    helper()
+
+
+def with_default(x=make_default()):
+    return x
+
+
+CONFIG = make_default()
+
+
+if __name__ == "__main__":
+    main()
+"""
+
+
+def _names(edges: set[tuple[str, ...]]) -> set[str]:
+    return {e.target_name for e in edges}
+
+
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile-apps

vitali87:
@greptile-apps

PR #540, opened 2026-06-28
feat(parser): emit module-load CALLS edges for bare decorators

vitali87:
## What

Emits `(Module)→decorator` `CALLS` edges for **bare** decorators (`@task`). Applying a decorator runs `task(func)` at module-load time, so it is a genuine module-level call — but a bare decorator is an identifier, not a call node, so the normal call pass never saw it. (Call decorators like `@register(...)` are call nodes and were already captured.)

This makes a function used only as a decorator reachable in the graph, removing a dead-code false positive (it becomes a real root once `cgr dead-code` consumes module-call roots).

## How

- `call_processor._ingest_decorator_calls`: for each decorated function/method, resolve each bare decorator's callee to a first-party function/method and emit a module-load `CALLS` edge.
- Scoped to module/class-body definitions via `_runs_at_module_load`: a decorator on a function nested inside another function runs at *that* function's call time, not at import, so it is not module-attributed.
- Python only (the L2-evaluated language); other languages' annotations are metadata, not calls.
- The L2 oracle (`evals/module_calls.py`) now also counts bare decorators, keeping the metric aligned.

## Tests

`test_decorator_call_edges.py`: bare decorator → module edge; call decorator → module edge; nested-function decorator → no module edge; undecorated → no edge. (The nested-function case was caught as an L2 eval false positive and fixed.)

## Verification

- Broad call/decorator/module/import sweep: 1022 passed, 0 failed.
- L2 module-call metric on `codebase_rag`: recall 0.74 → 0.76, precision 0.85 → 0.86 (correct decorator edges, no new spurious ones).
- ruff + ty + `check_no_docs` clean.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -148,6 +148,59 @@ class CallProcessor:
                 nested_starts.add(call.start_byte)
         return [c for c in all_call_nodes if c.start_byte not in nested_starts]
 
+    def _bare_decorator_name(self, decorator_node: Node) -> str | None:
+        # (H) A bare decorator `@task` / `@pkg.deco` (no call parens) is not a
+        # (H) `call` node, so the normal call pass misses it even though applying
+        # (H) it runs `task(func)` at module load. A call decorator `@deco(...)`
+        # (H) IS a call node and is already captured, so skip it here.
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #541, opened 2026-06-28
feat(dead-code): treat module-load callees as reachability roots

vitali87:
## What

Makes `cgr dead-code` treat **module-load callees** as reachability roots. A function/method called by a `Module` node runs when the module is imported — `if __name__ == "__main__": main()`, top-level calls, and bare-decorator applications — so it (and everything it transitively calls) is live.

One clause added to `build_dead_code_query`:

```cypher
OR exists((n)<-[:CALLS]-(:Module))
```

## Why

This is the payoff of the call-graph work in #535 (module-level call attribution) and #540 (decorator edges). Together they fix the two false positives the command shipped with:

- **`__main__` entry points**: `main()` invoked under `if __name__ == "__main__"` is a module-load call, so `main` and its callees are no longer reported dead.
- **Decorator definitions**: a function used only as `@deco` is called at module load, so the decorator function is no longer reported dead.

## End-to-end

Indexed a fixture with both patterns and ran `cgr dead-code`:

```
Dead Code Candidates
  Function  app.handler   (defined + decorated but never called)
  Function  app.orphan    (never referenced)
```

Only genuinely-unused functions are flagged; `main` (via `__main__`) and the decorator function are correctly live.

## Tests

- Unit: the query contains the module-load root clause.
- Integration (real Memgraph): a function reachable only via a `(:Module)-[:CALLS]->` edge — with no entry-point/decorator/export root — is not reported dead; its callees are live; a true orphan still is.
- 42 dead-code + cypher tests pass; ruff + ty + `check_no_docs` clean.

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -894,20 +894,20 @@ def _dead_code_params(
     project_name: str,
     entry_points: list[str],
     decorator_roots: list[str],
-    include_tests: bool,
 ) -> dict[str, PropertyValue]:
     root_decorators = sorted(
         {d.lower() for d in cs.DEFAULT_ROOT_DECORATORS}
         | {d.lower() for d in decorator_roots}
     )
-    params: dict[str, PropertyValue] = {
+    # (H) test_patterns is always passed: with tests included it makes test
+    # (H) functions roots; with tests excluded it filters test modules out of the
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #545, opened 2026-06-28
perf(query): reuse Memgraph connection in semantic search tools (#505)

### LINKED ISSUE #505: perf(query): semantic search creates new Memgraph connection per call instead of reusing
## Problem

In `codebase_rag/tools/semantic_search.py`, the `semantic_code_search()` function creates a new `MemgraphIngestor` context manager on every call rather than reusing the existing ingestor connection that the orchestrator already has. This adds unnecessary connection overhead per semantic search invocation.

Additionally, the code calls the private method `ingestor._execute_query()` directly instead of using the public `fetch_all()` API, which bypasses the public interface contract.

## Expected Behavior

1. Reuse the existing Memgraph connection/ingestor instance instead of creating a new one per search call
2. Use the public `fetch_all()` method instead of the private `_execute_query()`

## Affected Files

- `codebase_rag/tools/semantic_search.py` (lines 39-46)

ChetanyaRathi:
## Summary

- Reuse the orchestrator's existing `MemgraphIngestor` in `semantic_code_search` and `get_function_source_code` instead of opening a new Memgraph connection on every call.
- Switch the private `ingestor._execute_query()` calls to the public `fetch_all()` API.
- Thread the ingestor through the `create_semantic_search_tool` / `create_get_function_source_tool` factories and their call sites in `main.py` and `mcp/tools.py`. Behavior is unchanged — `fetch_all(query, params)` is the public wrapper around `_execute_query(query, params)`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Closes #505

## Test Plan

- Added `test_semantic_code_search_reuses_injected_ingestor`, which asserts the injected ingestor's `fetch_all()` is called and `_execute_query()` is never called — a regression guard against re-introducing a per-call connection.
- Migrated the existing `test_semantic_search.py` suite to pass a mock ingestor directly rather than patching `MemgraphIngestor`.
- Ran `uv run pytest -m "not integration"` locally: all 4019 non-integration tests pass. The only failures are the Java-oracle tests (`test_java_*_oracle.py`), which shell out to `javac` and fail solely because a JDK isn't installed in my local environment — unrelated to this change.

### DIFF
--- a/codebase_rag/main.py
+++ b/codebase_rag/main.py
@@ -1544,8 +1544,8 @@ def _initialize_services_and_agent(
     file_editor_tool = create_file_editor_tool(file_editor)
     shell_command_tool = create_shell_command_tool(shell_commander)
     directory_lister_tool = create_directory_lister_tool(directory_lister)
-    semantic_search_tool = create_semantic_search_tool()
-    function_source_tool = create_get_function_source_tool()
+    semantic_search_tool = create_semantic_search_tool(ingestor)
+    function_source_tool = create_get_function_source_tool(ingestor)
 
     confirmation_tool_names = ConfirmationToolNames(
         replace_code=file_editor_tool.name,
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -93,7 +93,7 @@ class MCPToolsRegistry:
                 create_semantic_search_tool,
             )
 
-            self._semantic_search_tool = create_semantic_search_tool()
+            self._semantic_search_tool = create_semantic_search_tool(self.ingestor)
             self._semantic_search_available = True
         else:
             logger.info(lg.MCP_SEMANTIC_NOT_AVAILABLE)
@@ -344,7 +344,7 @@ class MCPToolsRegistry:
                 self._file_editor_tool,
                 self._shell_command_tool,
                 self._directory_lister_tool,
-                create_get_function_source_tool(),
+                create_get_function_source_tool(self.ingestor),
             ]
             if self._semantic_search_tool is not None:
                 tools.append(self._semantic_search_tool)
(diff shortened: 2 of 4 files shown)


vitali87 (changes requested):
@ChetanyaRathi thank you for the PR. Please clear all greptile and gemini comments until the greptile Confidence Score becomes 5. 
I will review after.

ChetanyaRathi:
@greptileai please re-review. The previous pass ran against the first commit (`c74e94e`); the latest commit (`f95bd55`) addresses the open feedback:

- **Type hints:** the ingestor params now use `QueryProtocol` in `semantic_search.py` and both tool factories, matching how `main.py` already annotates it (resolves the mypy/pyright mismatch).
- **Defensive parsing:** the semantic-search result mapping uses `.get()` with fallbacks instead of direct key access.
- **Threading:** the blocking `fetch_all()` calls in the semantic-search and source-lookup tools are offloaded via `asyncio.to_thread`, so they no longer run on the event loop.

On the two "shared connection" P1s, I believe these are false positives: every query on the shared ingestor is serialized by `self._conn_lock` in `graph_service.py`. The `yield` in `_get_cursor()` sits inside the `with self._conn_lock:` block, so the lock is held across `execute()` and the fetch, not just cursor creation. `fetch_all()` / `execute_write()` / `list_projects()` all route through it, so concurrent calls queue on the lock rather than colliding on the connection. Happy to walk through it inline if the re-run still flags them.

ChetanyaRathi:
@vitali87 Greptile is now at 5/5 on the latest commit (f95bd55), and I've resolved the open review threads, also ready for your review whenever you get a chance. Thank you!

vitali87 (approved):

Issue #424, opened 2026-03-02
[Question]: Do we have plans to conduct a evaluation?


NoSugarCoffee:
### Question

I quickly glanced at the GitLab GKG evaluation results: https://gitlab.com/gitlab-org/rust/knowledge-graph/-/work_items/224. The results show an 8% performance improvement compared to using only the grep tool.

Do we have any plan to do benchmark.

### What I'm Trying to Achieve

No

### What I've Tried

No

### Operating System

Linux

### Python Version

3.12.0

### Code-Graph-RAG Version

master

### Installation Method

uv

### Context/Code

```shell

```

### Checklist

- [x] I have searched existing issues and discussions
- [x] This is a question, not a bug report (use Bug Report template for bugs)

vitali87:
Thanks for raising this @NoSugarCoffee — benchmarking is definitely planned.

I'm considering approaches like:
- Retrieval accuracy on code Q&A tasks (precision, recall, MRR)
- SWE-bench style evaluations comparing graph-augmented retrieval vs grep/LSP-only
- Code navigation tasks (e.g. "find all callers of X", "trace call chain A→B")
- Latency/throughput across repos of varying sizes

The GitLab GKG evaluation you linked is a great reference point.

If you have ideas on specific benchmarks, datasets, or methodology that would be useful, I'd love to hear your input!

vitali87:
## Yes, and there is now a benchmark for it

A deterministic retrieval benchmark has been added to the `evals/` harness (#546) to answer this directly: does graph-augmented retrieval find the code that calls a symbol better than grep? It runs with no LLM, no Docker, and no API keys.

```bash
uv run python -m evals.retrieval --target codebase_rag
```

For every first-party symbol `S`, the task is to find the files that call `S`. Three conditions (cgr graph, bare-name grep, call-tuned grep) are scored against one Python `ast` oracle over the same file and symbol universe, on the file-level call edge `(caller_file, callee_simple_name)`, which mirrors the GitLab GKG "did it open the right file" localization signal.

Result on this repo:

| condition | precision | recall | f1 |
|---|---|---|---|
| graph | 0.8457 | 0.9886 | 0.9116 |
| grep_call (call-tuned) | 0.3660 | 1.0000 | 0.5359 |
| grep_name (bare name) | 0.2350 | 1.0000 | 0.3806 |

[...]

PR #546, opened 2026-06-29
feat(evals): add graph-vs-grep retrieval benchmark

vitali87:
## What

Adds a deterministic retrieval benchmark to the `evals/` harness that answers the question raised in #424: does graph-augmented retrieval find the code that calls a symbol better than grep? It measures the retrieval layer in isolation, with no LLM, no Docker, and no API keys, which is the decoupled measurement the GitLab GKG evaluation flagged as out of scope.

Run it:

```bash
uv run python -m evals.retrieval --target codebase_rag
```

## Method

For every first-party symbol `S`, the task is to find the files that call `S`. The comparison unit is a file-level call edge `(caller_file, callee_simple_name)`, mirroring the GKG "did it open the right file" localization signal. Three conditions are scored against one Python `ast` oracle over the same file and first-party symbol universe:

- **graph**: cgr's resolved `CALLS`/`INSTANTIATES` edges, reduced to caller file plus callee simple name.
- **grep_name**: ripgrep for the bare token `\b(name)\b` (a user's first attempt).
- **grep_call**: ripgrep for the call-tuned `\b(name)\s*\(`.
- **oracle**: every `ast.Call` to a first-party, non-dunder name.

## Result (target: `codebase_rag`)

| label | tp | fp | fn | precision | recall | f1 |
|---|---|---|---|---|---|---|
| graph | 3217 | 587 | 37 | 0.8457 | 0.9886 | 0.9116 |
| grep_name | 3254 | 10590 | 0 | 0.2350 | 1.0000 | 0.3806 |
| grep_call | 3254 | 5637 | 0 | 0.3660 | 1.0000 | 0.5359 |

The resolved graph more than doubles the precision of even the call-tuned grep (0.85 vs 0.37) at near-perfect recall, for an F1 of 0.91 vs 0.54: a gain of roughly 0.38 absolute (about 70% relative) over the strongest grep baseline. Both greps reach recall 1.0 by construction (the oracle is name-based, so text search cannot miss), so the entire story is precision, which is where the graph wins.

## Note on the "8%" figure

The "8% over grep" mentioned in #424 does not appear in the GitLab GKG work item; that eval was an agentic SWE-bench-Lite pass rate (roughly 6 to 7 of 23 issues), not a retrieval-vs-grep number. This benchmark measures retrieval quality directly. The complementary agentic resolved-rate replication (hold agent and model fixed, vary only the tools) needs an LLM and a container harness, so it is documented as a follow-on rather than run inside this deterministic harness.

## Tests

`codebase_rag/tests/test_retrieval_eval.py` (red to green): oracle correctness, grep's two failure modes (a bare reference or import counts as a hit, and a definition site `def S(` is indistinguishable from a call), the scoring math, and a cgr smoke test.

Closes part of #424.


vitali87:
@greptile review

vitali87:
@greptile review

PR #549, opened 2026-06-29
feat(evals): add Java multi-language retrieval eval and fix three Java call-resolution bugs

vitali87:
## What

Adds a **Java multi-language retrieval eval** (`evals/java_retrieval.py`), mirroring the existing Go and Rust retrieval evals: for each first-party Java symbol, which files call it. cgr's Java `CALLS` edges, reduced to `(caller_file, callee_simple_name)`, are graded against method-invocation sites extracted by the JDK Compiler Tree API (`javac`), over the same first-party name universe. `javac` is independent of cgr's tree-sitter Java frontend.

The JDK structure oracle (`Oracle.java`) is extended to emit the trailing identifier of each `MethodInvocationTree`; `run_java_call_oracle` reduces it the same way the Go/Rust call oracles do.

## Bugs found and fixed

Running the eval on a real corpus (`java.util`, 349 files from the JDK `src.zip`) drove recall from **0** (every Java call dropped) to **0.52 at precision 1.0** (zero false positives) by fixing three real cgr bugs, each pinned by a RED→GREEN test:

1. **Caller-QN signature mismatch.** The definition pass registers a Java method node with its parameter signature (`Class.name(args)` — Java overloads), but the call pass built the enclosing-method caller qn without it (`Class.name`), so every Java method's `CALLS` from-endpoint matched no node and the edge would not attach in Memgraph. Fixed by routing the caller qn through the same signature build the definition pass uses. (`test_java_call_caller_qn.py`)
2. **Flat-layout cross-file resolution.** `find_package_start_index` returns `None` for any project not under a recognized `src/main/java` layout, so `_build_fqn_lookup_map` left the simple-name to module map empty and all cross-file resolution (instance dispatch in sibling files) silently failed. Fixed by falling back to the segment after the project root for flat/non-standard layouts.
3. **Static same-package receiver.** A static call on a bare class-name receiver in a sibling file (`T.make()`, same package, no import) never resolved because the receiver-type lookup only checked the current module and explicit imports. Fixed with a same-package class-name fallback in `_resolve_java_object_type`.

The remaining recall gap is interface/abstract dispatch (the name-based oracle counts a call whenever the callee name is any declared first-party method, but cgr emits an edge only when the concrete receiver type is statically knowable) and deep receiver-type inference (iterator/functional-interface/generic element types). Documented in `evals/README.md`.

## Validation

- Full suite: 4217 passed, 12 skipped, 0 failed
- Eval pinned by `codebase_rag/tests/test_java_retrieval_eval.py` (cgr matches the `javac` oracle on the fixture)


vitali87:
@greptile review

PR #550, opened 2026-06-29
feat(evals): add TypeScript multi-language retrieval eval and fix two TS call-graph bugs

vitali87:
## What

Adds a **TypeScript multi-language retrieval eval** (`evals/ts_retrieval.py`), mirroring the Go/Rust/Java retrieval evals: for each first-party TS symbol, which files call it. cgr's TS `CALLS` edges, reduced to `(caller_file, callee_simple_name)`, are graded against call sites extracted by the TypeScript compiler API, over the same first-party name universe. `tsc` is independent of cgr's tree-sitter TS frontend.

The tsc structure oracle (`ts_ast.js`) is extended to emit the callee of each `CallExpression` (bare identifier or property-access trailing name) and to name arrow / function expressions by their binding (`const foo = () => …` → `foo`), matching how cgr names them.

## Bugs found and fixed

Running the eval on a real library (`zod`, 88 non-test files from `packages/zod/src/v4`) drove recall from **0.45 to 0.75** at precision 1.0 (zero false positives) by fixing two real cgr bugs, each pinned by a RED→GREEN test:

1. **Exported-function duplication.** The definition pass already ingests an exported function / const-arrow at its natural qn, but the ES6-export pass (`ingest_exported_function`) re-registered it, and `register_unique_qn` minted a spurious `qn@line` duplicate node onto which call resolution then bound. ~54% of all TS call edges (1280/2370 on zod) carried a mangled `name@line` callee, and callers of the natural node were invisible. In an ESM codebase where most functions are exported, this pollutes the whole graph. Fixed by skipping the export-pass registration when the natural qn already exists. (`codebase_rag/tests/test_ts_export_no_duplicate.py`)
2. **Arrow-body calls unresolved.** The call pass derived a caller name with `child_by_field_name("name")`, which is empty for an arrow / function expression, so it skipped the entire `const f = () => …` body and never emitted its calls — and modern TS is mostly arrow functions. Fixed by recovering the binding name for the `variable_declarator` form so the body's calls attribute to the same qn the definition pass registered. (`codebase_rag/tests/test_ts_arrow_caller_calls.py`)

Bug 2 is pre-existing (verified by stashing the bug-1 fix); the bug-1 fix caused no regression.

Calls inside anonymous (returned/inline) arrows, method dispatch on typed receivers, and the name-based oracle's over-count of common method names (e.g. `error` on a receiver cgr cannot type). Documented in `evals/README.md`.

## Validation

- Full suite: 4223 passed, 12 skipped, 0 failed
- Eval pinned by `codebase_rag/tests/test_ts_retrieval_eval.py` (cgr matches the `tsc` oracle on the fixture)

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -58,6 +58,7 @@ _TYPED_LANGUAGES = frozenset(
 # (H) name lives in a nested declarator (no `name` field). Both need the libclang
 # (H) declarator-aware extractor rather than a plain child_by_field_name("name").
 _C_FAMILY_LANGUAGES = frozenset({cs.SupportedLanguage.C, cs.SupportedLanguage.CPP})
+_JS_TS_LANGUAGES = frozenset({cs.SupportedLanguage.JS, cs.SupportedLanguage.TS})
 
 
 class CallProcessor:
@@ -426,6 +427,8 @@ class CallProcessor:
                 func_name = cpp_utils.extract_function_name(func_node)
             else:
                 func_name = self._get_node_name(func_node)
+            if not func_name and language in _JS_TS_LANGUAGES:
+                func_name = self._js_ts_arrow_binding_name(func_node)
             if not func_name:
                 continue
             # (H) An out-of-line C++ method definition (`Ret Class::method() {...}`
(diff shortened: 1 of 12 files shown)


vitali87:
Disposition of the gemini-code-assist review:

- **[high] class-property arrow functions (`class T { helper = () => … }`)**: confirmed a real, separate gap — cgr does not model class-field arrow functions as nodes at all (no `T.helper` node is created by the definition pass, unlike `regular() {}`), so neither their declaration nor their body calls are captured. This is a definition-pass modeling change distinct from this PR's call-resolution fixes (`variable_declarator` arrows + exported-function de-duplication), with a broad JS/TS regression surface. Tracked as the next fix rather than expanded into this approved PR.
- **[medium] `OraclePayload(...)` → dict literal**: declined. The project standard mandates explicit TypedDict constructors over dict literals, and `ty` is clean on this code; the existing oracle helpers (e.g. `java_oracle`) use the same constructor form.
- **[medium] class-property arrow test**: will accompany the class-property-arrow fix above.

PR #551, opened 2026-06-29
fix(js-ts): model class-field arrow members and resolve their body calls

vitali87:
## What

Models **class-field arrow functions** (`class T { helper = () => … }`) in JS/TS, a gap surfaced by review feedback on #550 and confirmed with a repro: cgr created no node for such members (the arrow has no `name` field, so `ingest_method` returned early) and consequently dropped their declaration *and* their body's calls — unlike a normal `regular() {}` method.

## Fix

Two parts, mirroring how the definition and call passes already handle `const f = () => …`:

1. **Definition pass** (`parsers/utils.ingest_method`): when a JS/TS arrow / function expression has no `name` field, derive the binding name from the enclosing field definition (new `_js_ts_field_member_name`) and model it as `class_qn.helper` instead of dropping it.
2. **Call pass** (`call_processor`): `_js_ts_arrow_binding_name` is extended to the class-field case and used as a fallback in the class-method loop, so the arrow body's calls attribute to `class_qn.helper`.

Note: the helper compares the field's `value` node with `==` rather than `is`, because py-tree-sitter returns a fresh `Node` wrapper on each access (identity comparison always fails; `Node.__eq__` compares the node id).

## Validation

- New test `codebase_rag/tests/test_ts_class_field_arrow.py` asserts both the `T.helper` node exists and its body call resolves (with a regression guard that a normal method still works).
- Full suite: 4224 passed, 12 skipped, 0 failed.
- `ruff` and `ty` clean on changed files.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1632,11 +1632,11 @@ class CallProcessor:
         parent = func_node.parent
         if parent is None:
             return None
-        # (H) `==` not `is`: py-tree-sitter returns a fresh Node wrapper per access,
-        # (H) so identity comparison always fails; Node equality compares the id.
-        if parent.type != cs.TS_VARIABLE_DECLARATOR and (
-            parent.child_by_field_name(cs.FIELD_VALUE) != func_node
-        ):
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #552, opened 2026-06-29
feat(evals): add PHP multi-language retrieval eval and fix plain PHP function-call resolution

vitali87:
## What

Adds a **PHP multi-language retrieval eval** (`evals/php_retrieval.py`), mirroring the Go/Rust/Java/TS retrieval evals: for each first-party PHP symbol, which files call it. cgr's PHP `CALLS` edges, reduced to `(caller_file, callee_simple_name)`, are graded against call sites extracted by `php-parser`, over the same first-party name universe. `php-parser` is a pure-JS PHP parser, independent of cgr's tree-sitter PHP frontend (no PHP runtime needed).

The `php-parser` structure oracle (`php_ast.js`) is extended to emit the callee of each `call` (bare function `name`, or the trailing member of a `$obj->m()` / `Class::m()` lookup) and to carry real declaration names (it previously hardcoded `"decl"`, fine for the structure join but not for a name-based call universe).

## Bug found and fixed

Running the eval on a real library (`monolog`, 121 files from `src/Monolog`) drove recall from **0.97 to 0.99** at precision 1.0 (zero false positives) by fixing one real cgr bug:

**Plain PHP function calls were never resolved.** A PHP `function_call_expression` carries its callee as a `name`-type node under the `function` field — a type `_get_call_target_name`'s callee match arm did not handle — so no callee name was extracted and the `CALLS` edge was dropped. Only method (`$obj->m()`) and static (`Class::m()`) calls, which expose a `name` field directly (caught by the final fallback), resolved. Even a same-file `foo()` call produced no edge. monolog's almost-entirely-OOP code had masked the gap. Fixed by adding the PHP `name` type to the match arm. (`codebase_rag/tests/test_php_function_call.py`)

The remaining 3 monolog misses are the name-based oracle's over-count: monolog declares first-party methods named `substr`/`reset`/`fwrite` (e.g. a UTF-8-aware `Utils::substr`), so a bare call to the PHP builtin of the same name is counted as first-party; cgr's simple-name trie fallback links it to the same-named first-party symbol, which the file+name metric credits. Documented in `evals/README.md`.

## Validation

- Full suite: 4228 passed, 12 skipped, 0 failed
- Eval pinned by `codebase_rag/tests/test_php_retrieval_eval.py` (cgr matches the `php-parser` oracle on the fixture)

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -759,6 +759,7 @@ class CallProcessor:
                     | cs.CppNodeType.QUALIFIED_IDENTIFIER
                     | cs.TS_SCOPED_IDENTIFIER
                     | cs.TS_SELECTOR_EXPRESSION
+                    | cs.TS_PHP_NAME
                 ):
                     if func_child.text is not None:
                         return func_child.text.decode(cs.ENCODING_UTF8)
--- a/codebase_rag/tests/test_php_function_call.py
+++ b/codebase_rag/tests/test_php_function_call.py
@@ -1,0 +1,29 @@
+from pathlib import Path
+
+from evals.cgr_graph import _capture
+
+
+def _make(root: Path) -> None:
+    root.mkdir(parents=True, exist_ok=True)
+    (root / "util.php").write_text(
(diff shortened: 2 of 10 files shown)


vitali87:
@greptile review

PR #554, opened 2026-06-29
feat(evals): add C multi-language retrieval eval graded against an abstaining libclang oracle

vitali87:
## What

Adds a C dimension to the multi-language retrieval benchmark (mirrors Go/Rust/Java/TypeScript/PHP/Lua). cgr parses C with tree-sitter by default (`CPP_FRONTEND=libclang` is off), so a full libclang front end is an independent oracle.

### Eval

- `run_c_call_oracle` (in `evals/oracles/cpp_oracle.py`) + `evals/c_retrieval.py`: grades cgr's C `CALLS` reduced to `(caller_file, callee_simple_name)` against libclang call sites over the same first-party name universe.
- No `compile_commands.json` required: each `.c` is parsed directly, with the SDK sysroot and clang's builtin-header dir added so headers resolve. A translation unit that still errors (missing build-generated header) has a possibly-truncated AST, so the oracle **abstains** on it and the cgr side is held to the same covered files (graded-file count logged).
- Pinned by `codebase_rag/tests/test_c_retrieval_eval.py`.

## Validation on `jq` (18/21 src/*.c parsed cleanly)

precision **0.98**, recall **0.93**, F1 0.96. Every diff entry was classified against the source. Two causes:

1. **C preprocessor gap (inherent to tree-sitter, both directions).** All 11 FPs are calls cgr emits inside inactive `#ifdef`/`#if` branches that libclang compiles out (`USE_DECNUM`, `HAVE_PTHREAD`, `WIN32`, `YYDEBUG`) — cgr makes **zero misresolutions**. Most FNs are calls visible only after macro expansion (`jv_object_foreach` -> `jv_object_iter*`, jv_parse.c `token()`/`check_done()` macros, flex input macros, bison `api.prefix {jq_yy}` rename where cgr resolves the call under the textual name `yylex`).

2. **A genuine cgr coverage gap (NOT preprocessor).** In bison-generated `parser.c`, tree-sitter-c emits 20 `ERROR` nodes (cascading from an unrecovered parse error earlier in the generated file), and the normal `void yyerror(...)` definition (parser.c:406) lands inside one — so cgr never registers `yyerror` and cannot resolve calls to it. cgr silently drops real function definitions inside tree-sitter parse-error regions on machine-generated code. Rooted in the tree-sitter-c grammar (not cgr's resolution logic); reported, not hidden. Tracked for follow-up (grammar upgrade or routing C through the libclang frontend); a heuristic scrape of ERROR regions is deliberately avoided as a workaround.

## Checks

Full suite `4236 passed, 12 skipped, 0 failed`; ruff + ty clean.


vitali87:
@greptile review

vitali87:
@greptile review

PR #556, opened 2026-06-30
feat(evals): add C++ multi-language retrieval eval and fix namespace-dropped C++ caller qns

vitali87:
## Summary

Adds the C++ multi-language retrieval eval (cgr's C++ `CALLS` vs an independent `libclang` oracle) and fixes the dominant cgr bug it surfaced: the call pass dropped the enclosing `namespace` from C++ caller qualified names.

## The eval

For each first-party C++ function/member function, which files call it. cgr's C++ `CALLS` edges (reduced to `(caller_file, callee_simple_name)`) are graded against call sites extracted by `libclang`, over the same first-party name universe. cgr parses C++ with tree-sitter by default (`CPP_FRONTEND=libclang` off), so `libclang` is an independent oracle. No `compile_commands.json` is needed: each source is parsed directly with the SDK sysroot, the SDK's `libc++` headers, and first-party include dirs; a TU that still errors abstains (held out of the graded `covered` set on both sides). Both the C and C++ oracles now pin a system `libclang` whose clang version matches the SDK's `libc++` (the pip wheel's older clang cannot parse current C++ standard headers).

## The bug (root cause)

The definition pass binds a C++ free function or class inside a `namespace` to a namespaced qn (`module.ns.fn`, `module.ns.Class`), but the call pass built the enclosing caller's qn **without** the namespace (`module.fn`, `module.Class.method`). Every such `CALLS` edge's source pointed at a node that does not exist, so the call never attached. On `leveldb` (all in `namespace leveldb`), **904 of 1227 C++ call sources dangled**.

Fix: route both the free-function qn (`_build_nested_qualified_name` ignored `namespace_definition` ancestors) and the class qn through the same `cpp_utils.build_qualified_name` the definition pass uses, so caller and node qns always agree. Same family as the Go/Java/Rust caller-qn fixes.

RED → GREEN: `test_cpp_namespace_call_caller_qn.py` asserts a namespaced free function and a namespaced inline method attribute their calls to the namespaced caller node; it failed before the fix.

## Result on `leveldb` (40/42 core sources parse cleanly)

| | before | after |
|---|---|---|
| precision | 0.99 | **0.96** |
| recall | 0.54 | **0.82** |
| F1 | 0.70 | **0.88** |

Dangling C++ call sources: 904 → 251. The oracle grades a call only when libclang resolves its callee to a first-party declaration (`child.referenced`), so `std::` calls whose simple name collides with a first-party method are not counted.

## Remaining tail (documented in `evals/README.md`, not scoped away)

- Operator overloads (`operator=`, `operator[]`): `libclang` counts them as method calls; cgr models them as `builtin.cpp.*` — a metric difference, not a misresolution.
- Trie-fallback misresolution (the ~30 FP: `size`, `data`, `empty`, `clear`, `begin`, `end`): cgr's name-only fallback binds an external `std::` call to a same-named first-party method; the oracle correctly treats it as external.
- Receiver-type method dispatch and out-of-line static methods (`DB::Open`): needs C++ receiver type inference (C++ is not yet in the typed-language set), the same deeper gap as the Go/Java/Rust tails. Follow-on.

Full suite: 4242 passed, 12 skipped. ruff/ty clean.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #557, opened 2026-07-01
fix(graph): restore linear-scan fallback in find_ending_with

### LINKED ISSUE #513: fix: find_ending_with drops CALLS when suffix not in simple_name_lookup
## Summary

`FunctionRegistryTrie.find_ending_with()` silently drops call resolution results when `simple_name_lookup` is present but does not contain the searched suffix.

## Root cause

In `graph_updater.py`, the current code is:

```python
def find_ending_with(self, suffix):
    if self._simple_name_lookup is not None:
        if suffix in self._simple_name_lookup:
            return sorted(self._simple_name_lookup[suffix])
        return []          # <--- BUG: should fall through to linear scan
    return sorted(
        qn for qn in self._entries.keys() if qn.endswith(f".{suffix}")
    )
```

When `simple_name_lookup` exists but does not contain the suffix, it returns `[]` instead of falling through to the linear scan over `_entries`. The original code correctly fell through.

## Impact

14 CALLS relationships silently dropped on the axios repo (JavaScript). The bug affects any repo where functions are registered in the trie but their simple names are not indexed in `simple_name_lookup`. This can happen with generated names, IIFE patterns, or edge cases in JS/TS module systems.

## Fix

Restore the original fallback behavior:

```python
def find_ending_with(self, suffix):
    if self._simple_name_lookup is not None and suffix in self._simple_name_lookup:
        return sorted(self._simple_name_lookup[suffix])
    return sorted(
        qn for qn in self._entries.keys() if qn.endswith(f".{suffix}")
    )
```

## How it was found

Expanding the benchmark suite from 2 repos (requests, tree-sitter) to 10 repos across all supported languages revealed the discrepancy on the axios repo. The original 2 repos did not exercise the failing code path.

ChetanyaRathi:
## Summary

Fixes a bug where `FunctionRegistryTrie.find_ending_with()` silently dropped call-resolution results when `_simple_name_lookup` was populated but did not contain the searched suffix.

The method returned an empty result in that case instead of falling through to the linear scan over `_entries`, so any qualified name ending in `.<suffix>` was lost. This affects repos where a function is registered in the trie but its simple name is not indexed in `_simple_name_lookup` (generated names, IIFE patterns, some JS/TS module edge cases).

## Root cause

`find_ending_with` has two paths: a fast path over `_simple_name_lookup` and a
linear scan over `_entries`. `_simple_name_lookup` is optional — it is passed in
at construction and defaults to `None` (graph_updater.py:64-65), and when
present it is kept in sync by `insert` itself (graph_updater.py:117-118). So the
two failure-free cases are: lookup is `None` (scan runs) and lookup is present
and complete (fast path hits).

The pre-fix bug surfaced only in a third state — a lookup that is present but
does not contain the queried suffix. In that case the old code returned `[]`
instead of falling back to the `_entries` scan. The fix restores the fallback so
a present-but-incomplete lookup degrades to the scan rather than dropping the
result.

Note: because `insert` maintains a present lookup, this incomplete-lookup state
is not produced by `insert` alone; it requires the lookup to be populated by a
separate path. I verified the fallback behavior directly via the added unit
test, but I have not identified that construction path in the current tree, so
I'm treating the axios impact figure as reported in #513 rather than one I
traced.

## Impact

14 CALLS relationships were silently dropped on the axios repo (JavaScript). Any repo whose functions are registered in the trie without their simple names being indexed in `_simple_name_lookup` was affected.

## How it was found

Surfaced by expanding the benchmark suite from 2 repos (requests, tree-sitter) to 10 repos across the supported languages — the original two never exercised the failing path, but axios did.

## Test plan

Added a regression test (`test_find_ending_with_falls_back_when_suffix_missing_from_lookup`) covering all three branches: `_simple_name_lookup is None`, lookup present and containing the suffix, and lookup present but missing the suffix (the bug state). Verified it fails on the pre-fix code and passes with the fix. Full `test_trie_optimization.py` suite passes.

Fixes #513

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -257,11 +257,8 @@ class FunctionRegistryTrie:
         cached = self._ending_with_cache.get(suffix)
         if cached is not None:
             return cached
-        if self._simple_name_lookup is not None:
-            if suffix in self._simple_name_lookup:
-                result = sorted(self._simple_name_lookup[suffix])
-            else:
-                result = []
+        if self._simple_name_lookup is not None and suffix in self._simple_name_lookup:
+            result = sorted(self._simple_name_lookup[suffix])
         else:
             result = sorted(
                 qn for qn in self._entries.keys() if qn.endswith(f".{suffix}")
--- a/codebase_rag/tests/test_trie_optimization.py
+++ b/codebase_rag/tests/test_trie_optimization.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from pathlib import Path
 from unittest.mock import MagicMock
 
@@ -159,3 +160,19 @@ class TestTrieOptimization:
         registry["new.function.test"] = NodeType.FUNCTION
         assert "new.function.test" in registry
         assert registry["new.function.test"] == NodeType.FUNCTION
+
+    def test_find_ending_with_falls_back_when_suffix_missing_from_lookup(self) -> None:
+        trie_no_lookup = FunctionRegistryTrie(simple_name_lookup=None)
+        trie_no_lookup.insert("com.example.Logger.info", NodeType.FUNCTION)
+        assert trie_no_lookup.find_ending_with("info") == ["com.example.Logger.info"]
+
+        trie_with_lookup = FunctionRegistryTrie(simple_name_lookup=defaultdict(set))
+        trie_with_lookup.insert("com.example.Logger.debug", NodeType.FUNCTION)
+        assert trie_with_lookup.find_ending_with("debug") == [
+            "com.example.Logger.debug"
+        ]
+
+        trie_bug_state = FunctionRegistryTrie(simple_name_lookup=defaultdict(set))
+        trie_bug_state.insert("com.example.Logger.warn", NodeType.FUNCTION)
+        trie_bug_state._simple_name_lookup.pop("warn")
+        assert trie_bug_state.find_ending_with("warn") == ["com.example.Logger.warn"]


ChetanyaRathi:
The failing CI checks all error within ~2s, which looks like a shared setup
step failing rather than anything in this diff — the unit-test jobs don't run
long enough to have executed the suite, and the PR-title check is failing on a
title that matches the convention. On my end ruff and the
test_trie_optimization.py suite pass on the two changed files, and the T-Rex
review verified the fix by running the code. Happy to rebase onto latest main
if that would help CI go green.

vitali87:
Non-blocking: since `FunctionRegistryTrie.insert` always adds the simple name to the shared `simple_name_lookup`, the "registered in the trie but simple name not indexed" state is not reachable via `insert` alone, so could you cite the parser path that actually dropped the 14 axios CALLS (likely a direct dict mutation in `js_ts/ingest.py`)? The fix is correct either way; just want the root-cause note to match reality.

vitali87 (changes requested):
Read my comment aboave

ChetanyaRathi:
Traced this in graph_updater.py. The divergence isn't a parser mutating the
trie directly, I don't see any bulk ⁠ _entries[...] ⁠ write or ⁠ .insert() ⁠ call
under parsers/. It's narrower: insert only maintains the lookup when it's
non-None (graph_updater.py:117-118), and _simple_name_lookup is supplied at
construction (line 65) rather than always derived from _entries. So when a trie
holds names whose simple names aren't in the separately-populated lookup,
find_ending_with took the "lookup present" branch and returned [] instead of
falling back to the _entries scan, which the fix restores.

I've updated the root-cause note to reflect that. I verified the fallback
behavior itself via the direct-state unit test, but I haven't traced the exact
axios construction path that produced the 14-CALL drop, the impact figure is
from the issue. If you know the specific call site that builds the trie with the
mismatched lookup on the JS/TS path, happy to reference it precisely.

vitali87:
Thanks for this, @ChetanyaRathi — and for surfacing it by expanding the benchmark suite to 10 repos; that's exactly how the axios discrepancy got caught.

Closing as superseded: `main` already fixes this in `8ac54b3` (`fix(graph): fall back to linear scan for dotted suffixes in find_ending_with`).

The root cause turned out to be slightly different from the "present-but-incomplete lookup" framing here. `_simple_name_lookup` only ever holds a qualified name's **last segment** (see `insert`, `graph_updater.py:118-120`), so a **dotted** suffix like `Class.method` can never be a key in it — that miss was the real source of the dropped CALLS, which lines up with why the incomplete-lookup construction path couldn't be reproduced.

Because of that, falling back to the linear scan on *every* miss (as proposed here) would run a full O(n) scan over `_entries` on each dot-free miss too, which defeats the purpose of the `_simple_name_lookup` index. The version on main branches on it instead:

```python
if suffix in self._simple_name_lookup:      # fast path
    result = sorted(self._simple_name_lookup[suffix])
elif cs.SEPARATOR_DOT in suffix:            # dotted suffix always misses -> real fallback
    result = sorted(qn for qn in self._entries if qn.endswith(f".{suffix}"))
else:                                       # dot-free miss is authoritative, stays O(1)
    result = []
```

A dot-free miss stays O(1) (insert indexes every entry's last segment, so nothing can end in `.suffix`), and the scan only runs in the one genuinely ambiguous case. There's also a `_ending_with_cache` layer on main that this branch predates.

The regression coverage is in `test_find_ending_with_dotted_suffix_falls_back_to_scan`, which exercises the dotted-suffix fallback, the O(1) dot-free miss, and cache invalidation on both insert and delete. Appreciate the thorough writeup — #513 is closed.

PR #558, opened 2026-07-01
fix(cpp): skip forward-declaration class nodes that fragment classes and break graph determinism

vitali87:
## Summary

Fixes a C++ bug that both corrupted the call graph and made cgr's output **non-deterministic** across runs: forward declarations were registered as their own class nodes.

## The bug

A C/C++ forward declaration (`class VersionSet;`) is a bodyless `class_specifier`. The definition pass registered it as a `Class` node with zero methods. The real definition that followed then collided on the qualified name, so `register_unique_qn` suffixed it (`VersionSet@167`) — fragmenting **one** class into several same-named nodes across files. On `leveldb`, `VersionSet` had **5** class nodes (3 bodyless phantoms + a collision-split real definition).

Two consequences:

1. **Wrong-class edges.** Member-call resolution had multiple same-named class candidates to choose from.
2. **Non-deterministic graphs.** Candidate selection among the duplicates was hash-order dependent, so the same input produced different graphs each run. On the C++ retrieval eval this swung recall from ~0.69 to ~0.87 between runs (`PYTHONHASHSEED=0` made it reproducible — that was the tell).

## The fix

Skip bodyless class/struct/union specifiers in the definition pass: a forward declaration no longer creates a node; only the real (bodied) definition does. A type that is *only* ever forward-declared (opaque) correctly stays a non-node — cgr never knew its members anyway.

RED → GREEN: `test_cpp_forward_declaration.py` asserts a forward-declared-then-defined class produces exactly one class node (no `@line` fragmentation) with its method attached.

## Result on `leveldb`

- **Determinism restored**: cgr call edges are now identical across runs (763 / 763 / 763; previously 640–812).
- C++ retrieval eval, now stable: precision **0.96**, recall **0.83**, F1 **0.89**.

Full suite: 4243 passed, 12 skipped. ruff/ty clean.

Note: this non-determinism silently added ±0.1 variance to the C++ eval numbers reported in #554/#556; those were single-run snapshots. This makes the eval a reproducible instrument.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1826,6 +1826,13 @@ CPP_KEYWORD_CLASS = "class"
 CPP_KEYWORD_STRUCT = "struct"
 CPP_EXPORTED_CLASS_KEYWORDS = frozenset({CPP_KEYWORD_CLASS, CPP_KEYWORD_STRUCT})
 
+# (H) A C/C++ class/struct/union tag with no body is a forward declaration
+# (H) (`class Widget;`); it must not become its own node, or it collides with the
+# (H) real definition's qn and fragments one class into several same-named nodes.
+CPP_TYPE_SPECIFIER_NODE_TYPES = frozenset(
+    {"class_specifier", "struct_specifier", "union_specifier"}
+)
+
 CPP_FALLBACK_OPERATOR = "operator_unknown"
 CPP_FALLBACK_DESTRUCTOR = "~destructor"
 CPP_OPERATOR_TEXT_PREFIX = "operator"
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -525,6 +525,15 @@ class GraphUpdater:
         if go_methods:
             logger.info("Resolved {} Go receiver methods", go_methods)
 
+        kept_forwards = (
+            self.factory.definition_processor.resolve_deferred_forward_declarations()
+        )
+        if kept_forwards:
+            logger.info(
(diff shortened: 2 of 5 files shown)


vitali87:
@greptile review

vitali87:
On the template forward-declaration threads: scoped intentionally, and here is the evidence.

Extending the skip to `template_declaration`-wrapped forward declarations breaks valid C++ template metaprogramming. A primary template is frequently forward-declared and **only ever defined via partial specializations** (e.g. `Contains` in `test_cpp_templates.py::test_template_metaprogramming`), so its `template_declaration` node is the canonical class node, not a phantom. Skipping it drops the class entirely; I implemented that variant and it fails `test_template_metaprogramming` and `test_cpp_templates_comprehensive`.

The nondeterminism this PR fixes comes from **direct (non-template) forward declarations** colliding with same-file/cross-file definitions (leveldb's `VersionSet`, 5 fragmented nodes). With this fix the C++ call graph is fully deterministic (763/763/763 across hash seeds), which is verified. Template forward declarations do not contribute to that nondeterminism in the corpus.

A correct template-aware version would skip a template forward declaration only when a bodied definition of the same class also exists (a deferred two-pass check), which is a larger change out of scope for this determinism fix and does not affect it. Resolving these as won't-fix-here.

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #561, opened 2026-07-01
fix(cpp): suppress trie fallback for member calls on external-typed receivers

vitali87:
## What

A C++ member call whose receiver has a known external type (e.g. `std::string s; s.size();`) was mis-resolved to a same-named first-party method by the bare-name trie fallback. When receiver type inference resolves `s` to `std::string` (`std.string`), precise local-type resolution correctly fails (the type is not a first-party class), but the call then fell through to the simple-name trie which rebinds `size` to whichever first-party `size` method exists. This is the residual precision gap the C++ retrieval eval flagged on leveldb.

This adds a guard mirroring the existing external-import suppression (`_is_external_import`): for a two-part member call `obj.method`, if `obj` has an inferred local type that resolves to no first-party class (an unresolvable type, or a qn neither registered nor registered under the project prefix, rooted outside the project), the trie fallback is skipped and no edge is emitted.

Deliberately conservative to preserve recall:
- An untyped receiver (not in the local type map) keeps the fallback: its type is unknown, not known-external.
- A receiver whose type resolves to a first-party class is left alone, including bare-prefixed imports (`from models.user import User` -> `models.user.User`, stored project-prefixed) and cross-file imported-class calls the precise path missed.

## Test

`test_cpp_external_receiver_call_is_not_rebound_to_first_party`: a `std::string` receiver calling `.size()` must not produce a CALLS edge to a first-party `Widget.size`. RED before the guard, GREEN after.

Full suite: 4253 passed. Verified no regression across the Python imported-class/return-type inference suites (the guard's first cut over-suppressed those; the project-prefixed registry check restores them).

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -278,6 +278,17 @@ class CallResolver:
                 self._simple_resolution_cache[cache_key] = None
             return None
 
+        # (H) A member call `obj.method` whose receiver has a KNOWN inferred type that is
+        # (H) not a first-party class is a call on an external object (e.g. a
+        # (H) `std::string`). Precise local-type resolution above already failed, so the
+        # (H) method lives on the external type; do NOT let the simple-name trie fallback
+        # (H) rebind it to an unrelated first-party method of the same name. Untyped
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

Issue #555, opened 2026-06-29
tree-sitter-c drops function definitions after a #define whose body contains a block comment (affects C frontend on generated parsers)


vitali87:
## Summary

cgr's C frontend (tree-sitter, the default `CPP_FRONTEND=treesitter`) silently drops real function definitions that follow a `#define` macro whose body contains a block comment. The root cause is a bug in the `tree-sitter-c` grammar, present through the latest release (0.24.2). This issue tracks the investigation so we can file a well-supported upstream report against `tree-sitter/tree-sitter-c` after confirming the analysis.

## Root cause (minimal reproduction)

A block comment `/* */` embedded between tokens inside a `#define` macro body makes `tree-sitter-c` emit an `ERROR` node, and the parser fails to recover, so the next top-level declaration is dropped from the tree:

```c
#define M do { /*c*/; } while (0)
void after(void) {}
```

Parsing the above yields `(translation_unit (preproc_def ...) (ERROR ...))` with no `function_definition` for `after`. The same macro without the comment, or with the comment at the very end of the body, parses cleanly.

GNU bison emits exactly this shape. Its generated parser contains a macro such as:

```c
#define FAIL(loc, msg)                                             \
  do {                                                             \
    location l = loc;                                              \
    yyerror(&l, answer, errors, locations, lexer_param_ptr, msg);  \
    /*YYERROR*/;                                                   \
  } while (0)

void yyerror(YYLTYPE* loc, ...) { ... }
```

The `/*YYERROR*/` comment inside `FAIL` triggers the bug, and the `yyerror` definition immediately after it is never registered.

## Impact on cgr

Surfaced by the C multi-language retrieval eval (PR #554) on the `jq` codebase. In `jq`'s bison generated `parser.c`, tree-sitter produces 20 `ERROR` nodes, and the `void yyerror(...)` definition at `parser.c:406` falls inside one. cgr therefore never creates a `yyerror` function node and cannot resolve any call to it. Any function definition that follows a comment bearing `#define` macro in machine generated C is affected. Hand written C and most source is unaffected; the practical blast radius is generated parsers and lexers.

cgr's resolution logic is correct (zero misresolutions were found); this is purely a parse coverage hole inherited from the grammar.

## Version findings

* Reproduces on the pinned `tree-sitter-c` 0.24.1 and on the latest 0.24.2, so a version bump does not fix it.
[...]

PR #297, opened 2026-02-05
feat: add progress bar for indexing

tibisabau:
## Description

Added a progress bar during codebase indexing/parsing to provide visual feedback on processing status. The implementation uses the `rich` library (already in project dependencies) to display a spinner, descriptive text, and "M of N" completion counter. The progress bar is transient, automatically disappearing after completion to avoid interfering with logging output.

Closes #241.

## Checklist if Applicable

- [x] The tests passed 
```bash
make pre-commit
```
- [x] Linting passed 
```bash
ruff format .
 
ruff check --fix .

ty check --exclude codebase_rag/tests/
```
- [ ] Documentation has been added
- [ ] CHANGELOG.md has been updated

PR #564, opened 2026-07-01
docs(evals): cite tree-sitter-cpp #297 for the C++ member-init-list parse trigger

vitali87:
The C++ retrieval residual's dominant parse-corruption cause was reduced last session to an `#if`/`#endif` inside a constructor member-initializer list. That exact trigger is already reported upstream as [tree-sitter-cpp #297](https://github.com/tree-sitter/tree-sitter-cpp/issues/297) ("ifdef not supported inside constructor member initializer list") — its playground repro matches ours byte-for-byte.

This replaces the vague "same family as #555" note in the C++ retrieval section of `evals/README.md` with the precise upstream reference, and clarifies it is a sibling of (not the same as) the tree-sitter-c grammar bug tracked in #555.

Docs only; no code or behavior change.


vitali87:
@greptile review

PR #565, opened 2026-07-01
fix(cpp): attribute module-scope initializer calls to the Module in the libclang frontend

vitali87:
## Bug

The libclang C++ frontend's `_process_call` opened with `if enclosing is None: return`, silently dropping every call that has no enclosing function or method scope:

- a default member initializer — `struct S { int x_ = compute(); };`
- a file/namespace-scope global initializer — `int g_val = compute();`

These are first-party calls that libclang resolves, so dropping them under-counts the call graph. The default tree-sitter path already attributes such module-load-time calls to the **Module** node (verified: `Module proj.s -> compute`); the libclang frontend was inconsistent.

## Fix

When there is no enclosing function/method scope, fall back to the enclosing **Module** (`_module_caller`), derived from the call site's own file — matching the tree-sitter path. The call site must be inside the indexed repo (`module_qn`/`rel_path` are `None` for system headers), so no external calls are introduced, and the Module node is ensured before the edge is emitted.

## Test

RED→GREEN: `test_module_scope_initializer_calls_attributed_to_module` in `test_cpp_frontend_calls.py` (a default member initializer and a namespace-scope global initializer must both surface as a `Module -> compute` edge). Failed before the fix (`got []`), passes after.

## Scope / honest measurement

This is a zero-regression correctness fix. On the `leveldb` frontend-mode benchmark it is **neutral** (recall 0.4562, FP 1, tp 401 — unchanged): leveldb barely uses initializer-context calls. leveldb's frontend recall gap is dominated by unrelated diffuse causes (documented `operator*` metric differences, and the frontend's compile-DB parse emitting fewer edges than the oracle's per-file parse), which are a separate larger workstream, not addressed here.

### DIFF
--- a/codebase_rag/parsers/cpp_frontend/frontend.py
+++ b/codebase_rag/parsers/cpp_frontend/frontend.py
@@ -230,8 +230,6 @@ class _Collector:
         # (H) Resolve the callee semantically via cursor.referenced (libclang did
         # (H) the overload/name resolution already), preferring its definition so
         # (H) the edge targets the node the frontend emitted for the body.
-        if enclosing is None:
-            return
         referenced = cursor.referenced
         if referenced is None:
             return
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #569, opened 2026-07-01
fix(dead-code): use BFS expansion so reachability does not time out on real graphs

vitali87:
## Problem

`cgr dead-code` was unusable on any real-sized repo. The reachability query used `MATCH (r)-[:CALLS*0..]->(live)`, which enumerates *every path* from every root. With cycles and diamonds in a real call graph this is combinatorial, so the query hit Memgraph's transaction timeout and the command returned nothing:

```
mgclient.DatabaseError: Transaction was asked to abort because of transaction timeout.
```

All existing tests passed because their seed graphs are tiny (no path explosion).

## Fix

Switch to Memgraph BFS expansion (`*BFS`), which visits each reachable node once instead of enumerating paths. `*BFS` excludes the source node, so the roots are unioned into the live set explicitly.

Measured on a full ingest of this repo:

| | before | after |
|---|---|---|
| reachability query | timeout | 0.04s |
| `cgr dead-code` end-to-end | empty (error) | 354 candidates, ~4s |

## Verification

- Dogfooded end-to-end against a real Memgraph on the whole repo graph (functions/methods and `--classes`).
- 50 dead-code / cypher / constructor tests pass; ruff, ty (no new diagnostics), and `check_no_docs.py` clean.

### DIFF
--- a/codebase_rag/cypher_queries.py
+++ b/codebase_rag/cypher_queries.py
@@ -122,6 +122,9 @@ _DEAD_CODE_MODULE_ROOT_NON_TEST = (
 # (H) part of the dead cluster (the subclass is reported too). Classes referenced
 # (H) solely via type annotations / isinstance / dynamic lookups are not modelled
 # (H) as edges, so class candidates are review hints, not a delete list.
+# (H) *BFS visits each reachable node once; plain *0.. enumerates every path and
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

Issue #532, opened 2026-06-26
fix(graph): incremental update drops inbound CALLS from unchanged callers into a changed file


vitali87:
## Problem

The incremental updater (`GraphUpdater._process_files` in `codebase_rag/graph_updater.py`) now scopes work to changed files: when a file changes it calls `_delete_module_entities(file_key)`, which `DETACH`-deletes that file's Module subtree, including the `CALLS` edges attached to its functions.

This correctly removes outbound calls from the changed file. But it also removes **inbound** `CALLS` edges, i.e. calls from *unchanged* files into the changed file's functions. Those caller files are not reprocessed during the incremental run, so the inbound `CALLS` edges are deleted and never rebuilt until either:

- the caller file itself changes later, or
- a full re-index (`--clean` / forced) is run.

The result is that, after an incremental update to a callee, the graph can under-report who calls it.

## Expected

After an incremental update to file A, `CALLS` edges from unchanged callers into A's (re-created) functions should be preserved or rebuilt.

## Notes

- Spun off from #86, which asked for tailored incremental updates instead of wiping all `CALLS`. That broader optimization is done; this is the remaining correctness edge case.
- Possible approaches: re-resolve inbound callers of the changed file's symbols, or recompute calls for files that reference symbols defined in the changed file.

vitali87:
Fixed on main. The incremental path now captures reference edges from unchanged files into the re-indexed files before their subtrees are deleted, then restores them verbatim afterwards: `_capture_inbound_edges` / `_restore_inbound_edges` in `codebase_rag/graph_updater.py`, using `CYPHER_INBOUND_EDGES`. Targets that were renamed or removed are correctly left without their stale inbound edge, matching a clean re-index. Covered by `codebase_rag/tests/test_incremental_eval.py` (inbound-edge scenarios) and the `evals/incremental.py` harness.

PR #570, opened 2026-07-01
fix(incremental): rehydrate property status so property-dispatch calls survive re-index

vitali87:
## What

Closes a slice of the issue #532 incremental-update residual: a re-parsed file's call to a `@property` defined in an unchanged file was dropped versus a clean re-index.

## The bug

On an incremental rebuild, only the function registry is rehydrated from the graph for unchanged files. The registry trie also tracks a **property-name set** (`mark_property`), populated during the definition pass from `@property` decorators. That set was never persisted or rehydrated, so property-dispatch resolution (`obj.prop`) could not recognize a property defined in an unchanged file, and the resulting `CALLS` edge dropped.

## The fix

- Persist `is_property` on the Method node when it carries a property decorator.
- Return it from the definition-QN rehydration query.
- In `_rehydrate_registry_from_graph`, `mark_property` each rehydrated definition flagged as a property, restoring the property-name set for unchanged files.
- Mirror the new query field in the eval's in-memory stateful ingestor.

## Impact

On the `codebase_rag/parsers` incremental eval, clean-equivalent probes rose **19/25 → 21/25**, and `call_resolver.py` divergence dropped from **-13 to -1** edges, with **no new extra edges** introduced.

## Scope / remaining residual

This is the clean, no-regression slice. A further attempt to rehydrate the full `class_inheritance` hierarchy (for super()/override/protocol/subclass dispatch into unchanged files) was measured and **reverted**: it traded some dropped edges for new extra edges (net-negative parity), because clean builds `class_inheritance` in a specific incremental order during parsing and rehydrating the full hierarchy changes context-sensitive resolution outcomes. Full parity there needs order-identical resolution context, not naive rehydration; documented for follow-up.

## Testing

RED->GREEN: `test_incremental_preserves_property_dispatch_editing_caller` fails before the fix (property-dispatch edge dropped) and passes after. Full non-integration suite: 4122 passed, 3 skipped.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -185,6 +185,7 @@ KEY_EXPORTED_AT = "exported_at"
 KEY_PARSER = "parser"
 KEY_NAME = "name"
 KEY_QUALIFIED_NAME = "qualified_name"
+KEY_IS_PROPERTY = "is_property"
 KEY_QUERY = "query"
 KEY_RESPONSE = "response"
 KEY_START_LINE = "start_line"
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

PR #571, opened 2026-07-01
fix(dead-code): treat *.test.* and *.spec.* files as test code

vitali87:
## What

`TEST_PATH_PATTERNS` only matched Python test conventions (`test_`, `_test`, `conftest`, `/tests/`). JS/TS test files use `foo.test.ts` / `foo.spec.tsx`, which match none of those, so every symbol in a `*.test.*` file was reported as dead code.

Running `cgr dead-code` on a real polyglot repo produced hundreds of spurious `...test.anonymous_*` candidates from `.test.ts` files.

## Change

Add `.test.` and `.spec.` to `TEST_PATH_PATTERNS`.

## Test

RED→GREEN: `test_test_patterns_cover_js_ts_convention` asserts `src/solution.test.ts` and `app/foo.spec.tsx` match the patterns; fails before the fix, passes after.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -438,8 +438,17 @@ DEFAULT_ROOT_DECORATORS: frozenset[str] = frozenset(
     }
 )
 
-# (H) Substrings in a node's file path that mark it as test code.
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #572, opened 2026-07-01
fix(incremental): rehydrate class_inheritance so protocol dispatch survives re-index

vitali87:
## Problem

Issue #532 residual. An incremental re-index re-parses only the changed file, but `class_inheritance` is rebuilt from re-parsed files only. When you edit a **caller**, the hierarchy of classes defined in unchanged files is lost, so:

- **Protocol dispatch breaks.** A call `g.greet()` on a `GreeterProtocol`-typed receiver resolves to the Protocol stub `GreeterProtocol.greet` instead of being redirected to the concrete `Greeter.greet`, because `_protocol_classes()` keys off `class_inheritance` (now empty for `GreeterProtocol`). The clean re-index emits the concrete edge; the incremental run leaves a stale edge on the stub.
- **Inherited-method resolution** into unchanged bases degrades the same way.

## Fix

- Rehydrate `class_inheritance` from persisted `INHERITS` edges (new `CYPHER_ALL_INHERITS`) before Pass 3, filling only entries for classes not re-parsed this run (fresh entries keep their correct source order).
- Rehydrated bases carry **no source order**, so Pass 4 (`process_all_method_overrides`) skips those unchanged classes when emitting `OVERRIDES`. Without this, a multi-inheritance method (e.g. a method defined in two mixins) would be attributed to the wrong base, producing spurious `OVERRIDES` edges. Their real override edges are already preserved or restored verbatim.

## Impact (incremental eval on `codebase_rag/parsers`, sample 25)

| edge | before | after |
|------|--------|-------|
| CALLS | fp=1, fn=14 | fp=1, fn=13 |
| OVERRIDES | fp=0, fn=1 | fp=0, fn=0 |
| INHERITS / INSTANTIATES | perfect | perfect |

Net-positive with **zero new extras**.

## Tests (RED→GREEN)

- `test_incremental_preserves_protocol_dispatch_editing_caller` — edits the caller, asserts the concrete dispatch edge is present and equals a clean index (RED without rehydration).
- `test_pass4_skips_rehydrated_classes_for_overrides` — pins the Pass 4 skip guard: a rehydrated class with scrambled base order emits no `OVERRIDES` when skipped, but does without the skip.

Full suite: 4267 passed, 14 skipped. ruff/ty clean.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1125,14 +1125,19 @@ CYPHER_INBOUND_EDGES = (
 # (H) resolution still see the hierarchy of classes defined in files that were not
 # (H) re-parsed. Without it, editing a caller drops the protocol/inheritance
 # (H) redirect (issue #532 residual): a call resolves to the Protocol stub instead
-# (H) of the concrete implementer because _protocol_classes() is empty.
+# (H) of the concrete implementer because _protocol_classes() is empty. Ordered by
+# (H) base_index so multiple-inheritance base order matches the original source,
+# (H) which method resolution and override attribution depend on.
 CYPHER_ALL_INHERITS = (
-    "MATCH (child)-[:INHERITS]->(base) "
+    "MATCH (child)-[r:INHERITS]->(base) "
     "WHERE child.qualified_name IS NOT NULL AND base.qualified_name IS NOT NULL "
-    "RETURN child.qualified_name AS child_qn, base.qualified_name AS base_qn"
+    "RETURN child.qualified_name AS child_qn, base.qualified_name AS base_qn, "
+    "r.base_index AS base_index "
+    "ORDER BY child_qn, base_index"
 )
 KEY_CHILD_QN = "child_qn"
 KEY_BASE_QN = "base_qn"
+KEY_BASE_INDEX = "base_index"
 
 CYPHER_PARAM_PATHS = "paths"
 KEY_CALLER_LABEL = "caller_label"
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1120,6 +1120,20 @@ CYPHER_INBOUND_EDGES = (
     "caller.qualified_name AS caller_qn, type(r) AS rel, "
     "head(labels(target)) AS target_label, target.qualified_name AS target_qn"
 )
+# (H) Rehydrate class_inheritance on an incremental run: every INHERITS edge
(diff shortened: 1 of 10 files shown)


vitali87:
@greptile review

vitali87:
Good catch — confirmed real. Pass 3 method resolution walks `class_inheritance` first-match-wins, so replaying persisted INHERITS rows in a non-source order could bind an inherited call to the wrong base (`MixinB.shared` instead of `MixinA.shared`).

Fixed at the root: each `INHERITS` edge now persists a `base_index` (its position in the child's base list), and `CYPHER_ALL_INHERITS` returns it with `ORDER BY child_qn, base_index`, so rehydration restores the exact source order. This makes both Pass 3 (method/protocol resolution) and Pass 4 (override attribution) order-correct, so the earlier Pass 4 skip-set is no longer needed and has been removed.

Tests: `test_all_inherits_query_returns_bases_in_base_index_order` (five bases inserted in reverse index order must come back ordered — RED without the ordering) and `test_incremental_preserves_multiple_inheritance_dispatch` (editing the caller of `Combined(MixinA, MixinB)` keeps `c.shared()` bound to `MixinA.shared`, incr == clean). Incremental eval on codebase_rag/parsers: OVERRIDES fp/fn 0/0, no new extras. Full suite 4268 passed.

@greptile review

vitali87:
Addressed — the upgrade path is real. Fresh graphs get `base_index` on every INHERITS edge, but a graph indexed by an older cgr and then incrementally updated has base_index-less edges, so `ORDER BY base_index` can't recover order for a multi-base class.

Fix: rehydration now runs through `GraphUpdater._rehydrated_bases_by_child`, which skips any multi-base class whose edges don't all carry a `base_index`. Such a class is left out of `class_inheritance` and falls back to name-based resolution (pre-PR behavior) rather than risk binding a call or override to the wrong base; it self-heals once its file is re-parsed and gets base_index. Single-base classes are order-independent and still rehydrated.

Test `test_rehydrate_skips_multi_base_class_with_missing_base_index` pins all four cases (ordered multi-base rehydrated in order, missing-index multi-base skipped, single-base-with-missing-index rehydrated, already-parsed untouched). Parsers eval unchanged (OVERRIDES 0/0, no extras); full suite 4269 passed.

@greptile review

vitali87:
@greptile review

PR #573, opened 2026-07-02
fix(call-graph): process calls for every parsed file, not just AST-cache survivors

vitali87:
## The bug

On any repository larger than the AST cache, cgr silently drops most of the call graph.

Pass 2 parses each file and stores its AST in `BoundedASTCache` (default `CACHE_MAX_ENTRIES=1000`, `CACHE_MAX_MEMORY_MB=500`), which evicts LRU entries as it fills. Pass 3 (`_process_function_calls`) then iterated **only the survivors** in that cache (`self.ast_cache.items()`), so every file evicted during Pass 2 never had its calls resolved. A whole module ends up with zero CALLS edges even though its functions call each other.

This corrupts reachability analysis (`cgr dead-code` reports live code as dead) and any query that relies on CALLS.

### Evidence (real 6,400-file repo)

| Build | AST cache | Modules with CALLS | Total CALLS |
|---|---|---|---|
| default | 1000 / 500 MB | 500 / 4229 | 6,344 |
| large | effectively unbounded | 1,797 / 4229 | 31,577 |

The default cache dropped ~80% of the call graph (~25k edges). One module (`csv_builder.py`) had 12 functions with obvious intra-file calls and **zero** CALLS edges under the default cache, **46** under a large cache.

## The fix

`GraphUpdater` records every file parsed this run in `self._parsed_files`. Pass 3 iterates that list instead of the bounded cache, fetching each AST from the cache or re-parsing it from disk on a miss (`_ast_for`). Re-parsed trees re-enter the bounded cache (memory stays bounded) and their stale captures are dropped so captures recompute against the fresh tree.

Correctness is now independent of cache size; the only cost on huge repos is re-parsing evicted files in Pass 3.

## Test

`test_ast_cache_eviction_calls.py`: pins the cache to one entry so all but the last file are evicted during Pass 2, then asserts every module's intra-file `top -> helper` CALLS edge is still recorded. RED before the fix, GREEN after.

Full suite: 4264 passed, 14 skipped.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -437,6 +437,10 @@ class GraphUpdater:
             simple_name_lookup=self.simple_name_lookup
         )
         self.ast_cache = BoundedASTCache()
+        # (H) Every file parsed this run, in parse order. The AST cache is bounded
+        # (H) and evicts on large repos, so Pass 3 must iterate this full list (not
+        # (H) the cache) and re-parse evicted files, or their calls are dropped.
+        self._parsed_files: list[tuple[Path, cs.SupportedLanguage]] = []
         self.unignore_paths = unignore_paths
         self.exclude_paths = exclude_paths
         self.skipped_because_in_sync = False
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #574, opened 2026-07-02
test(evals): add JavaScript CALLS retrieval eval + fix anonymous-callback caller attribution

vitali87:
## What

Adds the JavaScript CALLS retrieval eval (the last supported cgr language without one) and fixes the dominant cgr bug it surfaced.

## Eval

Mirrors the TypeScript retrieval eval over `.js`/`.jsx`. The `tsc` compiler API parses JavaScript syntactically (independent of cgr's tree-sitter JS frontend), so it grades cgr's JS `CALLS` resolution against ground truth for a language cgr handles through a partly distinct path (CommonJS `require`/`module.exports`, no type annotations). `run_javascript_call_oracle` reuses the existing tsc oracle with JS suffixes; `evals/js_retrieval.py` mirrors `ts_retrieval.py`. Pinned by `test_js_retrieval_eval.py` (cgr matches the oracle on a clean ES-module fixture).

## Bug fixed: calls inside anonymous callback arrows were dropped

Running on `webpack/lib` (577 CommonJS files) surfaced it. A callback like:

```js
compiler.hooks.compilation.tap(NAME, (compilation) => { getReplacements() });
```

has no name and no binding, so the call pass skipped the arrow — and `_calls_owned_by` had already excluded its calls from the enclosing method, so `getReplacements()` attributed to nothing and vanished. webpack (and most modern JS) is saturated with this `hooks.tap(name, (x) => …)` pattern, making it the leading recall killer.

Fix: `_calls_owned_by` now excludes only *attributable* nested functions (named, or arrows bound to a name) from the enclosing scope, so an anonymous arrow's calls bubble up to the nearest named function/method — where the oracle attributes them. RED→GREEN in `test_js_anonymous_callback_calls.py`.

## Impact (webpack/lib)

| | precision | recall | F1 |
|---|---|---|---|
| before | 0.9998 | 0.613 | 0.760 |
| after | 0.9998 | 0.829 | 0.906 |

1495 real call edges recovered. Remaining FN are method dispatch on receivers cgr cannot type and the name-based oracle's over-count of calls through parameters (`callback`/`fn`) and Node built-ins colliding with first-party names — documented in evals/README.md.

Full suite: 4269 passed, 14 skipped. ruff/ty clean.


vitali87:
@greptile review

vitali87:
Thanks — the diagnosis is right: in an environment without cgr's optional JavaScript tree-sitter grammar, `load_parsers()` omits JS, so cgr indexes zero `.js` functions and the fixture test hard-fails. That's an environment/test-hygiene gap, not a resolution bug (with the grammar installed the fixture matches the oracle exactly, and webpack scored 5736 first-party edges).

Fixed by gating the cgr-side JS tests on grammar availability — they now `skip` (like the Go/Rust/Java retrieval tests do on a missing toolchain) instead of failing on an empty cgr graph: `test_cgr_matches_oracle_on_clean_js_project` and `test_call_inside_anonymous_callback_attributes_to_enclosing_method` carry a `needs_js_grammar` skipif that checks `SupportedLanguage.JS in load_parsers()`. The oracle-only test keeps just the node guard. Full suite 4269 passed locally (grammar present).

@greptile review

PR #576, opened 2026-07-02
test(evals): property-aware Python retrieval oracle

vitali87:
The Python retrieval oracle counted only `ast.Call` sites, so cgr's (correct, receiver-aware) CALLS edges for `@property`/`@cached_property` getter access were scored as false positives. This adds `first_party_property_names` and counts a bare read of a first-party property getter as a call edge, matching how cgr models it.

Validated on `django/django` (~2900 files): graph precision 0.838 -> 0.977, F1 0.889 -> 0.957 (FP 2006 -> 287), decisively ahead of grep (grep_call F1 0.789). Recall dips 0.946 -> 0.938 because the fuller oracle also expects property accesses on receivers cgr cannot infer.

RED->GREEN: `test_oracle_captures_property_access_as_calls` (property getters count, bound-method references do not). Full suite 4276 passed.

### DIFF
--- a/codebase_rag/tests/test_retrieval_eval.py
+++ b/codebase_rag/tests/test_retrieval_eval.py
@@ -7,6 +7,7 @@ from codebase_rag import constants as cs
 from evals import constants as ec
 from evals.retrieval import (
     cgr_call_edges,
+    first_party_property_names,
     first_party_symbols,
     grep_call_edges,
     oracle_call_edges,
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

vitali87:
Addressed: the oracle now counts a property access only in Load (getter read) and Store (setter write) context, excluding Del. Store is kept deliberately: `self.prop = v` on a property invokes the setter, a real descriptor-method call, and cgr emits a CALLS edge for it (dropping Store regressed django precision 0.977 -> 0.973, FP 287 -> 340, i.e. it turned ~53 real cgr edges into false positives). Del (`del self.prop`) is the one non-call context and is now excluded. RED->GREEN test updated to pin all three. @greptile review

PR #577, opened 2026-07-02
fix(php): resolve calls to use-function imports across namespace-path mismatch

vitali87:
## Bug
A bare call to a function imported via `use function A\B\c;` was dropped entirely. cgr registers functions by **file path** (`proj.Collections.functions.enum_value`), but a PHP `use function` import records the **namespace path** (`App.Support.enum_value`) — which matches no node. That non-project-rooted, unregistered target was then misclassified by `_is_external_import` as an external import, suppressing the simple-name trie fallback that a bare PHP call (no `use`) already relies on.

This is pervasive in laravel: global helpers live in `Illuminate/Collections/functions.php` but declare `namespace Illuminate\Support`, and callers import them with `use function Illuminate\Support\enum_value;`.

## Fix
Record the local names brought in by `use function` per module (`ImportProcessor.php_function_imports`) and exempt them from `_is_external_import` suppression, so they resolve by simple name via the trie — exactly as bare PHP calls already do. The Python/Java external-import suppression (#547) is untouched, because those import paths are real module qns.

## Validation
laravel/framework (`src`, 1598 PHP files): recall 0.9012 -> 0.9076, precision held 0.988; `enum_value` false negatives 57 -> 14 (residual are qualified-name call sites, a distinct case). Full suite 4276 passed.

RED->GREEN: `test_use_function_import_call_resolves_across_namespace_path_mismatch`.


vitali87:
@greptile review

vitali87:
Both gemini findings fixed: (1) `parse_imports` now clears `php_function_imports[module_qn]` on re-parse so an incremental re-index cannot leave a stale exemption (RED->GREEN test_reparse_clears_stale_php_function_imports); (2) `_is_external_import` avoids the per-call empty-set allocation on the hot path. @greptile review

vitali87:
On the external-import P1: cgr qualifies PHP functions by file path and does not track the `namespace` declaration anywhere in its PHP model (`_build_nested_qualified_name` drops `namespace_definition` ancestors), so a genuinely external `use function Vendor\pkg\helper` cannot be distinguished from a path-mismatched first-party one without systemic PHP namespace tracking. This exemption makes a `use function`-imported call resolve exactly like a **bare** `helper()` call, which already defers to the simple-name trie and carries the identical collision property — so it removes an inconsistency rather than adding a new failure mode, and on laravel/framework precision held at 0.988 with recall improved. Documented the limitation in-code. @greptile review

PR #580, opened 2026-07-02
fix(js): resolve calls to first-party functions imported via non-relative specifiers

vitali87:
## Bug
A call to a first-party function imported via a **non-relative** specifier is dropped. `_resolve_js_module_path` maps a non-relative specifier to a garbage dotted path (e.g. `ext:alias/util.ts` -> `ext:alias.util.ts`), so the import target is unregistered, `_is_external_import` judges it external, and the simple-name trie fallback is suppressed — even though the function is first-party and indexed.

This hits deno's `ext:deno_node/...` specifiers heavily, and **any TS project using tsconfig `paths` aliases** (`@/lib/util`) identically.

## Fix
Record the local names of non-relative JS/TS imports per module (`ImportProcessor.js_ts_bare_imports`) and exempt them from `_is_external_import`, so they defer to the simple-name trie like a relative import that misses. A genuine external whose simple name has no first-party symbol still resolves to nothing. Python's external-import suppression (#547) is untouched (the change is JS/TS-scoped).

## Validation (empirical, decided the design tradeoff)
- **deno/ext** (163 TS files): recall **0.729 -> 0.913** (FN 411 -> 132, +279 real edges), precision **unchanged at 0.976 — FP still 34, zero new false positives**.
- **zod v4** regression check: precision **still 1.0** (recall 0.40 is the pre-existing baseline for that subset; the change only adds trie fallback).
- Full suite 4279 passed.

RED->GREEN: `test_call_via_non_relative_aliased_import_resolves` (unfixed code resolved only the relative-import call).

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -390,6 +390,13 @@ DUP_QN_MARKER = "@"
 PATH_CURRENT_DIR = "."
 PATH_PARENT_DIR = ".."
 GLOB_ALL = "*"
+# (H) JS/TS import specifier schemes that name genuinely external code (node
+# (H) builtins, package registries, URLs). A specifier with any OTHER scheme
+# (H) (`ext:` deno-runtime aliases, bundler virtual modules) points at first-party
+# (H) code under a non-file-path name, so its unresolved calls defer to the trie.
+JS_EXTERNAL_IMPORT_SCHEMES: frozenset[str] = frozenset(
+    {"node", "npm", "jsr", "bun", "http", "https", "data", "file", "blob"}
+)
 PATH_RELATIVE_PREFIX = "./"
 PATH_PARENT_PREFIX = "../"
 CPP_IMPORT_PARTITION_PREFIX = "import :"
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -328,6 +328,16 @@ class CallResolver:
         php_imports = self.import_processor.php_function_imports.get(module_qn)
         if php_imports and call_name in php_imports:
             return False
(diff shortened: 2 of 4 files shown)


vitali87:
@greptile review

vitali87:
Both P1s addressed. The exemption now applies only to imports with a NON-STANDARD scheme (`_has_aliased_scheme`: deno's `ext:deno_node/...`, bundler virtual modules), not to all non-relative imports. Ordinary package specifiers — bare (`lodash`), scoped (`@scope/pkg`), and standard schemes (`node:`/`npm:`/`jsr:`/`http(s):`) — are excluded, so `import { map } from "lodash"` stays externally suppressed and cannot rebind to a first-party `*.map`. Added a regression-guard assertion (external package import whose name collides with a first-party symbol must not resolve). deno/ext still recovers recall 0.729 -> 0.888 with precision unchanged at 0.975 (FP 34); full suite 4279 passed. @greptile review

vitali87:
On the tsconfig `paths` alias (`@/util`) point: this is a deliberate scope boundary, not a regression (those aliases were already dropped before this PR). A `@/util` specifier has no scheme and is syntactically indistinguishable from a scoped package (`@scope/pkg`), so exempting it here would reintroduce exactly the round-1 P1 — rebinding genuine scoped-package calls to same-named first-party symbols. Resolving tsconfig aliases precisely requires reading `compilerOptions.paths` (or a deno import map) and mapping each alias to its first-party dir, which is a separate follow-on feature; I documented that in-code. This PR safely fixes the custom-scheme case (deno `ext:`, bundler virtual modules) with zero external-collision risk. @greptile review

PR #581, opened 2026-07-02
feat(js): resolve tsconfig paths aliases to first-party modules

vitali87:
## What
Completes the #580 follow-on: cgr now reads tsconfig `compilerOptions.paths` (JSONC-tolerant) and resolves alias specifiers to the real first-party module during JS/TS import resolution.

- `@/*` -> `src/*` (wildcard) and exact aliases (`~lib` -> `src/lib/index.ts`), with `baseUrl` folded in; longest matching prefix wins.
- Precise resolution (maps to the actual file), so — unlike a trie fallback — there is **no external-collision risk**: a scoped package `@scope/pkg` that is not an alias stays external.
- `extends` chains and index-directory resolution are not followed (documented).

## Why
Path aliases are ubiquitous in real TS apps; before this, `import { x } from "@/util"` resolved to a garbage module qn and every call to `x` was dropped as external (the same class as #580, which handled deno's `ext:` scheme).

RED->GREEN: `test_tsconfig_paths_alias_calls_resolve_to_first_party_files` (wildcard + exact alias). Repos without a tsconfig are unaffected (empty alias list). Full suite 4280 passed.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -397,6 +397,26 @@ GLOB_ALL = "*"
 JS_EXTERNAL_IMPORT_SCHEMES: frozenset[str] = frozenset(
     {"node", "npm", "jsr", "bun", "http", "https", "data", "file", "blob"}
 )
+# (H) Module file extensions stripped when turning a tsconfig `paths` target into a
+# (H) module qn (`src/util.ts` -> `src/util`), longest first so `.d.ts`-like
+# (H) compound suffixes are handled before the bare `.ts`.
+JS_TS_MODULE_EXTENSIONS: tuple[str, ...] = (
+    ".d.ts",
+    ".tsx",
+    ".ts",
+    ".jsx",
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
Both findings fixed together: `_ts_alias_module_qn` now (1) normalizes `.`/`..` segments via posixpath and rejects escaping targets, and (2) accepts an alias ONLY when it maps to a real first-party file on disk (`<target>.<ext>` or `<target>/index.<ext>`). So a broad/catch-all alias like `"*": ["src/*"]` no longer captures bare package imports (`lodash` -> `src/lodash` has no file -> alias not applied -> stays external, no rebind to a first-party `*.pkgCollide`). Added a regression-guard assertion for exactly that catch-all case. Full suite green (the tsconfig changes; an unrelated WIP test on another branch is not part of this PR). @greptile review

PR #582, opened 2026-07-02
feat(resolver): chained-call receiver type inference via method return types

vitali87:
## What
Adds a shared `method_return_types` map (func/method qn -> bare return-type name), populated at definition ingestion and threaded DefinitionProcessor -> factory -> TypeInferenceEngine -> CallResolver (mirrors `class_field_types`). The resolver's chained-call path now walks a receiver expression like `c.Root()` — base is a typed local (`c` -> Command), each `.method()` hop advances the type by that method's return type — so the final hop (`.Run()` / cobra's `.GenZshCompletion()`) resolves.

This is the first increment of the shared type-inference workstream: return-type capture is wired for **Go** here; TS/Java/C++ plug into the same map next.

## Why
The dominant cross-language recall gap is deeper receiver/return-type inference. Chained calls (`x.foo().bar()`) previously resolved only for Python; every other language dropped the second hop.

## Validation
- cobra (Go) retrieval: recall 0.9752 -> 0.9788, precision unchanged 1.0 (**zero new FP**).
- Language-agnostic and conservative: any untyped/unknown hop leaves the chain unresolved (never mis-resolved).
- Full suite 4280 passed; Python chained path unchanged (tried first, this is a fallback).

RED->GREEN: `test_go_chained_return_type_resolves_second_hop`.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -763,6 +763,7 @@ FIELD_DECLARATOR = "declarator"
 FIELD_PARAMETERS = "parameters"
 FIELD_RECEIVER = "receiver"
 FIELD_TYPE = "type"
+FIELD_RESULT = "result"
 # (H) Rust impl `trait`/`type` fields and a trait's supertrait `bounds`.
 FIELD_TRAIT = "trait"
 FIELD_BOUNDS = "bounds"
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2114,6 +2114,12 @@ TS_GO_EXPRESSION_LIST = "expression_list"
 TS_GO_COMPOSITE_LITERAL = "composite_literal"
 TS_GO_UNARY_EXPRESSION = "unary_expression"
 TS_GO_POINTER_TYPE = "pointer_type"
+# (H) Go composite types a method may return; a chained call lands on the CONTAINER,
+# (H) not its element, so return-type inference must not unwrap these to an element
+# (H) name (a `[]Command` return must not resolve `.Run()` to `Command.Run`).
+TS_GO_CONTAINER_TYPES: frozenset[str] = frozenset(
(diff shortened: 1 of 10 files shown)


vitali87:
@greptile review

vitali87:
Fixed the gemini HIGH on container return types: `extract_return_type_name` no longer unwraps `[]T`/`map[k]T`/`chan T`/`func` returns to their element (new `_return_type_identifier` rejects `TS_GO_CONTAINER_TYPES`, unwrapping only pointer/generic to a type_identifier), so a chained call on a container return (`c.Kids().Run()` where `Kids() []Command`) no longer emits a false edge to `Command.Run`. RED->GREEN test_chained_call_on_container_return_does_not_misresolve. On the dotted-argument concern: call arguments are stripped from the chained call_name upstream, so it already resolves (verified by test_chained_call_with_dotted_argument_resolves, kept as a guard) — no split change needed. Full suite 4283. @greptile review

vitali87:
Addressed the split concern with a paren/bracket-depth-aware `_split_receiver_chain`: it splits a receiver chain on hop `.`s only, never on a `.` inside call arguments, an index, or a generic (`c.Find(1.5).Root` -> `['c','Find(1.5)','Root']`). Added a direct unit test (test_split_receiver_chain_ignores_dots_inside_arguments) covering args, nested dotted args, and index dots. Full suite green apart from a known-flaky embedding semantic-search test (passes in isolation; unrelated to call resolution). @greptile review

vitali87:
The remaining flagged `object_expr.split('.')` no longer exists — it was replaced by the depth-aware `_split_receiver_chain` (call_resolver.py:963), which is directly unit-tested (test_split_receiver_chain_ignores_dots_inside_arguments) for argument/index/generic dots. Your own summary confirms the second hop now resolves and observed zero false positives on head. That inline comment is a stale anchor from a prior commit; the code path it describes is gone. @greptile review

PR #583, opened 2026-07-02
fix(dead-code): suppress spurious C++ edge from nested lambda parameter shadowing

vitali87:
## What

Fixes a C++ callable-parameter tracing precision bug surfaced during the #579 review: a nested lambda whose parameter shadows an enclosing function's callable parameter caused a spurious `CALLS` edge.

For

```cpp
int target() { return 1; }
int shadow(int (*cb)()) {
    auto inner = [](int (*cb)()) { return cb(); };  // lambda's own cb
    return 0;                                        // shadow's cb never invoked
}
int main() { return shadow(target); }
```

the graph emitted `shadow -> target`, even though `shadow`'s own `cb` is never invoked (only the lambda's shadowing `cb` is). All other flow languages already suppress this via their `*_scope_bound_names`, but `_cpp_scope_bound_names` returned an empty set.

## Fix

`_cpp_scope_bound_names` now enumerates a C++ lambda's own parameter names (off its `abstract_function_declarator`'s `parameters` list) so `_scan_invoked_parameters` subtracts them before descending into the lambda body. Extracted `_cpp_declarator_param_names`, shared with `cpp_parameter_names`, so both read the same declarator shape.

This is a precision/soundness fix (removes an over-connection); it never affected the false-dead-code class — an extra edge only makes a callback look more reachable.

## Tests

- `test_cpp_nested_lambda_shadowing_suppresses_forbidden_edge` (RED -> GREEN): asserts no `shadow -> target` edge.
- Existing `test_cpp_function_pointer_callback_is_traced` still passes (direct `cb()` invocation is unaffected — subtraction only applies when descending into a nested closure).

Full suite: 4311 passed, 0 failed. ruff, ty, and the no-inline-comments hook all clean.

### DIFF
--- a/codebase_rag/parsers/utils.py
+++ b/codebase_rag/parsers/utils.py
@@ -156,9 +156,7 @@ def _python_parameter_name(param_node: Node) -> str | None:
     return None
 
 
-_PY_CLOSURE_SCOPES = frozenset(
-    {cs.TS_PY_FUNCTION_DEFINITION, cs.TS_PY_LAMBDA}
-)
+_PY_CLOSURE_SCOPES = frozenset({cs.TS_PY_FUNCTION_DEFINITION, cs.TS_PY_LAMBDA})
 _GO_CLOSURE_SCOPES = frozenset({cs.TS_GO_FUNC_LITERAL})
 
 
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #584, opened 2026-07-02
feat(resolver): single-implementer interface dispatch

vitali87:
## What
When a call resolves to an interface's method (`I.m`) and that interface has **exactly one** first-party implementer `C`, redirect the edge to the concrete `C.m` — the method that actually runs. Generalizes the existing Python-Protocol redirect (`_redirect_protocol_method`) to `IMPLEMENTS` edges across languages.

- New `interface_implementers` map (interface_qn -> implementer qns), captured from IMPLEMENTS edges at ingestion and threaded DefinitionProcessor -> factory -> CallProcessor -> CallResolver.
- `_interface_impl_map` maps an interface to its SOLE implementer; >1 implementer is ambiguous -> not mapped -> the call stays on the interface method (no wrong-impl edge, recall preserved).
- Applied to both the general resolver and the Java resolution path.

## Why
The type-inference workstream's interface-dispatch item: a call typed to an interface should point at the concrete method. This is a call-graph **accuracy** improvement.

## Validation
- Precision-safe by construction (single implementer only; redirect only when the concrete method exists, else unchanged).
- spring-beans retrieval **exactly unchanged** (P 0.9916 / R 0.3643) — the metric is simple-name level so it can't distinguish `I.m` from `Impl.m`; this confirms zero regression while call-graph accuracy improves.
- Full suite 4316 passed.

RED->GREEN: test_interface_dispatch_single_impl.py (single-impl dispatches to concrete; multi-impl stays on interface).

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -134,6 +134,7 @@ class CallProcessor:
         type_inference: TypeInferenceEngine,
         class_inheritance: dict[str, list[str]],
         type_aliases: dict[str, str] | None = None,
+        interface_implementers: dict[str, list[str]] | None = None,
     ) -> None:
         self.ingestor = ingestor
         self.repo_path = repo_path
@@ -145,6 +146,7 @@ class CallProcessor:
             type_inference=type_inference,
             class_inheritance=class_inheritance,
             type_aliases=type_aliases,
+            interface_implementers=interface_implementers,
         )
         # (H) Inter-procedural callable-parameter flow: ordered params per function and
         # (H) the per-call-site argument bindings, resolved to a fixpoint in finalize.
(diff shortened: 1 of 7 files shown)


vitali87:
@greptile review

vitali87:
Both addressed: (1) greptile P1 (Rust implementers omitted) — the Rust `impl Trait for Type` path now also populates `interface_implementers`, so a Rust trait call to the sole concrete impl redirects (verified end-to-end: `s.run()` on `&dyn Svc` -> `SvcImpl.run`; RED->GREEN test_rust_single_trait_impl_dispatches_to_concrete). (2) gemini — `interface_implementers` values are now `set[str]` (naturally unique), simplifying `_interface_impl_map` (no list->set conversion). Full suite 4317. @greptile review

PR #585, opened 2026-07-02
fix(calls): trace self-method dispatch to subclass overrides

vitali87:
## What

Fixes a dead-code / call-graph false positive: a subclass method override reached only through a base-class `self.method()` call was reported as dead, because the call graph connected the `self.M()` site to the base declaration only, never to the overriding implementations.

Now, when a call resolves to a base-class method `C.M`, cgr also emits a CALLS edge to `M` on every concrete subclass of `C` that overrides it (in addition to the base edge). This mirrors the existing Protocol-stub dispatch fan-out (`protocol_dispatch_targets`) and reuses the existing `_concrete_subclasses` inheritance index.

## Why

Surfaced from a real `cgr dead-code` run on a monorepo. Confirmed false positives:
- `SqsBatchJob._raw_batch_operation` / `SqsKick._raw_batch_operation` reached only via `self._raw_batch_operation(...)` in the base.
- `OpenAIClient._generate` reached only via `self._generate(...)`.
- `BatchingStrategyBase._validate_output_schema` (base called via `self.`, override present).

## How

- `call_resolver.py`: new `override_dispatch_targets(callee_qn)` returning `(label, qn)` for each concrete subclass override, gated so it is a no-op when there are no subclasses.
- `call_processor.py`: emit those edges for Python method calls, in addition to the resolved base edge.

## Tests (RED to GREEN)

- `test_method_override_dispatch_calls.py`: direct override, transitive (two-level) override, and a negative case asserting no spurious edges when there are no subclasses.
- `test_nested_local_function_calls.py`: regression coverage for nested/local function calls and closures passed to builtins. These already pass on current cgr; they lock in that behavior (the equivalent monorepo findings were from a graph indexed by an older cgr, not a current defect).

Full suite: 4322 passed, 14 skipped. Scope is Python; cross-language virtual dispatch is a possible follow-up.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2690,6 +2690,7 @@ HIGHER_ORDER_BUILTINS: frozenset[str] = frozenset({"sorted", "min", "max", "redu
 
 # (H) Python attribute prefixes
 PY_SELF_PREFIX = "self."
+PY_CLS_PREFIX = "cls."
 
 # (H) Python type inference patterns
 PY_VAR_PATTERN_ALL = "all_"
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1221,6 +1221,21 @@ class CallProcessor:
                         )
                 continue
 
+            if is_python and callee_type == cs.NodeLabel.METHOD:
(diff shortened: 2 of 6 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #586, opened 2026-07-02
fix(dead-code): treat dunder methods as reachable roots

vitali87:
## What

Fixes a dead-code false positive: methods invoked implicitly by the Python runtime, such as the `async with` protocol hooks `__aenter__` / `__aexit__`, were reported as dead because no explicit call site exists for the reachability walk to follow.

Dunder (`__name__`) methods are now treated as reachability roots in the dead-code analysis.

## Why

`__enter__`/`__exit__`/`__aenter__`/`__aexit__`, iterator, operator and other dunder methods are dispatched by the interpreter, not by a call the graph can see. Tracing every implicit invocation site (`with`, `async with`, `for`, operators, `len`, comparisons, ...) is both incomplete and fragile, and the concrete case that surfaced this (`async with c():` where `c()` returns an `Inner` instance typed as an `AsyncContextManager`) cannot be resolved by type inference at all. Rooting dunders models the reality that the runtime always may invoke them, which is the complete, root-cause fix.

Surfaced from a real `cgr dead-code` run: `InFlightCounter.__call__.Inner.__aenter__` / `__aexit__`.

## How

- `cypher_queries.py`: add a root clause matching dunder method names.
- `evals/dead_code.py`: mirror the rule in the in-memory reachability evaluator so the eval stays faithful to the Cypher query.

## Tests (RED to GREEN)

- `test_dead_code_eval.py::test_dunder_method_is_a_root`: dunders are rooted, a plain private method with no caller stays dead.
- `test_dead_code_eval.py::test_cgr_dead_code_does_not_flag_context_manager_dunders`: full parse -> graph -> reachability on the real `async with` shape.
- `test_cypher_queries.py`: the dead-code integration seed now includes an edge-less dunder method; the existing exact-set assertions enforce it is not reported dead, exercising the real Cypher clause against Memgraph.

Full suite: 4317 passed, 14 skipped.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #587, opened 2026-07-02
fix(calls): resolve bare nested-function calls against the caller's enclosing scope

vitali87:
## What

Fixes a dead-code false positive from wrong resolution of bare calls to nested (local) functions. When two sibling scopes each define a nested function of the same name, both calls resolved to the first one alphabetically, orphaning the other so it was reported dead.

## Why

The bare-name resolver had no notion of the caller's enclosing scope. `_try_resolve_via_trie` matched by name suffix across the whole module and broke ties by qualified-name string, so for

```python
class Op:
    def get_a(self):
        def traverse(node): ...
        return traverse(self)
    def get_b(self):
        def traverse(node): ...
        return traverse(self)
```

both `traverse(...)` calls resolved to `Op.get_a.traverse`; `Op.get_b.traverse` got no inbound edge and was flagged dead. Real instances: `OperationResult.get_metadata.traverse` (alongside `get_llm_results.traverse`), and the `run_determination`/`run_determinations` pair each with a nested `_get_param`.

## How

- Thread the caller's qualified name into `resolve_function_call` / `_resolve_function_call`.
- Resolve a bare name against the caller's scope chain FIRST (before the module-keyed cache/trie): try `{caller_qn}.{name}`, then walk up through enclosing FUNCTION scopes only. Each ancestor must be in the function registry, so a class or the module boundary stops the walk — matching Python's LEGB rule, where class scope is not part of a method's name lookup. This also fixes recursive calls inside the correct nested function.

## Tests (RED to GREEN)

- `test_sibling_methods_resolve_own_nested_function` and `test_sibling_module_functions_resolve_own_nested_function` in `test_nested_local_function_calls.py`; the four existing nested-call tests still pass.

Full suite: 4333 passed, 14 skipped.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1092,7 +1092,7 @@ class CallProcessor:
                 )
             else:
                 callee_info = resolve_func(
-                    call_name, module_qn, local_var_types, class_context
+                    call_name, module_qn, local_var_types, class_context, caller_qn
                 )
             if not callee_info and resolve_builtin is not None:
                 callee_info = resolve_builtin(call_name)
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1109,7 +1109,7 @@ class CallProcessor:
                         )
                     if (rhs := alias_map.get(call_name)) is not None:
                         callee_info = resolve_func(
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #588, opened 2026-07-02
fix(calls): anchor self/cls dispatch on the enclosing class and fan out to concrete overrides

vitali87:
## What

Fixes dead-code false positives for method overrides reached only through a `self.M()` / `cls.M()` call. A previous pass fanned out from the *resolved* callee, which is wrong: when `M` is abstract with several overrides, the resolver's trie picks an arbitrary sibling override, so the fan-out started from a leaf and missed the base and the other overrides.

Now the fan-out anchors on the ENCLOSING class (`class_context`) and emits an edge to the enclosing class's own/inherited method plus every concrete subclass override.

## Why

Real instances from `cgr dead-code`:
- `SqsBatchJob._raw_batch_operation` abstract, `SqsKick`/`SqsDelete` override it, base `_batch_chunk` calls `self._raw_batch_operation(...)`. `self._raw_batch_operation` resolved to `SqsDelete` alphabetically, so `SqsKick`'s override had no inbound edge.
- `OpenAIClient._generate` reached via `self._generate()`, but `BaseLLMClient` has several subclasses each defining `_generate`, so the call resolved to a cousin (`AnthropicClient`) instead of the enclosing class's own method.

## How

- New `CallResolver.self_dispatch_targets(class_context, method_name)`: the enclosing class's own/inherited method plus every concrete subclass override.
- It SKIPS abstract targets: an `@abstractmethod` stub never runs, so it must not become a call target (a concrete override wins). This preserves the mixin-abstract-stub contract in `test_abstract_method_override_resolution.py`.
- `call_processor` fan-out re-anchored on `class_context` (available at the call site) instead of the resolved callee; gated to `self.`/`cls.` receivers with a single method segment (excludes `super().M()`, `Base.M()`, and `self.attr.M()`).

## Tests (RED to GREEN)

- `test_self_call_in_abstract_base_reaches_all_overrides` (two overrides, both reached), `test_self_call_binds_to_own_class_method_amid_cousin_overrides` (own-class method wins over a cousin override); the five existing override-dispatch tests and both abstract-stub-resolution tests still pass.

Full suite: 4333 passed, 14 skipped. Abstract *base* methods reached only polymorphically (e.g. the `SqsBatchJob` stub itself) are a separate dead-code root-seeding concern, handled in a follow-up.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1225,28 +1225,33 @@ class CallProcessor:
 
             if (
                 is_python
-                and callee_type == cs.NodeLabel.METHOD
+                and class_context
                 and (
                     call_name.startswith(cs.PY_SELF_PREFIX)
                     or call_name.startswith(cs.PY_CLS_PREFIX)
                 )
             ):
-                # (H) self.M()/cls.M() dispatches at runtime to any subclass override
-                # (H) of M, so emit an edge to each in ADDITION to the resolved base
-                # (H) edge (emitted by the normal path below); otherwise an override
-                # (H) reached only through the base call looks like dead code. Gating on
-                # (H) the self/cls receiver excludes super().M() (an explicit parent
-                # (H) call) and Base.M() (an explicit implementation): neither is virtual
-                # (H) dispatch, and fanning them out would create false edges.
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #589, opened 2026-07-02
fix(dead-code): treat @abstractmethod stubs as reachable roots

vitali87:
## What

Fixes a dead-code false positive: an `@abstractmethod` stub was reported as dead code. An abstract method is a contract invoked polymorphically through its concrete overrides, never by a direct call the reachability walk can see, so it must not be flagged.

`abstractmethod` is added to the default reachability-root decorator set.

## Why

`@abstractmethod` methods carry `abstractmethod` in their stored `decorators`, which the dead-code query already normalizes and matches against the root-decorator set, so this is a pure query/config change (no schema change, no re-index needed). Real instance from `cgr dead-code`: `SqsBatchJob._raw_batch_operation` and `BatchingStrategyBase._validate_output_schema`, both `@abstractmethod`, called via `self.M()` and implemented by concrete subclasses.

This complements the self/cls dispatch fan-out (which reaches the concrete overrides): the fan-out deliberately skips the abstract stub as a call target, and this seeds the stub as a root so the stub itself is not reported dead.

## Tests (RED to GREEN)

- `test_dead_code_eval.py::test_abstract_method_is_a_root`: `@abstractmethod` and `@abc.abstractmethod` stubs with no caller are not flagged.

Full suite: 4337 passed, 14 skipped.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -471,6 +471,7 @@ DEFAULT_ROOT_DECORATORS: frozenset[str] = frozenset(
         "root_validator",
         "field_serializer",
         "model_serializer",
+        "abstractmethod",
     }
 )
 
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #595, opened 2026-07-02
fix(imports): resolve Python absolute imports through nested source roots

vitali87:
## What

Fixes a whole class of dropped call/reference edges: Python absolute imports from packages under **nested source roots** (src-layout, monorepo `packages/*/src/*`, setuptools `package-dir` remaps, PEP 420 namespace packages) were misclassified as external imports, and the resolver then deliberately suppressed the fallback for them, so every cross-module edge through such an import silently vanished.

## Root cause (verified in code)

- `_is_local_module` (`import_processor.py`) judged an import first-party only if the top-level name exists **directly at the repo root**, so `import pkg...` with `pkg` at `packages/a/src/pkg/` was judged external.
- `_resolve_import_full_name` then left the target unprefixed.
- `_is_external_import` (`call_resolver.py`) checks `{project}.{target}` against the registry; the node is registered under the path-based QN (`packages.a.src.pkg...`), so the lookup misses, the import is confirmed "external", and the anti-rebind guard suppresses the trie fallback. Result: no edge at all.

Rust already gets crate-root detection and JS/TS gets tsconfig `paths`/`baseUrl` mapping; Python was the language missing source-root awareness.

Real-world impact: in a src-layout monorepo, `cgr dead-code` flagged every task handler registered through an absolute-import dispatch dict (`handlers = {"name": _impl}`) because the dict's reference edges were dropped.

## How

New `parsers/python_source_roots.py`:
- `discover_python_source_roots(repo_path)`: one repo walk (pruned by `IGNORE_PATTERNS`) mapping each importable top-level name to the dotted repo-relative directory holding it, from three signals: a package whose parent is not a package, children of a `src/` directory (namespace packages and single modules), and `pyproject.toml` `[tool.setuptools.package-dir]` remaps.
- `resolve_via_source_roots(...)`: translates `pkg.sub` to `packages.a.src.pkg.sub`; same-named roots are disambiguated by which candidate actually contains the imported submodule on disk.

`ImportProcessor._resolve_import_full_name` consults the map (lru-cached) after the existing root-level check, so flat-layout behavior is unchanged. No resolver changes needed: once the import target is the path-based QN, the existing registry lookups and the external-import guard pass naturally.

## Tests (RED to GREEN): full layout matrix

`test_python_source_root_imports.py`:
1. flat layout, absolute import (regression, passed before the change)
2. src-layout, absolute import
3. src-layout, module-level dispatch-dict value (the dead-code trigger shape)
4. monorepo, cross-package absolute import between two src roots
5. `package-dir` remap via pyproject.toml (import name != directory name)
6. PEP 420 namespace package under src (no `__init__.py`)
7. two same-named top-level packages disambiguated by submodule existence

Full suite: 4345 passed, 14 skipped.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -1225,6 +1225,9 @@ DIST_DIR = "dist"
 BYTES_PER_MB_FLOAT = 1024 * 1024
 
 PYPROJECT_PATH = "pyproject.toml"
+PYPROJECT_KEY_TOOL = "tool"
+PYPROJECT_KEY_SETUPTOOLS = "setuptools"
+PYPROJECT_KEY_PACKAGE_DIR = "package-dir"
 TREESITTER_EXTRA_KEY = "treesitter-full"
 TREESITTER_PKG_PREFIX = "tree-sitter-"
 
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -482,6 +482,7 @@ IMP_CPP_MODULE_IMPL = "C++20 module implementation: {name}"
 IMP_CPP_MODULE_IFACE = "C++20 module interface: {name}"
 IMP_CPP_PARTITION = "C++20 module partition import: {partition} -> {full}"
 IMP_GENERIC = "Generic import parsing for {language}: {node_type}"
+IMP_PY_SOURCE_ROOT = "Python source root: {name} -> {path}"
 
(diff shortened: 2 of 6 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #598, opened 2026-07-02
fix(java): resolve imported class qn for flat-layout cross-package calls

vitali87:
## What

Fixes dropped Java cross-package call edges in FLAT layouts (package directories at the repo root, e.g. `com/foo/...`). Both static (`Helper.help()`) and instance (`h.inst()`) calls through an `import com.foo.util.Helper;` produced no CALLS edge.

## Root cause

A local Java import is recorded in the import map as the imported FILE's module qn (`repo.com.foo.util.Helper`), but the registered class qn duplicates the class segment (`repo.com.foo.util.Helper.Helper`). The raw target then dead-ends: `_search_method_in_class` finds no member under the module qn, and because the target carries the project prefix, `_find_method_with_any_signature` skips the layout-independent fqn-map fallback (it is gated to non-project-prefixed types).

The irony: Maven/Gradle layouts work only by accident. There the import is misclassified as external (the locality check `_is_local_java_import` only looks at the repo root), which leaves the target unprefixed and routes it to the fqn-map fallback that resolves by package-suffix. Flat layout is the one case classified correctly, and it was the one that broke.

## How

New registry-guarded `_imported_class_qn(target, type_name)` in the Java type resolver: when the import-map target is not itself a registered Class/Interface but `target.{type_name}` is, return the class qn. Applied at both import-map consumption sites (`_resolve_java_type_name` for declared-variable types and `_resolve_java_object_type` for static receivers). Targets that are already class qns, nested-class imports, and external fqns pass through unchanged.

## Tests (RED to GREEN): Java layout matrix

`test_java_source_layout_calls.py`:
1. flat layout, static call (RED before)
2. flat layout, instance call via declared variable (RED before)
3. Maven `src/main/java` (regression)
4. package root nested under an unmarked directory (`lib/com/...`) (regression)
5. Gradle multi-module cross-module call (regression)

Full suite: 4355 passed, 14 skipped.

### DIFF
--- a/codebase_rag/parsers/java/method_resolver.py
+++ b/codebase_rag/parsers/java/method_resolver.py
@@ -38,6 +38,9 @@ class JavaMethodResolverMixin:
     def _resolve_java_type_name(self, type_name: str, module_qn: str) -> str: ...
 
     @abstractmethod
+    def _imported_class_qn(self, target: str, type_name: str) -> str: ...
+
+    @abstractmethod
     def _rank_module_candidates(
         self, candidates: list[str], class_qn: str, current_module_qn: str | None
     ) -> list[str]: ...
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #599, opened 2026-07-03
feat(graph): REFERENCES edges keep first-class function values reachable

vitali87:
## What

Functions are first-class values: they get passed as arguments, assigned to variables and object attributes, and stored by framework plumbing for later dynamic dispatch. cgr's call graph only recorded a keep-alive edge when a callback was passed to an EXTERNAL callee, so every other first-class use produced a dead-code false positive.

This adds a `REFERENCES` relationship: an edge from the referencing scope to a function used as a value, walked by dead-code reachability (`CALLS|REFERENCES`).

## The three mechanisms it fixes (all reproduced from a real monorepo)

1. **Stored-not-invoked callbacks to first-party callees.** `with_brrr_from_cfg(cfg, handlers, create_context=create_context)` stores the callback in config; callable-param flow only emits an edge when the callee visibly invokes the parameter, so the callback had no inbound edge. Also covers first-party CONSTRUCTORS (`AnteriorCodec(create_context=create_context)`).
2. **Misbound external methods.** `df.apply(build_suggestion, axis=1)` is pandas, but a same-named first-party method (`BatchingStrategyBase.apply`) makes the trie bind the call to it, so the external-callee reference path never ran and the callback was dropped.
3. **Function-value assignments.** `http_callback = llm_http_task_closure_with_context` and monkeypatch-style `mock_client.post = handle_post` reference a function without any call expression at all.

## How

- `RelationshipType.REFERENCES` (+ schema entry, inbound-edges query template).
- Call arguments: the existing argument scan now also runs for first-party callees, emitting `REFERENCES` (external/builtin callees keep their historical `CALLS` reference edges; precise invocation edges still come from callable-param flow).
- Assignments: a new per-scope walker emits `REFERENCES` for a plain name/attribute RHS that resolves to a registered callable, stopping at nested-scope boundaries.
- Dead-code reachability (Cypher query + the in-memory eval mirror) traverses `CALLS|REFERENCES`, including module-root seeding.

## Tests (RED to GREEN)

- `test_callback_reference_edges.py` (6): first-party kwarg callback, constructor kwarg callback, misbound external method, local-variable assignment, object-attribute assignment (with the transitive `CALLS` chain), and a negative (plain arguments produce no REFERENCES noise).
- `test_dead_code_eval.py`: REFERENCES-only reachability keeps the callback live while an orphan stays dead; full-pipeline test for the stored-callback shape.
- Dead-code integration tests run the updated query against real Memgraph.

Full suite: 4386 passed, 14 skipped.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -553,6 +553,7 @@ class RelationshipType(StrEnum):
     IMPLEMENTS = "IMPLEMENTS"
     OVERRIDES = "OVERRIDES"
     CALLS = "CALLS"
+    REFERENCES = "REFERENCES"
     INSTANTIATES = "INSTANTIATES"
     DEPENDS_ON_EXTERNAL = "DEPENDS_ON_EXTERNAL"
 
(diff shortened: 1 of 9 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
Verified false positive: tree-sitter-python parses `handler: Callable = handle_event` as the same `assignment` node type with the identical `right` field (the annotation is an extra `type` child, there is no separate annotated-assignment node type). The walker therefore already scans it. Added `test_annotated_assignment_is_referenced` proving the REFERENCES edge is emitted for the annotated shape with no code change.

vitali87:
@greptile review

PR #600, opened 2026-07-03
fix(types): infer attribute-constructor receiver types to stop bare-name misbinds

vitali87:
## What

Fixes a false CALLS edge: a member call on a variable constructed by an ATTRIBUTE constructor (`df = pd.DataFrame(results); df.apply(...)`) was rebound by bare name to an unrelated first-party method of the same name (`BatchingStrategyBase.apply`), polluting the call graph with a wrong cross-package edge.

## Root cause

Python type inference only recognized BARE constructors (`DataFrame(...)`, via the uppercase-initial heuristic). An attribute callee like `pd.DataFrame(...)` was routed to method-return-type inference, which knows nothing about external modules, so the variable stayed UNTYPED. The resolver's known-external suppression (`_receiver_type_is_external`) only fires for typed receivers, so the untyped `df.apply` fell to the last-resort trie fallback, which matches by trailing method name and picked the first-party `apply` alphabetically.

## How

`py/expression_analyzer.py`: when an attribute call's method-return inference yields nothing and the final segment is uppercase-initial, treat `alias.ClassName(...)` as a constructor and record the dotted class (`pd.DataFrame`) as the variable's type — the same heuristic the bare-name branch already uses, extracted into `_attribute_constructor_type` and applied in both the simple-pass and complex-pass inference variants.

Effects:
- External receivers (`pd.DataFrame`) are now typed, so the existing suppression kills the bare-name rebind: no more false CALLS edge, and the passed callback is kept reachable via the external-callee reference path.
- First-party attribute constructors (`models.Helper()`) gain a usable receiver type, so `helper.process()` resolves precisely (regression-tested).

One pre-existing test updated: its fixture relied on the now-suppressed misbind to exercise the first-party REFERENCES arg path; it uses an untyped parameter receiver instead (still misbinds, still covered), with the typed-receiver behavior asserted in the new test file.

## Tests (RED to GREEN)

`test_external_receiver_misbind.py`:
1. external attribute constructor: no CALLS edge to the same-named first-party method; the callback keeps a keep-alive edge.
2. first-party attribute constructor: the member call resolves precisely.

Full suite: 4391 passed, 14 skipped.

### DIFF
--- a/codebase_rag/parsers/py/expression_analyzer.py
+++ b/codebase_rag/parsers/py/expression_analyzer.py
@@ -67,14 +67,28 @@ class PythonExpressionAnalyzerMixin(_ExprBase):
                 and func_node.type == cs.TS_PY_ATTRIBUTE
                 and (method_call_text := self._extract_full_method_call(func_node))
             ):
-                return self._infer_method_call_return_type(
+                if inferred := self._infer_method_call_return_type(
                     method_call_text, module_qn, None
-                )
+                ):
+                    return inferred
+                return self._attribute_constructor_type(method_call_text)
 
         elif node.type == cs.TS_PY_LIST_COMPREHENSION:
             if body_node := node.child_by_field_name(cs.TS_FIELD_BODY):
                 return self._infer_type_from_expression(body_node, module_qn)
 
+        return None
+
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #609, opened 2026-07-04
feat(references): emit REFERENCES for JSX component usage

vitali87:
## Summary
Follow-up to #608: with .tsx finally parsing, `<Card />` usage still produced no edge -- React invokes components through the element, never by a call the graph can see -- so components used only in markup still reported as dead.

- `_ingest_jsx_component_references`: walks each scope (plus module scope before the no-calls early return, like the assignment pass) for `jsx_self_closing_element` / `jsx_opening_element`, resolves the name through the standard callback-edge path, and emits REFERENCES from the rendering scope.
- Uppercase names only: lowercase tags are HTML elements and must not misbind to same-named locals. Closing elements are excluded so paired elements emit once. Member-expression names (`<Menu.Item />`) resolve like any dotted callable reference.
- Applies to both .tsx (tsx grammar) and .jsx (the javascript grammar parses JSX natively).

## Tests (RED -> GREEN)
5 tests in `test_jsx_component_references.py`: self-closing usage, paired element, .jsx parity, lowercase-tag negative, namespaced member component. 4 were RED.

Validation: FastAPI full-stack template dead-code 177 -> 168 with the ui/* component cluster cleared (proper .tsx parsing also registers more nested functions, so remaining candidates shifted to the useMutation-callback and generated-client classes).

Full suite: 4439 passed, 14 skipped. ruff + format clean; ty at the pre-existing baseline.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2290,6 +2290,11 @@ CPP_NESTED_SCOPE_NODE_TYPES = frozenset(
     )
 )
 JS_TS_PARENT_REF_TYPES = (TS_IDENTIFIER, TS_MEMBER_EXPRESSION)
+# (H) JSX element nodes that carry a component name (javascript and tsx
+# (H) grammars share these); the closing element repeats the name and must not
+# (H) double-emit.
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile-apps

PR #612, opened 2026-07-04
feat(references): emit REFERENCES for inline object-literal callback values

vitali87:
Inline arrow/function-expression values in object literals (`useMutation({ mutationFn: () => {}, onSuccess: () => {} })`) are registered by the JS/TS definition pass as their own nodes (`scope.mutationFn`, `scope.onSuccess`) but had zero incoming edges, so every TanStack Query / React callback reported as dead code.

`_ingest_collection_function_references` now detects inline function values at the object-literal pair level and emits a REFERENCES edge from the enclosing scope to each registered variant (`_emit_inline_value_function_ref` reconstructs the `{scope}.{key}` qn). Applies at both function and module scope; classic function-expression object values covered too.

This was the dominant remaining gap on the FastAPI full-stack template dead-code run (~75 of 168 candidates).

Tests: 2 new RED to GREEN cases in test_object_literal_callback_references.py; full suite 4457 passed.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -131,6 +131,13 @@ _ASSIGNMENT_RHS_REF_TYPES = frozenset(
 _JSX_NAMED_ELEMENT_TYPES = frozenset(
     {cs.TS_JSX_SELF_CLOSING_ELEMENT, cs.TS_JSX_OPENING_ELEMENT}
 )
+# (H) Inline function values in an object literal (`{ onSuccess: () => {} }`): the
+# (H) JS/TS definition pass registers these as their own nodes named by the key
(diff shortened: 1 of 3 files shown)


vitali87:
Addressed the string-literal-key case in 630c2a7, though differently than suggested: string-literal keys (`{'onSuccess': () => {}}`) have no property name, so the definition pass registers the inline arrow as `scope.anonymous_<row>_<col>` (position-based), not `scope.onSuccess`. Reconstructing `scope.<strippedKey>` would therefore match no registered node and emit nothing. The fix now references both the key-based name (identifier keys) and the value's position-based anonymous name (string keys), with a registry-existence guard so only real nodes are linked. Added a dedicated test.

vitali87:
@greptile review

PR #620, opened 2026-07-05
fix(go): treat init and main functions as dead-code roots

vitali87:
## What

Go's `func init()` is auto-run by the runtime at package load (any number per package) and `func main()` is the program entry point — neither is ever called explicitly, so both were reported as dead code. They are now reachability roots, exactly analogous to the existing Python-dunder root exemption, gated by the `.go` extension so same-named symbols in other languages are unaffected.

Applied to both dead-code surfaces: the eval reachability (`evals/dead_code.py`) and the production `cgr dead-code` Cypher query (`_DEAD_CODE_QUERY_TEMPLATE`).

## Impact

Measured on `gin-gonic/gin`: dead-code candidates **16 → 8**. The eliminated 8 were all `func init()` (config/codec registration, appengine, generated protobuf inits). The remaining 8 are a separate, harder Go method-receiver resolution gap (`node.addChild` etc. *are* called but don't resolve) — out of scope here.

## Test

RED→GREEN `test_go_init_and_main_are_roots`: `init`/`main` in `.go` files are not flagged, while a same-file uncalled helper still is (the exemption is name-scoped).

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -484,6 +484,12 @@ DEFAULT_ROOT_DECORATORS: frozenset[str] = frozenset(
     }
 )
 
+# (H) Go functions the runtime invokes with no explicit call site: `func init()`
+# (H) runs at package load (any number per package), `func main()` is the program
+# (H) entry. Both are reachability roots (like Python dunders), gated by the .go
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #621, opened 2026-07-05
fix(rust): treat main and trait-impl methods as dead-code roots

vitali87:
## What

Rust trait-impl methods are dispatched by the language and standard library, never through an explicit call the graph can see: `Display::fmt` via `format!`, `PartialEq::eq` via `==`, `Iterator::next` via `for`, operator traits, `Drop::drop`, serde, etc. And `fn main()` is the binary entry point. All were reported as dead. They are now reachability roots — the Rust analog of the existing Python-dunder root exemption — rooted by name and gated by the `.rs` extension so same-named symbols elsewhere are unaffected. Trait-method rooting is scoped to methods; `main` to any `.rs` function.

Applied to both the eval reachability (`evals/dead_code.py`) and the production `cgr dead-code` Cypher query.

## Impact

Measured on `tokio-rs/mini-redis`: dead-code candidates **67 → 59**. Eliminated: `Frame.fmt`/`Error.fmt`/`ParseError.fmt` (Display), `Frame.eq` (PartialEq), `SubscriberIterator.next` (Iterator), and the `main` entry points. The remaining candidates are separate classes out of scope here: a Rust method-receiver resolution gap (`Command.apply`, `into_frame`, `Listener.run` *are* called but don't resolve), Rust closures (`anonymous_*`), and integration-test functions under a top-level `tests/` dir (a language-agnostic test-path-exclusion gap).

## Test

RED→GREEN `test_rust_trait_methods_and_main_are_roots`: `fmt`/`eq`/`next` methods and `main` in `.rs` files are not flagged; a custom method (`push_int`) still is (the exemption is name-scoped to trait method names).


vitali87:
@greptile review

PR #622, opened 2026-07-05
fix(ts): unwrap as/satisfies/non-null casts when resolving function references

vitali87:
## What

A function aliased through a TypeScript cast — `export const persist = persistImpl as unknown as Persist` (the zustand middleware pattern), or `handler!`, or `fn satisfies T` — was never referenced: the assignment-reference walker saw an `as_expression` RHS (not a bare identifier) and skipped it, so the real implementation (and everything only it reaches) reported as dead.

A new `_unwrap_ts_cast` helper peels the transparent cast wrappers (`as_expression`, `satisfies_expression`, `non_null_expression`, nesting like `x as unknown as T`) to the wrapped value. It is applied in the assignment-reference walker's RHS and at the top of `_emit_callback_edge`, so casts resolve wherever a function reference is expected (assignments, call arguments, object values, JSX handlers, returns).

## Impact

Measured on `pmndrs/zustand`: dead-code candidates **319 → 269**, and specifically the middleware public API (`persistImpl`, `devtoolsImpl`, `immerImpl`, `reduxImpl` — each exported via `impl as unknown as Public`) is now live (`src/` 88 → 55). The remaining candidates are separate classes out of scope here: a language-agnostic top-level `tests/` path-exclusion gap, demo example apps, and deeper inline-callback patterns.

## Test

RED→GREEN `test_ts_as_cast_assignment_references_target` (`persist = persistImpl as unknown as Persist`) and `test_ts_non_null_cast_assignment_references_target` (`cb = handleEvent!`).

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2462,6 +2462,16 @@ TS_SPREAD_PARAMETER = "spread_parameter"
 TS_LOCAL_VARIABLE_DECLARATION = "local_variable_declaration"
 TS_FIELD_DECLARATION = "field_declaration"
 TS_ASSIGNMENT_EXPRESSION = "assignment_expression"
+# (H) TS "cast" wrappers that are transparent for reference resolution: `x as T`,
+# (H) `x satisfies T`, and the non-null assertion `x!`. Their first named child is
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

PR #627, opened 2026-07-06
fix(rust): reconstruct receiver chains from macro token streams for call resolution

vitali87:
## Problem

A `receiver.method()` call buried inside a macro (e.g. `tokio::select! { res = server.run() => ... }`) is tokenized into loose tokens (`server . run ( )`) — the `field_expression` structure is gone. Only the bare method identifier `run` was captured, so it mis-resolved to the same-module free function `server::run` — a **false self-edge**. In mini-redis this severed `Listener.run`'s only incoming edge, cascading the entire server/command cluster to dead: `Handler.run`, `Listener.accept`, `DbDropGuard.db`, `Shutdown::new/.is_shutdown`, `Command.apply`, `Unknown.apply`.

## Fix

`_macro_call_name` reconstructs the `<recv>.method` chain by walking the captured identifier's preceding sibling tokens over `("." <ident|self>)*`:
- `server . run` → `server.run` → resolves via the local's type to `Listener.run`
- `self . shutdown . recv` → `self.shutdown.recv` → hops `self` → field type → `Shutdown.recv`

Resolution itself is unchanged — this just hands the receiver-qualified name to the existing type-inference machinery. A receiverless macro call (`foo()`) still returns the bare name, so non-macro and other-language behavior is untouched.

## Impact

- mini-redis dead-code false positives **36 → 28** — exactly the 8 nodes predicted by tracing the cascade, zero regressions.
- gin (Go) unchanged at 2 — no cross-language regression.

This is the first of three verified root causes behind the remaining mini-redis false positives (the others: match-arm variable shadowing, and closures — filed separately).

## Tests

Two new RED→GREEN tests in `test_rust_receiver_resolution.py` (`test_macro_buried_receiver_call_dispatch`, `test_macro_buried_field_hop_call_dispatch`), each with a sort-first decoy type to defeat the trie coincidence. Full suite passes (the 2 tool-calling integration tests that fail are the live local Ollama model not emitting tool calls, unrelated).

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -3077,6 +3077,10 @@ TS_RS_TYPE_ARGUMENTS = "type_arguments"
 TS_RS_TRY_EXPRESSION = "try_expression"
 TS_RS_FIELD_EXPRESSION = "field_expression"
 TS_RS_FIELD_PATH = "path"
+TS_RS_TOKEN_DOT = "."
+# (H) Nodes that can be a receiver token preceding `.method` in a macro token
+# (H) stream: a plain identifier or the `self` keyword.
+RS_MACRO_RECEIVER_TYPES = (TS_IDENTIFIER, KEYWORD_SELF)
 # (H) Rust `Self` return type resolves to the enclosing impl target.
 RS_SELF_TYPE = "Self"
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #628, opened 2026-07-06
fix(rust): scope match-arm variant bindings per arm so shared names dispatch correctly

vitali87:
## Problem

When every arm of a `match` reuses one binding name for different variant types — the idiomatic Rust dispatch pattern:

```rust
match self {
    Get(cmd)     => cmd.apply(db, dst).await,
    Set(cmd)     => cmd.apply(db, dst).await,
    Ping(cmd)    => cmd.apply(dst).await,
    Unknown(cmd) => cmd.apply(dst).await,
}
```

the local-variable type map is flat (`{name: type}`), so `cmd` is overwritten by each arm and only the **last** arm's type survives. Every `cmd.apply()` then resolved to the last variant's method — in mini-redis, `Command.apply` emitted 6 CALLS edges **all to `Unknown.apply`**, leaving `{Get,Ping,Publish,Set,Subscribe}.apply` (and their cascade) with no incoming edge → dead-code false positives.

## Fix

`collect_match_arm_bindings` returns each variant binding scoped to its arm's byte range `(start, end, name, type)`. During call processing, `_overlay_match_arm_binding` overlays the binding whose arm range contains the call (innermost wins) onto a per-call copy of the type map, so each `cmd.apply()` resolves against **its own arm's** variant type. The flat map is untouched as a fallback; the overlay only fires for a receiver that is an in-scope arm binding.

## Impact

- mini-redis dead-code false positives **28 → 15** — exactly the 13 nodes predicted by the cascade trace (`{Get,Ping,Publish,Set,Subscribe}.apply`, `handle_command`, `make_*_frame`, `subscribe_to_channel`, `Frame.push_int`, `Command.get_name`, `Unknown.get_name`), zero regressions.
- The remaining 15 are all closures (a separate follow-up).
- gin (Go) unchanged at 2.

## Tests

`test_enum_match_multiarm_binding_shadowing` (RED→GREEN): three arms binding the same name to distinct variant types, asserting each dispatches to its own method. Full suite passes (the 2 tool-calling integration failures are the live local Ollama model, unrelated).


vitali87:
@greptile review

PR #629, opened 2026-07-06
fix(rust): reference closures from their enclosing scope so they are reachable

vitali87:
## Problem

A Rust `closure_expression` (`.map(|x| ...)`, a spawn body, a `let` binding) receives a `DEFINES` edge from its enclosing function but **no** `CALLS`/`REFERENCES` edge. Dead-code reachability walks only `CALLS`/`REFERENCES`, so every closure is an orphan and reports as a false positive — 14 of mini-redis's remaining candidates were closures.

## Fix

A closure is a value constructed at its definition site, so it is used wherever it is written. In `_create_function_relationships`, for a Rust `closure_expression`, emit a `REFERENCES` edge mirroring the `DEFINES` (enclosing scope → closure). The closure is then reachable exactly when its enclosing function is — the same treatment JS/TS already give inline callbacks. Gated to Rust closures; no other language or node type is affected.

## Impact

- mini-redis dead-code false positives **15 → 2** — all 13 orphaned closures (and their reachable bodies) revived.
- gin (Go) unchanged at 2.
- The remaining 2 (`State.next_expiration` and its own closure) are a **separate** finding: `state.next_expiration()` has `state` typed as `MutexGuard<State>` from `.lock().unwrap()` — a smart-pointer/guard deref receiver, not a closure issue.

This is the last of the three verified root causes behind mini-redis's post-receiver-resolution false positives (macro-buried calls #627, match-arm shadowing #628, and now closures).

## Tests

New `test_rust_closure_references.py` (RED→GREEN): asserts a closure in a free fn and in an impl method each carries a `REFERENCES` edge from its enclosing scope. Existing closure DEFINES/containment oracle tests still pass. Full suite green (the 2 tool-calling integration failures are the live local Ollama model, unrelated).

### DIFF
--- a/codebase_rag/parsers/function_ingest.py
+++ b/codebase_rag/parsers/function_ingest.py
@@ -532,6 +532,25 @@ class FunctionIngestMixin:
             (cs.NodeLabel.FUNCTION, cs.KEY_QUALIFIED_NAME, resolution.qualified_name),
         )
 
+        # (H) A Rust closure is a value constructed at its definition site (a `.map`
+        # (H) arg, spawn body, `let` binding), so it is used wherever it is written.
+        # (H) Dead-code reachability walks CALLS/REFERENCES (not DEFINES), so mirror the
+        # (H) DEFINES with a REFERENCES from the enclosing scope -- else every closure is
+        # (H) an orphan and reports dead. (JS/TS emit the analogous inline-callback ref.)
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #630, opened 2026-07-06
fix(rust): type guard/smart-pointer deref receivers via field-hop chain inference

vitali87:
## Problem

A local bound through a lock guard / smart pointer could not be typed, so a later method call on it fell back to the ambiguous name-only trie:

```rust
let state = self.shared.state.lock().unwrap();  // Db.shared: Arc<Shared>, Shared.state: Mutex<State>
state.next_expiration();                         // should resolve to State.next_expiration
```

The chain roots in a **field access** (`self.shared.state`), which the Rust chain collector didn't handle (it only handled `Type::assoc()` and bare-identifier bases), and `Mutex<T>`/`RwLock<T>` weren't treated as deref-to-inner.

## Fix

- `Mutex`, `RwLock`, `RefCell`, `Cell` join the deref-wrapper set, so a field/param/return `Mutex<State>` reduces to `State` (as it behaves for method dispatch — you always lock to access the inner value).
- `lock`, `read`, `write`, `try_lock` join the identity-method set (guard accessors that yield a guard dereferencing to the inner type).
- `_callee_chain_segments` now flattens **any** value chain to ordered segments, including field-access bases: `self.shared.state.lock().unwrap()` → `['self','shared','state','lock','unwrap']`.
- `_infer_rust_call_return_type` resolves the base via the local var-type map (`self` → impl target, or a typed local) or as a type name, then walks each hop trying field-type → method-return → passthrough identity. An unknown hop drops the binding (never guessed).

## Impact

- Adds guard/smart-pointer deref receiver typing, proven by an isolated RED→GREEN test.
- gin (Go) unchanged at 2; full Rust/resolution suite green.
- Note: mini-redis's two remaining candidates are **not** cleared by this alone — there the `state.next_expiration()` call sits inside a closure that *captures* `state` from the parent scope, and closure bodies don't yet inherit captured locals (a separate follow-up). This PR is the guard-deref capability; the capture-inheritance piece lands next.

## Tests

`test_guard_deref_field_hop_binding_dispatch` (RED→GREEN): `self.shared.state.lock().unwrap()` in an impl method, asserting `state.next_expiration()` dispatches to `State.next_expiration` and not a decoy. Full suite passes (the 2 tool-calling integration failures are the live local Ollama model, unrelated).

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -3081,13 +3081,20 @@ TS_RS_FIELD_PATH = "path"
 TS_RS_TOKEN_DOT = "."
 # (H) Nodes that can be a receiver token preceding `.method` in a macro token
 # (H) stream: a plain identifier or the `self` keyword.
-RS_MACRO_RECEIVER_TYPES = (TS_IDENTIFIER, KEYWORD_SELF)
+# (H) A receiver/chain base that is a plain identifier or the `self` keyword (used
+# (H) both for macro-token receiver reconstruction and value-chain base flattening).
+RS_IDENT_OR_SELF = (TS_IDENTIFIER, KEYWORD_SELF)
+RS_MACRO_RECEIVER_TYPES = RS_IDENT_OR_SELF
 # (H) Rust `Self` return type resolves to the enclosing impl target.
 RS_SELF_TYPE = "Self"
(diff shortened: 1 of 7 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #631, opened 2026-07-06
fix(rust): bubble anonymous closure calls to enclosing scope so captured-local receivers resolve

vitali87:
## Problem (Class E: closure-captured local inheritance)

A call inside an anonymous Rust closure on a captured local was dropped from the call graph:

```rust
let state = self.shared.state.lock().unwrap();      // typed State via guard-deref (Class D)
expire.map(|_when| state.next_expiration())         // state.next_expiration() -> NO edge
```

Root cause: an anonymous closure (`closure_expression`) has no name, so the call loop skips it as a caller. But `_calls_owned_by` still treated the closure as owning its body's calls and excluded them from the enclosing method. With no caller to attribute them to, those calls dropped entirely, orphaning their targets (e.g. `State.next_expiration` reported dead in mini-redis).

## Fixes

**1. Anonymous closures bubble (Class E).** `_js_ts_attributable_nodes` (renamed `_attributable_func_nodes`, now multi-language) excludes anonymous Rust `closure_expression` nodes from the ownership set, exactly as it already does for JS/TS anonymous arrows. The closure's calls now bubble up to the enclosing fn/method, which holds the captured locals' types, so `state.next_expiration()` resolves against the enclosing scope's `state: State`.

**2. Named nested Rust `fn`s own their calls (Rust-scoped).** A named nested `fn inner()` inside a free `fn outer()` previously had its body calls double-counted onto `outer` (a pre-existing bug: the free-function path used `_filter_calls_in_node` without the descendant-ownership exclusion the method path already applies). The Rust free-function path now uses `_calls_owned_by`, so `inner`'s calls belong to `crate.lib.outer.inner` only, not also to `crate.lib.outer`. Scoped to Rust to preserve other languages' flow-tracing behavior (a broad application regressed zustand's TS dead-code count 268 -> 282).

## Validation

- RED->GREEN tests: closure-captured local receiver dispatch in a method; named nested fn calls not bubbled to the enclosing fn.
- All 22 Rust receiver + closure tests pass.
- Dead-code evals (head vs base 30221b9): **mini-redis 2 -> 0**, **gin 2 (unchanged)**, **click 5 (unchanged)**, **zustand 268 (unchanged)**.
- Full suite: 4499 passed (5 Rust oracle tests + 2 testcontainer integration tests verified green in isolation; 1 pre-existing live-Ollama tool-calling flake, all unrelated to this change).
- ruff + ty clean.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -912,7 +912,7 @@ class CallProcessor:
         # (H) Only functions that get their own caller node exclude their calls from
         # (H) the enclosing scope; anonymous arrows (skipped below) must not, so
         # (H) their calls bubble up instead of dropping.
-        owned_func_nodes = self._js_ts_attributable_nodes(method_nodes, language)
+        owned_func_nodes = self._attributable_func_nodes(method_nodes, language)
         for method_node in method_nodes:
             # (H) The body byte-range slice also captures functions of a NESTED
             # (H) class (Outer body contains Inner.run); those belong to the
@@ -2950,7 +2950,7 @@ class CallProcessor:
             return None
         return safe_decode_text(name_node)
 
-    def _js_ts_attributable_nodes(
+    def _attributable_func_nodes(
         self, func_nodes: list[Node], language: cs.SupportedLanguage
     ) -> list[Node]:
         # (H) The func nodes that will get their own caller node: named functions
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #634, opened 2026-07-06
feat(cpp): fan template-parameter receiver calls out to all interface implementers

vitali87:
## Problem

Second class from the nlohmann/json C++ dogfood (after operator roots, #632). The JSON parser dispatches SAX events through a template-parameter receiver:

```cpp
template<typename SAX>
bool sax_parse_internal(SAX* sax) {
    ...
    sax->start_object(elements);   // SAX is a template param -> no concrete type here
}
```

`SAX` has no concrete type at the call site, so precise resolution fails and the external-receiver guard drops the edge. Every structural implementer of the SAX concept (`json_sax_dom_parser`, `json_sax_acceptor`, `json_sax_dom_callback_parser`) reports as dead. The concrete classes do NOT inherit the abstract `json_sax` base (duck-typed via an `is_sax` trait), so there is no inheritance edge or override-resolution to leverage.

## Fix

When a C++ member call's receiver is typed to a **template type parameter** of the enclosing scope (`template<typename SAX>`), fan the call out to the method on **every** class defining it (mirrors the existing `protocol_dispatch_targets` / `self_dispatch_targets` multi-target pattern):

- `CppTypeInferenceEngine.collect_template_param_names()` gathers the template type-parameter names in scope (function's own + enclosing class templates).
- `CallResolver.cpp_dispatch_targets()` returns the method on every class when the receiver is typed to one of those params.
- `call_processor` emits these edges before the primary resolution, so they land even when the external-receiver guard would drop the single edge.

**Guardrails (why this is precise, not a flood):**
- Only template-parameter-typed receivers fan out. A concrete first-party receiver still dispatches exactly; an external-typed receiver (a `std::string` field's `buffer_.clear()`) is left untouched and must NOT rebind to a first-party `clear`. An untyped receiver is left to the single-best trie fallback rather than flooding the graph.

## Validation

- RED->GREEN test `test_template_param_receiver_fans_out_to_all_implementers`; guardrail test `test_typed_receiver_does_not_fan_out`; the existing external-field-receiver precision test still passes.
- Dead-code evals (head vs base 1f673cc): **nlohmann/json 447 -> 436**, **fmt 1305 -> 1287** (both modest, principled reductions -- interface implementers revived, not a false-edge flood).
- 275 C++ unit tests pass; ruff + ty clean.

**Note:** nlohmann/json's `json_sax_dom_parser` SAX methods remain dead by a separate *cascade* -- their driver `parser::sax_parse_internal` is itself unreachable (the public `parse()` -> `parser` construction/dispatch is unresolved). That upstream `parser`-reachability gap is a distinct follow-up finding, not a dispatch problem this PR can address.


vitali87:
@greptile review

PR #636, opened 2026-07-07
fix(cpp): register a bodied templated class once under its natural qn

vitali87:
## Problem

A bodied templated C++ class was captured twice by the class query: once as the `template_declaration` wrapper and once as the inner `class_specifier`. Both reached `_process_class_node`, so the class registered **two** nodes. The wrapper (no `body` field) claimed the natural qn but attached no members; the inner specifier registered with a `@line` suffix and received all the methods, fields, and base clause.

Result: every method lived under `Foo@N.*` while callers reference the natural `Foo.*`, so the whole class was orphaned from the call graph. In `nlohmann/json`, 156 of 335 class nodes had an empty natural twin, keeping the entire `parser`/`json_sax_*` cluster unreachable in dead-code analysis.

## Fix

The `template_declaration` is the canonical node (it carries the template parameters and always registers under its natural qn). Drop the inner `class_specifier` outright for bodied templates too, not just bodyless forward decls, and route member extraction (base clause, field-type map, methods) through the inner specifier so members bind to the class's natural qn.

This adds **no new edges** — it only makes the method/field nodes exist under the qn that callers already reference, so previously-dangling edges land.

## Impact

- Dead-code eval: `nlohmann/json` 436 -> 427, `fmt` 1287 -> 877 (heavily templated; all revived nodes come from pre-existing edges).
- 280 C++ tests pass; full non-integration suite green.
- No cross-language effect: the drop is gated on C++ node types, and for non-C++ (and non-template C++) the member node equals the class node, so behavior is unchanged.

## Test

`test_cpp_template_class_registration.py` (RED before, GREEN after): a templated class registers exactly one Class node under its natural qn, its methods attach to that qn, and an in-class member call resolves to it.

### DIFF
--- a/codebase_rag/parsers/class_ingest/mixin.py
+++ b/codebase_rag/parsers/class_ingest/mixin.py
@@ -281,6 +281,21 @@ class ClassIngestMixin:
         # (H) solely via specializations) has no other node, so it must be kept. We
         # (H) cannot tell which until every file's definitions are in, so defer the
         # (H) forward declaration and decide in resolve_deferred_forward_declarations.
+        # (H) The class query captures a templated class twice: the inner
+        # (H) class_specifier AND its template_declaration wrapper. The wrapper is the
+        # (H) canonical node (it carries the template params and always registers with
+        # (H) its natural qn); the inner specifier is redundant. Drop the inner one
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

PR #645, opened 2026-07-07
fix(parsers): key query cache by Language object to prevent id reuse collisions

vitali87:
## Problem

The JS/TS assignment-function ingestion tests (`test_assignment_arrow_functions_are_ingested`, `test_object_literal_methods_are_ingested`, `test_export_const_function_expression`, ...) flake in CI, almost exclusively on py3.13 jobs, across all three OSes. This is what failed `Unit Tests (ubuntu-latest, py3.13)` on #643 and several other recent PR runs.

## Root cause

`get_cached_query` keyed its module-level cache on `(id(language_obj), query_text)` without retaining the `Language` wrapper. `Query` does not hold a Python reference to its `Language` either (refcount stays at baseline after construction). So once a test's wrapper is garbage collected, a later `Language` allocation (possibly for a different grammar) can land at the same address, and the cache serves a `Query` compiled for the wrong grammar. Captures silently come back empty (the code path swallows the error into a debug log), and the assignment-arrow tests assert False.

Whether an address gets reused depends on allocator and GC timing, which is why the failures are nondeterministic and skew toward py3.13.

## Fix

Key the cache on the `Language` object itself. `Language.__hash__`/`__eq__` follow the underlying grammar pointer, so distinct wrappers of the same grammar share one cache entry, the dict pins the key alive, and an address can never be reused into a stale hit.

RED: `test_query_cache.py::test_cache_keyed_by_grammar_not_wrapper_address` fails on main (two wrappers of the same grammar get separate cache entries). GREEN with the one-line key change.

Full non-integration suite: 4399 passed, 5 skipped.

### DIFF
--- a/codebase_rag/parsers/utils.py
+++ b/codebase_rag/parsers/utils.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, NamedTuple
 
 from loguru import logger
-from tree_sitter import Node, Query, QueryCursor
+from tree_sitter import Language, Node, Query, QueryCursor
 
 from .. import constants as cs
 from .. import logs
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

PR #651, opened 2026-07-08
fix(js): reference new-expression constructor callbacks and nested arrow-const caller qns

vitali87:
## What

Two root-cause fixes that eliminate dead-code false positives for JS/TS constructor callbacks and nested arrow-const callbacks. Measured on the full-stack-fastapi-template, this takes the frontend dead-code count from 7 to 4 (the whole session ran 333 to 4).

### 1. `new X(callback)` was never treated as a call

`SPEC_JS_CALL_TYPES` captured only `call_expression`, so `new_expression` was invisible to the call pass (C++ and Python already include it). A Promise executor `new Promise((resolve, reject) => {...})` or `new CancelablePromise(async (...) => {...})` therefore got no incoming edge and reported as dead.

- Added `new_expression` to the JS/TS call query.
- Taught `_get_call_target_name` to read the `new_expression` `constructor` field (guarded to JS/TS so it does not intercept the identically-named C++ node).

This also wires the previously-missing INSTANTIATES edge for `new Foo()` to a first-party class.

### 2. Nested arrow-const callers got a flattened qualified name

`_build_nested_qualified_name` read only an ancestor's `name` field. An arrow-const ancestor (`getQueryString = () => {...}`) has no `name` field, so its segment was dropped: a nested `encodePair`'s callbacks were attributed to `request.encodePair` instead of `request.getQueryString.encodePair`, and the inline `.forEach(...)` arrow arguments never matched their registered nodes. Added the same nameless-binding fallback the definition pass already uses.

## Tests

Three new tests in `test_object_literal_callback_references.py`, each verified RED before the fix and GREEN after:

- inline arrow argument to `new X(...)` is referenced (Promise executor)
- `new Foo()` records INSTANTIATES to a first-party class
- inline `.forEach(...)` callback in a nested arrow-const is referenced

Full suite: 4559 passed, 0 failed. ruff and ty clean.


vitali87:
@greptile review

PR #656, opened 2026-07-08
fix(go): resolve inline field-hop calls and build-variant package siblings

vitali87:
Drives the gin dead-code eval from 2 residual false positives to 0. Both are root-cause fixes with RED->GREEN tests; cobra and mini-redis stay at 0 (no regressions).

## 1. Field-receiver direct call (`responseWriter.reset`)

Two composing gaps behind gin's `c := engine.pool.Get().(*Context); c.writermem.reset(w)`:

- **Type assertion binding**: `c` was bound via a Go type assertion (`x := expr.(*T)`), which `_infer_value_type` did not handle, so `c` stayed untyped and the field-hop fell to the bare-method trie, mis-binding to the same-named `Context.reset`. Added a `type_assertion_expression` case (reads the `type` field; `type_identifier_text` unwraps `*T`).
- **Inline field-hop resolution**: even once typed, a paren-free field-hop called inline with no intermediate local (`c.writermem.reset`) was never walked; `_resolve_multi_part_call` only joined the tail as a method name. Added `_resolve_field_hop_method`: resolve the base local type, walk each middle segment as a struct field via `class_field_types`, then resolve the final method on the last field's type. Precise and revive-only (no trie fallback), distinct from the stored-local form (`root := e.field.get(); root.m()`) already handled.

## 2. Build-tag duplicate function (`binding_nomsgpack.validate`)

cgr keys each file as its own module (`pkgdir.file.name`), but Go package-level functions are package-scoped. gin defines `validate` in both `binding.go` (`//go:build !nomsgpack`) and `binding_nomsgpack.go` (`//go:build nomsgpack`); cross-file callers all resolve to one copy, orphaning the other build's copy.

In valid Go, two same-package same-name top-level functions can only be mutually-exclusive build-tag variants (the compiler rejects real duplicates), so a fan-out is safe. Added `go_package_sibling_targets`, which returns same-package same-name FUNCTION-labeled siblings; the call processor emits a CALLS edge to each, gated to `language == GO and callee_type == FUNCTION`. Revive-only; a same-name function in a different package is excluded by the package-directory equality guard.

## Tests

- `test_go_chained_return_type.py::test_direct_field_hop_method_call_resolves` (type assertion + field-hop + `Context.reset` decoy to defeat trie coincidence).
- `test_go_build_variant_functions.py` (sibling fan-out + different-package guard).


vitali87:
@greptile review

vitali87:
@greptile review

PR #658, opened 2026-07-08
fix(java): link new-expression to constructor and class instantiation

vitali87:
Found by dogfooding the dead-code eval on google/gson (Java's first dead-code dogfood). Drives gson core from 114 to 92 false positives by fixing the single largest class.

## The class: Java `new X(...)` never linked to constructors

`SPEC_JAVA_CALL_TYPES` collected only `method_invocation`, never `object_creation_expression`, so `new X(...)` produced **no INSTANTIATES edge to the class and no CALLS edge to the constructor**. Every constructor reached only via `new` therefore looked dead (45 of gson's 114 false positives).

## Fix (mirrors the JS `new_expression` / Python class-call path)

1. Add `object_creation_expression` to `SPEC_JAVA_CALL_TYPES`.
2. `_get_call_target_name` names a Java `object_creation_expression` by its `type` field (strips generic args, e.g. `new ArrayList<T>()` -> `ArrayList`; a scoped `Outer.Inner` is left for the resolver), routing construction through the normal resolve loop.
3. At the emit site, a resolved class node gets INSTANTIATES plus a CALLS edge to each declared constructor via the new `java_constructor_targets` helper. A Java constructor is registered as a method named like its class (`Foo.Foo(int)`); the helper matches only constructors DIRECTLY on the class (a nested class's constructor has an extra qn segment and is excluded). Argument-type overload selection is not attempted, which is unnecessary for reachability and never fabricates a call to a non-constructor.

## Tests

`test_java_constructor_calls.py`: a `new X(...)` both instantiates the class and calls the constructor; an overloaded class has every constructor reached.

Verified: gson 114 -> 92, full suite green (4427 passed), 522 Java tests pass, no cross-language effect (every new path is `language == JAVA`-gated).


vitali87:
@greptile review

PR #660, opened 2026-07-08
fix(java): resolve nested-class type names referenced by simple name within a file

vitali87:
Third Java dead-code class from dogfooding google/gson. Fixes nested-class type-name resolution so a receiver typed by a nested class resolves its method calls.

## The class: nested-class-typed receiver never resolves

`_resolve_java_type_name(type_name, module)` resolved a simple type name only against `module.Type` (same-package top-level). A nested class is registered as `module.Outer.Nested`, so a field/variable typed by a nested class (gson's `private static final RecordHelper RECORD_HELPER`, where `RecordHelper` is nested in `ReflectionHelper`) never resolved. Every `RECORD_HELPER.isRecord(...)` / `getRecordComponentNames(...)` call dropped, so the whole nested `RecordHelper` hierarchy (base plus its strategy subclasses) was reported dead by cascade.

## Fix

When the direct `module.Type` lookup misses, search this module for a CLASS/INTERFACE whose last qualified-name segment is the simple type name, and use it only when unambiguous (so a same-named nested type elsewhere cannot mis-resolve). This resolves `RecordHelper` to `...ReflectionHelper.ReflectionHelper.RecordHelper`.

## Tests

`test_java_nested_type_receiver.py`: a static field typed by a nested abstract class resolves its method call to the nested class's method (the gson shape).

Verified: gson core 79 -> 72 (the RecordHelper base cluster revived), full suite green (4438 passed), Java suite 525 passed, no regressions.

### DIFF
--- a/codebase_rag/parsers/java/type_resolver.py
+++ b/codebase_rag/parsers/java/type_resolver.py
@@ -162,15 +162,17 @@ class JavaTypeResolverMixin:
         # (H) A nested class referenced by its simple name from within the same file
         # (H) (`RECORD_HELPER` typed by the nested `RecordHelper`): the qn is
         # (H) `module.Outer.Nested`, not `module.Nested`, so the direct check above
-        # (H) misses it. Search this module for a CLASS/INTERFACE whose last segment is
-        # (H) the simple name, and use it only when unambiguous so a same-named nested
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #661, opened 2026-07-08
fix(dead-code): treat overrides of reachable methods as reachable

vitali87:
Stacked on #660. Fourth dead-code fix from dogfooding google/gson: an override of a reachable method is itself reachable.

## The class: overrides reached only by virtual dispatch look dead

A call to a base or interface method dispatches at runtime to any override, but the reachability walk follows `CALLS`/`REFERENCES` only. The graph records `OVERRIDES` (overrider -> overridden), so an override reached solely through dispatch on the base type had no incoming walked edge and was reported dead (gson's `RecordHelper` strategy subclasses `RecordSupportedHelper` / `RecordNotSupportedHelper`).

## Fix

An override of a REACHABLE method is a potential runtime dispatch target, so it is reachable (sound, conservative virtual dispatch, mirroring the existing Python self-dispatch handling).

- `evals/dead_code.py`: add a reverse edge `overridden -> overrider` for each `OVERRIDES` edge, so the walk reaches every override (and transitively its callees) of any live base method.
- `codebase_rag/cypher_queries.py`: an `OVERRIDES` expansion stage mirroring the existing decorated-closure expansion, adding overriders of any live method (plus their forward closure) to the live set.

Revive-only: an override of a DEAD base stays dead (verified in the test), so this only removes false positives, never masks a genuinely dead method whose base is unreachable.

## Tests

`test_dead_code_eval.py::test_override_of_reachable_method_is_reachable`: a called base method's override is revived; an override of an uncalled base stays dead.

Verified: combined with #660, gson core drops 72 -> 67 (the RecordHelper override cluster revived); full suite green (4439 passed), no cross-language regression (revive-only, gated on real OVERRIDES edges).

### DIFF
--- a/codebase_rag/cypher_queries.py
+++ b/codebase_rag/cypher_queries.py
@@ -229,7 +229,7 @@ UNWIND (
 OPTIONAL MATCH (cr)-[:{traversal}*BFS]->(cr_reached)
 WITH live_set, closure_roots, collect(DISTINCT cr_reached) AS cr_reached_set
 WITH live_set + closure_roots + cr_reached_set AS live_set
-OPTIONAL MATCH (ov:Function|Method)-[:OVERRIDES]->(base)
+OPTIONAL MATCH (ov:Function|Method)-[:OVERRIDES*]->(base)
 WHERE base IN live_set AND NOT ov IN live_set
 WITH live_set, collect(DISTINCT ov) AS override_roots
 UNWIND (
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
Superseded: after #659/#660 squash-merged, the stacked branches conflicted. The override-reachability commits from this PR are included, verified, in #662 (rebased clean onto main). Landing via #662.

PR #662, opened 2026-07-08
fix(java): anonymous-class override dispatch (OVERRIDES edges + lexical-scope calls)

vitali87:
Stacked on #661. Bug 1 from dogfooding google/gson: Java anonymous-class override dispatch. Clears the JavaTimeTypeAdapters cluster (gson 67 -> 49).

## Two composing bugs behind the cluster

`new IntegerFieldsTypeAdapter<T>(...){ @Override create(); integerValues(); }` — 8 anonymous adapters whose overrides looked dead.

**1. Anonymous-class methods had no OVERRIDES edge.** An anonymous class is not modelled as a subclass, so its override methods register under the enclosing class with no link to the base. Fix: `ingest_method` now returns the registered qn; `_ingest_class_methods` detects a method inside an anonymous `class_body` and records `(anon_method_qn, name, base_type, module_qn)`; a deferred pass (after all types are registered) resolves the base type by unique global simple-name match and emits an `OVERRIDES` edge to each matching base method. With #661's override-reachability, the override is live when the base is.

**2. Unqualified calls ignored lexical scope.** `create(values)` inside `IntegerFieldsTypeAdapter.read` (i.e. `this.create()`) resolved to the *outer* class's same-named `create` via a module-wide name scan, so the base's own abstract `create` never got the edge and stayed dead. Fix: an unqualified call resolves against the ENCLOSING class hierarchy (own + inherited + interface, via `_lexical_class_qn`) before the module-wide fallback.

Both are revive-only / precision-improving (an ambiguous base or a genuinely dead base leaves the override dead).

## Tests

`test_java_anonymous_override_reachable.py`: an anonymous override of a called base method is not reported dead.

Verified: gson core 67 -> 49 (JavaTime/IntegerFields cluster 19 -> 1), full suite green (4441 passed), Java suite 527 passed, no cross-language regression.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -3405,6 +3405,14 @@ FQN_JAVA_SCOPE_TYPES = (
     TS_INTERFACE_DECLARATION,
     TS_ENUM_DECLARATION,
     TS_PROGRAM,
+    # (H) An enclosing method/constructor is a qn scope for a function nested in its
+    # (H) body (a method-body anonymous class's method): the call pass builds
+    # (H) `Class.method.nested`, so the definition pass must include the method too --
+    # (H) else the two disagree and every edge from the nested function dangles onto a
+    # (H) phantom `Class.method.nested` node, orphaning its callees. A direct method is
+    # (H) the func itself, not an ancestor, so its own qn is unaffected.
+    TS_METHOD_DECLARATION,
+    TS_CONSTRUCTOR_DECLARATION,
 )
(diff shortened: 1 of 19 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #664, opened 2026-07-08
fix(java): cast-expression receivers and anonymous-class own-method calls

vitali87:
Follow-up to the Java dead-code series (#658/#659/#660/#662). Fixes two more real resolution bugs found by tracing gson's residual dead-code; gson core 21 -> 19.

## Cast-expression receivers
A method call on a cast receiver `((T) x).m()` (gson's `((JsonTreeReader) in).nextJsonElement()`) was dropped: the Java call-info extractor ignored `parenthesized_expression`/`cast_expression` receivers, so `object` was null and the call fell to the unqualified path — which never finds a cross-file/sibling `T`. Fix: extract the cast's target type as the receiver so `m` resolves on `T`.

## Anonymous-class own-method calls
An unqualified call to an anonymous class's OWN (non-inherited) method — gson's `delegate().read()` where `delegate()` is defined in the same anon — didn't resolve, because those methods register as `Function` nodes and the module-wide fallback (`_resolve_static_or_local_method`) only considered `Method`/`Constructor`. Fix: include `Function` in that fallback's callable set (it is a last-resort scan after precise class/anon-base/enclosing lookups).

## Tests
`test_java_cast_receiver.py` (cross-file cast with a decoy) and `test_java_anon_method_caller_qn.py::test_anon_own_method_unqualified_call_resolves`.

Verified: gson 21 -> 19, full suite green (4462), no over-match regression from the Function inclusion.

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -2417,6 +2417,7 @@ TS_ERROR = "ERROR"
 TS_EXPRESSION_STATEMENT = "expression_statement"
 TS_STATEMENT_BLOCK = "statement_block"
 TS_PARENTHESIZED_EXPRESSION = "parenthesized_expression"
+TS_JAVA_CAST_EXPRESSION = "cast_expression"
 TS_ATTRIBUTE = "attribute"
 
 FIELD_OPERATOR = "operator"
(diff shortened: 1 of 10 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #667, opened 2026-07-08
fix(js-ts): reference functions consumed via casts, currying, member assignment, call args, param defaults, and shorthand methods

vitali87:
## Summary

Dogfooding `cgr dead-code` against pmndrs/zustand surfaced six systematic JS/TS false-positive classes. All are "a function is consumed by syntax the call/reference pass did not scan," fixed with revive-only reference emission. zustand goes 32 -> 0 dead-code false positives.

## Fix classes (each RED -> GREEN)

1. **Cast-wrapped arrow bindings** (`test_ts_cast_wrapped_arrow_binding.py`): `export const createStore = ((s) => ... createStoreImpl) as CreateStore` lost ALL body call/ref edges. `_js_ts_arrow_binding_name` required the arrow to be the declarator's direct value; it now climbs paren/cast wrappers, matching the def pass.
2. **Curried inner arrows** (`test_ts_curried_arrow_references.py`): `(config) => (set, get, api) => {...}` has no `return_statement`; the expression body IS the implicit return. The returned-function pass now references function-typed bodies, including through unowned anon layers.
3. **Member-assigned inline functions** (`test_ts_member_assigned_function_references.py`): `api.setState = (state, replace) => {...}`. Two composing gaps: the assignment walk stopped at every function boundary (assignments inside anonymous curried arrows scanned by nobody), and only the position-named registration was referenced, never the property-named one (`persistImpl.setState`).
4. **Inline args to unresolved callees** (`test_ts_inline_arg_to_unresolved_callee.py`): an arrow passed to a param callee (`set((state) => ...)`) or cast-wrapped callee (`;(set as NamedSet)(...)`) was consumed invisibly. Every inline function argument is now referenced from the calling scope, and calls nested purely in anonymous scopes bubble to module scope instead of nobody.
5. **Default parameter values + shorthand methods** (`test_ts_default_param_and_shorthand_method_refs.py`): `useStore(api, selector = identity as any)` and persist's thenable `{ then(x) {...}, catch(x) {...} }` (method_definition nodes, which the dispatch-table scan skipped as non-pairs).
6. **Nested pair-key paths** (`test_ts_nested_pair_key_value_refs.py`): `configure({ onCreated: (s) => { s.setEvents({ compute: ... }) } })` registers as `Canvas.onCreated.compute`, but the scan built `Canvas.compute`; candidates now include ancestor pair keys.

Also guards `_emit_callback_edge` to callable targets only (a default value can resolve a cast's TYPE name to an Interface node, which is schema-invalid as a reference target).

## Verification

- Full suite: 4628 passed, 0 failed
- zustand: 32 -> 0
- gson unchanged (52), no cross-language drift (all new passes JS/TS-gated; shared walks no-op for other languages)

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -2688,7 +2688,12 @@ class CallProcessor:
         resolve_func = self._resolver.resolve_function_call
         ensure_rel = self.ingestor.ensure_relationship_batch
         for param in params.named_children:
-            value = param.child_by_field_name(cs.FIELD_VALUE)
+            # (H) TS carries a param default in required_parameter's `value` field;
+            # (H) plain JS wraps the param in an assignment_pattern whose default
+            # (H) sits under `right`. Scan both forms.
+            value = param.child_by_field_name(
+                cs.FIELD_VALUE
+            ) or param.child_by_field_name(cs.FIELD_RIGHT)
             if value is None:
                 continue
             value = self._unwrap_ts_value(value)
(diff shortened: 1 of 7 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #668, opened 2026-07-08
fix(py,ts): click and zustand dead-code residuals plus same-line and overload-variant caller attribution

vitali87:
## Summary

Second half of the cross-repo dead-code residual sweep: pallets/click goes 7 -> 0 and pmndrs/zustand returns to 0 (two regressions from recently merged work fixed at root). Every fix is RED -> GREEN.

### click (Python)
1. **Subscripted generic bases** (`test_py_subscripted_generic_base.py`): `class IntRange(_NumberRangeBase[int, int])` produced NO INHERITS edge -- the superclass walk skipped `subscript` nodes. Without the edge the subclass had no OVERRIDES, so `self._clamp(...)` dispatch never reached the override.
2. **Duplicate-variant nested scope** (`test_py_overload_variant_nested_scope.py`): click's real `command` registers as `command@168` behind its `@t.overload` stubs; its `return decorator(func)` probed `command@168.decorator` (never exists), fell to the module trie, and mis-bound to the alphabetically-earlier `argument.decorator` -- a false edge AND a false-dead. The enclosing-scope walk now probes the variant-stripped scope.
3. **Conditional-import fallback** (`test_py_conditional_import_fallback.py`): `if WIN: from ._winconsole import X ... else: def X(...)` is statically undecidable; the call resolved only to the import, orphaning the local fallback. Mirrors the Go build-variant fan-out: registry-guarded CALLS to the local same-named twin.
4. **External stdlib override roots** (`test_py_external_override_flag.py` + eval test): click's `TextWrapper(textwrap.TextWrapper)` overrides `_wrap_chunks`/`_handle_long_word`, invoked only by the stdlib parent's `wrap()`. New `overrides_external` node property (set when a method name exists on an external stdlib base, stdlib-gated import), rooted in both the eval and the Cypher dead-code surfaces.

### Cross-PR interaction regressions (zustand was 0, drifted to 2; 7 TS tests red on main)
5. **Same-line caller attribution** (fixes 7 red tests from the #665 x #667 interaction): `function_locations` is line-keyed, and a one-line curried arrow records both the outer and inner arrow at the same line -- `_recorded_caller` adopted the WRONG function's qn, silently dropping the outer arrow's whole flow pass. `FunctionLocation` now carries `start_col`; a column mismatch falls back to name-based attribution. The C++ template-child entry records the child's column.
6. **Def-pass naming for cast-wrapped function expressions**: `export const api = (function (x) {...}) as unknown as Api` registered as `anonymous_2_20` -- the declarator-name climb only handled arrows, while the call pass accepts both. Aligned.
7. **Overload-variant anon scope** (`test_ts_overload_variant_inline_refs.py`): a TS overload implementation registers as `useStore@27`, but its inline `useCallback` arrows register under the natural `useStore.anonymous_R_C`; positional reference candidates now include the variant-stripped scope.

## Verification

- Full suite: 4570 passed, 0 failed (91 errors all memgraph-integration, environment-only)
- click 7 -> 0; zustand 0 (restored); gson 55 on this branch == 55 on main (neutral)
- ruff clean; ty clean on all changed files

## Note for reviewers

gson moved 52 -> 55 ON MAIN (before this branch): `FieldNamingStrategy.translateName`, `ToNumberStrategy.readNumber`, and `TypeAdapters.FactorySupplier.get()` lost their interface-method binding (calls through interface-typed receivers now bind the concrete impl instead of the interface node). Introduced by #665/#666; tracked separately, not addressed here.


vitali87:
@greptile review

vitali87:
@greptile review

PR #670, opened 2026-07-09
fix(js): express dead-code residuals via named fn-expr references, member-call precision, twin fan-out, var-require imports, and test dir exclusion

vitali87:
## Summary

Fresh-repo dead-code dogfood on expressjs/express (first CommonJS/prototype-idiom target): 49 -> 1. The remaining 1 (`GithubView.render`) is genuinely dynamic (`app.set('view', GithubView)` -- a string-keyed settings registry express invokes reflectively), correct to report. Every fix RED -> GREEN; zustand 0, click 0, gson unchanged (22, after `/test/` correctly excludes its `src/test/java` tree).

## Fix classes

1. **Singular `test/` dir** (`test_dead_code_eval.py::test_singular_test_dir_is_excluded`): the Node.js/mocha convention (34 of express's 49 reports were test helpers). Previously excluded on purpose; the pattern is segment-anchored by the leading-slash normalization so `contest/`/`latest/` cannot match, and `/spec/` stays out (documented collision with domain "spec" modules).
2. **Named function expressions as call args** (`test_js_named_function_expression_args.py`): `this.on('mount', function onmount(parent) {...})`, `defineGetter(req, 'path', function path() {...})` register by their OWN name; positional candidates never matched. Name candidates (scope + module-flat) added.
3. **Named fn-exprs as pair values + logical defaults** (`test_js_named_pair_value_and_logical_default.py`): `get: function getrouter() {...}` (own-name candidates in the dispatch-table scan) and `done = done || function (err, str) {...}` (binary-RHS operands scanned).
4. **Object-value fn-exprs stealing the declarator name** (`test_js_object_value_function_expression_naming.py`): `var pets = { list: function(req, res) {...} }` climbed past the pair and registered phantom `pets`/`pets@50` Function nodes (a regression risk from #668's fn-expr climb). The climb now stops at pair/object boundaries.
5. **Untyped-receiver member calls** (`test_js_untyped_receiver_member_call.py`): `view.render(opts, cb)` inside `tryRender(view, cb)` mis-bound to the same-module free function `application.render` (false edge + dead View.render). A JS/TS dotted call now binds a same-module free function only through module-ish receivers (`exports`/`module`/`this`); otherwise it resolves to the unique member-like candidate, disambiguated by import-visibility, or drops.
6. **Prototype twin fan-out** (`test_js_prototype_assignment_single_registration.py`): `View.prototype.lookup = function lookup()` registers two nodes for one method (member `View.lookup` + module-flat own-name twin); a call that binds either now edges the same-module same-name twin too (the established duplicate-QN keep-both design; revive-only).
7. **CommonJS `var` requires** (`test_js_untyped_receiver_member_call.py::test_var_require_visibility_disambiguates_member_call`): `var View = require('./view')` is a `variable_declaration`, which the require-mapping walk never visited (only const/let `lexical_declaration`) -- import visibility was empty for classic CJS files.

## Verification

- Full suite: 4593 passed, 0 failed (91 errors all memgraph-integration, environment-only)
- express 49 -> 1 (the 1 is genuinely dynamic); zustand 0; click 0; gson 22 (55 -> 22 is its `src/test/java` tree now correctly excluded by `/test/`; main-src unchanged)
- ruff clean; ty clean on changed paths

### DIFF
--- a/codebase_rag/constants.py
+++ b/codebase_rag/constants.py
@@ -589,15 +589,18 @@ PROTOCOL_BASE_QNS: tuple[str, ...] = ("typing.Protocol", "typing_extensions.Prot
 
 # (H) Substrings in a node's file path that mark it as test code. Covers Python
 # (H) (test_, _test, conftest, /tests/), the JS/TS filename convention
-# (H) (foo.test.ts, foo.spec.tsx), and the Jest __tests__/ directory so those
-# (H) files are not reported as dead. Singular /test/ and /spec/ directories are
-# (H) intentionally excluded: they collide with product code (a domain "spec"
-# (H) module), which would misclassify live code as test.
+# (H) (foo.test.ts, foo.spec.tsx), the Jest __tests__/ directory, and the
+# (H) Node.js/mocha singular /test/ dir (express: 34 of 49 dead-code reports
+# (H) were test/ helpers). Matching is segment-anchored via the leading-slash
+# (H) normalization, so contest/ and latest/ do not match. Singular /spec/
+# (H) stays excluded: it collides with product code (a domain "spec" module),
+# (H) which would misclassify live code as test.
 TEST_PATH_PATTERNS: tuple[str, ...] = (
     "test_",
     "_test",
     "conftest",
     "/tests/",
+    "/test/",
     ".test.",
     ".spec.",
     "__tests__",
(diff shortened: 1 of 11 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

Issue #46, opened 2025-07-07
Add diffusion models?


vitali87:
## 📋 Task Description
Those models are very fast and might be suitable for cyper generation and other agentic framework requireing no latency.

## 🎯 Acceptance Criteria
- [ ] Fast
- [ ] Cheap
- [ ] Adequate reasoning

## 📝 Additional Context
PydanticAI does not directly support diffusion models but it supports OpenRouter which has Inception Labs Mercury model, so this might work.
HOWEVER, Mercury DOES NOT yet support tool calling as of 23.07.2025 so it is useless to consider them now. Revisit later.

## 🚦 Priority
- [ ] High
- [ ] Medium
- [x] Low

## 📊 Estimated Effort
- [x] Small (< 2 hours)
- [ ] Medium (2-8 hours)
- [ ] Large (> 8 hours)

vitali87:
Update: not closing this but tracking: dLLMs are not yet there

PR #671, opened 2026-07-09
feat(cpp-frontend): emit IMPORTS edges for within-repo includes

vitali87:
## Summary

C++ libclang frontend follow-up (parked Option A from the frontend work, Codeberg #46 series): `#include` is the C++ import, so the frontend now emits `IMPORTS` Module -> Module edges for every within-repo inclusion.

## Implementation

- `_Collector.process_includes(tu)` iterates `tu.get_includes()` after each translation unit's cursor walk: `FileInclusion.source` is the INCLUDING file, so transitive inclusions attribute correctly (calc.h including util.h emits `calc.h IMPORTS util.h`, not an edge from the TU's main file).
- System headers resolve outside the repo (`module_qn` is None) and emit nothing.
- A `source != include` guard keeps the tree-sitter path's known self-import bug out of the frontend.
- Modules discovered only through inclusion chains are ensured (`_add_module`), so header-only modules get their node and CONTAINS_MODULE parent like every other module.

No schema change: the IMPORTS Module -> Module shape is the one the tree-sitter path already emits.

## Verification (RED -> GREEN)

- `test_cpp_frontend_includes.py`: within-repo direct + transitive includes emit IMPORTS with the resolver's qn shapes; no self-imports; no system-header edges. Confirmed failing before the change.
- All 9 cpp_frontend tests pass; full suite 4594 passed, 0 failed (91 errors are memgraph-integration environment-only; the one parallel-run failure is the known `test_rust_containment_oracle` scheduling flake, passes isolated).

### DIFF
--- a/codebase_rag/parsers/cpp_frontend/frontend.py
+++ b/codebase_rag/parsers/cpp_frontend/frontend.py
@@ -268,6 +268,38 @@ class _Collector:
         self._add_module(module_qn, rel, file_name)
         return (fc.LABEL_MODULE, module_qn)
 
+    def process_includes(self, tu) -> None:
+        # (H) `#include` is the C++ import: emit IMPORTS Module -> Module for every
+        # (H) within-repo inclusion, at any depth (calc.h including util.h counts,
+        # (H) attributed to calc.h -- FileInclusion.source is the INCLUDING file).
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #678, opened 2026-07-09
fix(deps): patched grammar forks for comment-in-define declaration drop

vitali87:
## Summary

tree-sitter-c drops real declarations that follow a `#define` whose value embeds a block comment (issue #555, upstream [tree-sitter-c#319](https://github.com/tree-sitter/tree-sitter-c/issues/319)). The trigger is token, comment, token inside a macro value: `preproc_arg` is a single token whose regex stops before `/*`, and the grammar allows only one value token, so the text after the comment has no rule to attach to. With the common bison `do { ... /*YYERROR*/; } while (0)` shape, error recovery turns the trailing `} while (0)` into a real `while_statement` that consumes the next top-level declaration, which is how `yyerror` vanishes from every bison-generated parser. tree-sitter-cpp inherits the identical bug through the shared C grammar.

Upstream review is slow, so cgr now sources both grammars from patched forks pinned by commit in `[tool.uv.sources]`:

* [vitali87/tree-sitter-c@fa327ad](https://github.com/vitali87/tree-sitter-c/commit/fa327ad98f847f8173fc46926be4165c7971f960), branch `fix/preproc-arg-embedded-comment-0.24.1`: the v0.24.1 release plus a one-line-per-site grammar change, `optional($.preproc_arg)` to `repeat($.preproc_arg)` at the three value/argument sites, so comment `extras` can sit between value segments. Regenerated with the same CLI line as the tag (0.25.4) and carrying a new corpus test; all 86 corpus tests pass.
* [vitali87/tree-sitter-cpp@b4a7be2](https://github.com/vitali87/tree-sitter-cpp/commit/b4a7be2f62f36dd460a14b2d1728d42d81d2c486), branch `fix/preproc-arg-embedded-comment-0.23.4`: the v0.23.4 release regenerated against a 0.23.1-based patched C grammar (matching the lockfile pin of the tag), with a corpus test; all 199 corpus tests pass.

Both forks are based on the exact release tags cgr already pinned, not on upstream master, so the only behavioral delta versus the previous wheels is the fix. A first attempt based on master regressed three C++20 module tests because master carries unreleased grammar changes; the tag-based rebuild keeps the full suite green. The overrides are documented in `pyproject.toml` for removal once upstream lands the fix.

## TDD

The first commit adds the failing tests: C and C++ projects containing a bison-style macro with an embedded `/*YYERROR*/` comment followed by `void yyerror(...)`, asserting the function node is ingested. Both fail on the pinned PyPI grammars and pass with the forks.

## Validation

* Full suite: 4711 passed, 10 skipped (environment), zero failures.
* `ruff check` and `ruff format` clean; `ty` at the 352 baseline.
* Real-world check on jq's bison-generated `parser.c`: `yyerror` at line 406 is now a proper `function_definition`; ERROR nodes drop from 20 to 12, and the remainder are attribute-macro constructs that only preprocessing can resolve (the documented libclang frontend territory).
* `evals/README.md` updated where the limitation was documented.

Refs #555

### DIFF
--- a/codebase_rag/tests/test_macro_comment_function_capture.py
+++ b/codebase_rag/tests/test_macro_comment_function_capture.py
@@ -1,0 +1,55 @@
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+from codebase_rag import constants as cs
+from codebase_rag.tests.conftest import get_node_names, run_updater
+
+BISON_STYLE_MACRO = (
+    "#define FAIL(loc, msg)   \\\n"
+    "  do {                   \\\n"
+    "    handle(loc, msg);    \\\n"
+    "    /*YYERROR*/;         \\\n"
+    "  } while (0)\n"
+    "\n"
+    "void yyerror(int* loc, const char* msg) { }\n"
+)
+
+
+@pytest.fixture
+def c_macro_comment_project(temp_repo: Path) -> Path:
+    project_path = temp_repo / "c_macro_comment_project"
+    project_path.mkdir()
+    (project_path / "Makefile").write_text("all:\n\tgcc -o parser parser.c\n")
+    (project_path / "parser.c").write_text(BISON_STYLE_MACRO)
+    return project_path
+
+
+@pytest.fixture
+def cpp_macro_comment_project(temp_repo: Path) -> Path:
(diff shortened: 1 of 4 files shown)


vitali87:
@greptile review

PR #681, opened 2026-07-09
fix(rust): type free-function factory results from recorded return types

vitali87:
## Problem

Two adjacent gaps left open by #679 meant a free-function trait-object factory still did not type its result:

1. A free Rust fn's return type was never recorded in `method_return_types`; only impl methods were (in class ingest). `fn make() -> Box<dyn Svc>` left no entry.
2. `_infer_rust_call_return_type` bailed on single-segment chains, so `let s = make()` could never be typed even with a recorded return type.

With `s` untyped, `s.run()` fell to the name-only trie fallback and its lexicographic tie-break bound an arbitrary same-named impl method (`Alpha.run`), leaving the trait method and the other impls reported dead.

Fixing those exposed a third defect on the cross-module path: `_RS_TYPE_NODE_TYPES` in the call resolver only matched `{Class, Enum, Type}`, so an imported trait-typed receiver (`Svc` is registered as `Interface`) resolved to nothing and `_receiver_type_is_external` suppressed the call edge entirely.

## Fix

- `function_ingest.py`: record a free Rust fn's return type after registration, mirroring the existing C++ block directly above it (`Self` stays unrecorded: no impl target).
- `type_inference.py`: resolve an untyped chain base as a free fn called by bare name (same module first, then a `use`-imported fn through its raw `::` path via `_resolve_rust_import_path`, now parameterized by node type), and stop bailing on single-segment chains.
- `rs/type_inference.py`: gate the binding collector so a single bare identifier (`let f = make;`, a move or fn-pointer binding) is not treated as a call; only an invoked base yields a value of its return type.
- `call_resolver.py`: add `Interface` to `_RS_TYPE_NODE_TYPES` so a trait-typed receiver resolves like any other Rust type that carries methods.

## Validation

RED first (commit `edcc8e1`, all 5 tests failing with the arbitrary `Alpha.run` binding), GREEN in `78a70b8`. Covered shapes: same-module `let s = make()`, `try_make().unwrap()` through Result stripping and identity hops, a `use`-imported factory in another module, dead-code keeping all impls alive, and the fn-pointer non-binding guard. Full suite 4730 passed, ruff and ty clean.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -20,8 +20,11 @@ _QN_SPLIT_CACHE: dict[str, tuple[list[str], int]] = {}
 _CHAIN_OPEN_BRACKETS = "([{"
 _CHAIN_CLOSE_BRACKETS = ")]}"
 # (H) Node labels a Rust receiver type name may resolve to: a struct (Class), an
-# (H) enum, or a type alias -- all can carry impl methods.
-_RS_TYPE_NODE_TYPES = frozenset({NodeType.CLASS, NodeType.ENUM, NodeType.TYPE})
+# (H) enum, a type alias, or a trait (Interface, when the receiver is typed to a
+# (H) `dyn`/`impl` trait or a trait-returning factory) -- all can carry methods.
+_RS_TYPE_NODE_TYPES = frozenset(
+    {NodeType.CLASS, NodeType.ENUM, NodeType.TYPE, NodeType.INTERFACE}
+)
 
 
 def _split_receiver_chain(expr: str) -> list[str]:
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

vitali87:
Codecov note: the 2 uncovered patch lines are the `except ImportError` / `RUST_AVAILABLE = False` grammar guard in the new test file. They execute only in an environment without tree-sitter-rust installed, which CI (and any environment that can run the guarded tests) never is. This is the standard guard pattern used across the repo's grammar-dependent test files (same conclusion as PR #679), so the lines are left as-is rather than mocking the import machinery to fabricate coverage.

Issue #689, opened 2026-07-10
fix(embedder): semantic embedding pass wedges in PyTorch MPS on monorepo-scale batches


vitali87:
Found dogfooding on kubernetes (17,469 Go files, 139k graph nodes).

`cgr start --update-graph` finished all graph passes, then the semantic embedding pass (pass 4) queried 107,696 functions and froze inside a single Metal GPU sync. Process sampling showed the main thread parked in `at::mps::MPSStream::copy_and_sync -> dispatch_sync_with_rethrow -> -[_MTLCommandBuffer waitUntilCompleted]` (~89% of samples), with ~1% CPU, zero qdrant points stored across minutes of observation, and stable RSS (~4 GB). This is a known PyTorch MPS hazard with long-running command buffers on large batches.

Impact: on big repos indexing appears to hang forever after the graph is actually complete; the process had to be killed (the graph itself was intact and usable).

Environment: macOS (Apple Silicon), torch with MPS backend, `_select_device()` in `codebase_rag/embedder.py` picks cuda -> mps -> cpu with no override.

Possible directions: chunk embedding batches with explicit `torch.mps.synchronize()`/`empty_cache()` between them, add a watchdog that falls back to CPU when a batch makes no progress, or cap MPS batch sizes.

Related: the lack of any off switch for the pass is tracked separately.

vitali87:
Validated at the scale that originally wedged: reran the full kubernetes sync on Apple Silicon with the merged fix (v0.0.288). The embedding pass completed all 52,108 embeddings in ~52 minutes with steady qdrant point growth throughout (monitored every 2 minutes, zero stalls) and the process exited cleanly. Previously this run parked permanently in Metal's waitUntilCompleted with zero points stored.

PR #691, opened 2026-07-10
feat(cli): flag and settings to skip the semantic embedding pass

### LINKED ISSUE #690: feat(cli): flag to skip the semantic embedding pass during graph sync
Found dogfooding on kubernetes, alongside #689.

The semantic embedding pass runs whenever `has_semantic_dependencies()` is true (torch + transformers + qdrant importable) — there is no CLI flag, config setting, or environment variable to skip it, and the device choice in `codebase_rag/embedder.py::_select_device` is hardcoded cuda -> mps -> cpu with no escape hatch.

Consequences:

- Any workflow that only needs the graph (dead-code, stats, structural queries) still pays the full embedding cost of the repo — on kubernetes that is 107k functions, and with #689 it never completes at all.
- The only workaround today is uninstalling the semantic extras or killing the process after the graph passes finish.

Proposal: a `--no-embeddings` flag on `cgr start` (and workspace sync), plus a settings/env equivalent, and optionally an embeddings device override (e.g. `CGR_EMBEDDING_DEVICE=cpu`) so MPS-affected machines can still embed.

vitali87:
Closes #690

Adds an off switch for the semantic embedding pass and an escape hatch for the device it runs on:

- `cgr start --no-embeddings` skips pass 4 after graph sync, for both single-repo sync and workspace sync. The graph itself still builds fully.
- `CGR_SKIP_EMBEDDINGS=1` (settings/env equivalent) does the same for env-configured callers; an explicit constructor argument on `GraphUpdater` overrides it in either direction.
- `CGR_EMBEDDING_DEVICE=cpu|cuda|mps` overrides the hardcoded cuda -> mps -> cpu auto-selection in the embedder. An unavailable requested device logs a warning and falls back to auto-selection instead of crashing.
- Device strings are now a `cs.EmbeddingDevice` StrEnum; `get_model` uses one `.to(device)` path instead of separate cuda/mps branches.

Motivation: on monorepo-scale repos the embedding pass dominates sync time, and with #689 it can wedge entirely on Apple Silicon. Workflows that only need the graph (dead-code, stats, structural queries) previously had no way to opt out short of uninstalling the semantic extras.

TDD: the first commit adds the failing tests (flag threading through the CLI, settings fallback and override precedence, env parsing, device override honored and unavailable-device fallback), the second makes them pass.

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -130,6 +130,7 @@ def _sync_workspace(
     config: WorkspaceConfig,
     batch_size: int,
     exclude: list[str] | None,
+    skip_embeddings: bool | None = None,
 ) -> None:
     total = len(config.repos)
     if total == 0:
@@ -162,6 +163,7 @@ def _sync_workspace(
             batch_size=batch_size,
             exclude=exclude,
             interactive_setup=False,
+            skip_embeddings=skip_embeddings,
         )
 
 
@@ -192,6 +194,7 @@ def _run_graph_sync(
     interactive_setup: bool,
     clean: bool = False,
     output: str | None = None,
+    skip_embeddings: bool | None = None,
 ) -> None:
     cgrignore = load_ignore_patterns(repo)
     cli_excludes = frozenset(exclude) if exclude else frozenset()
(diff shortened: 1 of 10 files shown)


vitali87:
@greptile review

PR #707, opened 2026-07-11
feat(flow): add opt-in FLOWS_TO data-flow edges for intra-procedural taint

vitali87:
## What

Adds `FLOWS_TO`, an opt-in relationship type that records intra-procedural value flow (taint/provenance) between resources and callables. Turns questions like "does a value read here reach a write over there?" into graph reachability, which coarse `CALLS` co-occurrence cannot answer.

Stacked on #638 (base branch `feat/relationship-capture-io`): `FLOWS_TO` builds on that PR's `Resource` nodes and I/O source/sink registry, so it must land after #638. When #638 merges this PR auto-retargets to `main`.

## Behavior

`FLOWS_TO` is emitted in three shapes, distinguished by a `kind` edge property (plus a `via` conduit label):

- **resource → resource** (`kind=resource`): a value read from one resource reaches a write to another within a body, e.g. `x = os.getenv("K"); print(x)` yields `Resource(ENV::K) -FLOWS_TO-> Resource(STDOUT)`.
- **caller → callee** (`kind=arg`, `via=arg:<i>` / `kw:<name>`): a tainted local is passed as an argument to a first-party callee.
- **callee → caller** (`kind=return`, `via=return`): a callee whose return value is tainted flows it back to the assignment in its caller.

Taint propagates through plain `x = y` assignments.

## Opt-in / zero default cost

`FLOWS_TO` joins the existing `io` capture group, which is excluded from the default capture set. A default build emits zero `FLOWS_TO` edges and skips the body pass entirely. Enable with `--capture io` (or `CGR_CAPTURE=io`).

## Design

Self-contained `codebase_rag/parsers/flow_access/` submodule mirroring `io_access/`: one scope-pruned body DFS per caller, invoked once beside `process_io_for_caller`, reusing the resolver for callee resolution. Pure extractors shared with `io_access` were promoted to `io_access/extract.py` (no behavior change; the io suite is unchanged). No new node types: `FLOWS_TO` connects existing `Function`/`Method`/`Module`/`Resource` nodes.

Phase-1 scope is deliberately conservative and documented inline: Python-only, single function body plus one level of argument/return hand-off, direct source/sink calls, no SSA/parameter nodes. Upgrade paths are noted at each ceiling.

## Tests

RED → GREEN: failing tests committed first (`2c3f1fe`), then the implementation (`2cc78d2`). Eight `test_flow_edges.py` cases cover each edge shape, taint propagation, the anti-false-positive negative case (co-occurrence is not flow), default-off gating, and the dangling-endpoint guard, plus capture-selection membership. Full suite green (4853 passed); `ty` and `ruff` clean; relationship `properties` confirmed to round-trip the protobuf path.

## Docs

`docs/architecture/graph-schema.md` gains the `Resource` node row, the `READS_FROM`/`WRITES_TO`/`FLOWS_TO` relationship rows, and an I/O and data-flow section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,6 +771,7 @@ The knowledge graph uses the following node types and relationships:
 | Module, Function, Method | INSTANTIATES | Class |
 | Module, Function, Method | READS_FROM | Resource |
 | Module, Function, Method | WRITES_TO | Resource |
+| Module, Function, Method, Resource | FLOWS_TO | Module, Function, Method, Resource |
 <!-- /SECTION:relationship_schemas -->
 
 ## 🔧 Configuration
--- a/README.md
+++ b/README.md
@@ -715,13 +715,13 @@ The knowledge graph uses the following node types and relationships:
 | Folder | `{path: string, name: string, absolute_path: string}` |
 | File | `{path: string, name: string, extension: string?, absolute_path: string}` |
 | Module | `{qualified_name: string, name: string, path: string, absolute_path: string, start_line: int?, end_line: int?}` |
-| Class | `{qualified_name: string, name: string, decorators: list[string], path: string, absolute_path: string, start_line: int?, end_line: int?, docstring: string?, is_exported: boolean?}` |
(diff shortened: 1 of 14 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

Issue #102, opened 2025-08-01
Support C# (fully)


aricwang88:
No C# found in the supported list. Why?

vitali87:
You can add any language grammar yourself including C#. Please follow the instructions in the README.

However, it will not be an official support and will not function fully. The reason is that every language has its own quirks and needs a detailed coding and testing to cover all failure cases.

vitali87:
I might add it later depending on demand...

aricwang88:
Thx

ZlobnyiSerg:
Any progress on this? C# has out-of-the-box Roslyn analyzer that allows to easily "understand" code - may be it's worth to use separate C# server to process C# code and pass to core?

vitali87:
Wow, it seems there is a demand. I will add this next. Stay tuned.

PR #716, opened 2026-07-12
feat(csharp): tree-sitter C# support for definitions, imports, and calls (Phase 1 of #102)

vitali87:
## Summary

First phase of full C# support (issue #102). Registers C# as a tree-sitter language and extracts core structure into the graph. C# is marked **In Development** in the support matrix; it reaches **Fully Supported** at the end of the phased plan below.

Grammar: `tree-sitter-c-sharp` 0.23.5 (latest on PyPI and upstream), added to the `treesitter-full` extra.

## What Phase 1 covers

- **Definitions**: namespaces (block and file-scoped), classes, structs, records (incl. `record struct`/`record class`), interfaces, enums, methods, constructors, destructors, operators, conversion operators, properties, local functions. Interfaces and enums receive their proper `Interface`/`Enum` labels; structs and records map to `Class`.
- **Imports**: every `using` form (plain, `using X = Y` alias, `using static`, `global using`) becomes an IMPORTS edge.
- **Calls**: intra-file and unambiguous cross-file calls resolve via the existing trie and simple-name lookup.
- **Generics**: type parameters do not leak into names, so definition and call FQNs agree.

## FQN design

C# namespaces are lexical and decoupled from directory layout, so the namespace is folded into the qualified name (the C++ model, not the Java path-only model). The load-bearing detail: with a **file-scoped namespace** (`namespace N;`) the grammar makes the type declarations **siblings** of the namespace node, not its children, so a plain ancestor walk would drop the namespace entirely. This is handled by making `compilation_unit` a FQN scope whose resolver folds in the file-scoped namespace; block namespaces remain ordinary ancestors. Both spellings produce identical namespace-qualified FQNs.

## Testing

- New: `test_csharp_definitions.py`, `test_csharp_imports.py`, `test_csharp_calls.py` (7 tests) covering both namespace spellings, all type kinds, generics, nested types, operators, `using` forms, and call resolution.
- Full suite green (4947 passed, 10 skipped); ruff and ty clean.
- Dogfooded on Polly (416 `.cs` files): clean parse with no error forest; 467 Class, 40 Interface, 10 Enum, 2215 Method nodes; 292 IMPORTS, 4629 CALLS; namespace-folded FQNs verified.

## Deferred to later phases (as planned in #102)

- **Phase 2**: INHERITS / IMPLEMENTS / OVERRIDES.
- **Phase 3**: type inference, receiver-typed call resolution, `new X()` to constructor, overload disambiguation, extension methods, partial-class merge.
- **Phase 4**: dead-code roots, stdlib prefixes, exports, flip to Fully Supported.
- **Follow-up**: optional Roslyn semantic frontend (analogous to `CPP_FRONTEND=hybrid`).

Closes part of #102.


vitali87:
@greptile review

Issue #524, opened 2026-03-28
feat: add symbol definition/reference tracking via tags.scm queries


vitali87:
## Background

Tree-sitter's `tags.scm` query files define patterns for identifying symbol definitions (where classes, functions, methods are declared) and references (where they are used). These are the same queries that power GitHub's "Go to Definition" and "Find References" features.

## Problem

Our current definition tracking is built from custom tree-sitter queries in `language_spec.py` and `constants.py`. While functional, this approach:

- Misses some definition types (e.g., TypeScript type aliases, Rust trait implementations, Java interface implementations)
- Doesn't track references at all — we only discover references through call resolution, missing type annotations, variable declarations, and inheritance clauses
- Requires manual maintenance per language when grammars update

The upstream `tags.scm` files are community-maintained and already capture definition/reference pairs we're missing.

## Approach

Vendor `tags.scm` files from official tree-sitter grammar repos for all 12 supported languages. These are already available in the pip packages for 9/12 languages; we supplement the remaining 3 (TypeScript, Scala, C#) from the grammar repos.

### What tags.scm captures

Capture types with semantic meaning:

- `@definition.class` — class/struct/enum declarations
- `@definition.function` — function declarations
- `@definition.method` — method declarations
- `@definition.interface` — interface declarations (Java, TS, Go)
- `@reference.call` — function/method call sites
- `@reference.class` — class references (inheritance, instantiation)
- `@reference.implementation` — interface implementation references
- `@name` — the identifier name at each definition/reference site
- `@doc` — associated documentation comments

Example from Java's tags.scm:
```scheme
(class_declaration
  name: (identifier) @name) @definition.class

(interface_declaration
  name: (identifier) @name) @definition.interface

(type_list
  (type_identifier) @name) @reference.implementation

(superclass (type_identifier) @name) @reference.class
```

### Implementation

1. Create `codebase_rag/queries/tags/` directory with .scm files per language
2. Run tags queries alongside existing structural queries during parsing
3. Cross-validate our definitions against `@definition.*` captures — log warnings for any definitions we miss
4. Use `@reference.*` captures to create new relationship types or supplement existing CALLS/IMPORTS edges
5. Use `@doc` captures to associate documentation comments with their definitions
6. Add tests per language verifying tag capture completeness

## Languages

All 12: Python, JavaScript, TypeScript, Rust, Java, C, C++, Go, Lua, Scala, PHP, C#

## References

[...]

PR #725, opened 2026-07-12
feat(parsers): warn on definition drift via tags cross-check in process_file (#524)

vitali87:
## Why

Follows #717 / #718 / #721. Those built the cross-validation diff and the community `tags.scm` oracle but left them as pure functions, unit-tested in isolation. Nothing ran during a real graph build, so issue #524's actual value (surfacing when cgr's hand-written definition queries drift from the grammar) was not delivered. This wires the definition check into the parse pipeline.

## What

- `DefinitionProcessor.process_file` now calls `_crossvalidate_definitions(...)` just before it returns. It diffs cgr's own function/class query captures (already computed in scope) against the community `tags.scm` oracle and logs genuine drift.
- **Module-level constants are excluded.** The community oracle marks `@definition.constant` for every top-level assignment (`X = 1`), which cgr deliberately does not model as a graph node. Without a filter, `missed` would be non-empty on almost every real file, drowning real signal in known noise. `crossvalidate` gained an `exclude_kinds` parameter (default empty, so existing behavior is unchanged); the pipeline passes `cs.TAGS_UNMODELED_DEF_KINDS = {"constant"}`. With constants filtered, any remaining disagreement is real, so it logs at **warning**.

## Scope

- **Validates the extraction queries, not the emitted graph.** The cgr side is `combined_captures` (cgr's own func/class queries), which is exactly what #524 worries about drifting, and needs no interception of the scattered `ensure_node_batch` emits.
- **Definitions only.** `crossvalidate_calls` (from #721) is not wired here; calls resolve in `call_processor.py`, a separate seam and a separate follow-up.

## Tests (RED first)

- `bef1d31f` commits the tests alone: `test_definition_crosscheck_warns_on_drift_not_constants` hands the diagnostic a capture set missing a real function `g` (so the oracle's `g` is genuine drift) plus a constant `X`, and asserts a warning names `g` but **not** `X`. Fails with `AttributeError` before the implementation (RED, verified with the impl stashed).
- `6f45b4b0` adds the implementation (GREEN). `test_process_file_runs_crosscheck_without_noise` spies the seam to prove `process_file` invokes the check and stays silent on a clean file.

Full suite green (4984 passed, 26 skipped), ruff / ty / no-docs clean.

Closes the "wire the diff into the parse pipeline" item on #524. Two items remain there (per-language parity tests, fixing surfaced `language_spec.py` gaps).

### DIFF
--- a/codebase_rag/constants/ast_nodes.py
+++ b/codebase_rag/constants/ast_nodes.py
@@ -125,6 +125,10 @@ CAPTURE_KEYWORD = "keyword"
 CAPTURE_ATTRIBUTE = "attribute"
 CAPTURE_FUNCTION_DECORATOR = "function.decorator"
 
+# (H) tags.scm @definition.* kinds cgr does not model as graph nodes; excluded from the
+# (H) definition cross-check so they are not reported as false drift (issue #524).
+TAGS_UNMODELED_DEF_KINDS = frozenset({"constant"})
+
 # (H) Modifier extraction
 EXCLUDED_KEYWORDS = frozenset(
     {
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -52,6 +52,7 @@ DEF_PARSING_AST = "Parsing and Caching AST for {language}: {path}"
 DEF_UNSUPPORTED_LANGUAGE = "Unsupported language '{language}' for {path}"
 DEF_NO_PARSER = "No parser available for {language}"
 DEF_PARSE_FAILED = "Failed to parse or ingest {path}: {error}"
+DEF_CROSSVAL = "  tags cross-check {path}: missed={missed} extra={extra}"
 DEF_PARSING_DEPENDENCY = "  Parsing dependency file: {path}"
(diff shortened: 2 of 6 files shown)


vitali87:
Closing unmerged. After wiring the definition cross-check in, review (Greptile P1 + Gemini) surfaced a real name-extraction bug for C/C++, and fixing it exposed a deeper problem: cgr and the community oracle model definitions at different granularities (constants, methods, nesting) and every language's `tags.scm` anchors differently. Making the warning genuinely noise-free would require reconciling all 12 languages by hand.

That cost is not justified. The oracle is strictly weaker than cgr (it marks capture sites; cgr already resolves references to real targets and models INHERITS/IMPLEMENTS/CALLS/etc.), so it can only validate a narrow slice, and on that slice the diff is dominated by known-expected disagreement rather than real drift. cgr's extraction is already guarded by the test suite, the fixture dangling-edge gates, and the parser-fingerprint staleness warning.

The reviewers were right on the specifics; the specifics are precisely why the per-parse warning is not worth carrying. Parking the idea as an occasional audit script instead (see #524).

PR #730, opened 2026-07-12
feat(io): recognize httpx/aiohttp and pathlib.Path I/O in the registry (#723)

### LINKED ISSUE #723: I/O registry misses httpx/aiohttp and pathlib.Path.* (zero recall on modern-idiom repos)
## Summary

The Python I/O registry recognizes `requests.*`, `urllib.request.urlopen`, `open`, `json.load/dump`, `os.getenv`, `print`, `socket.socket`. It does **not** recognize two idioms that are now standard in modern Python codebases:

- **`httpx` / `aiohttp`** network calls (`httpx.Client().get/post`, `client.get(...)`, `session.get(...)`).
- **`pathlib.Path` file I/O** (`Path.read_text`, `read_bytes`, `write_text`, `write_bytes`, `open`).

As a result, repos that use these idioms produce **zero** `READS_FROM` / `WRITES_TO` / `FLOWS_TO` edges despite heavy I/O.

## Repro (real, dogfooded)

`suno_ai_cli` indexed with `--capture io` → **0 READS_FROM, 0 WRITES_TO, 0 FLOWS_TO**, even though it does substantial network + file I/O:

```python
# store/store.py
self._store_path.write_text(store.model_dump_json(indent=2), encoding="utf-8")  # FILE write, missed
...
with httpx.Client(timeout=60.0, follow_redirects=True) as client:
    resp = client.get(clip.audio_url)                                            # NETWORK read, missed
    target.write_bytes(resp.content)                                             # FILE write, missed

# suno/api_backend.py
self._http = httpx.Client(timeout=30.0)
resp = self._http.post(...)                                                      # NETWORK write, missed
```

Contrast: `genai-stack` (uses `requests.get` + `os.getenv`) and `llm-shell` (uses `open` + `requests.post`) both produce correct edges. The gap is specifically the unregistered libraries/idioms, not the flow engine.

## Registry locations

- `codebase_rag/parsers/io_access/registry.py` — sinks (`requests.*`, `urllib...`, `print`, `json.*`) and handle constructors (`open`, `socket.socket`).

## Suggested direction (not prescriptive)

1. **`httpx` / `aiohttp`**: register `httpx.get/post/put/patch/delete/head` and the client-method forms (`httpx.Client(...).get`, `AsyncClient`, `aiohttp.ClientSession().get`) as NETWORK sinks, mirroring the existing `requests.*` rows. Direction mapping same as requests (GET/HEAD = read, POST/PUT/PATCH/DELETE = write). Client-instance methods flow through the existing cross-scope handle resolution.
2. **`pathlib.Path`**: register `Path.read_text`/`read_bytes` (FILE read) and `write_text`/`write_bytes` (FILE write). These are method calls on a `Path` handle; the identity is the path literal where statically known, else `<dynamic>` (same as `open`).

## Notes

This is a recall/coverage gap on standard Python, distinct from the non-Python tracking in #714. Modern Python code frequently uses `httpx` over `requests` and `pathlib` over `open`, so this materially affects real-repo recall.

vitali87:
## Summary

Fixes #723. The Python I/O registry recognized `requests.*`, `urllib`, `open`, `json`, `os.getenv`, `print`, `socket`, but **not** two idioms standard in modern Python, so repos using them produced **zero** `READS_FROM` / `WRITES_TO` / `FLOWS_TO` edges:

- **`httpx` / `aiohttp`** network calls.
- **`pathlib.Path`** file I/O.

## Change (registry only)

`codebase_rag/parsers/io_access/registry.py`:

1. **httpx module-level sinks** — `httpx.get/head/post/put/patch/delete` mirror `requests.*` (GET/HEAD read, rest write, URL from arg 0 / `url=`). These flow through both the io and flow processors, so they also produce `FLOWS_TO`.
2. **NETWORK handle constructors + methods** — `httpx.Client`, `httpx.AsyncClient`, `aiohttp.ClientSession` register as NETWORK handles; a new `ResourceKind.NETWORK` method table maps `get/head/options` → read and `post/put/patch/delete` → write. Client-instance calls (`client.get(...)`, `self._http.post(...)`) resolve through the existing cross-scope handle resolution.
3. **pathlib.Path handle** — `pathlib.Path` registers as a FILE handle (identity = the path literal); `read_text/read_bytes` → read and `write_text/write_bytes` → write join the FILE method table. Non-I/O `Path` methods (`.exists`, `.parent`) aren't in the table, so no false edges.

Import normalization already covers both `import httpx` / `from httpx import Client` and `import pathlib` / `from pathlib import Path`, so a single canonical registry key handles each idiom.

## Test (RED → GREEN)

New integration test `test_io_registry_modern_idioms_e2e.py`: httpx module-level get/post, httpx client-instance methods (via a `self._http` handle), and pathlib read/write. Confirmed RED before the registry rows (0 edges) and GREEN after.

## Known ceiling

Handle-method calls (`client.get`, `path.write_text`) are recognized by the **io** processor (READS_FROM/WRITES_TO). The **flow** processor models taint only through direct sinks, so direct `httpx.get/post` participate in `FLOWS_TO` but handle-method network/file taint does not yet. Noted in code; a separate flow-engine change if it matters.

Full suite: 5024 passed, 10 skipped. ruff + ty clean.

### DIFF
--- a/codebase_rag/parsers/io_access/registry.py
+++ b/codebase_rag/parsers/io_access/registry.py
@@ -77,6 +77,49 @@ _PYTHON_SINKS: tuple[IOSink, ...] = (
         target_arg=0,
         target_kw="url",
     ),
+    # (H) httpx module-level calls mirror requests.*: GET/HEAD read, the rest write.
+    IOSink(
+        "httpx.get",
+        ResourceKind.NETWORK,
+        IODirection.READ,
+        target_arg=0,
+        target_kw="url",
+    ),
+    IOSink(
+        "httpx.head",
+        ResourceKind.NETWORK,
+        IODirection.READ,
+        target_arg=0,
+        target_kw="url",
+    ),
+    IOSink(
+        "httpx.post",
+        ResourceKind.NETWORK,
+        IODirection.WRITE,
+        target_arg=0,
+        target_kw="url",
+    ),
+    IOSink(
+        "httpx.put",
+        ResourceKind.NETWORK,
+        IODirection.WRITE,
+        target_arg=0,
+        target_kw="url",
+    ),
+    IOSink(
+        "httpx.patch",
+        ResourceKind.NETWORK,
+        IODirection.WRITE,
+        target_arg=0,
+        target_kw="url",
+    ),
+    IOSink(
+        "httpx.delete",
+        ResourceKind.NETWORK,
+        IODirection.WRITE,
+        target_arg=0,
+        target_kw="url",
+    ),
 )
 
 IO_SINKS: dict[cs.SupportedLanguage, tuple[IOSink, ...]] = {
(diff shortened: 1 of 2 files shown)


vitali87:
Addressed both suggestions. Added `aiohttp.request` (target_arg=1) + a generic network `request` method (both READ_WRITE, since direction depends on the HTTP-verb arg), and the extra pathlib methods (`iterdir`/`glob`/`rglob`→read, `touch`/`unlink`→write). One deviation: I mapped `Path.open` to **READ** rather than READ_WRITE — the handle-method path can't inspect the mode arg, so I defaulted to read to match the existing top-level `open()` sink's no-mode default; emitting a false WRITES_TO on the common read case would hurt flow precision more than a coarse READ.

vitali87:
@greptile review

Issue #712, opened 2026-07-12
FLOWS_TO: cross-file/forward return-taint is missed by the single-pass ordering


vitali87:
## Summary

`FLOWS_TO` return-taint propagation is a single source-order pass, so a callee whose body is processed **after** its caller is not yet known to return a tainted value at the caller's site. The caller then misses the return edge and the downstream resource flow. In practice this happens across files (and within a file when a helper is defined below its user).

Found while pushing I/O/data-flow recall on real repos.

## Root cause (grounded)

- `FlowProcessor._returns_taint` (`codebase_rag/parsers/flow_access/processor.py:101`) maps `callee_qn -> frozenset` of the source bindings that callee can return. It is populated **at the end of walking each caller body**: `self._returns_taint[caller_qn] = frozenset(...)` (`processor.py:170`).
- When a caller `C` contains `v = D()` (`_apply_assignment`, `processor.py:~199`) or `return D()` (`_return_taint_source`, `processor.py:~258`), the code checks `callee[1] in self._returns_taint`. If `D`'s body has not been walked yet, `D` is absent, so no return edge or return-taint is propagated.
- This is the same single-pass, order-dependent limitation as `CallProcessor._returned_callables`, and is called out in the in-code note at `processor.py:98-100`.

## Reproduction

```python
# a.py
from .b import build
def caller():
    v = build()       # build() defined in b.py, processed later
    print(v)          # expect ENV::K -> STDOUT resource flow + build->caller return edge

# b.py
import os
def build():
    return os.getenv("K")
```

If `a.py` is processed before `b.py`, `build` is not yet in `_returns_taint` when `caller` is walked, so the return edge (`b.build -FLOWS_TO-> a.caller`) and the `ENV::K -> STDOUT` resource flow are both dropped.

## Why the naive fix is wrong

Re-walking every caller after the pass would need the caller AST nodes held live across the whole calls pass. The AST cache (`BoundedASTCache`) evicts trees, so buffering `Node` objects is unsafe — exactly why the current design emits during the per-file walk while the tree is live.

## Suggested direction

A fixpoint over **serialized per-function facts**, not buffered nodes:

1. First pass (or fold into the definition pass): for each function, record a compact return-taint summary — `returns_tainted: bool`, `origins: set[ResourceKind/identity]`, and `returned_callee_qns` (callees whose return it forwards, e.g. `return helper()`).
[...]

PR #734, opened 2026-07-13
feat(flow): path-sensitive MAY taint analysis for FLOWS_TO (#713)

### LINKED ISSUE #713: FLOWS_TO/READS_FROM: path-insensitive walk drops taint killed on only one branch
## Summary

Data-flow taint (`FLOWS_TO`) and I/O handle resolution (`READS_FROM`/`WRITES_TO`) are computed by a single flat, source-order walk with one live taint/handle map. There is no notion of control-flow branches, so a reassignment on **one** branch of an `if`/`else` kills taint globally, and taint introduced on one branch is treated as unconditional. Taint should be a MAY analysis (union across branches): keep taint that survives on **any** path, kill only when killed on **all** paths.

This causes both under-reporting (a real flow on the not-taken branch is dropped) and imprecise attribution.

## Root cause (grounded)

- `FlowProcessor.process_flow_for_caller` walks the body as a flat pre-order DFS with a single `tainted: dict[str, frozenset[HandleBinding]]` map (`codebase_rag/parsers/flow_access/processor.py:141-167`). Branch structure (`if`/`elif`/`else`, `try`/`except`, `while`) is not modeled — nodes are visited in raw source order.
- `_apply_assignment` (`processor.py:172-204`) unconditionally retaints or **kills** (`tainted.pop(lhs, None)`) the target, regardless of the branch it sits in.
- `IOAccessProcessor.process_io_for_caller` has the same source-order, non-path-sensitive handle map (`codebase_rag/parsers/io_access/processor.py:74-79`, 92-104).
- The current behavior is noted in-code at `flow_access/processor.py:138-140` ("a KILL on one branch drops taint conservatively").

## Reproduction

```python
import os
def f(cond):
    x = os.getenv("K")   # tainted
    if cond:
        x = "safe"       # kill on ONE branch only
    print(x)             # real flow on the else-path: ENV::K -> STDOUT is MISSED
```

Source-order walk sees `x = getenv`, then `x = "safe"` (kill), then `print(x)` → `x` untainted → no `ENV::K -> STDOUT` edge, even though the else-path leaks.

## Suggested direction

Model taint as a MAY (union) analysis over the block structure:

- Build (or traverse) a lightweight CFG / block tree for the body.
- Compute taint per block and **union at merge points**; a variable is killed only if reassigned to an untainted value on **every** incoming path.
- Handle resolution (io_access) should follow the same merge semantics.

This is inherently more work than the flat walk and needs care to avoid false positives (e.g. loop back-edges). A regression suite should cover: kill-on-one-branch (must still flow), taint-on-one-branch (must flow), and kill-on-all-branches (must NOT flow).

vitali87:
## Summary

Fixes #713. `FLOWS_TO` taint was computed by a flat, source-order walk with one live taint map, so a reassignment on **one** branch of an `if`/`else` killed taint globally — dropping a real flow on the not-taken path.

```python
import os
def f(cond):
    x = os.getenv("K")   # tainted
    if cond:
        x = "safe"       # kill on ONE branch only
    print(x)             # else-path leak ENV::K -> STDOUT was MISSED
```

## Change

`FlowProcessor.process_flow_for_caller` is now a **path-sensitive MAY analysis**. Taint is still threaded statement by statement, but each `if`/`elif`/`else`, `try`/`except`/`else`/`finally`, and `for`/`while` is evaluated **per branch against a copy of the incoming state** and **unioned at the merge point** (`_merge`):

- A variable is tainted after a join if it is tainted on **any** incoming path (origins unioned).
- A variable is killed only when reassigned to an untainted value on **every** path.
- No `else` → the implicit skip path preserves the incoming state.
- Loops evaluate the body twice (skip ∪ body, then re-walk from that merge) so loop-carried taint reaching an earlier statement is caught. Edges are `MERGE`-idempotent, so the re-walk never duplicates graph edges.
- `except` handlers are seeded with `union(pre, try-body-exit)`; `finally` applies to the merged state (runs on all paths).

Straight-line code takes a single path with nothing to merge, so it behaves exactly as the old flat walk — every existing flow/io test is unchanged.

**Scope:** this is flow_access (`FLOWS_TO`) only. io_access's handle map is **add-only** within a body (it never pops a handle on reassignment), so it has no kill-on-one-branch false negative — no change needed there.

## Test (RED → GREEN)

New `test_path_sensitive_flow_e2e.py` covers the issue's required matrix plus the novel loop/try logic:
- kill-on-one-branch → **must still flow** (was RED, now GREEN)
- taint-on-one-branch → must flow
- kill-on-all-branches → must **not** flow
- loop-carried taint reaches an earlier statement
- taint survives an except-only kill

Full suite: 5044 passed, 10 skipped. ruff + ty clean.

## Known ceilings (noted in code)

- Loops use a two-pass approximation, not a full fixpoint (upgrade path noted).
- `elif`/`while` conditions are processed in the incoming state (conditions rarely mutate taint).
- The single-pass forward return-taint summary (a callee defined after its caller) is unchanged — that ordering limitation is #712.


vitali87:
Thanks — I checked this empirically before applying, and `child_by_field_name(cs.FIELD_BODY)` returns **None** for both `except_clause` and `finally_clause`: unlike `else_clause` (whose block *is* under a `body` field), the block child of an except/finally clause is **unnamed** in tree-sitter-python. Verified:

```
except_clause:  child_by_field_name('body')=None, by-type-search=BLOCK
finally_clause: child_by_field_name('body')=None, by-type-search=BLOCK
```

So switching to the field lookup would make both handlers skip their block entirely (no taint processing in except/finally). Keeping the by-type block search for those two clauses is required; the `FIELD_BODY` lookup is used only where the field actually exists (`else_clause`, loop/if bodies). Leaving as-is.

PR #735, opened 2026-07-13
feat(flow): resolve forward/cross-file return-taint via a fixpoint over summaries (#712)

### LINKED ISSUE #712: FLOWS_TO: cross-file/forward return-taint is missed by the single-pass ordering

vitali87:
## Summary

Fixes #712. `FLOWS_TO` return-taint was a single source-order pass: `_returns_taint[callee]` was only known once the callee's body had been walked, so a caller processed **before** its callee (a helper defined below its user, or in a later-processed file) missed the return edge **and** the resource flow the returned value carries.

```python
def caller():
    v = build()   # build() defined/processed later
    print(v)      # build -> caller return edge AND ENV::K -> STDOUT were dropped
def build():
    return os.getenv("K")
```

## Change (serialized summaries + fixpoint, no AST re-walk)

Taint now carries **resolved origins** *and* **pending callee QNs** (`Taint(origins, pending)`). During the walk, `v = D()` / `return D()` marks the value *pending* on `D` and records a deferred fact instead of requiring `D`'s summary now. Nothing but plain data is buffered — never a tree-sitter `Node` (the AST cache evicts trees, which is why buffering nodes is unsafe).

A new `FlowProcessor.finalize()` runs once after every function is walked (wired next to the existing `finalize_callable_param_flow` in Pass 3):

1. **Fixpoint** over the per-function summaries — a function's resolved origins are its own plus every pending callee's, and it is *tainted* iff it has an origin or a pending callee is tainted. Monotonic, so it converges through recursion.
2. Drain the deferred facts: emit a **return edge** for each candidate whose callee resolves to a tainted return; emit **resource flows** `origin -> sink` for each pending callee's resolved origins; emit **arg edges** only when a pending callee resolves to tainted.

Directly-tainted values (from read sinks) are still emitted inline. All emitted edges are `MERGE`-idempotent, so the backward case (callee already known) produces identical edges whether emitted inline or at finalize — every existing flow/io test is unchanged.

**Precision:** an untainted forward callee (`v = compute(); sink(v)` where `compute` returns nothing tainted) resolves to *not tainted* at finalize, so **no** return edge and **no** arg edge are fabricated.

## Test (RED → GREEN)

New `test_forward_return_taint_e2e.py`:
- forward same-file: caller defined before build → return edge + resource flow (was RED)
- transitive forward: top→mid→low, each defined before its callee → both return edges + resource flow
- untainted forward callee → **zero** FLOWS_TO edges (precision guard)

Full suite: 5061 passed, 10 skipped. ruff + ty clean.

Builds on #713 (path-sensitive walk); the two compose — the fixpoint resolves pending taint that the MAY walk threads through branches.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1558,6 +1558,7 @@ class GraphUpdater:
                 func_class_captures_cache=captures_cache,
             )
         self.factory.call_processor.finalize_callable_param_flow()
+        self.factory.call_processor.finalize_flow()
 
     def _prune_orphan_nodes(self) -> None:
         """Remove graph nodes whose files/folders no longer exist on disk."""
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -3566,6 +3566,12 @@ class CallProcessor:
                     )
                 )
 
+    def finalize_flow(self) -> None:
+        # (H) Resolve deferred FLOWS_TO return-taint once every function body has
+        # (H) been walked, so a callee processed after its caller still contributes
+        # (H) its return edge and resource flow (issue #712).
+        self._flow_processor.finalize()
+
     def finalize_callable_param_flow(self) -> None:
         # (H) Resolve the recorded call-site argument bindings to a fixpoint and emit a
         # (H) CALLS edge from every function that invokes a callable parameter to each
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections import defaultdict, deque
 from typing import NamedTuple
 
 from tree_sitter import Node
@@ -472,9 +473,13 @@ class FlowProcessor:
             if is_tainted.get(callee_qn):
                 self._emit_return_edge((callee_type, callee_qn), caller_spec)
         for pending, sink_kind, sink_identity in self._deferred_resource_flows:
+            # (H) Union origins across pending callees first so a shared origin is
+            # (H) emitted once, not per callee.
+            origins: set[HandleBinding] = set()
             for callee_qn in pending:
-                for origin in resolved.get(callee_qn, frozenset()):
-                    self._emit_resource_flow(origin, sink_kind, sink_identity)
+                origins |= resolved.get(callee_qn, frozenset())
+            for origin in origins:
+                self._emit_resource_flow(origin, sink_kind, sink_identity)
         for (
             pending,
             caller_spec,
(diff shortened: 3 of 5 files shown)


vitali87:
Both applied in 187c223c: replaced the naive nested-loop fixpoint with a worklist over a reverse-dependency graph (callee→callers), re-queuing only a callee's callers when its resolved origins/taint change — O(V+E). Also deduped the deferred resource-flow emission by unioning origins across pending callees before emitting. Full suite green.

vitali87:
@greptile review

Issue #714, opened 2026-07-12
READS_FROM/WRITES_TO/FLOWS_TO: extend I/O and data-flow edges beyond Python


vitali87:
## Summary

`READS_FROM` / `WRITES_TO` / `FLOWS_TO` are emitted for **Python only**. Every other indexed language parses into the graph with full structure/calls but produces zero I/O or data-flow edges, because the source/sink/handle registry and the processors' language gate are Python-specific.

## Root cause (grounded)

- Both processors early-return for non-Python:
  - `IOAccessProcessor.process_io_for_caller` — `if language != cs.SupportedLanguage.PYTHON: return` (`codebase_rag/parsers/io_access/processor.py:64`).
  - `FlowProcessor.process_flow_for_caller` — same gate (`codebase_rag/parsers/flow_access/processor.py:116`).
- The registries have only Python entries: `IO_SINKS`, `IO_HANDLE_CONSTRUCTORS`, `IO_HANDLE_METHODS` are keyed `{SupportedLanguage.PYTHON: (...)}` (`codebase_rag/parsers/io_access/registry.py`).
- The AST-node handling (assignment / call / return / handle-method detection, e.g. `_binding_from_node`, `_arg_names`, `_return_value_nodes`) is written against Python tree-sitter node types (`TS_PY_*`).

## Scope

Add per-language source/sink/handle tables plus per-language node-type handling. Rough per-language surface:

- **Go** — `os.Getenv`/`os.Open`/`os.Create`, `net/http` (`http.Get`, `http.Post`), `database/sql` (`db.Query`, `db.Exec`), `fmt.Println`.
- **JavaScript / TypeScript** — `fs.readFile`/`writeFile`, `fetch`/`axios`, `process.env`, `console.log`.
- **Java** — `java.io`/`java.nio` streams, `java.sql` (`Statement.execute*`), `System.getenv`, `System.out.print*`.
- **Rust** — `std::fs`, `std::net`, `std::env::var`, `println!`.
- **C++** — `fstream`, sockets, `getenv`, `std::cout`.

Each language also needs its own node-type constants for assignment/call/return and handle-method receiver parsing, mirroring the Python paths.

## Suggested direction

- Generalize the registry maps to `{SupportedLanguage: (...)}` (the dict shape already exists) and populate per language.
[...]

PR #736, opened 2026-07-13
feat(io): READS_FROM/WRITES_TO for JavaScript/TypeScript I/O sinks (#714)

vitali87:
## Summary

First increment of #714 (I/O + data-flow beyond Python). Until now `READS_FROM`/`WRITES_TO`/`FLOWS_TO` were emitted for **Python only** — every other language parsed into the graph but produced zero I/O edges. This lands **JavaScript / TypeScript / TSX** direct-call I/O sinks behind the existing opt-in `io` capture group.

## What's covered

`console.log/info/warn/error` → STDOUT write; `fetch`, `axios.get/head/post/put/patch/delete`, `http.get`/`https.get` → NETWORK (GET/HEAD read, rest write, url = arg 0); `fs.readFile[Sync]` → FILE read, `fs.writeFile[Sync]`/`fs.appendFile[Sync]` → FILE write (path = arg 0).

```js
function leak(data) {
  console.log(data);                        // WRITES_TO STDOUT
  fetch("https://api.example.com/data");    // READS_FROM NETWORK::…/data
  fs.writeFileSync("out.txt", data);        // WRITES_TO FILE::out.txt
  axios.post("https://api.example.com/up");  // WRITES_TO NETWORK::…/up
}
```

## Design

The Python and JS/TS grammars share the **field** names the walk needs (`function`/`arguments`/`body`); only the **node types** differ (`call_expression` vs `call`, `string_fragment` vs `string_content`). So instead of rewriting the processor, a small **`LanguageDescriptor`** (`io_access/descriptor.py`) captures just those differing node types, and:

- `string_literal` / `literal_target` (`extract.py`) take optional per-language node-type args (Python defaults unchanged — every existing Python path is byte-for-byte the same).
- The processor's hard `if language != PYTHON: return` gate is replaced with a lean `_emit_direct_sinks` walk for languages that have a descriptor: DFS the caller body for call sinks, pruning nested function/arrow/method definitions (their body is a separate caller). Python keeps its full handle-aware walk untouched.

This gives each future language a bounded landing: one descriptor entry + registry rows + tests.

## Scope / follow-ups (noted in code)

- **io only** — JS/TS `FLOWS_TO` (taint) still gates on Python; data-flow for JS is a separate increment.
- **call sinks only** — `process.env.X` (member-access ENV read) and stream handles (`fs.createReadStream`, `http` client objects) are not yet covered.
- Other languages (Go, Java, Rust, C++) remain follow-ups under #714.

## Test (RED → GREEN)

New `test_jsts_io_e2e.py`: a `.js` and a `.ts` project each asserting the STDOUT/NETWORK/FILE edges appear (both RED before, GREEN after).

Full suite: 5067 passed, 10 skipped. ruff + ty clean. Every existing Python io/flow test unchanged.

### DIFF
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -169,6 +169,8 @@ JS_ASSIGNMENT_FUNCTION_QUERY = """
 
 # (H) JS/TS module system node types
 TS_OBJECT_PATTERN = "object_pattern"
+TS_ARRAY_PATTERN = "array_pattern"
+TS_REST_PATTERN = "rest_pattern"
 TS_SHORTHAND_PROPERTY_IDENTIFIER_PATTERN = "shorthand_property_identifier_pattern"
 TS_PAIR_PATTERN = "pair_pattern"
 TS_FUNCTION_DECLARATION = "function_declaration"
--- a/codebase_rag/constants/ast_nodes.py
+++ b/codebase_rag/constants/ast_nodes.py
@@ -167,6 +167,9 @@ TS_STRING = "string"
 # (H) JS/TS string literals hold their text in a string_fragment child (the
 # (H) counterpart of Python's string_content), used for I/O target extraction.
 TS_STRING_FRAGMENT = "string_fragment"
+# (H) Modern Node builtin imports carry a node: scheme (`import fs from 'node:fs'`);
+# (H) stripped when checking whether an imported name is the genuine builtin module.
+NODE_BUILTIN_PREFIX = "node:"
 TS_IMPORT_CLAUSE = "import_clause"
 TS_LEXICAL_DECLARATION = "lexical_declaration"
 TS_VARIABLE_DECLARATION = "variable_declaration"
--- a/codebase_rag/constants/ast_nodes.py
+++ b/codebase_rag/constants/ast_nodes.py
@@ -164,6 +164,9 @@ TS_WILDCARD_IMPORT = "wildcard_import"
 
 # (H) Tree-sitter JS/TS import node types
 TS_STRING = "string"
+# (H) JS/TS string literals hold their text in a string_fragment child (the
+# (H) counterpart of Python's string_content), used for I/O target extraction.
+TS_STRING_FRAGMENT = "string_fragment"
 TS_IMPORT_CLAUSE = "import_clause"
 TS_LEXICAL_DECLARATION = "lexical_declaration"
 TS_VARIABLE_DECLARATION = "variable_declaration"
--- a/codebase_rag/parsers/io_access/descriptor.py
+++ b/codebase_rag/parsers/io_access/descriptor.py
@@ -26,6 +26,9 @@ class LanguageDescriptor:
     identifier_type: str
     declarator_type: str
     params_field: str
+    # (H) A nested block introduces a new lexical scope: const/let declared inside
+    # (H) it do not shadow the enclosing scope, so declarator collection stops here.
+    block_scope_type: str
 
 
 _JS_TS_DESCRIPTOR = LanguageDescriptor(
@@ -45,6 +48,7 @@ _JS_TS_DESCRIPTOR = LanguageDescriptor(
     identifier_type=cs.TS_PY_IDENTIFIER,
     declarator_type=cs.TS_VARIABLE_DECLARATOR,
     params_field=cs.TS_FIELD_PARAMETERS,
+    block_scope_type=cs.TS_STATEMENT_BLOCK,
 )
 
 # (H) Non-Python languages with a direct-sink descriptor. Python keeps its own
--- a/codebase_rag/parsers/io_access/descriptor.py
+++ b/codebase_rag/parsers/io_access/descriptor.py
@@ -21,6 +21,11 @@ class LanguageDescriptor:
     # (H) Nested definitions whose body is a separate caller: the walk prunes them
     # (H) so a nested function's I/O is not credited to the enclosing one.
     nested_scope_types: frozenset[str]
+    # (H) Local-binding detection so a name declared in the caller scope (a local
+    # (H) `const fs`, `function fetch`, or a parameter) shadows the builtin sink.
+    identifier_type: str
+    declarator_type: str
+    params_field: str
 
 
 _JS_TS_DESCRIPTOR = LanguageDescriptor(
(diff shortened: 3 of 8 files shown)


vitali87:
Addressed in eeae829e:
- **P1 module-scope sinks** — fixed: the walk now seeds from the caller's own `named_children` when there's no `body` field (the module root), so top-level `console.log`/`fetch`/`fs.*` are captured. Added a RED→GREEN `test_module_level_js_io_sinks`.
- **named_children optimization** — applied throughout the walk (skips the anonymous punctuation JS/TS emit in bulk).
- **TS test parity** — added the `axios.post` NETWORK-write assertion to the TS test.

One deliberate deviation: I kept **full pruning** of nested definitions rather than traversing their headers/default-args. Python evaluates default args and decorators at *definition* time in the enclosing scope (hence its header traversal), but **JS/TS default arguments evaluate at call time in the nested function's own scope** — crediting a `function f(x = fetch(u)) {}` to the enclosing caller would be a false positive. Decorators are the only definition-time part and I/O in a decorator is vanishingly rare; I've left that as a noted follow-up rather than add header-walking that would misattribute the common default-arg case.

vitali87:
@greptile review

vitali87:
Fixed in the latest commit (RED→GREEN). Added a `_head_is_shadowed` guard for JS/TS dotted sinks: a real builtin import maps the head to itself (`fs` → `fs` for require, `fs.default` for ESM), but `import fs from './fake'` maps it to a project-qualified base, so the raw-dotted registry fallback is suppressed when the head resolves to a different module. Verified with two tests: `import fs from './fake'` emits **no** FILE edge, while the genuine `import fs from 'fs'` still does. A bare/unimported head (e.g. global `console`) is trusted.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). Extended the shadow check to the caller's own scope: `_local_bindings` collects the caller's parameters, local `const/let/var` declarators, and hoisted `function` declarations (nested scopes pruned), and `_is_shadowed` now rejects a sink whose head (dotted) or bare name is bound locally, in addition to the import-shadow check. New test covers a local `const fs`, a local `function fetch`, and an `http` parameter — none emit edges, while genuine builtins still do.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_is_shadowed` now strips the `node:` scheme from the imported module base before comparing, so `import fs from 'node:fs'` (and node:http / node:https) is recognized as the genuine builtin and still emits its edge. Added a test asserting `node:fs` → `fs.writeFileSync` emits the FILE write. (Named-import forms like `import { writeFileSync } from 'node:fs'` remain a separate coverage follow-up — the registry keys on the `module.method` dotted form.)

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_local_bindings` now excludes names present in `import_map`: a `const fs = require('fs')` (even function-local) is an import alias tracked by the import processor, not a shadowing local, so it's removed from the local-shadow set and correctly resolved by the import branch of `_is_shadowed`. Added a test with a function-local `require('fs')` that must still emit the FILE write. A genuine local `const fs = {}` (not in import_map) still shadows.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). Restructured matching into `_resolve_sink`: import-**normalised** names are matched first (so `const myfs = require('fs')` → `myfs.writeFileSync` resolves to `fs.writeFileSync` and emits), and the raw-dotted fallback is the only path gated by the builtin check. So aliased builtins match, genuine builtins (incl `node:`, require, ESM) match, and shadowing locals / local modules are rejected. Added a test for `const myfs = require('fs')`.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_local_bindings` no longer descends into nested `{ }` blocks, which are their own lexical scope: a `const fs` inside a nested block no longer shadows a `fs.writeFileSync` outside it. Top-level declarations, parameters, and hoisted function declarations still shadow. Added a test for the block-scoped case.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN) with a proper lexical-scope walk. Instead of one flat per-caller name set, `_walk_scope` recurses block by block: each call is checked against the names in scope for ITS block (enclosing scopes + this block's own const/let/function declarations). So a block-local `const fs` shadows within its own block and nested blocks, but never a sibling/outer use. Both directions are now covered by tests: block-local `const fs` used inside its block → no edge; used outside that block → edge stands. Parameters remain function-wide; import aliases still resolve as the genuine module.

vitali87:
@greptile review

vitali87:
This one is a documented ceiling rather than a code defect. The trigger — `function f(){ if(ok){ const fs = require('fs'); } fs.writeFileSync('out', d); }` — is a runtime **ReferenceError**: `fs` is block-scoped to the `if`, so the outer `fs.writeFileSync` references an undefined name (dead/broken code), where I/O-edge precision isn't meaningful. The underlying cause is that `import_map` is module-scoped (the import processor records `const fs = require(...)` module-wide regardless of the block it sits in), so this is a limitation of the import layer, not the shadow walk. Properly modeling it needs per-block import tracking, which is disproportionate for unreachable code — noted as a ponytail ceiling in the walk with the upgrade path. All realistic shadowing cases (module/function/param/block-local declarations, import aliases, node:/require/ESM builtins, both block-scope directions) are handled and tested.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_declarator_names` now extracts the leaf bindings of a destructuring declaration — `const { writeFileSync } = …`, `const { get: g } = …` (binds the value `g`, not the key), `const [fetch] = …`, and rest patterns — so a destructured local shadows a same-named builtin. Added a test: `const { fetch } = obj; fetch(...)` emits no NETWORK edge. (Note: destructured builtin *imports* like `const { readFile } = require('fs')` are a separate recall gap — the registry keys on the `module.method` dotted form — tracked as a follow-up, not a false positive.)

vitali87:
Also fixed the destructured-parameter case from this thread: `_param_names` now runs each parameter (unwrapping the TS `required_parameter` wrapper) through the same pattern extractor, so `function run({ http }) { http.get(...) }` binds `http` locally and emits no NETWORK edge. Both the declarator and parameter destructuring examples are now covered by tests.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_emit_direct_sinks` now handles three body shapes: a statement-block body (walk its statements), the module root (walk its own children), and an **expression-bodied arrow** (`() => fetch(url)`) where the body IS the expression — walked directly. The scope walk was generalized to take a statement list so all three feed the same lexical-scope logic. Added a test: `const load = () => fetch(...)` now emits the NETWORK read.

vitali87:
@greptile review

PR #739, opened 2026-07-13
feat(io): process.env.X environment reads for JavaScript/TypeScript (#714)

vitali87:
## Summary

Follow-up to #714 (JS/TS I/O). The first JS/TS increment (#736) covered **call** sinks; this adds the other idiomatic JS read: **`process.env.X`** environment access, which is a `member_expression` / `subscript_expression`, not a call.

```js
function boot() {
  const a = process.env.SECRET;      // READS_FROM ENV::SECRET
  const b = process.env['TOKEN'];    // READS_FROM ENV::TOKEN  (subscript)
  fetch(process.env.URL);            // READS_FROM ENV::URL  (in an argument)
}
```

## Design

- A small per-language member-read registry `IO_MEMBER_READS` maps an object prefix to a `ResourceKind` — for JS/TS: `process.env → ENV`. Each future language/prefix is one row.
- The `LanguageDescriptor` gains the member/subscript node types + `object`/`property` fields.
- The scope-aware JS walk now also inspects `member_expression` / `subscript_expression` nodes: when the object text matches a prefix, it emits a `READS_FROM` to `ENV::<key>` (the member property, or the string subscript index; else `<dynamic>`).

**Shadowing:** the prefix head (`process`) is shadow-checked exactly like a call sink — a local `const process` / `function process` / a `process` **parameter**, or an `import process from './x'`, suppresses the read (it isn't the Node global). Reuses the same lexical `in_scope` set and `node:`-aware import check from #736.

## Test (RED → GREEN)

New `test_jsts_env_e2e.py`: member access, string subscript, and env-in-argument all emit `ENV::<key>`; a TS file works the same; a `process` parameter emits nothing.

Full suite: 5081 passed (one Rust-eval flake, passes in isolation), 10 skipped. ruff + ty clean; Python I/O unchanged.

Next #714 follow-up: JS `FLOWS_TO` (taint), which will use `process.env` as a source.

### DIFF
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -173,6 +173,9 @@ TS_ARRAY_PATTERN = "array_pattern"
 TS_REST_PATTERN = "rest_pattern"
 TS_SHORTHAND_PROPERTY_IDENTIFIER_PATTERN = "shorthand_property_identifier_pattern"
 TS_PAIR_PATTERN = "pair_pattern"
+# (H) `process.env.X` is a member_expression; `process.env['X']` a subscript, used
+# (H) to detect environment-variable reads (issue #714 process.env follow-up).
+TS_SUBSCRIPT_EXPRESSION = "subscript_expression"
 TS_FUNCTION_DECLARATION = "function_declaration"
 TS_GENERATOR_FUNCTION_DECLARATION = "generator_function_declaration"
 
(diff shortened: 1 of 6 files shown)


vitali87:
Good catch — fixed, and better than a comparison: the subscript node has a dedicated `index` field, so `_member_identity` now reads `child_by_field_name('index')` directly instead of iterating and excluding the object node. That removes the unreliable `is not` identity check entirely (py-tree-sitter reuses/reinstantiates Node wrappers, so identity isn't stable — `!=` compares node IDs, but the field lookup sidesteps it).

vitali87:
@greptile review

PR #740, opened 2026-07-13
feat(flow): straight-line FLOWS_TO taint for JavaScript/TypeScript (#714)

vitali87:
## Summary

Follow-up to #714. `FLOWS_TO` (data-flow / taint) was Python-only; this adds it for **JavaScript / TypeScript / TSX** behind the opt-in `io` capture, so a tainted value that reaches a sink is tracked — the killer case being an environment secret leaking to stdout.

```js
function boot() {
  const secret = process.env.SECRET;   // source
  console.log(secret);                 // FLOWS_TO  ENV::SECRET -> STDOUT
}
```

## What's covered

- **resource → resource** flows: a read source (`process.env.X`, `fetch`, `fs.readFile`, `axios.get`, `http(s).get`) whose value reaches a write sink (`console.*`, `fs.writeFile*`, `axios.post`), whether via a variable or passed directly (`console.log(process.env.TOKEN)`).
- **arg edges**: a tainted value passed to a function → `caller -FLOWS_TO-> callee` (via=arg:i).
- **return edges + fixpoint**: a function returning a tainted value feeds the **shared** `#712` summary/fixpoint, so `const s = getSecret(); console.log(s)` emits both the `getSecret -> boot` return edge and the `ENV::SECRET -> STDOUT` flow — including forward/cross-file callees.
- `await fetch(...)` is unwrapped.

## Design

A lean **straight-line** JS walk (`_process_lean_flow`) on the flow processor: a single taint map over `variable_declarator` / `assignment_expression` / `call_expression` / `return_statement`, nested functions pruned. It reuses the descriptor from #736/#739 for node types, the shared `Taint`/`_emit_resource_flow`/`_resolve`/`finalize` machinery, and the same shadow-aware sink matching (`_js_match_sink`: import-normalised name first so aliased builtins resolve, raw dotted only when the head is the genuine `node:`/require/ESM module, never a shadowing local; a local `const console` / `function fetch` / parameter shadows). Python keeps its full path-sensitive walk untouched.

## Test (RED → GREEN)

New `test_jsts_flow_e2e.py` (7 tests): env→stdout, direct env arg, network→file, arg edge, untainted→no flow, cross-call return-taint, shadowed `console`.

Full suite: 5089 passed, 10 skipped. ruff + ty clean; all Python flow/io tests unchanged.

## Ceilings (noted in code, JS follow-ups)

- **Straight-line, not path-sensitive** — no if/try/loop MAY merge for JS yet (Python has it via #713).
- Local-name shadowing is flat (not block-scoped) and skips destructuring — the io walk has the precise version; refine if JS-flow precision needs it.


vitali87:
All four addressed (RED→GREEN):
- **Comments as named children** — arg indexing (`_js_arg_taints`), await unwrap, and return now filter `comment` nodes via `_named_no_comments`/`_js_first_expr`, so `console.log(/* c */ x)`, `await /* c */ fetch()`, and `return /* c */ x` keep correct positions/taint.
- **TS parameter wrappers + destructuring** — `_js_local_names` now runs each parameter (and each declarator name) through `_js_binding_names`, which unwraps `required_parameter`/`optional_parameter` (pattern field), `assignment_pattern` defaults, and `object_pattern`/`array_pattern`/`rest_pattern`/`pair_pattern` destructuring — so a typed `process` param, a destructured `{ fetch }`, etc. correctly shadow the builtin.

New tests: a comment inside `console.log(...)` still emits the flow; a TS typed `process` parameter shadows the env source.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `literal_target` now excludes `comment` nodes (not just keyword args) from the positional argument list, so a comment before a sink's target argument no longer shifts the index used for the resource identity. Added a test: `fs.writeFileSync(/* path */ 'out.txt', data)` still resolves the FILE::out.txt identity. This is shared with Python and universally correct (comments are never real arguments).

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN) in **both** flow and io (same latent bug). Instead of subtracting every `import_map` key from the local-shadow set — which wrongly dropped a genuine function-local `const fs = {}` when `fs` is also imported module-wide — a declarator is excluded only when it is *itself* an import alias, i.e. its value is a `require(...)` call (new shared `is_require_alias` in extract.py). A module-level `import fs from 'fs'` no longer cancels a local `const fs = {}`. Parameters (never import aliases) also no longer get import-subtracted. New tests in both suites: `import fs from 'fs'; function f(){ const fs = {}; fs.writeFileSync(...) }` emits no edge, while `const fs = require('fs')` still does.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). `_js_expr_taint` now unwraps `parenthesized_expression` alongside `await` (both via `_js_first_expr`, which also skips comments), so `console.log((process.env.SECRET))` and `const x = (await fetch(u))` keep their taint. Added a test.

vitali87:
@greptile review

PR #741, opened 2026-07-13
feat(io): recognize destructured and node:-scheme builtin imports for JS/TS (#714)

vitali87:
## Summary

Follow-up to #714. Closes a recall gap for two common Node import idioms that produced no I/O edges even though the underlying call is a known sink:

```js
const { writeFileSync } = require('fs');   // destructured CommonJS
import { writeFileSync } from 'node:fs';   // node:-scheme named import
writeFileSync('a.txt', data);              // now: WRITES_TO FILE::a.txt
```

(ESM named imports `import { writeFileSync } from 'fs'` and aliases `{ x as y }` already worked — the import processor records those.)

## Change

1. **Destructured `require`** — `_parse_js_require` now handles an `object_pattern` target, binding each destructured name to `module.imported` in `import_map` (`{ writeFileSync }` → `fs.writeFileSync`, `{ x: y }` → `y → fs.x`), mirroring how ESM named imports already resolve. Non-destructured `const fs = require('fs')` is unchanged.
2. **`node:` scheme on named imports** — a named import from `node:fs` records `writeFileSync → node:fs.writeFileSync`, but the registry keys on the bare module. A new shared `match_normalised` helper (extract.py) tries the import-normalised name and its `node:`-stripped form, and replaces the inline normalise-match in both the io `_resolve_sink` and the flow `_js_match_sink`, so **both READS_FROM/WRITES_TO and FLOWS_TO** benefit.

## Test (RED → GREEN)

New tests in `test_jsts_io_e2e.py`: ESM named (regression), `node:` named, destructured `require`, and renamed destructured `require` — all emit the FILE write; the first was already green, the other three were RED.

Full suite: 5103 passed (2 CLI-smoke subprocess-timeout flakes, pass in isolation), 10 skipped. ruff + ty clean; Python I/O unchanged.


vitali87:
Fixed — `match_normalised` now only attempts the `node:`-stripped lookup when the normalised name actually starts with the `node:` scheme, avoiding the redundant dictionary lookup for the common (unprefixed) case.

vitali87:
@greptile review

PR #744, opened 2026-07-13
feat(io): READS_FROM/WRITES_TO for Go I/O sinks (#714)

vitali87:
## Summary

Follow-up to #714: the first **non-JavaScript** language. Go source now produces `READS_FROM`/`WRITES_TO` edges behind the opt-in `io` capture, validating the `LanguageDescriptor` abstraction across a structurally different grammar.

```go
func leak() {
    s := os.Getenv("SECRET")                       // READS_FROM  ENV::SECRET
    fmt.Println(s)                                  // WRITES_TO   STDOUT
    http.Get("https://api.example.com/data")        // READS_FROM  NETWORK::…/data
    os.WriteFile("out.txt", []byte(s), 0644)        // WRITES_TO   FILE::out.txt
    http.Post("https://api.example.com/upload", …)  // WRITES_TO   NETWORK::…/upload
}
```

## What's covered

`os.Getenv`/`os.LookupEnv` → ENV read; `os.ReadFile`/`os.Open`/`ioutil.ReadFile` → FILE read, `os.WriteFile`/`os.Create`/`os.Remove`/`ioutil.WriteFile` → FILE write; `fmt.Print`/`Println`/`Printf` → STDOUT; `http.Get`/`Head` → NETWORK read, `http.Post`/`PostForm` → NETWORK write.

## Design

One new `_GO_DESCRIPTOR` entry + a `_GO_SINKS` table — no processor changes to the walk itself. Go's `call_expression` uses a `selector_expression` callee (`os.Getenv`), so the shared `call_name` already yields the dotted name, and the string node types (`interpreted_string_literal` / `…_content`) come from the descriptor.

**One shared refactor was needed:** Go imports map the package by its path (`import "net/http"` → `http → net/http`), so the raw-fallback "is this the genuine module?" check couldn't be JS-only (`base == head`). Extracted `head_is_genuine_module`, which strips a `node:` scheme *and* accepts the last path segment (`net/http` → `http`), and reused it in the io `_resolve_sink`/member-read check and the flow `_js_match_sink`/`_js_import_shadowed`. JS behavior is unchanged (no slashes → the segment check is a no-op).

## Ceilings (noted in code)

- **Direct sinks only** — no data-flow (`FLOWS_TO`) for Go yet, and handle-returning calls (`os.Open`) are treated as a direct path read (stream handles are a follow-up).
- Go's `:=` / `parameter_declaration` shapes don't match the JS-shaped local-name collection, so Go **local shadowing** isn't modeled — benign, since shadowing a package name (`os`, `fmt`) with a local doesn't compile in Go.
- `log.*` and writer-targeted `fmt.Fprint*`, plus raw (backtick) string targets, are follow-ups.

## Test (RED → GREEN)

New `test_go_io_e2e.py` asserts all five edges from a `main.go`.

Full suite: 5106 passed, 10 skipped. ruff + ty clean; Python and JS/TS I/O unchanged.

### DIFF
--- a/codebase_rag/constants/ast_go.py
+++ b/codebase_rag/constants/ast_go.py
@@ -34,6 +34,8 @@ TS_GO_ASSIGNMENT_STATEMENT = "assignment_statement"
 # (H) the selector/subscript node shapes (only used by member-access reads, which Go
 # (H) has none of, so they are inert placeholders here).
 TS_GO_IDENTIFIER = "identifier"
+TS_GO_CONST_SPEC = "const_spec"
+TS_GO_RANGE_CLAUSE = "range_clause"
 TS_GO_BLOCK = "block"
 TS_GO_INTERPRETED_STRING = "interpreted_string_literal"
 TS_GO_INTERPRETED_STRING_CONTENT = "interpreted_string_literal_content"
--- a/codebase_rag/constants/ast_go.py
+++ b/codebase_rag/constants/ast_go.py
@@ -33,6 +33,7 @@ TS_GO_ASSIGNMENT_STATEMENT = "assignment_statement"
 # (H) `interpreted_string_literal_content` child; `index_expression`/operand+field are
 # (H) the selector/subscript node shapes (only used by member-access reads, which Go
 # (H) has none of, so they are inert placeholders here).
+TS_GO_IDENTIFIER = "identifier"
 TS_GO_BLOCK = "block"
 TS_GO_INTERPRETED_STRING = "interpreted_string_literal"
 TS_GO_INTERPRETED_STRING_CONTENT = "interpreted_string_literal_content"
--- a/codebase_rag/constants/ast_go.py
+++ b/codebase_rag/constants/ast_go.py
@@ -28,6 +28,18 @@ TS_GO_VAR_DECLARATION = "var_declaration"
 TS_GO_VAR_SPEC = "var_spec"
 TS_GO_SHORT_VAR_DECLARATION = "short_var_declaration"
 TS_GO_ASSIGNMENT_STATEMENT = "assignment_statement"
+# (H) I/O detection (issue #714): a function body is a `block`; string arguments are
+# (H) `interpreted_string_literal` (double-quoted) whose text lives in a
+# (H) `interpreted_string_literal_content` child; `index_expression`/operand+field are
+# (H) the selector/subscript node shapes (only used by member-access reads, which Go
+# (H) has none of, so they are inert placeholders here).
+TS_GO_BLOCK = "block"
+TS_GO_INTERPRETED_STRING = "interpreted_string_literal"
+TS_GO_INTERPRETED_STRING_CONTENT = "interpreted_string_literal_content"
+TS_GO_INDEX_EXPRESSION = "index_expression"
+TS_GO_FIELD_OPERAND = "operand"
+TS_GO_FIELD_FIELD = "field"
+TS_GO_FIELD_INDEX = "index"
 TS_GO_EXPRESSION_LIST = "expression_list"
 TS_GO_COMPOSITE_LITERAL = "composite_literal"
 TS_GO_LITERAL_VALUE = "literal_value"
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -468,13 +468,17 @@ class FlowProcessor:
             return None
         if (sink := match_normalised(raw, jc.flow.import_map, sink_map)) is not None:
             return sink
-        if not sep or not head_is_genuine_module(jc.flow.import_map.get(head), head):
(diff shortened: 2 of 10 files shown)


vitali87:
Addressed all four:

**Path-segment match broadening JS (P1 + HIGH, threads 1 & 4):** gated the last-path-segment fallback in `head_is_genuine_module` behind a new `path_based_imports` descriptor flag — **True for Go, False for JS/TS**. So it only ever applies to Go (whose import bases keep the path, `net/http`); JS is back to strict `module == head`. Worth noting the JS false positive did not actually reproduce even before this — the JS import processor normalises `/` to `.`, so `import fs from 'memfs/fs'` records `fs → memfs.fs.default` (module `memfs` ≠ `fs`) — but the flag makes it structurally impossible regardless. Added a regression test (`memfs/fs` emits no edge).

**Go identifier constant (MEDIUM, threads 2 & 3):** added `cs.TS_GO_IDENTIFIER` and used it in the Go descriptor instead of `cs.TS_PY_IDENTIFIER`.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). For path-based (Go) imports, `_resolve_sink` now reconstructs the registry key from the **resolved package name** (last segment of the import path) rather than the local head: `import h "net/http"; h.Get(...)` resolves `h → net/http → http` → matches `http.Get`. This covers aliased *and* plain imports uniformly, so `head_is_genuine_module` no longer needs a path-segment branch — I reverted it to the simple `module == head` (JS-only) and the flow calls back to two args. Added a test with `import h "net/http"`.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN), and it simplified the design. Go sinks are now keyed on the **full import path** (`net/http.Get`, `io/ioutil.ReadFile`, `os.Getenv`). A Go call `http.Get` resolves through import_map to its path (`http → net/http`), and the existing `match_normalised` keys on that — so:
- stdlib `net/http` matches, but a third-party `import "example.com/foo/http"` resolves to `example.com/foo/http.Get` and does **not** match (new test);
- import aliases (`import h "net/http"`) still work for free (h also resolves to net/http).

This let me drop the last-path-segment reconstruction, the `path_based_imports` descriptor flag, and revert `head_is_genuine_module` to the simple `module == head`. Net: less code, more precise.

vitali87:
@greptile review

vitali87:
You're right — I was wrong that it doesn't compile; a local **can** shadow a package name in Go. Fixed (RED→GREEN): the local-name collection is now Go-aware — `_declarator_names` reads a `short_var_declaration`'s `left` field (an `expression_list` of identifiers) in addition to JS's `name` field, and `_pattern_names` handles Go's `expression_list` and `parameter_declaration` (its `name` field(s)). So a local `os` (whether `:=` or a parameter) shadows the package and `os.Getenv` no longer emits. Two new tests cover the parameter and `:=` cases.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). Added an `extra_declarator_types` descriptor field (empty for JS/TS; `{var_spec, const_spec, range_clause}` for Go), and `_block_declarations` now collects names from those too. `_declarator_names` reads all `name` field children (Go `var`/`const` can bind several: `var a, b = ...`) plus the `left` field (`:=` and `range`). So `var os = ...`, `const os = ...`, and `for os := range …` all shadow the package. New test covers `var` + `range`.

vitali87:
@greptile review

vitali87:
On the remaining summary concern (block-level precollection lets a later Go local suppress an earlier valid package call in the same block): this is a documented ceiling rather than a defect I'd fix here. It's a **false negative** (a missed edge, never a wrong one), triggered only by the unusual pattern of *using* a package and then *redeclaring its exact name* as a local later in the same block. Making it order-sensitive is not free: block precollection is actually **correct** for JS/TS, where `function`/`var` hoist (a local shadows the whole block) and `const`/`let` before their declaration is a TDZ error — so an order-sensitive walk would have to encode per-language hoisting rules and risks regressing the shipped JS/TS I/O for the same contrived shape. I've marked it as a ponytail ceiling in `_walk_scope` with the upgrade path. All realistic Go shadowing (params, `:=`, `var`/`const`/`range`) is handled and tested.

vitali87:
@greptile review

vitali87:
Fixed (RED→GREEN). Added a `sinks_require_import` descriptor flag (**True for Go, False for JS/TS**). Go stdlib is always imported, so a dotted sink whose package head is not in `import_map` (a package-scope `var os Config`, a same-file type named `os`, etc.) is rejected. JS keeps matching unimported globals (`console`, `fetch`) since its flag is False. New test: a package-scope `var os` + `os.Getenv` emits no ENV edge.

vitali87:
@greptile review

PR #761, opened 2026-07-16
fix(csharp): discover .slnx solutions and surface Roslyn frontend failures

vitali87:
## Problem

Running the C# hybrid frontend on Polly produced **zero Roslyn facts, silently**: scores were byte-identical to pure tree-sitter and nothing in the logs said why. Three defects stacked:

1. **`.slnx` solutions were never discovered.** Polly migrated to the XML solution format (`Polly.slnx`, no `.sln`), so `find_csharp_project` fell back to a single `.csproj` and the tool loaded one project instead of the whole solution. Additionally, Roslyn 4.x cannot open `.slnx` at all ("No file format header found"); Microsoft.CodeAnalysis 5.0.0 is the floor.
2. **A repo `global.json` SDK pin mismatch loads 0 projects.** Polly pins an SDK patch version newer than the installed one; MSBuildWorkspace then loads nothing and emits an empty (but well-formed) payload.
3. **Every failure path was silent.** Workspace diagnostics were gated behind `CGR_FE_DEBUG`, Python ignored the tool's stderr whenever the JSON parsed, and both the build-failure and subprocess-exception paths returned empty facts without a log line.

## Fix

- `find_csharp_project` discovers `*.sln`, then `*.slnx`, then `*.csproj` (shortest-path-first within each format); the Roslyn tool's `FindProjectOrSolution` mirrors this and `LoadProjectsAsync` routes `.slnx` through `OpenSolutionAsync`.
- Bump `Microsoft.CodeAnalysis.CSharp.Workspaces` / `Microsoft.CodeAnalysis.Workspaces.MSBuild` to 5.0.0 (the `.slnx` floor). Roslyn 5.0 obsoletes the `Workspace.WorkspaceFailed` event, so the tool now uses `RegisterWorkspaceFailedHandler`.
- The tool always reports load health on stderr: hard workspace failures (capped at 5 lines outside debug mode) plus a final `[projects] N loaded, M workspace failure(s)` line; `[LoadProjects]` exceptions are no longer debug-gated.
- Python surfaces every degradation: a well-formed but fact-empty payload logs the tool's stderr as a warning (this is the SDK-pin case), a tool build failure warns instead of silently returning, and a subprocess timeout/OS error warns with the error.

## Validation

RED → GREEN in commit history: `test(csharp)` commit adds the failing tests first (slnx discovery, empty-payload warning with tool stderr, quiet-on-success, and a dotnet-gated end-to-end test that opens a two-project `.slnx` and requires facts from **both** projects, which the single-csproj fallback cannot produce).

On Polly with `CSHARP_FRONTEND=hybrid`, the frontend now loads the full solution: 318 type base lists, 12,447 resolved call sites, 25 partial-type groups merged (previously all zero). Full suite: 5225 passed, 10 skipped.

### DIFF
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -52,6 +52,13 @@ CSHARP_FRONTEND_PARSE_FAILED = (
     "C# Roslyn frontend produced no parseable JSON; using tree-sitter.\n"
     "stdout: {stdout}\nstderr: {stderr}"
 )
+CSHARP_FRONTEND_RUN_FAILED = (
+    "C# Roslyn frontend tool did not finish ({error}); using tree-sitter"
+)
+CSHARP_FRONTEND_NO_FACTS = (
+    "C# Roslyn frontend produced no facts; every join falls back to "
+    "tree-sitter. Tool diagnostics:\n{stderr}"
+)
 GRAPH_ALREADY_IN_SYNC = (
     "Knowledge graph already in sync (hash cache matches every file). Skipping passes."
 )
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

PR #770, opened 2026-07-16
feat(io): handle-aware I/O for the lean non-Python walk plus Python cursor derive

vitali87:
## Problem

The lean non-Python I/O walk (issue #714) only matched **direct sink calls**. Handle-based I/O, which is the dominant idiom in most of these languages, produced nothing:

- Go: `f, _ := os.OpenFile("data.txt", os.O_WRONLY, 0644); f.WriteString(s)` — no edge (OpenFile's direction depends on flags, so it cannot be a direct sink at all).
- JS/TS: `const ws = fs.createWriteStream('out.txt'); ws.write(data)` — no edge.
- Java: `new FileWriter("out.txt")`, `new BufferedReader(new FileReader("in.txt"))`, `Files.newBufferedReader(Path.of("cfg.txt"))` — Java had **zero** handle coverage since every file/DB touch goes through a handle.
- Rust: `let mut f = File::open("in.txt")?; f.read_to_string(&mut s)?` — no edge.
- C++: `std::ifstream in("in.txt"); in >> word;` — no edge.
- Python: `cur = conn.cursor(); cur.execute("SELECT ...")` — the cursor was invisible even in the full Python handle walk.

`IO_HANDLE_CONSTRUCTORS` had only Python entries; the registry comment marked handles as a follow-up.

## Fix

The lean walk now tracks handle bindings in source order (flat, mirroring Python's handle walk) and attributes handle-method calls to the constructor's resource:

- **Call-shaped constructors** per language (`os.Open/Create/OpenFile`, `database/sql.Open`, `net.Dial`, `fs.createReadStream/createWriteStream`, `Files.newBufferedReader/...`, `DriverManager.getConnection`, `std::fs::File::open/create`, `std::net::TcpStream::connect`), resolved with the same shadow-aware, import-expanding matching as sinks.
- **`new`-shaped constructors** (Java `new FileWriter` etc.) plus **wrapper types** (`new BufferedReader(new FileReader(p))`, `new Scanner(new File(p))`) that delegate identity to arg0; PrintWriter keeps both its writer-wrapping and filename overloads.
- **Call-shaped wrappers**: Rust `BufReader::new(f)` / `BufWriter::new(f)`, Go `bufio.NewReader/NewWriter/NewScanner`.
- **Type-declaration constructors**: C++ `std::ifstream in("x.txt")` (init_declarator) and the most-vexing-parse form `std::ifstream dyn(path)` (function_declarator, `<dynamic>` identity). `<<` on a bound handle writes and `>>` reads its resource; the existing cout/cerr stream sink behaviour is unchanged.
- **Rust Result unwrapping**: bindings resolve through `?` (try_expression) and `.unwrap()` / `.expect(..)`.
- **Identity unwrapping**: `Files.newBufferedReader(Path.of("cfg.txt"))` and `new Scanner(new File("data.csv"))` carry the literal.
- **Derive methods** (`IO_HANDLE_DERIVES`): `conn.createStatement()` / `conn.prepareStatement()` (java.sql) and `conn.cursor()` (Python DB-API) yield same-resource sub-handles — this also closes the long-standing Python cursor gap.
- Aliases track (`g := f`), rebinds kill, unbound receivers and arithmetic `x << 2` stay silent.
- Per-language handle-method tables (`IO_LEAN_HANDLE_METHODS`) map methods to direction; java.sql `execute(sql)` refines READ_WRITE by the SQL first keyword, like Python.

The (targets, values) extraction the flow walk already used for taint binding is factored into a shared `binding_targets_values` helper in `io_access/extract.py`, used by both walks.

## Validation

RED → GREEN in commit history: the `test(io)` commit adds 26 failing specs first (30 after review-driven additions), covering every language plus negatives; the feature commit turns them green. Five e2e integration tests added to the per-language `test_*_io_e2e.py` files.

Full suite: 4986 passed, 5 skipped (`pytest -n auto`). No default-capture behaviour changes: everything stays behind the opt-in `io` capture group, and all existing direct-sink semantics (including Go `os.Open`/`os.Create` as direct sinks) are preserved.

Part of #714; the remaining lean-flow path-sensitivity depth (loops/try for non-hoisted languages, Rust `if`/`match`) follows in the next PR.


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #782, opened 2026-07-17
fix(csharp): dead-code dogfood round 1 on Polly, 190 false flags down to 23

vitali87:
## What

First round of the C# dead-code dogfood on Polly (the #726 dead-code support had only ever seen fixture tests). Polly's report drops **190 → 23** via three root-cause fixes, each RED → GREEN:

1. **Interface members are implicitly public** (`export_detection`): C# interface members carry no visibility modifier, so the modifier scan left them `is_exported=False` and every interface member became a dead-code candidate. All 190 initial findings had `is_exported=false`, dominated by `IAsyncPolicy.ExecuteAsync` overloads. Interface members now export as the API surface they are (190 → 65).
2. **Explicit interface implementations are API surface** (`export_detection`): `IAsyncPolicy IAsyncPolicy.WithPolicyKey(...)` has no modifier and lives in a class, but is invocable from outside via the interface. Detected by the `explicit_interface_specifier` child (65 → 49).
3. **Method-group delegate arguments get REFERENCES edges** (`call_processor`): `WaitAndRetry(retryCount, sleepDurationProvider, EmptyHandler)` passes a private method as a method group — never an `invocation_expression`, so no CALLS edge existed and Polly's whole `EmptyHandler` default-handler family (16 findings) reported dead. C# now runs the argument-function-reference half of the first-class-callable machinery (the interprocedural callable-param flow stays off for C#), with C#'s `argument` wrapper nodes unwrapped to their expressions (49 → 23).

4. **Method-group passes are REFERENCES, never CALLS** (`call_processor`): the first version of fix 3 reused the flow-language convention of emitting CALLS when the callee is external, which put 282 phantom edges into Polly's call graph (C# retrieval precision 1.0 -> 0.92, caught by re-running the eval before shipping). A method-group pass is not an invocation: C# now records REFERENCES at every site, flow languages keep their historical CALLS form, and the retrieval eval is back to precision 1.0.

## Remaining 23, classified

Real residual classes, each its own follow-up: local functions shadowed by same-name method overloads (`PredicateBuilder.HandleInner`), property-read references (private `Context.WrappedDictionary`), extension-method call resolution (the known #102 follow-up; `GetPipelineDescriptor`), generic/non-generic same-name resolution across partial files (`GetAsyncContext`), cast-expression receivers (`((ReloadableComponent)s!).Reload()`), and method-group references binding one overload of a family. The rest are plausible true positives (legacy `TimedLock`, docs snippets, bench helpers).

## Validation

Strict RED → GREEN per fix in commit history. Polly reindexed and re-measured after each fix (190 → 65 → 49 → 23). Full suite green.

### DIFF
--- a/codebase_rag/constants/ast_csharp.py
+++ b/codebase_rag/constants/ast_csharp.py
@@ -14,6 +14,7 @@ TS_CSHARP_FILE_SCOPED_NAMESPACE_DECLARATION = "file_scoped_namespace_declaration
 # (H) one. Used to tell a top-level type (default `internal`) from a nested type
 # (H) or member (default `private`) for export detection.
 TS_CSHARP_DECLARATION_LIST = "declaration_list"
+TS_CSHARP_EXPLICIT_INTERFACE_SPECIFIER = "explicit_interface_specifier"
 
 # (H) Type declarations -> Class nodes.
 TS_CSHARP_CLASS_DECLARATION = "class_declaration"
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -1935,7 +1935,9 @@ class CallProcessor:
                 # (H) or otherwise yields no name still consumes its arguments through
                 # (H) whatever callable it produced; reference first-party functions
                 # (H) passed to it or dead-code flags every django-style view
-                # (H) decorator wrapper.
+                # (H) decorator wrapper. REFERENCES (not arg_ref_rel) is
+                # (H) deliberate and predates the C# split: with no callee name
+                # (H) there is no invocation to assert, for ANY language.
                 if is_arg_ref_lang:
                     self._ingest_argument_function_references(
                         call_node,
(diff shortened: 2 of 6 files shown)


vitali87:
Pushback on the single P1 ("Preserve flow edge type", call_processor.py:1949): this is factually incorrect.

The claim is that this branch "used to call `_ingest_argument_function_references` with its default `CALLS` relationship for flow languages". It did not. On current main, that exact branch already passes `cs.RelationshipType.REFERENCES` explicitly:

```
$ git show main:codebase_rag/parsers/call_processor.py | sed -n '1925,1936p'
                if is_flow_lang:
                    self._ingest_argument_function_references(
                        call_node,
                        caller_spec,
                        module_qn,
                        local_var_types,
                        class_context,
                        resolve_func,
                        ensure_rel,
                        caller_qn,
                        cs.RelationshipType.REFERENCES,
                    )
```

This PR's diff touches only the gate on that branch (`is_flow_lang` to `is_arg_ref_lang`); the argument list, including the explicit `REFERENCES`, is untouched. For Python/Go/JS/TS/C++ the path is byte for byte identical to main. The rationale on main is the same one this PR extends: a callee with no resolvable name means we cannot assert an invocation, so the argument pass is recorded as a reference, not a call. The `arg_ref_rel` split applies only to the resolved external/builtin callee sites, which historically emitted CALLS for flow languages and keep doing so.

Note this finding again cites the custom-context memory `d4240b05`, the same stale memory behind the repeated false positive on PR #771. Please drop that memory; it keeps generating incorrect P1s.

@greptile review

vitali87:
@greptile review

PR #794, opened 2026-07-18
feat(csharp): Roslyn external-site facts eliminate untypeable-receiver phantom calls

vitali87:
# C# hybrid: Roslyn external-site facts kill the untypeable-receiver phantom class

## Context

The C# recall campaign (#789) measured the pure tree-sitter frontend's honest ceiling on Polly: precision 0.9815 with 55 false positives, all one class. A receiver the syntax walk cannot type (pattern variables, `var` from unresolvable chains, reflection locals) falls to the name-trie fallback, which binds the member name to an unrelated first-party method (`disposable.Dispose()` onto a first-party `Dispose`). Three gating experiments each traded roughly 300 true edges for 30 false ones and were reverted: the pure frontend cannot see which of those sites actually leave the repo.

The hybrid frontend can. Measured on Polly before this change, `CSHARP_FRONTEND=hybrid` lifts recall 0.8473 to 0.9575 but leaves all 55 false positives standing, because the Roslyn tool only emitted facts for calls it resolved to in-source declarations. A site the compiler resolves to a metadata method produced no fact at all, so the resolver fell back to the same trie.

## Changes

**1. External-site facts (the headline).** The Roslyn tool now also reports external sites: invocations whose symbol resolved successfully but whose declaring method has no first-party declaration. Only resolved symbols count; error sites keep the heuristics, and a site that is first-party under another target framework's compilation still wins because the resolver consults the positive call facts before the external set.

- `Frontend.cs`: `CollectCall` records `(file, line, col, name)` into a new `externals` payload section when the resolved method lives in metadata.
- `frontend.py`: `CSharpSemanticFacts.external_sites` parses the new section; empty on any fallback path.
- `csharp/type_inference.py`: the semantic path returns the `CSHARP_EXTERNAL_TARGET` sentinel for external sites, so the dispatcher emits nothing and skips the name-trie fallback. Key computation is shared with the positive-fact lookup (`_call_site_key`).
- `graph_updater.py`: external sites reset and update in place alongside the call-site facts (watch-mode discipline), and the frontend log line reports the external count.

**2. LoggerMessage partials stay first-party.** The first measurement run exposed a regression: a `[LoggerMessage]` partial's implementation part lives in a source-generated tree, so `FirstPartyDecl` concluded the method was external and suppressed real edges into Polly's `Log.cs`. The tool now probes the partial definition part before declaring a site external. The dotnet end-to-end test pins both directions: `System.Console.WriteLine` must surface as external, and a LoggerMessage call must be a positive fact anchored at the attribute line (the same line tree-sitter's `method_declaration` starts on, which is what the location join matches).

**3. Local-function method groups survive without the phantoms.** With the phantoms gone, Serilog's dead-code briefly reported `LoggerConfiguration.CreateLogger`'s `Dispose`/`DisposeAsync` local functions, which are passed to `return new(...)` and had NO legitimate incoming edge; the phantom trie edges had been masking two pre-existing bugs:

- `implicit_object_creation_expression` exposes no fields in the tree-sitter grammar, so `_parse_call_arguments`'s `arguments`-field lookup silently dropped every target-typed `new(...)` argument list. A typed fallback now finds the unfielded `argument_list` child.
- Method-group arguments never probed in-scope local functions; `_emit_callback_edge` went straight from the enclosing type's member family to the trie, whose lexicographic pick can land on an unrelated class. The new `csharp_local_function_group` walks the same scope chain as bare-call resolution (locals shadow members, arity-blind because a method group carries no call arity) and references the whole group.

## Validation (every fix RED to GREEN in commit history)

- `test_external_site_fact_suppresses_name_trie_fallback`: pattern-variable receiver phantom, suppressed by a synthetic external fact.
- `test_roslyn_tool_emits_semantic_facts`: extended for both the external-site emission and the LoggerMessage positive fact (real dotnet build, LoggerMessage source generator restored from nuget).
- `test_local_function_method_group_argument_referenced`: local function passed to target-typed `new(...)` referenced, decoy member untouched.

Polly retrieval eval, hybrid frontend (semantic oracle):

| frontend | tp | fp | fn | precision | recall | f1 |
|---|---|---|---|---|---|---|
| pure (main) | 2914 | 55 | 525 | 0.9815 | 0.8473 | 0.9095 |
| hybrid before this PR | 3293 | 55 | 146 | 0.9836 | 0.9575 | 0.9704 |
| hybrid with this PR | 3272 | 0 | 167 | 1.0000 | 0.9514 | 0.9751 |

The 21 suppressed pairs that moved from tp to fn were hand-verified as eval-oracle artifacts, not regressions: sites like `FakeTimeProvider.Advance`, `loggerFactory.CreateLogger`, configuration `.Add`, and Shouldly-lambda `StackTrace.Contains` resolve to nuget metadata under the frontend's real MSBuild references, but the TPA-only eval oracle cannot resolve those receivers and counts them via its name-collision fallback. The frontend is the more accurate side of that disagreement; hybrid recall is correspondingly understated.

Dead-code dogfood (regression guard): Polly hybrid 1 (the verified `private PolicyBuilder()` true positive, unchanged), Serilog hybrid 0, Serilog pure 0. Pure-frontend eval and graphs are untouched: external sites are empty when the frontend is off. Full suite green.

### DIFF
--- a/codebase_rag/constants/ast_csharp.py
+++ b/codebase_rag/constants/ast_csharp.py
@@ -175,6 +175,10 @@ TS_CSHARP_MEMBER_BINDING_EXPRESSION = "member_binding_expression"
 TS_CSHARP_FIELD_EXPRESSION = "expression"
 TS_CSHARP_THIS = "this"
 TS_CSHARP_ARGUMENT = "argument"
+# (H) implicit_object_creation_expression (`new(...)`) exposes NO field names;
+# (H) its argument_list is an unfielded named child, so argument parsing needs
+# (H) a typed fallback where the `arguments` field lookup returns nothing.
+TS_CSHARP_ARGUMENT_LIST = "argument_list"
 
 # (H) Nested scopes that own their own locals; the variable-type walk stops at
 # (H) these so a lambda/local-function local cannot leak into (or shadow) the
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -649,6 +649,7 @@ class GraphUpdater:
         dp = self.factory.definition_processor
         dp.csharp_base_kinds = {}
         dp.csharp_call_sites.clear()
+        dp.csharp_external_sites.clear()
         self._csharp_partial_decls = []
         self._csharp_query_calls = []
         if settings.CSHARP_FRONTEND == cs.CSharpFrontend.TREESITTER:
@@ -665,6 +666,7 @@ class GraphUpdater:
         facts = run_csharp_frontend(self.repo_path)
         dp.csharp_base_kinds = facts.base_kinds
         dp.csharp_call_sites.update(facts.call_sites)
+        dp.csharp_external_sites.update(facts.external_sites)
         self._csharp_partial_decls = facts.partial_groups
         self._csharp_query_calls = facts.query_calls
         logger.info(ls.CSHARP_FRONTEND_TYPES.format(count=len(facts.base_kinds)))
@@ -673,6 +675,7 @@ class GraphUpdater:
                 calls=len(facts.call_sites),
                 partials=len(facts.partial_groups),
                 queries=len(facts.query_calls),
+                externals=len(facts.external_sites),
             )
         )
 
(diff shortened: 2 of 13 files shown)


vitali87:
The failing Unit Tests matrices are inherited main breakage, not from this branch: #793 merged without the Unit Tests workflow running, and its two CLI help smoke tests assert plain substrings while rich colorizes help output in the GitHub Actions environment. Fix in #795; this branch will pick up main once that merges and CI reruns.

vitali87:
@greptile review

vitali87:
@greptile review

PR #795, opened 2026-07-18
fix(tests): strip ANSI from CLI help smoke asserts so rich-colorized CI logs pass

vitali87:
## Problem

PR #793 merged without the Unit Tests workflow running (no Unit Tests entries in its check rollup), and its two new help smoke tests assert plain substrings on raw stdout. rich colorizes help output when it detects an ANSI-capable log sink, which includes the GitHub Actions environment, so `test_help_command_shows_task_grouped_index` and `test_help_command_shows_detailed_command_page` fail with escape codes on every subsequent PR's CI (first seen on #794) while passing locally.

Reproduction on main: `GITHUB_ACTIONS=true CI=true uv run pytest codebase_rag/tests/test_cli_smoke.py` fails the two tests; with this fix both env and plain runs pass all 14.

## Fix

Apply the file's existing `_ANSI_RE` strip (already used by `test_help_command_works`) to both tests before asserting.

### DIFF
--- a/codebase_rag/tests/test_cli_smoke.py
+++ b/codebase_rag/tests/test_cli_smoke.py
@@ -12,7 +12,17 @@ from codebase_rag import constants as cs
 from codebase_rag.cli import app
 
 _ANSI_RE = re.compile(r"\x1b\[[0-9;]*m")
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #796, opened 2026-07-18
fix(tests): strip ANSI SGR codes from CLI help assertions broken by CI color forcing

vitali87:
## Problem

Main has been red since #793 merged: `test_help_command_shows_task_grouped_index` and `test_help_command_shows_detailed_command_page` fail on every CI platform while passing locally. rich force-enables terminal styling when it detects `GITHUB_ACTIONS`, even though CliRunner's capture stream is not a terminal, so on CI (and only there) SGR escape codes split the asserted substrings (`Usage: \x1b[0m\x1b[1mcgr ...`). Reproduced locally with `GITHUB_ACTIONS=true`: same two failures.

`NO_COLOR` is not a fix — it strips color but keeps bold, which is exactly the code class in the CI output.

## Fix

Strip SGR codes from CliRunner output before asserting, via a `_plain_stdout` helper reusing the file's existing `_ANSI_RE` (the same guard `test_help_command_works` already applies to its subprocess output). Applied to all four CliRunner-based help tests — the two that only pass by escape-placement luck are hardened too.

## Validation

- `GITHUB_ACTIONS=true TERM=xterm-256color pytest test_cli_smoke.py`: 2 failed before, 14 passed after.
- Plain env: 14 passed before and after.

### DIFF
--- a/codebase_rag/tests/test_cli_smoke.py
+++ b/codebase_rag/tests/test_cli_smoke.py
@@ -5,6 +5,7 @@ from importlib.metadata import version as get_version
 from pathlib import Path
 
 import pytest
+from click.testing import Result
 from typer.testing import CliRunner
 
 from codebase_rag import cli_help as ch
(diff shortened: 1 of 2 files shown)


vitali87:
Superseded by #795, which fixed the same CI breakage on main first (ANSI stripping plus box-drawing normalization) and is confirmed green on main's merge-commit run. Closing; the COLUMNS width pin here is unnecessary now that #795's normalization handles narrow-console reflow.

PR #822, opened 2026-07-20
feat(io): model Java java.net/http NETWORK sinks and Socket streams

vitali87:
## What

Expands the Java I/O sink catalog to cover the **NETWORK** resource surface, which was previously invisible (only `java.net.Socket` was modelled, as a SOCKET handle with no method table).

Concretely, flow/IO edges now emit for:

- **`new URL("http://..").openStream()`** → `READS_FROM NETWORK::<url>` — the URL literal is the resource identity.
- **`HttpClient.newHttpClient().send(..)` / `.sendAsync(..)`** (java.net.http, Java 11+) → `NETWORK::<dynamic>` `READ_WRITE` (the verb rides on the `HttpRequest`, not the client, so direction is the honest "either" label, matching the DB `execute()` and Python `client.request()` stance).
- **`new Socket(host, port).getInputStream()/getOutputStream()`** → `READS_FROM`/`WRITES_TO SOCKET::<host>` — `new Socket` was already a registered handle but had **no method table**, so its reads/writes emitted nothing; that dangling capability is now closed.

## Why

Java's HTTP surface is a core I/O modality that the catalog didn't touch, so any Java taint/flow analysis silently dropped network reads and writes. This continues the sink-catalog line from #820 (C/libc) into Java networking.

## Tests

`codebase_rag/tests/test_io_java_network.py` — RED→GREEN for URL stream reads, HttpClient send, and Socket streams. Full suite: 3074 passed.

Builder form `HttpClient.newBuilder().build()` and `URLConnection` derive-chains are noted as follow-ups (the chained `build()` / `openConnection()` do not bind here).

### DIFF
--- a/codebase_rag/parsers/io_access/registry.py
+++ b/codebase_rag/parsers/io_access/registry.py
@@ -748,6 +748,8 @@ IO_LEAN_HANDLE_METHODS: dict[
         # (H) and Python client.request() stance.
         ResourceKind.NETWORK: {
             "openStream": IODirection.READ,
+            # (H) URL.getContent() opens a connection and retrieves the resource.
+            "getContent": IODirection.READ,
             "send": IODirection.READ_WRITE,
             "sendAsync": IODirection.READ_WRITE,
         },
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

Issue #652, opened 2026-07-08
fix(graph): ~19k emitted edges reference nonexistent nodes and are silently dropped (dangling-edge campaign)


vitali87:
## Background

While fixing #650, the structural audit gained a dangling-relationship check: an emitted edge whose endpoint (label, key) matches no emitted node. Memgraph MERGEs relationships by MATCHing both endpoints, so every such edge is silently dropped at flush time. The in-memory audit previously counted these edges as connectivity, hiding them from the test gate.

The check is implemented (`graph_audit.find_dangling_relationships`, part of `collect_violations`) but excluded from the conftest test gate for now, because the debt is pervasive: enabling it fails 302 existing fixture tests across every language (132 cpp, 52 javascript, 33 rust, 26 typescript, 22 java, 9 python, 5 lua).

## Scale on a real corpus

Indexing souffle-lang/souffle and simulating drop semantics over the emitted batches: roughly 19k emitted edges never materialize. By family:

```
 5676  CALLS     FROM Function -> Function   (phantom out-of-line caller qns, e.g. Function:...souffle::ast.Aggregator for the constructor Aggregator::Aggregator)
 3908  IMPORTS   TO   Module                 (placeholder module qns for unresolved #includes, e.g. Module:souffle.souffle.TypeAttribute)
 3787  CALLS     FROM Function -> Method     (same phantom-caller family)
 1437  CALLS     FROM Method   -> Function   (callers resolved to the .cpp-module class qn instead of the header-anchored node)
  910  CALLS     FROM Method   -> Method
  745  CALLS     TO   Function               (synthetic builtin targets, e.g. Function:builtin.cpp.operator_logical_and, no node ever created)
  581  CALLS     TO   Function               (same builtin family from Function callers)
  398  INHERITS  TO   Class                  (base classes resolved to the referencing file's module qn, e.g. Class:...Aggregator.h.souffle::ast.Argument instead of ...Argument.h...)
  394  REFERENCES FROM Function              (phantom-caller family)
  105  REFERENCES FROM Function -> Method
   85  CALLS     TO   Module -> Function     (builtin family)
```

Also present: CALLS targets pointing at `qn@line` overload variants that were never created as nodes.

## Suggested attack order

[...]

PR #823, opened 2026-07-20
fix(incremental): disambiguate cross-language module qn on incremental add

vitali87:
## Problem

On a full build, two files that strip to the same module qn (`shapes.rs` / `shapes.cpp`) are disambiguated: the second gets its extension appended, because both pass through `_disambiguate_module_qn` in one run.

On an **incremental** run that ADDS one of them next to an already-indexed sibling of another language, the disambiguator only sees files processed in the current run (`_process_files` runs before `_rehydrate_registry_from_graph`). The added file re-claims the bare qn and **silently overwrites the existing module** under the `qualified_name` uniqueness constraint (issue #652 class). Index `shapes.rs` -> `proj.shapes`; later add `shapes.cpp` -> the rust module is erased.

This was invisible to the mock test harness, which never rehydrates. It surfaced via the new polyglot cross-language eval.

## Fix

`_seed_module_qns_from_graph()` seeds `module_qn_to_file_path` from `CYPHER_ALL_MODULE_PATHS_INTERNAL` before the incremental file loop, so disambiguation sees already-taken qns. A re-parsed (changed) file still matches its own path and keeps its bare qn.

## Tests

RED -> GREEN with the query-capable `_StatefulIngestor` (the mock harness can't reproduce it). Full non-integration suite: 5313 passed / 5 skipped; incremental set incl. the real-Memgraph integration test green.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1087,6 +1087,28 @@ class GraphUpdater:
                 self._rehydrated_module_qns.add(qn)
         self._rehydrate_class_inheritance_from_graph()
 
+    def _seed_module_qns_from_graph(self) -> None:
+        # (H) Cross-language module-qn disambiguation (definition_processor.
+        # (H) _disambiguate_module_qn) only sees files processed in the current
+        # (H) run. On an incremental ADD of a file whose basename collides with an
+        # (H) already-indexed sibling of another language (shapes.rs already owns
+        # (H) proj.shapes, then shapes.cpp is added), the added file would re-claim
+        # (H) the bare qn and silently overwrite the existing module under the
+        # (H) qualified_name constraint. Seed the qn->file map from the graph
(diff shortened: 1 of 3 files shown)


vitali87:
Addressed the P1: `_seed_module_qns_from_graph()` now skips modules whose file no longer exists on disk (mirrors `_prune_orphan_nodes`), so a same-cycle delete+add of a colliding basename gives the survivor the bare qn, matching a clean index. Added RED->GREEN test `test_incremental_delete_then_add_sibling_matches_clean_index`.

@greptile review

vitali87:
Addressed gemini's three notes together: `_seed_module_qns_from_graph` now seeds only modules whose file is still **eligible** this run (eligible set computed before seeding). This covers both deletion and same-cycle exclusion by construction — an excluded/gitignored file is already dropped from `_collect_eligible_files`, so a same-basename add gets the bare qn a clean index would. Simpler than threading `deleted_keys` through, and the delete+add regression test stays green.

@greptile review

PR #824, opened 2026-07-20
test(evals): add polyglot cross-language ingestion eval and regression guard

vitali87:
## What

Every other eval indexes a single language (or a Python-dominant tree). None checks what happens when cgr builds **one** graph over files from all 14 supported languages at once — the mixed-language repo it is actually pointed at in the wild.

`evals/polyglot.py` ingests a corpus spanning every `SupportedLanguage`, with a deliberate three-way basename collision (`shapes.rs` / `shapes.cpp` / `shapes.ts`), and grades cross-language integrity invariants that need no external oracle (`codebase_rag/tests/test_polyglot_eval.py`):

1. every available language contributes at least one module and one definition (no language silently dropped when mixed in),
2. files that strip to the same module qn get **distinct** qns (collision disambiguated, not overwritten — issue #652 class),
3. no `CALLS` / `INHERITS` / `OVERRIDES` / `IMPLEMENTS` / `INSTANTIATES` edge crosses a language boundary (cross-language qn bleed),
4. the recorded graph is dangling/orphan free, and the report is deterministic.

cgr satisfies all of these on a clean build; the eval stands as a regression guard for polyglot ingestion.

## Note

Building this surfaced a real incremental-path bug (cross-language module-qn collision on incremental add), fixed separately in #823.

## Tests

5 new tests pass; full non-integration suite green (5313 passed / 5 skipped).

### DIFF
--- a/codebase_rag/tests/test_polyglot_eval.py
+++ b/codebase_rag/tests/test_polyglot_eval.py
@@ -48,7 +48,13 @@ def test_every_available_language_is_represented(polyglot_corpus: Path) -> None:
     )
     # (H) each present language must also contribute at least one definition,
     # (H) not just an empty module node.
-    empty = [lang.value for lang in available if not report.defs_by_language.get(lang)]
+    # (H) sorted() so iteration order is deterministic across runs (StrEnum
+    # (H) members hash by string, whose order varies with hash randomization).
+    empty = [
+        lang.value
+        for lang in sorted(available)
+        if not report.defs_by_language.get(lang)
+    ]
     assert not empty, f"languages with no definitions: {sorted(empty)}"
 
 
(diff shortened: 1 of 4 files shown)


vitali87:
Addressed gemini's note: iterate `sorted(available)` for deterministic order (StrEnum hash order varies across runs). The assertion outcome was already order-independent, but this makes iteration deterministic.

@greptile review

vitali87:
Addressed the P1: the test now asserts all three collision files (`shapes.rs`/`shapes.cpp`/`shapes.ts`) produced a module before checking qn uniqueness, so a real collapse (one path dropped from the qn-keyed map) fails the cardinality check instead of passing on survivors. Skip guard now also requires the typescript parser.

@greptile review

Issue #118, opened 2025-08-23
Support Ruby (fully)


yoda:

vitali87:
Thank you for submitting a request for this beautiful language...

vitali87:
Partial progress: Ruby now has **structural** support via the ast-grep pattern tier shipped in PR #830 (#414). Indexing a `.rb` file produces Module/Function/Class nodes and DEFINES/IMPORTS edges (`def`, `def self.`, `class`, `module`, `require`, `require_relative`), driven by `codebase_rag/parsers/ast_grep_patterns/ruby.yaml`.

This is a basic structural tier, so it does **not** yet cover: CALLS edges / call-graph resolution, nested-namespace-qualified names (names are flat), `require_relative` local-path resolution (every require is treated as an external module), or visibility/dead-code analysis. Full support would need a dedicated Ruby call-extractor on top of the tier. Leaving this open to track that remaining work.

PR #830, opened 2026-07-20
feat(parsers): ast-grep pattern-driven language tier with Ruby support

### LINKED ISSUE #414: feat(parsers): accelerate new language support using ast-grep patterns
## Summary

Use ast-grep's pattern syntax to simplify adding new language support to CGR, reducing per-language implementation effort by creating a generic handler that uses YAML pattern configs instead of hand-written tree-sitter traversal code.

## Motivation

CGR fully supports 7 languages but has 4 more "in development" (Go, Scala, C#, PHP) and 20+ languages that ast-grep supports but CGR does not (Ruby, Kotlin, Swift, Bash, C, etc.). Currently each language needs a `LanguageSpec`, `LanguageHandler`, and `FQNSpec` with hundreds of lines of tree-sitter traversal code.

ast-grep patterns like `def $FUNC($$$ARGS): $$$BODY` can replace complex tree-sitter query + traversal logic for basic extraction. A "basic language support" tier using ast-grep patterns could cover function/class/import extraction without a full handler.

## Implementation

- `codebase_rag/parsers/handlers/ast_grep_handler.py` (~200 lines) implementing `LanguageHandler` protocol using ast-grep patterns instead of tree-sitter traversal
- Language pattern configs in YAML (one per language) defining function/class/import patterns
- Register new handler as fallback in `registry.py` for languages without specialized handlers
- Use this approach to finish Go, C#, PHP, Scala support
- Provide a template for community contributors to add new languages easily

## Acceptance Criteria

- [ ] Generic `AstGrepHandler` implementing `LanguageHandler` protocol
- [ ] YAML-based pattern config format documented and validated
- [ ] At least one language (Go or C#) fully supported via this approach
- [ ] Handler registered as fallback in parser registry
- [ ] Extracted entities (functions, classes, imports) match quality of hand-written handlers for basic use cases
- [ ] Template/guide for adding new language support via YAML config
- [ ] Unit tests comparing ast-grep handler output vs existing handlers for supported languages

## Related

Part of the ast-grep integration initiative:
- #411 structural search tool
- #412 structural rewrite tool
- #413 graph enrichment via ast-grep rules
- #415 combined search + rewrite toolkit

vitali87:
## Summary

Adds a generic **ast-grep pattern-driven language tier** (`AstGrepTier`) that gives basic structural support to languages with no tree-sitter `LanguageSpec`, driven by per-language YAML pattern configs instead of hand-written traversal. Proven end to end on **Ruby**.

Closes #414.

## Why this shape

The issue originally proposed an `AstGrepHandler` implementing the `LanguageHandler` protocol. That protocol is only a name/FQN-extraction helper (the Python handler is ~25 lines); languages like Go, C#, Scala already ride the `BaseLanguageHandler` fallback and are fully supported via tree-sitter **queries**, not the handler. So a handler-shaped `AstGrepHandler` would accelerate nothing.

The real per-language cost is the tree-sitter query authoring that produces `Function`/`Class`/`Import` nodes. This PR targets exactly that: a tier that emits those nodes from ast-grep patterns, registered as a fallback for files the tree-sitter path skips.

## What it does

- `codebase_rag/parsers/ast_grep_tier.py` — loads `ast_grep_patterns/*.yaml`; for a matching file extension emits `Module` + `Function`/`Class` nodes and `DEFINES`/`IMPORTS` edges, matching the tree-sitter path's node/relationship contract. Wired into `graph_updater._process_single_file` as a fallback below the tree-sitter and dependency-file paths.
- `ast_grep_patterns/ruby.yaml` — Ruby patterns (methods incl. `def self.x`, `class`, `module`, `require`/`require_relative`).
- `ast_grep_patterns/README.md` — contributor template: YAML format, `$NAME`/`$PATH` metavar conventions, first-match-per-line ordering rule.
- `pyyaml` added to the existing `ast-grep` optional extra. The tier degrades to a no-op when the extra is absent, so base installs and all existing languages are unaffected.

## Scope

This is the **basic structural tier** the issue describes: flat qualified names (no nested-namespace qualification) and no `CALLS` resolution. Languages needing full analysis still get a tree-sitter `LanguageSpec`.

Ruby is a partial step toward #118 (Support Ruby fully); it is **not** full Ruby support.

## Testing

- TDD RED→GREEN (RED commit verified failing at its own tree state).
- 9 tests in `tests/test_ast_grep_tier.py`: end-to-end Ruby extraction (functions, classes/modules, module node, DEFINES, IMPORTS) plus unit tests for config validation and quote-stripping.
- Full suite: 5645 passed, 10 skipped.

## Acceptance criteria (#414)

- [x] Generic tier over the ingest node-writing layer
- [x] YAML pattern config format, documented and validated
- [x] One language (Ruby) supported end to end
- [x] Registered as a fallback in the parse dispatch
- [x] Emits Module/Function/Class nodes + DEFINES/IMPORTS edges
- [x] Template/guide for adding a language
- [x] Unit + end-to-end tests

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -21,6 +21,7 @@ from .language_spec import (
 )
 from .parser_fingerprint import compute_parser_fingerprint
 from .parser_loader import COMBINED_FUNC_CLASS_IMPORT_QUERIES
+from .parsers.ast_grep_tier import AstGrepTier
 from .parsers.cpp.preproc_recovery import parse_with_preproc_recovery
 from .parsers.cpp_frontend import (
     cpp_frontend_available,
@@ -568,6 +569,9 @@ class GraphUpdater:
             exclude_paths=self.exclude_paths,
             capture=self.capture,
         )
+        # (H) Fallback structural tier for languages with no tree-sitter
+        # (H) LanguageSpec (e.g. Ruby), driven by ast-grep pattern configs.
+        self.ast_grep_tier = AstGrepTier(self._sink, self.repo_path, self.project_name)
 
     def _run_cpp_frontend(self) -> None:
         # (H) Optional libclang C++ pre-pass when a compile_commands.json is
@@ -1684,6 +1688,10 @@ class GraphUpdater:
                 self._parsed_files.append((filepath, language))
         elif self._is_dependency_file(filepath.name, filepath):
             self.factory.definition_processor.process_dependencies(filepath)
+        elif self.ast_grep_tier.handles(filepath.suffix):
+            self.ast_grep_tier.process_file(
+                filepath, self.factory.structure_processor.structural_elements
+            )
 
         self.factory.structure_processor.process_generic_file(filepath, filepath.name)
 
--- a/codebase_rag/parsers/ast_grep_patterns/README.md
+++ b/codebase_rag/parsers/ast_grep_patterns/README.md
@@ -1,0 +1,59 @@
+# ast-grep language patterns
+
+Basic structural support for a language that has **no tree-sitter `LanguageSpec`**
+in cgr can be added with a single YAML file here, instead of a hand-written
+tree-sitter traversal. The `AstGrepTier` (`../ast_grep_tier.py`) loads every
+`*.yaml` in this directory and, for files whose extension matches, emits
+`Module`, `Function`, and `Class` nodes plus `DEFINES` and `IMPORTS`
+relationships, using [ast-grep](https://ast-grep.github.io/) patterns.
+
+This is a **basic** tier: names are flat (no nested-namespace qualification) and
+there is no call-graph (`CALLS`) resolution. Languages that need that get a full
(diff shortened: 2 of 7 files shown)


vitali87:
Addressed all review feedback in 2a70cc06:
- **Gemini (high)** per-label `claimed` set — moved inside the label loop so a class and a method sharing a start line (one-liners) both land; added a RED→GREEN test `test_class_and_method_on_same_line_both_land`.
- **Gemini (medium)** robust `extensions` parsing — a comma-separated string is now split/stripped.
- **Greptile (P1)** added `pyyaml>=6.0` to the `test` extra so `uv sync --extra test` can run the tier tests.
- Also hardened `AstGrepTier.__init__` to disable the tier (not crash `GraphUpdater`) on a malformed shipped config.

@greptile review

vitali87:
Addressed SonarCloud in 1248b8c3: split `process_file` into `_extract_definitions`/`_extract_imports` (cognitive complexity 23→under 15) and unnested the module-parent ternary into an if/elif chain. Behavior unchanged; the 9 tier tests stay green.

@greptile review

Issue #413, opened 2026-03-02
feat(graph): enrich knowledge graph with pattern/smell/vulnerability detection via ast-grep rules


vitali87:
## Summary

Run ast-grep YAML rules during graph indexing to detect design patterns, code smells, and security issues, then store them as new node types and relationships in the knowledge graph.

## Motivation

This enables queries like "find all Singleton classes" or "show functions with SQL injection risk." It adds a quality/security dimension to the graph that pure structure analysis misses. Rules are YAML files, easy to maintain and extend, and detection is deterministic (unlike LLM-based analysis).

### New Graph Schema

- New node types added to `NodeLabel` enum: `Pattern`, `CodeSmell`, `SecurityIssue`
- New relationships added to `RelationshipType` enum: `IMPLEMENTS_PATTERN`, `HAS_SMELL`, `HAS_VULNERABILITY`

### New Analyzer

- `codebase_rag/analyzers/ast_grep_analyzer.py` (~200 lines) running YAML rules via CLI or Python API
- `.ast-grep-rules/` directory with categorized YAML rules:
  - `patterns/` (Singleton, Factory, Observer, etc.)
  - `smells/` (long method, deeply nested, unused imports, etc.)
  - `security/` (SQL injection, XSS, hardcoded secrets, etc.)

### Integration Points

- Post-processing step in `graph_updater.py` after tree-sitter indexing
- Update `cypher_queries.py` with query templates for new node types

## Acceptance Criteria

- [ ] New node types and relationships added to graph schema
- [ ] ast-grep analyzer runs YAML rules and produces structured findings
- [ ] Findings stored as graph nodes linked to source code nodes
- [ ] Cypher query templates for querying patterns, smells, and vulnerabilities
- [ ] At least 5 rules per category (patterns, smells, security) for Python
- [ ] At least 3 rules per category for JavaScript/TypeScript
- [ ] Integration with graph indexing pipeline (runs automatically on index)
- [ ] Rules are additive and do not affect existing graph functionality

## Related

Part of the ast-grep integration initiative:
- #411 structural search tool
- #412 structural rewrite tool
- #414 accelerated language support
- #415 combined search + rewrite toolkit

vitali87:
## Triage: not resolved, keep open

Confirmed unimplemented against current code:

- No `codebase_rag/analyzers/` directory and no `ast_grep_analyzer.py`.
- No `.ast-grep-rules/` directory of YAML rules.
[...]

PR #837, opened 2026-07-21
fix(findings): flag loose inequality (!=) and regression-lock every rule fires

vitali87:
## What

Exhaustive dogfood of the opt-in `FINDINGS` analyzer (#413) across the shipped rule set surfaced one real defect and one coverage gap.

**Bug: `loose_equality` missed `!=`.** The JS/TS/TSX smell rule matched only `$A == $B`, so loose *inequality* (`x != y`) went unflagged even though it does the same type coercion as `==`. Broadened the rule to `any: [== , !=]`. Verified the strict forms stay clean: `===` and `!==` are still not flagged (they are the correct forms).

**Coverage: most rules had no firing test.** Only 5 of ~50 rules were exercised, so a rule whose `kind`/`pattern` is wrong for ast-grep's bundled grammar would silently match nothing and ship dead (the `f_string`-invalid-kind class of bug that already bit this feature once). Added `test_every_rule_fires_on_positive_fixture`, which asserts every shipped rule fires on a hand-crafted positive fixture per language.

## Testing

- RED→GREEN: `test_loose_equality_flags_inequality_not_strict` fails on old rule (`[3]`), passes after (`[3, 4]`), and guards that `===`/`!==` stay unflagged.
- `test_every_rule_fires_on_positive_fixture` locks all py/js/ts/tsx rules as live.
- Full non-integration suite: 5408 passed, 5 skipped.

Dogfood also confirmed cgr's own code is clean on the dangerous smells (bare-except 0, mutable-default 0); the parser `except: pass` sites all have a real tree-sitter-query → manual-walk fallback, not silent swallows.

### DIFF
--- a/codebase_rag/analyzers/ast_grep_rules/smells/javascript.yaml
+++ b/codebase_rag/analyzers/ast_grep_rules/smells/javascript.yaml
@@ -7,9 +7,11 @@ rules:
     rule:
       kind: variable_declaration
   - id: loose_equality
-    message: "Loose equality (==); prefer strict ==="
+    message: "Loose equality/inequality (== or !=); prefer === / !=="
     rule:
-      pattern: "$A == $B"
+      any:
+        - pattern: "$A == $B"
+        - pattern: "$A != $B"
   - id: console_log
     message: "Leftover console.log"
     rule:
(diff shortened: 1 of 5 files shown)


vitali87:
@greptile review

vitali87:
Addressed both: rule-fire coverage is now keyed by ast-grep grammar id and iterates every loaded extension (so a new language with no fixture fails the test), plus a per-extension rule-id uniqueness assertion so no duplicate id can hide a dead rule.

vitali87:
@greptile review

PR #838, opened 2026-07-21
chore: drop (H) comment markers and the no-docs enforcement hook

vitali87:
## What

Retires the legacy anti-AI-slop guard rail.

- Removes every `# (H)` human-author marker from comments across the codebase (~11.8k occurrences).
- Deletes the `check-no-docs` pre-commit hook and its script (`scripts/check_no_docs.py`) plus its test.
- Drops the now-unused supporting constants (`ALLOWED_COMMENT_MARKERS`, `QUOTE_CHARS`, `TRIPLE_QUOTES`, `COMMENT_CHAR`, `ESCAPE_CHAR`) and log strings (`NO_DOCS_*`).

## Why

The marker system existed to gate hand-written comments through a pre-commit check. That rule is no longer needed, so the markers and the machinery that enforced them come out together.

## Notes

Comment text and code are otherwise unchanged; every file still compiles and the full non-integration suite (5372 tests) passes. A follow-up will tighten the comments themselves (why/how over what).


vitali87:
@greptile review

PR #841, opened 2026-07-21
chore: contract comments and drop (H) markers in core library

vitali87:
## What

Second wave of the comment cleanup (follows #838). Contracts inline comments across the core library so they explain **why/how** rather than restate **what**, and removes the last `# (H)` author markers.

Scope: 135 non-test Python files (`codebase_rag/`, `evals/`, `scripts/`, `benchmarks/`, `optimize/`, `realtime_updater.py`). Test files land in a follow-up PR to keep each change set under Greptile's 500-file review limit.

## Rules applied

- Dropped pure "what" comments that only echo the adjacent code (section banners, name restatements).
- Contracted multi-line why/how comments to their essence, keeping causal content and concrete examples.
- Kept verbatim: every comment carrying an issue/PR number or bug rationale, tool directives (`# type:`, `# noqa`, `# ty:`, `# nosec`), `# ponytail:` / TODO notes, commented-out code, shebangs.
- British spelling in touched prose; no dashes as punctuation.

## Verification

- Non-comment token streams proven identical to base for all 135 files (no code, string, identifier, or docstring changed).
- `ruff check` + `ruff format --check` clean in pre-commit scope.
- Full pre-commit run green.
- `pytest -n auto -m "not integration"`: 5377 passed, 5 skipped.

### DIFF
--- a/codebase_rag/analyzers/ast_grep_analyzer.py
+++ b/codebase_rag/analyzers/ast_grep_analyzer.py
@@ -174,8 +174,7 @@ class FindingAnalyzer:
         start_line = node_range.start.line + 1
         end_line = node_range.end.line + 1
         # qn scopes the finding to file+line+column+rule so two matches of one
-        # rule on the same line stay distinct, while re-indexing merges the
-        # same site idempotently.
+        # rule on one line stay distinct, while re-indexing merges the site.
         qualified_name = cs.SEPARATOR_DOT.join(
             [module_qn, str(start_line), str(node_range.start.column), rule.rule_id]
         )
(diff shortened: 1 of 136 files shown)


vitali87:
@greptile review

vitali87:
On the P2 (COMMENT-only token-stream verification): agreed, and it is a wording issue in the verification note rather than a code defect. My filter excluded only `tokenize.COMMENT`, so consolidating multi-line comments changes the number of comment-only physical lines and therefore the `NL` count; the 81 mismatches are all `NL` deltas. The guarantee that matters, no change to code, identifiers, strings, or docstrings, holds: re-running the comparison excluding both `COMMENT` and `NL` yields 135/135 identical, which your T-Rex run independently reproduced. No code change needed.

PR #847, opened 2026-07-21
fix(findings): eliminate Go sqli_concat and hardcoded_secret false positives

vitali87:
## What

Dogfooding the `+findings` ast-grep rules (issue #413) over real repos surfaced two false-positive classes in the Go security rules. Both are fixed here, RED→GREEN.

### 1. `sqli_concat` fired on `url.QueryEscape` + unrelated concatenation

Running the rules over **gin** flagged:

```go
c.Writer.Header().Set("Content-Disposition", `attachment; filename*=UTF-8''`+url.QueryEscape(filename))
```

Two problems compounded: `url.QueryEscape` matches the `^(Query|Exec)` selector regex (a `net/url` false friend), and the `+` concatenation lived in the `Header().Set(...)` call, structurally unrelated to it. The old rule only checked "a `.Query*` selector exists somewhere **and** a `+` exists somewhere" in the same call.

Fix: sink-anchor. The `Query`/`Exec` selector must be **this** call's function (a direct child, not nested), and the `+` must live in **this** call's `argument_list`. Also exclude `Query(Escape|Unescape)` explicitly to cover a direct `url.QueryEscape("a" + b)` (the Rust regex engine has no lookaround).

### 2. `hardcoded_secret` matched a string buried in a call argument (latent)

Same buried-value class:

```go
token := getEnv("SOME_LONG_DEFAULT_VALUE")   // NOT a hardcoded secret
```

The old rule used `stopBy: end`, so any string literal anywhere in the RHS subtree counted. Fix: the string literal must be the assigned **value** itself (a direct child of the right-hand `expression_list`), across all four declaration forms (`:=`, `=`, `var`, `const`). A real literal `token := "literalsecretvalue"` still fires.

## Verification

- RED→GREEN tests: `test_go_sqli_concat_requires_concat_inside_the_query_call`, `test_go_hardcoded_secret_requires_literal_value`.
- Re-dogfooded gin: `sqli_concat` FP gone; `hardcoded_secret` now flags only the real `password := "my-super-secret-password"`.
- Full findings suite green (21 passed); `test_every_rule_fires_on_positive_fixture` still passes (both rules still fire on true positives).

### DIFF
--- a/codebase_rag/analyzers/ast_grep_rules/security/go.yaml
+++ b/codebase_rag/analyzers/ast_grep_rules/security/go.yaml
@@ -50,11 +50,12 @@ rules:
     # var/const declarations. Ceiling: without entropy scoring, plain word/URL
     # literals still surface as review prompts; that residual is the known limit
     # of regex-only secret detection, not a defect.
-    # The string literal must be the ASSIGNED VALUE itself (a direct child of
-    # the right-hand expression_list), not merely present somewhere in the RHS.
-    # Dropping the old `stopBy: end` kills the false-positive class where the
-    # value is a function call carrying a long string argument, e.g.
-    # `token := getEnv("SOME_LONG_DEFAULT")`.
+    # The string literal must be the ASSIGNED VALUE itself, reached from the
+    # right-hand expression_list without crossing a call boundary
+    # (`stopBy: {kind: call_expression}`). This traverses through
+    # parenthesized_expression wrappers so `token := ("literal")` is caught,
+    # while killing the false-positive class where the value is a function call
+    # carrying a long string argument, e.g. `token := getEnv("SOME_LONG_DEFAULT")`.
     message: "Hardcoded secret assigned to a credential-named variable"
     rule:
       any:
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
Addressed the review, latest commits pushed:

- **Parenthesized literal secrets** (P1): fixed. The four declaration forms now reach the string via `stopBy: {kind: call_expression}`, which traverses `parenthesized_expression` wrappers but stops at a call boundary. So `token := ("literalsecretvalue")` is caught, while `token := getEnv("...")` (buried in a call) stays clean. New assertion in `test_go_hardcoded_secret_requires_literal_value` (line 5).
- **Parenthesized SQL concatenation** (P1): no change needed, and I added a regression test to prove it. The `sqli_concat` rule searches for the `binary_expression` with `stopBy: end` *within the anchored call's `argument_list`*, so `db.Query(("select "+id))` already matches. `test_go_sqli_concat_requires_concat_inside_the_query_call` now asserts the parenthesized form fires (line 5) while the `url.QueryEscape` false friend still does not.

Full findings suite green (21 passed); `test_every_rule_fires_on_positive_fixture` still passes.

vitali87:
@greptile review

PR #848, opened 2026-07-21
fix(findings): tighten Lua findings-rule precision (secret/factory/module-return/global)

vitali87:
## What

Dogfooding the `+findings` ast-grep rules (issue #413) over **luarocks** surfaced four over-broad Lua rules. Each is tightened to match what it claims to, RED→GREEN. Real luarocks counts before → after:

| rule | before | after | fix |
|---|---|---|---|
| `hardcoded_secret` | 3 (all FP) | 0 | value must be a literal, not buried in a call |
| `factory_function` | 107 | 19 | decide on the function NAME, not the body |
| `module_return` | 339 | 90 | top-level export only, not every `return x` |
| `global_assignment` | 1838 | 1113 | field/index writes are never global leaks |

### `hardcoded_secret`
`user, password = creds:match("([^:]*):(.*)")` was flagged: the `stopBy: end` string search dove into the `match()` argument. Now the string must be a direct child of the value `expression_list`, so only a real `local secret = "..."` fires.

### `factory_function`
`stopBy: end` scanned the whole body, so any function *calling* `results.new(...)` matched. Now it checks only the name: a bare identifier (`function create()`) or the last segment of a dotted name (`function results.new()`).

### `module_return`
Matched every `return <ident>` anywhere. Now constrained to the idiomatic top-level trailing `return M` by excluding returns nested inside a `function_declaration`/`function_definition`.

### `global_assignment`
Flagged field writes like `M.const = {}` and `mt.__index = f`, which target an existing table and never leak a global. Now requires a bare-identifier target. Telling a local reassignment from a true global still needs scope analysis ast-grep cannot do; that residual is a documented recall-first ceiling (noted in the rule).

## Verification

- RED→GREEN tests: `test_lua_hardcoded_secret_requires_literal_value`, `test_lua_factory_function_matches_name_not_body`, `test_lua_module_return_only_top_level`, `test_lua_global_assignment_ignores_field_writes`.
- Full findings suite green (23 passed); `test_every_rule_fires_on_positive_fixture` still passes, so each tightened rule still fires on true positives.


vitali87:
@greptile review

vitali87:
Addressed the review, latest commits pushed. Each fix was validated against ast-grep's own bundled Lua grammar (which differs from tree-sitter-lua kind names) before applying.

**Fixed:**
- **Colon-method factories** (P1): `function Factory:new()` parses as a `function_declaration` whose name is a `method_index_expression`. Added that branch (`has: {field: method, regex: ...}`) alongside the bare-identifier and `dot_index_expression` forms. Test now asserts `function Factory:new(a)` fires.
- **Parenthesized literal secrets** (P1/security): the string is now reached with `stopBy: {kind: function_call}`, which traverses `parenthesized_expression` but stops at a call. `local token = ("...")` is caught; `password = creds:match("...")` and `password = get_default_password("...")` stay clean. Test asserts the paren form fires.
- **Mixed-target global leaks** (P1): switched from "variable_list has no index write" to "variable_list has a bare identifier". `leaked, M.field = 1, 2` now correctly reports the bare `leaked` while `M.const = {}` / `t[k] = v` stay clean. Test covers the mixed case.

**Pushed back (with grammar evidence):**
- **`local_function_declaration` exclusion**: not applicable to this grammar. ast-grep's bundled Lua grammar parses `local function f() ... end` as a `function_declaration` (with a `local` keyword child), not `local_function_declaration` (a tree-sitter-lua kind that does not exist here and would raise "Kind is invalid"). Returns inside a local function are therefore already excluded. Added `test_lua_module_return_only_top_level` with a `local function helper() return q end` proving no false module_return.
- **`stopBy: neighbor` for the secret string**: `neighbor` keeps the search to direct children and would still miss the parenthesized literal above. `stopBy: {kind: function_call}` is the correct constraint: it allows paren wrappers through while rejecting call-buried strings.

Full findings suite green (23 passed); `test_every_rule_fires_on_positive_fixture` and `test_every_language_has_all_three_categories` still pass.

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #851, opened 2026-07-21
refactor(calls): collapse duplicated filtered-calls blocks into one helper

vitali87:
## What

The caller-attribution loop in `call_processor._process_calls_in_functions` computed the `filtered` call slice in four near-identical inline blocks (recorded caller, C++ out-of-line method, Go receiver method, nested/C++ free function). Two were the full Rust-aware shape and two were the Rust-less ternary.

Since the C++ and Go sites are never Rust, one helper covers all four:

```python
def _filtered_calls_for(self, func_node, language, all_call_nodes, call_starts, owned_func_nodes):
    if all_call_nodes is None or call_starts is None:
        return None
    if language == cs.SupportedLanguage.RUST:
        return self._calls_owned_by(func_node, owned_func_nodes, all_call_nodes, call_starts)
    return self._filter_calls_in_node(all_call_nodes, call_starts, func_node)
```

## Why

Behaviour-preserving deduplication: one place to reason about how a caller's own calls are sliced, instead of four copies that could drift. Net removal of code.

## Verification

- `ruff check`, `ruff format --check`, `ty check` clean
- 173 tests pass (call_resolver, dead_code_eval, rust, trie_optimization)

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -295,6 +295,26 @@ class CallProcessor:
         call_starts = [n.start_byte for n in call_nodes]
         return call_nodes, call_starts
 
+    def _filtered_calls_for(
+        self,
+        func_node: Node,
+        language: cs.SupportedLanguage,
+        all_call_nodes: list[Node] | None,
+        call_starts: list[int] | None,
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #861, opened 2026-07-21
fix(vector): close the vector store client at exit instead of interpreter shutdown

vitali87:
## Summary
Every CLI run that touches the local vector store ended with an `Exception ignored in: QdrantClient.__del__` traceback on stderr. Nothing closes the module-global client on process exit, so teardown falls to the garbage collector during interpreter shutdown, where qdrant's `close()` imports portalocker and dies with `ImportError: sys.meta_path is None`. Found while dogfooding `cgr start --clean --no-embeddings` on real Dart repositories: #858 routes the clean path through the vector store, so the traceback now also fires on runs that never generate embeddings.

## Changes
- Register `close_vector_store_client` with `atexit` in `codebase_rag/vector_store.py`, so the client closes while imports still work. The function is idempotent and a noop when no client was ever created.

## Testing
- RED: new `test_process_exit_closes_local_client_cleanly` spawns a subprocess that opens the local client and exits, asserting no "Exception ignored" on stderr; it failed before the fix.
- GREEN after the fix; full unit suite passes (5477 passed) and a live `cgr start --clean --no-embeddings` run now exits with empty stderr.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown behavior by proactively closing the local vector store client during interpreter termination.
  * Prevented cleanup-related “Exception ignored” traceback output on exit.
* **Tests**
  * Added a regression test to verify the subprocess exits successfully without stderr cleanup tracebacks.

### DIFF
--- a/codebase_rag/tests/test_vector_store.py
+++ b/codebase_rag/tests/test_vector_store.py
@@ -511,7 +511,12 @@ def test_process_exit_closes_local_client_cleanly(temp_qdrant_path: Path) -> Non
         check=False,
         capture_output=True,
         text=True,
-        env={**os.environ, "QDRANT_DB_PATH": str(temp_qdrant_path)},
+        env={
+            # An inherited QDRANT_URL would open a remote client and skip
+            # the local-path shutdown this test exists to exercise.
+            **{k: v for k, v in os.environ.items() if k != "QDRANT_URL"},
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

PR #862, opened 2026-07-21
feat: show per-turn and session token usage and cost (#80)

### LINKED ISSUE #80: Show token consumption and cost for proprietary models

vitali87:
Closes #80.

Shows how many tokens each turn consumed and what it cost, plus running session totals, in the interactive chat.

## What it does
- After every assistant turn the REPL prints a compact dim line:
  `tokens · turn 1,523→488 · session 12,045→3,902 · $0.0105 turn · $0.1837 session`
- Token counts come from the model's own reported usage (`response.usage()`), so they are exact, not estimated.
- Cost is computed from the bundled `genai-prices` data, so there is no hand-maintained price table to go stale.
- The cost segment appears only for models with a known public price. Local or unknown models (Ollama and similar) show token counts with no cost, which keeps cost reporting to proprietary models, exactly what the issue asked for.

## How
- `services/usage_cost.py`: a small pure `price_run(usage, provider, model_id)` that returns the USD cost for one run, or `None` when the model has no known price. It tries the configured provider first, then falls back to price data auto-detection from the model id, so a proxy provider name still prices a recognizable model.
- `SessionState` gains cumulative token and cost accumulators.
- `_run_agent_response_loop` accumulates each run's usage across a turn (including tool-approval rounds) and prints the line once the turn completes.

Pricing is display only and is fully guarded, so a missing price or a pricing error never interrupts a turn.

## Tests
- `test_usage_cost.py`: prices a known proprietary model, strips a `provider:model` prefix, returns `None` for a local model, and falls back to auto-detection on an unknown provider.
- `test_turn_usage_display.py`: accumulation across turns, cost hidden until a turn is priced, cost shown once priced.
- All unit tests here run with no database or model call. Full suite: 5775 passed, 10 skipped.

Note: the one-shot `--ask-agent` path is left untouched so its stdout stays machine-parseable; a usage figure there would fit better as a JSON payload field and can be a follow-up.

## Summary by CodeRabbit

- **New Features**
  - Added per-turn token usage reporting (input/output) with session-wide cumulative totals.
  - Added per-turn USD cost display when pricing can be computed, plus accumulated session cost.
  - Supports pricing across recognized providers and honors model overrides.
- **Improvements**
  - Cost details remain hidden for turns that can’t be priced, preventing cost “leakage” across turns.
  - Usage/cost reporting stays consistent during interactive runs, including model switching.
- **Tests**
  - Added/extended unit coverage for pricing logic and per-turn/session display behavior, including partial-session totals.

### DIFF
--- a/codebase_rag/constants/cli.py
+++ b/codebase_rag/constants/cli.py
@@ -161,6 +161,9 @@ UI_MODEL_USAGE = "[bold yellow]Usage: /model <provider:model> (e.g., /model goog
 # appended only for proprietary models with a known price.
 UI_TURN_USAGE_TOKENS = "tokens · turn {ti:,}→{to:,} · session {si:,}→{so:,}"
 UI_TURN_USAGE_COST = " · ${tc:.4f} turn · ${sc:.4f} session"
+# When an earlier turn had no known price (e.g. a local model), the running
+# session total understates the true spend, so it is shown as a partial floor.
+UI_TURN_USAGE_COST_PARTIAL = " · ${tc:.4f} turn · ${sc:.4f}+ session (partial)"
 UI_HELP_COMMANDS = """[bold cyan]Available commands:[/bold cyan]
   /model <provider:model> - Switch to a different model
   /model                  - Show current model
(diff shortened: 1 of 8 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #863, opened 2026-07-21
fix(updater): discard the hash cache when the graph no longer holds the project

vitali87:
## Summary
Found while dogfooding on real Dart repositories: `--clean` wipes the whole shared database but deletes only the current repo's `.cgr-hash-cache.json`, which lives inside that repo's working tree. Re-indexing any other repo afterwards trusts its now-orphaned cache, logs "no changes detected", and leaves that project silently empty; two of three indexed projects ended up with zero Module nodes after a successful-looking sync. MCP `wipe_database` and pointing cgr at a fresh Memgraph instance void caches the same way, so deleting other repos' cache files at clean time cannot fix this: the caches are unreachable from the cleaning process.

## Changes
- Before trusting a hash cache, `GraphUpdater.run` asks the graph whether the project still has Module nodes (`CYPHER_COUNT_PROJECT_MODULES`). An explicit zero means the cache describes a graph that no longer exists, so the cache and dir-mtimes files are deleted and a full rebuild runs.
- The guard fails open: sinks that cannot answer the query (offline writers, unit-test fakes returning no rows) keep the cache, since Cypher's `count()` always yields exactly one row when it really ran.

## Testing
- RED: new integration test `test_reindex_after_external_wipe_rebuilds_fully` (index, external `DETACH DELETE`, re-index must rebuild) failed before the fix; a control test asserts an intact graph still short-circuits to an incremental skip.
- GREEN after the fix; full unit suite (5476 passed) and full integration suite (293 passed) both green.
- Live re-run of the original three-repo scenario rebuilds the wiped projects instead of skipping them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically detects when the current project has no indexed modules and discards stale local cache files to trigger a full rebuild.
  * Improves cache-validation behavior in multi-file incremental runs by checking graph state before treating data as “in sync.”
  * Fails safely: if the project-graph check can’t be completed, it preserves the existing cache and continues syncing.
* **Tests**
  * Added end-to-end integration coverage for recovery after an external graph wipe and for continued incremental syncing when the graph is intact.
  * Added a guard test to ensure cache files remain when the graph check fails.

### DIFF
--- a/codebase_rag/constants/graph.py
+++ b/codebase_rag/constants/graph.py
@@ -348,6 +348,10 @@ CYPHER_DELETE_CALLS = "MATCH ()-[r:CALLS]->() DELETE r"
 CYPHER_DELETE_ORPHAN_EXTERNAL_MODULES = (
     "MATCH (m:ExternalModule) WHERE NOT (m)<--() DETACH DELETE m"
 )
+CYPHER_COUNT_PROJECT_MODULES = (
+    "MATCH (m:Module) WHERE m.qualified_name STARTS WITH $project_prefix "
+    "RETURN count(m) AS count"
+)
 
 # Queries for orphan pruning: return all paths stored in the graph
 CYPHER_ALL_FILE_PATHS = (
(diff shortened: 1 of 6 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #903, opened 2026-07-22
fix: harden shell tests against slow Windows spawns and surface timeout cause

### LINKED ISSUE #902: flaky: test_safe_awk_allowed returns -1 on windows-latest runner
## What happened

On PR #898 (head 8212baf7, CI run 29945276330), the Windows unit test job failed with:

```
FAILED codebase_rag/tests/test_shell_command.py::TestAwkSedXargsIntegration::test_safe_awk_allowed[asyncio] - assert -1 == 0
```

The test writes a file and runs `cat data.txt | awk '{print $1}'` through ShellCommander, expecting return code 0. On that runner the pipeline came back with -1, which ShellCommander uses for rejected or failed spawns.

## Why this looks like a flake

The PR touches only graph service constraint migration and pruning code, nothing near shell command handling. The same test passed on every earlier CI run of the same PR, and a plain rerun of the failed job (`gh run rerun --failed`) went green with zero code changes.

## Suggested investigation

Determine what makes the awk spawn fail intermittently on windows-latest (PATH resolution of awk under Git Bash vs the runner shell, or a transient spawn error), and either harden the test's tool discovery or surface the underlying spawn error so the assertion message shows the real cause instead of a bare -1.

vitali87:
Fixes #902.

## What happened

On CI run 29945276330 (PR #898), `test_safe_awk_allowed` failed on windows-latest with a bare `assert -1 == 0`. Return code -1 is `SHELL_RETURN_CODE_ERROR`, which ShellCommander returns for every rejection, timeout and spawn failure alike, and the assertion discarded `result.stderr`, so the actual cause never reached the CI log. A plain rerun went green with zero code changes.

## Root cause

The shared test fixture gave ShellCommander a 5 second budget for the whole pipeline. On a cold, loaded Windows runner, spawning two Git Bash executables (`cat`, then `awk`) under antivirus scanning can occasionally exceed that, and the TimeoutError path returns exactly -1. The tight budget was a test-only value with no test depending on it.

## Fix

- The fixture now uses the production default of 30 seconds, so process spawn latency can no longer masquerade as a command failure. Rejection-path tests never spawn anything and are unaffected.
- Every success assertion (`return_code == 0`) in the file now carries `result.stderr` as its message, so any future failure shows the real reason directly in the CI log instead of a bare -1.
- New test `test_timeout_reports_reason_in_stderr` pins the previously uncovered timeout path: an exhausted budget must return -1 with the timeout message in stderr. The RED evidence for this change is the observed CI failure itself, which was undiagnosable precisely because nothing surfaced the reason.

## Verification

Full unit suite (5675 passed), integration suite (297 passed) and pre-commit all green locally.

## Summary by CodeRabbit

* **Tests**
  * Improved command execution test coverage and diagnostics.
  * Added explicit success-status validation across shell command, pipeline, filtering, and approval scenarios.
  * Enhanced timeout checks to verify a clear “timed out” error message.
  * Increased the test timeout threshold to reduce false failures.

### DIFF
--- a/codebase_rag/tests/test_shell_command.py
+++ b/codebase_rag/tests/test_shell_command.py
@@ -40,7 +40,10 @@ def temp_project_root(tmp_path: Path) -> Path:
 
 @pytest.fixture
 def shell_commander(temp_project_root: Path) -> ShellCommander:
-    return ShellCommander(str(temp_project_root), timeout=5)
+    # Real-spawn tests share this budget; cold Windows CI runners can take
+    # seconds per process spawn, so keep it at the production default
+    # rather than a tight test-only value (issue #902).
+    return ShellCommander(str(temp_project_root), timeout=30)
 
 
 class TestShellCommanderInit:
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile review

PR #918, opened 2026-07-22
ci: pin split-score actions by hash and attach SLSA provenance to releases

vitali87:
Two OpenSSF Scorecard remediations:

1. Pinned-Dependencies (9/10): split-score.yml was the only workflow with unpinned actions. actions/checkout is now pinned to the same v6.0.2 hash the other workflows use, and vitali87/pr-split is pinned to the commit behind its v1.0.0 tag.

2. Signed-Releases (8/10): releases carry cosign signatures but no provenance. The sign-release job now generates GitHub build provenance for all binaries (excluding the signature bundles) via actions/attest-build-provenance and uploads it to the release as multiple.intoto.jsonl, the asset name Scorecard recognises. Takes effect from the next release.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -92,6 +92,7 @@ jobs:
     permissions:
       contents: write
       id-token: write
+      attestations: write
     steps:
       - name: Install cosign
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

PR #919, opened 2026-07-22
feat: prefix-tolerant endpoint matching for gateway and proxy rewrites

### LINKED ISSUE #911: feat: prefix-tolerant endpoint matching for gateway and proxy path rewrites
## Problem

In real deployments the path a client writes and the path a service exposes differ by an infrastructure prefix:

- ingress or gateway mounts: client calls `/y/some-service/review`, the service exposes `POST /review`;
- dev proxies with rewrites: `http-proxy-middleware` forwards `/api/*` to a backend after stripping `^/api`, so the client's `/api/cases` must match the backend's `/cases`;
- clients constructed around a base URL plus sub-path (`new URL("/user/", baseUrl)`) where the registered route carries the sub-path but the outgoing literal does not.

`url_matches_template` requires exact segment-count agreement (the unknown-lead marker only helps when the SERVER template has an unresolved mount), so all of these produce zero RESOLVES_TO edges even when both sides are in the graph.

## Proposed direction

Add a suffix-match mode on the URL side: a candidate matches when the full template matches a tail of the URL path, gated to keep evidence strong:

- the template must contain at least one literal segment (existing rule);
- the read/write direction and method compatibility gates stay;
- bound the number of stripped lead segments (for example at most two) and require the stripped lead to be all-literal;
- when several endpoints tie, prefer the exact-length match and drop ambiguous ties instead of guessing.

Optionally record the matched variant on the edge (for example `lead_prefix: "/y/some-service"`) so consumers can audit the inference. Proxy-rewrite awareness (reading `pathRewrite` tables or gateway mount calls) can come later; the suffix mode alone unlocks the common cases.

## Notes

Found while dogfooding a monorepo where every single frontend-to-backend HTTP call crossed such a prefix; combined with the extraction gaps filed separately, cross-service RESOLVES_TO was zero across roughly 260 client sinks.

vitali87:
## Problem

In real deployments the path a client writes and the path a service exposes differ by an infrastructure prefix: ingress mounts (client calls `/y/some-service/review`, the service exposes `POST /review`) and dev proxies with rewrites (`/api/cases` forwarded as `/cases`). `url_matches_template` requires exact segment-count agreement, so all of these produce zero RESOLVES_TO edges even when both sides are in the graph. Dogfooding a monorepo where every frontend-to-backend call crossed such a prefix: zero edges across roughly 260 client sinks.

## Fix

A bounded suffix mode in `link_endpoints`:

- `url_suffix_match_lead(url, template)` returns the stripped lead when the full template matches a proper tail of the URL path, stripping at most two lead segments; templates carrying the unknown-lead marker stay out (they already tail-match in the exact mode).
- Exact matches always win: the suffix mode only runs when a URL produced no exact link.
- The inference must be unambiguous: a tie between distinct candidate endpoints is dropped instead of guessed.
- The matched variant is recorded on the edge as `lead_prefix` (for example `/y/some-service`) so consumers can audit the inference.

All existing gates stay: at least one literal template segment, direction and method compatibility, and host-aware candidate scoping.

## Tests

RED first (commit 3d7e60d4): ingress-prefix write linking with the lead recorded, proxy-strip read linking, an absolute gateway URL, ambiguity dropping, exact-match suppression, and the two-segment bound. Full unit (5979) and integration (307) suites pass.

Fixes #911

## Update after merging main

Main now carries #915 (rootful same-origin candidate scoping), which deliberately blocks cross-project links for rootful URLs, while an ingress-prefixed rootful call crosses projects by design. Reconciliation: exact matching keeps the caller-project scope; when a rootful URL has no exact match and no in-scope suffix match, the suffix search widens to all endpoints, still gated by unambiguity. A tie now resolves only when the stripped lead itself names exactly one candidate's project (`/y/service-b/cases` picks service-b), otherwise it is dropped.

This restructuring also extracts `_candidate_endpoints`, `_match_candidates` and `_resolve_suffix_match` from `link_endpoints`, clearing the python:S3776 cognitive-complexity finding SonarCloud raised on #915.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved endpoint linking to handle client URLs with extra leading infrastructure/gateway path segments (supports exact then bounded-suffix matching).
  * Records a `lead_prefix` relationship property when the inferred match is unambiguous.
  * Avoids incorrect or ambiguous inferred links, preserving exact matches as highest priority.
  * Refuses linking when the matched leading prefix exceeds the supported limit.

* **Tests**
  * Added coverage for proxy rewrites, absolute gateway-prefixed URLs, ambiguous suffix candidates, exact-match precedence, and unsupported long prefixes.

### DIFF
--- a/codebase_rag/parsers/endpoints.py
+++ b/codebase_rag/parsers/endpoints.py
@@ -144,6 +144,41 @@ def url_matches_template(url: str, template: str) -> bool:
     )
 
 
+# Bounded suffix mode (issue #911): ingress mounts (`/y/<service>/review` to
+# `POST /review`) and proxy rewrites (`/api/cases` to `GET /cases`) put
+# infrastructure lead segments on the client path. At most this many are
+# stripped, and only when exactly one candidate endpoint matches.
+_MAX_SUFFIX_LEAD_SEGMENTS = 2
+KEY_LEAD_PREFIX = "lead_prefix"
+
+
+def url_suffix_match_lead(url: str, template: str) -> str | None:
+    """The stripped lead when ``template`` matches a proper tail of ``url``.
+
+    ``/y/some-service/review`` against ``/review`` yields
+    ``/y/some-service``. Gated: one or two stripped lead segments only, and
+    a template carrying the unknown-lead marker already tail-matches in
+    :func:`url_matches_template`, so it stays out of this mode.
+    """
+    parsed = urlparse(url)
+    is_absolute = bool(parsed.scheme and parsed.netloc)
+    is_rooted = not parsed.netloc and url.startswith("/")
+    if not (is_absolute or is_rooted):
+        return None
+    url_segments = [s for s in parsed.path.split("/") if s]
+    template_segments = [s for s in template.split("/") if s]
+    if not template_segments or template_segments[0] == UNKNOWN_LEAD_SEGMENT:
+        return None
+    lead = len(url_segments) - len(template_segments)
+    if not 1 <= lead <= _MAX_SUFFIX_LEAD_SEGMENTS:
+        return None
+    matches = all(
+        _TEMPLATE_PARAM_RE.match(expected) or expected == actual
+        for actual, expected in zip(url_segments[lead:], template_segments, strict=True)
+    )
+    return "/" + "/".join(url_segments[:lead]) if matches else None
+
+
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #921, opened 2026-07-23
fix: accept object method shorthand as options-object handler evidence

### LINKED ISSUE #920: Options-object routes: method-shorthand handler is not recognised as evidence
Dogfooding a large private polyglot monorepo after #907: services registering routes through an options object extract fine when the handler is an arrow property (`handler: async (ctx) => ...`), but registrations written with the object method shorthand yield nothing:

```ts
app.endpoint({
  method: "GET",
  route: "/users",
  async handler(ctx) {
    return ctx.json([]);
  },
});
```

In the tree-sitter grammar this handler is a `method_definition` child of the object literal, not a `pair`, so `_object_entries` skips it and the registration fails the handler-evidence gate. In the dogfooded repo roughly 130 of 330 options-object registrations across six packages use this shape and currently produce no ENDPOINT resources.

Fix: `_object_entries` should map a method definition to itself under its name, and the options handler evidence should treat a method definition as an inline handler, exactly like an arrow or function expression.

vitali87:
Fixes #920.

Options-object route registrations written with the object method shorthand (`async handler(ctx) {...}`) produced no ENDPOINT resources: the handler is a `method_definition` child of the object literal, not a `pair`, so `_object_entries` skipped it and the registration failed the handler-evidence gate. Found while re-dogfooding a large private polyglot monorepo, where roughly 130 of 330 options-object registrations across six packages use this shape.

- `_object_entries` now maps a method definition to itself under its declared name.
- The options handler evidence accepts a method definition as an inline handler, exactly like an arrow or function expression.
- RED test first (method-shorthand registration extracts), plus a negative guarding the member-name gate (`client.request({...})` with a method shorthand stays out).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved endpoint detection for JavaScript/TypeScript options objects that define handlers as object-literal methods.
  * Async method-style handlers (e.g., `async handler(ctx) { ... }`) are now correctly recognized for endpoint registrations.
  * Avoids mistakenly treating unrelated client request objects as exposed endpoints.
* **Tests**
  * Added new test coverage for parsing method-definition handlers inside endpoint option objects.
  * Added a negative test to ensure non-route client requests do not produce endpoint/expose relationships.

### DIFF
--- a/codebase_rag/parsers/endpoint_routes.py
+++ b/codebase_rag/parsers/endpoint_routes.py
@@ -337,6 +337,13 @@ def _object_entries(obj: Node) -> dict[str, Node]:
             if name:
                 entries[name] = child
             continue
+        if child.type == cs.TS_METHOD_DEFINITION:
+            # `async handler(ctx) {...}`: an object method maps to itself
+            # (issue #920).
+            name = _decode(child.child_by_field_name(cs.TS_FIELD_NAME))
+            if name:
+                entries[name] = child
+            continue
         if child.type != cs.TS_PAIR:
             continue
         key = child.child_by_field_name(cs.FIELD_KEY)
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

PR #948, opened 2026-07-23
fix: keep escape sequences in string-literal resource identities

### LINKED ISSUE #944: fix: string escape sequences are dropped from sink identities
Found during adversarial review of #943, but pre-existing and shared: `string_literal` in `codebase_rag/parsers/io_access/extract.py` joins only content-fragment children, so `escape_sequence` nodes are silently dropped from every sink identity it renders. A JS/TS literal like `'/a\tb'` yields the identity `/ab` instead of a faithful rendering, which can silently mislink resources whose paths contain escapes.

Fix direction: include `escape_sequence` children in the join (verbatim source text is fine; interpreting the escape is optional), in both the plain-string and template-literal paths, with RED coverage for an escaped path. Applies to every language whose descriptor routes identities through this helper.

vitali87:
Closes #944.

## Problem

Most grammars split a string literal into content fragments **around** an `escape_sequence` sibling. The shared `string_literal` reader joined only the content children, so the escape and everything it separated fused: `'/logs\tdaily'` rendered as the identity `/logsdaily`, a path that never existed. The same reader feeds every language whose descriptor routes identities through it, and the route reader in `endpoint_routes.py` had the identical defect, so a server route and a client URL written from the same literal rendered differently and could not link.

## Fix

Join the escape back in as written, in the plain-string path, the template-literal path, and the route-path reader. The rendering is the source text, so both sides of one literal still meet, and it matches what Python and Lua already produced (their grammars nest the escape inside the content node, so those were never truncated).

An escape sequence is literal text like any other fragment, so it also counts as identity: a template whose only literal is an escape keeps that identity rather than collapsing to the dynamic node, exactly as `` `${host}/${path}` `` already does for its `/`. A string with no literal text at all still stays dynamic.

## Coverage

Twelve tests, all demonstrated RED before the fix: client sinks in JS, TS templates, Go, Rust, C#, Java and C++; route registrations in JS and Go; the Go raw-string arm (no escape children, must be left alone); and the two identity-vs-dynamic boundary cases.

`TS_PY_ESCAPE_SEQUENCE` was unused; it becomes the shared `TS_ESCAPE_SEQUENCE`, whose comment now records where each grammar puts the node, including the two (Dart, Scala) that expose the escape as the only named child and would break a naive reuse.

## Known trade

Rendering the source text means a raw and an interpreted spelling of the same escape (Go `` `/a\tb` `` and `"/a\tb"`) now share one identity, where before the change the interpreted form silently became `/ab`. Python has always rendered both spellings alike, so this makes the other languages consistent rather than introducing a new class of collision, and it replaces a fabricated path that could link to an unrelated real endpoint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of escaped characters in string and template literals so resource identities preserve the complete source text.
  * Fixed route path reconstruction when escaped characters occur within paths.
  * Preserved dynamic behavior for placeholders that contain no literal text.

* **Tests**
  * Added cross-language coverage for escaped strings, template literals, and route patterns.


vitali87:
@greptile review

Issue #909, opened 2026-07-22
fix: Go route strings built from module constants are invisible (router.Get(BaseURL+"/x", wrapper.Handler))


vitali87:
## Problem

oapi-codegen and similar generators emit Go registrations of the shape:

```go
router.Get(BaseURL+"/api/v1/me", wrapper.GetMe)
```

Two current gates in `endpoint_routes.py` reject this:

1. the path argument is a binary concatenation, not a plain literal, and
2. the handler is an attribute expression (`wrapper.GetMe`), not a bare identifier or inline function, while the receiver is a parameter rather than a factory bound in the module.

Result observed while dogfooding a monorepo whose public REST surface is entirely generated: not a single generated route produced an ENDPOINT resource, while a hand-written registration inside a test file did (see companion test-module issue).

## Proposed fix

- Resolve `IDENT + "literal"` (and chained concatenations) when `IDENT` is a module-level `const` string in the same file; generated files always keep `BaseURL` adjacent to the routes.
- Accept attribute-expression handlers as registration evidence when the call otherwise looks like a server registration (verb method name and resolved literal path opening with `/`); attribute handlers keep EXPOSES attribution on the enclosing function, matching the existing ladder.

## Acceptance

- Fixture modelling the generated shape above yields `GET /api/v1/me` anchored to the registering function.
- A client-style call (`client.Get(baseURL+"/api/v1/me")` with no handler argument) stays ignored.

vitali87:
Being tackled in PR #917: module-const concatenation resolution (file scope only) plus attribute-expression handler evidence on Go verb methods.

vitali87:
Fixed and merged via PR #917.

PR #952, opened 2026-07-23
fix: register generated Go routes whose mount prefix cannot be resolved

### LINKED ISSUE #950: fix: Go route registrations through a router parameter are invisible
Found while dogfooding a large private polyglot monorepo (follow-up to #909 / #912).

## Symptom

The generated server file of the standard OpenAPI Go codegen registers every route inside a function that receives the router and a mount prefix:

```go
func RegisterHandlersWithOptions(router fiber.Router, si ServerInterface, options FiberServerOptions) {
	wrapper := ServerInterfaceWrapper{Handler: si}
	router.Post(options.BaseURL+"/v2/things", wrapper.CreateThing)
	router.Get(options.BaseURL+"/v2/things/:id", wrapper.GetThing)
}
```

Not one of those registrations becomes an ENDPOINT. In the dogfooded repo an entire generated server surface (50 registrations) is missing from the graph, so its clients, which do have their URL sinks, have nothing to resolve to.

## Cause

`options.BaseURL` is a struct field, not a module-level const, so `_go_path_value` in `codebase_rag/parsers/endpoint_routes.py` resolves neither side of the concatenation and returns None, and the registration is dropped before any evidence check runs.

(An earlier version of this issue also blamed the router arriving as a PARAMETER. That part is wrong: a router parameter already works through the generated-wrapper evidence from #909, as `test_const_concat_with_attribute_handler_registers` shows with `router chi.Router`. The path is the only blocker.)

## Fix direction

A concatenation whose prefix is a field access keeps its rooted literal suffix under the unknown-lead marker the template vocabulary already has (`UNKNOWN_LEAD` / `/**` in `endpoint_prefixes.py`), so the endpoint states exactly what is known and links across a mount prefix of any length. Restricted to a field access, because any other unresolvable prefix may be knowable (a function-scoped literal) or truncated (the recursion depth valve), and a wrong template is worse than none. With no resolved lead, the generated wiring shape must be the only admissible server evidence, or a client request handing a callback to `client.Get(base+"/x", check)` would look like a route.

RED coverage: the codegen shape above yields one ENDPOINT per registration under `/**`; a client callback on the same shape yields none; a function-scoped literal prefix is never truncated into a shorter route.

vitali87:
## Summary

- The standard OpenAPI Go codegen registers every route through a mount prefix held in a struct field (`router.Post(options.BaseURL+"/v2/things", wrapper.CreateThing)`). Nothing could resolve that field, so the whole concatenation read as unresolvable and each registration was dropped: an entire generated server surface was missing from the graph while its clients had their URL sinks.
- A concatenation whose prefix is a field access now keeps its rooted suffix under the unknown-lead marker the template vocabulary already has (`/**`, from `endpoint_prefixes`), so the endpoint states exactly what is known and links across a mount prefix of any length instead of asserting a lead it does not have.
- Restricted deliberately to a field access. Any other unresolvable prefix may be knowable (a function-scoped literal) or truncated (the recursion depth valve), and a wrong template is worse than none.
- With no resolved lead, the generated wiring shape becomes the only admissible server evidence, so a client request handing a declared function or an inline closure to `client.Get(base+"/x", check)` still registers nothing.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Closes #950.

## Test Plan

- [x] Unit tests pass (`uv run pytest codebase_rag/tests -n auto`: 6133 passed, 10 skipped)
- [x] New tests added
- [x] Integration tests pass (run as part of the suite above)

Five tests, four of them RED before the fix and each pinning a distinct rule: the codegen shape registers both routes under `/**`; a declared-function callback and an inline callback on the same path shape both register nothing; a function-scoped literal prefix is never truncated into a shorter route; and an unrooted suffix stays out of the `VERB /path` pattern reader.

Manual validation: the shape was first reproduced from a private monorepo whose generated Go handlers hold 50 such registrations, none of which reached the graph.

## Checklist

- [x] All pre-commit checks pass (`uv run pre-commit run --all-files`, plus `uv run ruff check`)
- [ ] No new comments or docstrings (code should be self-documenting)

The last item is not met: this module documents every evidence rule it applies in comments, and the constraint here (why only a field access qualifies, and why the marker rather than a bare path) cannot be read off the code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved endpoint detection for Go Fiber route registrations with prefixes that cannot be resolved.
  - Preserved complete route templates instead of incorrectly dropping or truncating their prefixes.
  - Prevented callback-like code and relative paths from being incorrectly identified as endpoints.
  - Improved matching for `Handle` and `HandleFunc` routes, including validation of rooted URL suffixes.

- **Tests**
  - Added coverage for generated Go Fiber routes, unresolved prefixes, callback patterns, and route matching edge cases.

### DIFF
--- a/codebase_rag/parsers/endpoint_routes.py
+++ b/codebase_rag/parsers/endpoint_routes.py
@@ -41,6 +41,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from .. import constants as cs
+from .endpoint_prefixes import UNKNOWN_LEAD
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -457,7 +458,14 @@ def _go_server_evidence(
     args: list[Node],
     evidence: _ModuleEvidence,
     concat_path: bool,
+    unknown_lead: bool = False,
 ) -> bool:
+    if unknown_lead:
+        # Nothing resolved the lead, so the generated wiring shape below is
+        # the ONLY evidence that this is a route rather than a client
+        # request: a callback argument alone would make an assertion helper
+        # passed to `client.Get(base+"/x", check)` look like a server.
+        return _go_wrapper_selector_evidence(fn, field, args, evidence, concat_path)
     if _receiver_is_framework(fn, evidence):
         return True
     if (
@@ -473,7 +481,21 @@ def _go_server_evidence(
     # (`wrapper := ServerInterfaceWrapper{...}`) in the SAME function (or at
     # file scope). A client passing `opts.Header` after a concat path has no
     # in-scope binding and stays out.
-    if (
+    if _go_wrapper_selector_evidence(fn, field, args, evidence, concat_path):
+        return True
+    return _handler_evidence(
+        args[1:], evidence, frozenset({cs.TS_GO_FUNC_LITERAL}), cs.TS_GO_IDENTIFIER
+    )
+
+
+def _go_wrapper_selector_evidence(
+    fn: Node,
+    field: str,
+    args: list[Node],
+    evidence: _ModuleEvidence,
+    concat_path: bool,
+) -> bool:
+    return (
         concat_path
         and field in _GO_VERB_METHODS
         and any(
@@ -486,10 +508,6 @@ def _go_server_evidence(
             )
             for arg in args[1:]
         )
-    ):
-        return True
-    return _handler_evidence(
-        args[1:], evidence, frozenset({cs.TS_GO_FUNC_LITERAL}), cs.TS_GO_IDENTIFIER
     )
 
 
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

vitali87:
On the Linked Issues check: the router-parameter half of #950 was a mis-diagnosis in my original issue text, and I have now corrected the issue body. A router received as a parameter already registers routes today through the generated-wrapper evidence from #909, which `test_const_concat_with_attribute_handler_registers` demonstrates with `router chi.Router`; I verified it in a fixture before writing this PR. The only blocker was the path, and that is what this change fixes.

Adding declared-type detection for router parameters would therefore add a second, redundant evidence route into the endpoint extractor and widen what counts as a server registration for no recovered edges, so it is deliberately not here.

PR #953, opened 2026-07-23
feat: anchor generated artefacts to the operations their contracts declare

### LINKED ISSUE #912: feat: link services through generated API clients and contract files (OpenAPI, protobuf)
## Problem

In contract-first architectures no HTTP path literal ever appears in application code, so URL matching cannot see the links at all:

- protobuf services: one `.proto` file generates a Go connect/gRPC server binding, a Go client, a Python stub and sometimes a TS client. Application code calls generated methods (`NewSomeServiceClient(...).CreateThing(...)`); the server implements a generated base class. The only join key is the proto service/method, reachable through the generated symbols.
- OpenAPI services: the same spec generates the server router (for example oapi-codegen, with routes as string literals inside `gen.go`) and typed clients in other languages (for example `@hey-api/openapi-ts`), whose method names derive from `operationId`. Client and server never share a symbol; the join key is the spec file.

Dogfooding a monorepo built this way: every service-to-service RPC call (Go to Go, Go to Python, TS to Go) was invisible, even though the CALLS graph inside each language was healthy.

## Proposed direction

Phase 1, symbol-level heuristic without parsing contracts: recognise generated client/server pairs by their shared naming stem. A call to `<Stem>Client.<Method>` in project A and a handler `<Stem>` implementing `<Method>` in project B (or a generated `New<Stem>Handler(impl)` wiring) yield a `RESOLVES_TO`-like edge keyed on `(stem, method)`. The stems come from codegen and are identical across languages by construction.

Phase 2, contract-file anchoring: parse `.proto` service definitions and OpenAPI `operationId`s into contract Resource nodes, then link generated artefacts to their contract by path or well-known codegen markers. This gives one canonical endpoint node per operation and lets clients and servers in any language meet at it.

## Notes

This is the dominant omission by traffic volume in the dogfooded repo: REST endpoint extraction improvements (filed separately) cover the browser-facing edge, but the interior service mesh is entirely codegen-based.

vitali87:
## Summary

- In a contract-first repo neither side of a service call is written by hand: the client stub and the server handler are both generated, they share no symbol, and often no URL literal either. What they do share is the operation their contract declares, so every OpenAPI `operationId` and every protobuf `rpc` now becomes one CONTRACT resource.
- The artefacts already in the graph resolve into it: an RPC resource by the `<Service>.<Method>` identity the proto declares, and an ENDPOINT resource by the method and path template it serves, reusing the same template matching the URL linking uses so `:id` and `{id}` are one segment and an unresolved mount lead still matches.
- Each operation is anchored by the file that declares it, so the node survives the unanchored-resource prune and answers "which contract declares this" as well as "who implements it"; the identity is the declaring file's path, because `openapi.json` is a filename, not a namespace.
- Bounded deliberately: only this project's endpoints are candidates (endpoint nodes are project-scoped), a template with no literal segment of its own anchors nothing, and a template matching more than one operation anchors none of them, since a parameter segment swallows its literal siblings.
- Discovery is deliberately narrow: a JSON or YAML document counts as a spec only when it declares `openapi`/`swagger` and a `paths` mapping, and a `.proto` yields operations only from inside a `service` block, so the manifests and fixtures a repo is full of contribute nothing. YAML is read when PyYAML is installed and skipped with a debug line otherwise, so no new required dependency.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Closes #912 (phase 2; phase 1 shipped in #939 and #943).

## Test Plan

- [x] Unit tests pass (`uv run pytest codebase_rag/tests -n auto`)
- [x] New tests added
- [x] Integration tests pass (run as part of the suite above)

34 tests. Discovery: JSON and YAML specs, an operation with no id, non-spec JSON, ignored directories, protobuf services including a single-line one, `rpc` outside a service, two services declaring the same method, an rpc and an HTTP operation of the same name (which crashed the sort), the same spec filename in two directories, `basePath` and `servers` prefixes, disagreeing servers, and protobuf block comments, string literals and a brace inside a string. Linking: a node per operation, the contract kind, rpc and endpoint anchoring, parameter spelling, the unknown mount lead, method mismatch, unrelated path, the method-agnostic registration, another project's endpoints, an endpoint matching several operations, a lead-only template, the capture gate, sweep ownership, an unindexed contract file, and no contract files at all.

Manual validation on a large private polyglot monorepo: 122 declared operations (85 HTTP across 8 specs, 37 rpcs across 4 services), 83 artefact links (51 endpoints, 32 rpcs) reaching 82 distinct operations, no duplicate identities, and 40 operations with no implementation in the repo, which the graph could not previously distinguish from an implemented one. The full cross-language chain resolves: a generated TypeScript client method, its URL sink, the server route registered in a different language, and the operation the contract names.

## Checklist

- [x] All pre-commit checks pass (`uv run pre-commit run --all-files`, plus `uv run ruff check`)
- [ ] No new comments or docstrings (code should be self-documenting)

The last item is not met: the two new modules document what makes a document a contract and why an operation is anchored by its file, neither of which can be read off the code.

## Summary by CodeRabbit

* **New Features**
  * Added contract discovery from OpenAPI/Swagger and protobuf specs, including operation extraction from `paths`/`operationId` and protobuf `service`/`rpc`.
  * Introduced a new contract resource type and automatic linking of live RPC/endpoint artifacts to discovered contract operations.
  * Enforced contract spec parsing limits (including max file size) and optional YAML support when available.
* **Bug Fixes**
  * Improved endpoint/RPC anchoring accuracy (method/path matching, ambiguity handling, and project isolation), plus cleanup of stale contract links.
* **Tests / Chores**
  * Added end-to-end contract operation discovery and linking tests; expanded relationship targeting to support file-to-flow and introduced recursive JSON typing.

### DIFF
--- a/codebase_rag/constants/languages.py
+++ b/codebase_rag/constants/languages.py
@@ -165,6 +165,11 @@ CONTRACT_OPERATION_METHODS: frozenset[str] = frozenset(
     {"get", "post", "put", "patch", "delete", "head", "options", "trace"}
 )
 CONTRACT_OPERATION_ID_KEY = "operationId"
+# The mount a spec states its paths are relative to: `basePath` in Swagger 2,
+# the path component of each server URL in OpenAPI 3.
+CONTRACT_BASE_PATH_KEY = "basePath"
+CONTRACT_SERVERS_KEY = "servers"
+CONTRACT_SERVER_URL_KEY = "url"
 # A spec is a document, not a data dump; anything larger is not one, and the
 # cap keeps the walk from parsing a huge generated file.
 CONTRACT_MAX_FILE_BYTES = 8 * 1024 * 1024
--- a/codebase_rag/constants/languages.py
+++ b/codebase_rag/constants/languages.py
@@ -150,6 +150,25 @@ TSCONFIG_FILENAMES: tuple[str, ...] = (
 TS_ALIAS_SKIP_DIRS: frozenset[str] = frozenset(
     {"node_modules", "dist", "build", "out", ".git"}
 )
+# Contract files that declare service operations (issue #912): OpenAPI specs
+# in JSON or YAML, and protobuf service definitions. The markers gate parsing
+# so the JSON and YAML a repo is otherwise full of is never read as a spec.
+CONTRACT_JSON_EXTENSION = ".json"
+CONTRACT_SPEC_EXTENSIONS: frozenset[str] = frozenset({".json", ".yaml", ".yml"})
+CONTRACT_PROTO_EXTENSION = ".proto"
+CONTRACT_SPEC_VERSION_KEYS: tuple[str, ...] = ("openapi", "swagger")
+CONTRACT_SPEC_MARKERS: tuple[str, ...] = ("openapi", "swagger")
+CONTRACT_PATHS_KEY = "paths"
+# The keys under an OpenAPI path item that name an operation; every other key
+# there (parameters, servers, summary) describes the path, not an operation.
+CONTRACT_OPERATION_METHODS: frozenset[str] = frozenset(
+    {"get", "post", "put", "patch", "delete", "head", "options", "trace"}
+)
+CONTRACT_OPERATION_ID_KEY = "operationId"
+# A spec is a document, not a data dump; anything larger is not one, and the
+# cap keeps the walk from parsing a huge generated file.
+CONTRACT_MAX_FILE_BYTES = 8 * 1024 * 1024
+
 JS_INDEX_STEM = "index"
 TS_COMPILER_OPTIONS_KEY = "compilerOptions"
 TS_PATHS_KEY = "paths"
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -27,6 +27,7 @@ from .language_spec import (
 from .parser_fingerprint import compute_parser_fingerprint
 from .parser_loader import COMBINED_FUNC_CLASS_IMPORT_QUERIES
 from .parsers.ast_grep_tier import AstGrepTier
+from .parsers.contract_linking import link_contracts
 from .parsers.cpp.preproc_recovery import parse_with_preproc_recovery
 from .parsers.cpp_frontend import (
     cpp_frontend_available,
@@ -1034,6 +1035,12 @@ class GraphUpdater:
         created = link_endpoints(self.ingestor)
         if created:
             logger.info("Resolved {} client request URLs to endpoints", created)
+            self.ingestor.flush_all()
+        # Contract files name the operations the generated artefacts on both
+        # sides implement, so they anchor after the artefacts exist.
+        anchored = link_contracts(self.ingestor, self.repo_path)
+        if anchored:
+            logger.info("Anchored {} artefacts to contract operations", anchored)
             self.ingestor.flush_all()
 
     def _rehydrate_registry_from_graph(self) -> None:
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1038,7 +1038,13 @@ class GraphUpdater:
             self.ingestor.flush_all()
         # Contract files name the operations the generated artefacts on both
         # sides implement, so they anchor after the artefacts exist.
-        anchored = link_contracts(self.ingestor, self.repo_path)
+        anchored = link_contracts(
+            self.ingestor,
+            self.repo_path,
+            self.project_name,
+            self.exclude_paths,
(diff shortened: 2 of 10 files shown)


vitali87:
@greptile review

vitali87:
Two of the three findings are real and fixed in the new commit, each pinned by a test:

- **RPC links crossing projects.** RPC resources are unscoped on purpose, so a client in one project meets a server in another on one node, but a contract is declared by THIS repo. The query now returns only RPC resources this project's own code participates in (`(f)-[:EXPOSES|READS_FROM|WRITES_TO]->(r)` with `f.qualified_name STARTS WITH $project_prefix`), so a same-named service in an unrelated project is no longer reported as this contract's implementation. Field check on the dogfooded monorepo: unchanged at 32 rpc links, so every one of them was already local.
- **A file that stops declaring operations.** The file-scoped EXPOSES cleanup now runs before discovery and covers every contract file in the repo rather than only the ones that still declare something, so a spec edited down to nothing sheds its operations: the nodes lose their anchor and the unanchored-resource prune takes them and their edges.

The third finding does not hold. Watch mode is not a separate path: it calls `GraphUpdater.run()` with `_single_file` set, and `_link_endpoint_resources()` (which is where contract anchoring runs) is called from `run()` at graph_updater.py:794, exactly as the endpoint URL linking it sits beside is. A watched edit therefore recomputes contract relationships on the same schedule as every other resource link.

vitali87:
@greptile review

vitali87:
SonarCloud flagged one Cognitive Complexity smell in the spec reader. Fixed by extraction rather than by raising the threshold: `_spec_document` performs the is-this-a-spec gate, `_path_operations` reads one path item, and `_is_operation_method` names the key test. No behaviour change; all 36 contract tests still pass.

vitali87:
Both P1 findings are fixed in the new commit, each pinned by a test:

- **Contract identities merging across projects.** The qualified name is now scoped by declaring project (`resource::CONTRACT::<project>::<path>.<operation>`), exactly as an ENDPOINT resource is, so two repositories holding the same spec at the same relative path keep separate nodes and neither one's relink can delete the other's links. The cross-project meeting points remain the deliberately unscoped RPC and NETWORK resources.
- **Cleanup crossing sibling repository roots.** The prefix now ends with a path separator, so `/work/api` no longer claims `/work/api-backup`.

Field check on the dogfooded monorepo after both changes: 122 operations, 83 links (51 endpoints, 32 rpcs), zero unscoped identities, all unchanged otherwise.

vitali87:
@greptile review

vitali87:
Good catch on the test: it asserted only that a cleanup query is issued, not that the transition works. It now links a repo that declares operations, empties both specs, and relinks, asserting that nothing is re-emitted and the file-scoped EXPOSES cleanup still fires.

vitali87:
@greptile review

vitali87:
Fixed the attribution and pinned it: RPC participation is now matched on the qualified name's first SEGMENT (`head(split(f.qualified_name, '.')) = $project_name`), which is the idiom the endpoint linking already uses (endpoints.py:38), so a project named `projector` can no longer be claimed by `proj`. The test fails against a naive string-prefix match. I also made the test double model each query's own scoping rather than returning rows unfiltered, which is what let the earlier version pass.

On the specific `api` / `api.worker` example: that pair is not separable by any qualified-name rule. A project whose NAME contains a dot produces qualified names indistinguishable from a submodule of a shorter-named project, so `head(split(...))` yields `api` for both, and so does every other project-attribution site in the codebase, including the URL-to-endpoint linking. Tightening it further would need a project property on function nodes, which they do not carry; that is a repo-wide property of the qualified-name scheme rather than something this pass introduces, and this pass is now consistent with the rest of it.

vitali87:
@greptile review

vitali87:
Fixed: a protobuf service is now identified the way protobuf itself does, by `package.Service`, so two packages declaring `Health` are two contracts rather than one node anchored by both files.

The RPC resource is keyed by the bare `Service.Method` the codegen produced, so the match drops the package, and when two packages declare the same service name that key is ambiguous and claims nothing, the same uniqueness rule the endpoint side uses.

Field check on the dogfooded monorepo: unchanged at 122 operations and 83 links, with rpc names now resolving to package-qualified contracts (`UserService.GetUser` to `user.v1.UserService.GetUser`).

vitali87:
@greptile review

vitali87:
@greptile review

PR #954, opened 2026-07-24
ci: only cancel superseded runs on PRs, never on main

vitali87:
## Summary

- The CI badge intermittently showed **failing** on `main` even though CI passed.
- Root cause: every merge to `main` fires a `ci.yml` run, then `version-bump.yml` pushes a `chore: bump version` commit that fires a second run; with `cancel-in-progress: true` keyed on `github.ref` (identical for both), the second run **cancels** the first, and the branch badge latches onto the cancelled run and renders red.
- Fix: PRs keep ref-grouping so superseded runs still cancel (saves minutes); non-PR events (main push, version-bump push, dispatch) group by `github.sha`, so each commit gets its own group and can neither cancel nor be cancelled, even in a burst.

## Type of Change

- [x] CI/CD or tooling

## Related Issues

None.

## Test Plan

- [x] Greptile verified the concurrency expressions with actionlint and per-context evaluation (PR, main-push, dispatch), exit code 0.
- No application code changed; unit/integration suites unaffected.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,10 @@ permissions: read-all
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  # Cancel superseded PR runs to save minutes, but never cancel a main push:
+  # each main commit (incl. the auto version-bump) must finish so the CI badge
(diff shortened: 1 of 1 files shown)


vitali87:
@greptile review

Issue #970, opened 2026-07-28
Methods of a class expression assigned to a field (static X = class {...}) get no caller pass


vitali87:
## Summary
Methods of an anonymous class expression bound to a field/variable (`private static readonly Proxy = class { ... }`) are attributed to the OUTER class (losing the `Proxy` binding name) and receive no caller pass, so their entire bodies (calls, inline callbacks) go unscanned and `cgr dead-code` false-flags every callback inside them.

## Repro (verified in-process)
```ts
export class Outer {
  private static readonly Proxy = class {
    public exclude(routes: any[]) {
      return routes.reduce((a, r) => { useX(); return a; }, []);
    }
  };
}
```
`exclude` registers as `Outer.exclude` (not `Outer.Proxy.exclude`) and has ZERO outgoing edges; the `reduce` callback is reported dead. Found dogfooding NestJS `MiddlewareBuilder.ConfigProxy` (6 false positives: `exclude`, `getRoutesFlatList`, `removeOverlappedRoutes` callbacks + `isOverlapped`).

## Root cause / fix direction
Same family as the arrow-property scope fix (#960): the binding-name recovery must extend to CLASS expressions so their methods attribute to `Outer.Proxy.<method>` and are enumerated as callers. Simplified callbacks (not in a class expression) are already referenced correctly.

vitali87:
Fixed in #975 (merged as ee9a40c7). Anonymous class expressions now recover their field/declarator binding name and are treated as a named FQN scope, so their methods attribute to `Outer.Proxy.<method>` and get a caller pass; nested callbacks get their REFERENCES edges. Review also surfaced and fixed a related gap: a JS/TS file with no call expressions now still runs the class pass, so a method whose only body is a callback is no longer dead-flagged.

PR #974, opened 2026-07-29
fix: reference a bare-identifier expression-body arrow's target

### LINKED ISSUE #972: Expression-body arrow returning a bare function name (() => noop) does not reference it
## Summary
An expression-body arrow whose body is a bare function-name reference (`get: () => noop`) does not emit a reference to that function, so a function used only this way is reported dead.

## Repro
NestJS `noop-graph-inspector.ts`: `const noop = () => {}; new Proxy(GraphInspector.prototype, { get: () => noop })`. `noop` is returned by the expression-body arrow `() => noop` but reported dead.

## Root cause / fix direction
The expression-body-return handling references an inline function body but not a bare identifier body that names a function. Found dogfooding NestJS.

AmirF194:
## Summary

- `dead_code` false-flags a function whose only use is `() => noop` (a bare-identifier
  expression-body arrow), because `_emit_expression_body_return` only recognized an
  arrow/function-expression body and skipped a bare identifier/attribute/member/selector.

## Type of Change

- [x] Bug fix

## Related Issues

Fixes #972

## Test Plan

- [x] Unit tests pass (`uv run pytest -n auto -m "not integration"`, 5876 passed, 85 skipped;
  the one unrelated failure, `test_changes_when_csharp_frontend_setting_changes`, also fails
  on unmodified `main` in the same environment)
- [x] New tests added: `test_expression_body_arrow_returning_bare_identifier_is_referenced`
  fails on `main` (`assert False`) and passes on this branch
- [ ] Integration tests (requires Memgraph, not run)
- `ruff check`, `ruff format --check`, `ty check codebase_rag --exclude codebase_rag/tests/` all pass

## Checklist

- [ ] No new comments or docstrings: kept one 3-line comment explaining the non-obvious
  reason (a bare identifier is a reference to an existing function, not an inline
  declaration), per CONTRIBUTING's comment policy

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relationship tracking for functions returned via concise arrow expressions, correctly resolving implicit (non-braced) returns that reference an existing callable.
  * Prevents incorrect reference links when the returned identifier is shadowed by the arrow’s own parameters.

* **Tests**
  * Added coverage for `() => noop` producing the expected reference edge.
  * Added a negative test ensuring `(noop) => noop` does not generate a false reference edge due to parameter shadowing.


vitali87 (approved):
Reviewed and verified against current main (post-#970). This correctly fixes #972.

- **No false edges**: reuses `_emit_callback_edge`, which returns early unless the name resolves (call_processor.py:4712) and only emits for FUNCTION/METHOD targets (4730), so `() => somethingUnknown` produces nothing.
- **Right node-type set**: `_ASSIGNMENT_RHS_REF_TYPES` (identifier/attribute/member/go_selector) is the conservative reference set, no inline-declaration types.
- **Merge-compatible**: the 4 threaded params (module_qn, local_var_types, class_context, resolve_func) are all in scope at both call sites in the current `_ingest_returned_function_references`; no collision with the recent #970/#971 changes.
- **Proper TDD**: the new test fails on main and passes on the branch.

Optional (non-blocking): the test only covers the bare-identifier case (`() => noop`); a member-expression case such as `() => obj.method` would also exercise the `member_expression`/selector entries in the set.

LGTM.

vitali87:
Pushed a fix commit (285525e2) addressing both review points, so we can land this:

- **module_qn not forwarded (Minor):** `_emit_expression_body_return` now passes `module_qn=module_qn` into `_emit_inline_arg_function_ref`, so a named function-expression body (`() => function named(){}`) registers against module scope.
- **Parameter shadowing (Major):** a body rooted at one of the arrow's OWN parameters (`(noop) => noop`, `(x) => x.foo`) now skips emission via a new `_body_root_identifier` + `js_ts_parameter_names` check, so it no longer fabricates a false REFERENCES edge to a same-named module function. Added `test_expression_body_arrow_returning_own_parameter_is_not_referenced` (RED→GREEN); the original `() => noop` case still references.

Full suite green, ruff/ty clean. Thanks @AmirF194 for the fix.

vitali87 (approved):
Re-approving after pushing the fix commits (module_qn forwarding, parameter-shadow guard) and merging current main. All checks green, CodeRabbit clean, threads resolved.

AmirF194:
Thanks for taking it the rest of the way, the parameter shadowing case was exactly the kind of thing easy to miss from outside. Appreciate the quick turnaround.

PR #975, opened 2026-07-29
fix: walk anonymous class-expression methods so their callbacks are not dead-flagged

### LINKED ISSUE #970: Methods of a class expression assigned to a field (static X = class {...}) get no caller pass

vitali87:
## Problem
Closes #970.

An anonymous class expression bound to a field or variable
(`private static readonly Proxy = class {...}`, `const Proxy = class {...}`)
has no `name`, so it was skipped in the caller pass. Its methods were never
enumerated as callers, their bodies went unscanned, and every inline callback
inside them (plus named locals) got no incoming edge, so `cgr dead-code`
false-flagged them. Found dogfooding NestJS `MiddlewareBuilder.ConfigProxy`
(6 false positives: `exclude` / `getRoutesFlatList` / `removeOverlappedRoutes`
callbacks + `isOverlapped`).

## Root cause
The class-expression node type (`class`) was never treated as a named scope.
Both passes dropped it, so the definition pass registered nothing for its
methods while nested arrows attributed to a phantom prefix.

## Fix
Recover the field/declarator binding name for a class expression and thread it
through both passes so they agree on `Outer.Proxy.<method>`:

- `js_ts/utils.py`: new shared `class_binding_name` (and a `_value_binding_name`
  helper factored out of `arrow_binding_name`, a byte-for-byte move).
- `language_spec._js_get_name` + `fqn_specs` scope tuples: a class expression is
  now a named FQN scope, so methods and nested arrows carry the `Proxy` segment.
- `function_ingest._handle_class_ancestor` and the `call_processor` class loop:
  recover the binding name in the definition and call passes respectively.

After the fix, methods register as `Outer.Proxy.exclude` and block-body
callbacks get a `REFERENCES` edge from their method, matching the normal-method
baseline exactly.

## Tests
`test_ts_class_expression_methods.py` (RED->GREEN): field-bound, var-bound, and
the block-body callback shape (asserts the Proxy-scoped callback node exists and
is referenced). Full unit suite: 5958 passed, 5 pre-existing env skips, 0 failed.
ruff + ty + pre-commit clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JavaScript and TypeScript handling for anonymous class expressions by correctly deriving their binding names.
  * Anonymous class expression methods are now properly walked and associated with the correct class context.
  * Call and callback relationships inside these methods are captured more reliably, including when a file contains no call expressions.

* **Tests**
  * Added end-to-end test coverage for field- and variable-bound class expression method bodies, plus callback/reference edge cases.


vitali87:
Addressed CodeRabbit's review (commit 7ef52f24):

- **Major — no-call files skipped the class pass (call_processor.py):** Confirmed and fixed. A JS/TS file with no call expressions early-returned before `_process_calls_in_classes`, so a method whose only body is a callback never got its `REFERENCES` edge and was dead-flagged. JS/TS now proceeds through that guard like C#/C++/Dart. Verified it fixes both class-expression and plain-class no-call cases; added `test_callback_referenced_when_file_has_no_call_nodes` (RED→GREEN).
- **Minor — tests only asserted the callee leaf:** Applied. The two body-walked tests now assert the exact binding-qualified caller edges (`Outer.Proxy.exclude`/`Proxy.run` → `useX`).

- **Docstring Coverage pre-merge warning:** Not applicable here. This module documents functions with leading `#` comment blocks, not docstrings (see the surrounding code the diff sits in); the new `class_binding_name` / `_value_binding_name` follow that same convention. Adding docstrings would be inconsistent with the file.

Full unit suite: 5959 passed, 5 pre-existing env skips, 0 failed. ruff + ty + pre-commit clean.

PR #976, opened 2026-07-29
fix: resolve this.method() self-dispatch through nested arrows to the enclosing class

### LINKED ISSUE #971: this.method() self-call inside a nested arrow function not linked (dead-code false positive)
## Summary
A `this.method()` self-dispatch call made inside a NESTED arrow function is not linked to the method, so a method called only from such a position is reported dead.

## Repro
NestJS `ListenersController.getContextId` is called via `this.getContextId(...)` at two sites, but both are inside `const requestScopedHandler = async (...args) => { ... this.getContextId(...) ... }` (an arrow assigned to a local const). The self-dispatch through the nested arrow isn't resolved, so `getContextId` is reported dead though it is called.

## Root cause / fix direction
Arrow functions bind `this` lexically to the enclosing method, so `this.getContextId` inside a nested arrow refers to the class instance. The self-dispatch resolver likely anchors `this` only at method-body level, not through nested arrows. Found dogfooding NestJS.

vitali87:
## Summary
Closes #971.

A JS/TS `this.method()` call whose `this` is bound lexically to the enclosing
class instance (a method body, or a nested arrow that inherits `this`) was not
resolved when the call sat inside a nested arrow AND several classes declared a
same-named method. Ordinary name resolution climbs one scope from the caller qn
(landing on the class for a direct method, but on the METHOD for a nested
arrow), then falls back to a bare-name lookup that, under cross-class ambiguity,
is dropped. The method then reports dead.

Found dogfooding NestJS `ListenersController.getContextId`: 5 same-named
`getContextId` methods across classes, and the real self-call sits inside the
`requestScopedHandler` nested arrow, so it was reported dead. Full-repo re-index
confirms the fix drops it from the dead list (25 -> 24 findings).

## Fix
In the unresolved-call branch, for JS/TS `this.M()` with an enclosing class,
anchor on the class via `self_dispatch_targets` (the same class-anchored resolver
Python uses for `self.M()`) and emit the self-dispatch edge(s).

Crucially, this is gated by `_js_this_is_class_instance`: arrows inherit `this`
lexically, but a plain nested `function(){}` / generator / function expression
REBINDS `this` per call, so `this.M()` there does NOT refer to the instance. The
gate walks up from the call, treating arrows as transparent and stopping at the
first `this`-binding scope (method/class = instance; plain function = dynamic),
so no false edge is minted for the plain-function case (caught in adversarial
review; the JS same-syntax/different-semantics trap).

## Tests
`test_ts_this_dispatch_nested_arrow.py` (RED->GREEN):
- `this.M()` in a nested arrow under cross-class ambiguity resolves to the
  enclosing class (and NOT the sibling class).
- `this.M()` in a plain nested `function(){}` is NOT anchored (no false edge).

Verified matrix: direct method body, nested arrow, arrow-in-arrow, class-field
arrow -> anchored; plain nested function, arrow-inside-plain-function -> not
anchored. Full unit suite: 5961 passed, 5 pre-existing env skips, 0 failed.
ruff + ty + pre-commit clean; adversarial diff review (which caught the plain-
function defect) re-verified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JavaScript and TypeScript call resolution for `this.method()` inside class methods and nested arrow functions.
  * Prevented incorrect method links when `this` is rebound inside nested regular functions.

* **Tests**
  * Added coverage for `this`-based calls in nested arrow functions and regular nested functions.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -372,6 +372,16 @@ _JSX_NAMED_ELEMENT_TYPES = frozenset(
 # (scope.onSuccess), so a passed object of callbacks must reference each or
 # every TanStack-style callback reports as dead.
 _INLINE_FUNC_VALUE_TYPES = frozenset({cs.TS_ARROW_FUNCTION, cs.TS_FUNCTION_EXPRESSION})
+# JS/TS function scopes that REBIND `this` per call (a plain function /
+# generator / function expression), as opposed to an arrow_function (inherits
+# `this` lexically) or a method_definition / class-field arrow (binds `this` to
+# the instance). Used to tell whether a `this.M()` call site lexically refers to
+# the enclosing class instance.
+_JS_THIS_REBINDING_FUNC_TYPES: frozenset[str] = frozenset(
+    t
+    for t in cs.JS_TS_FUNCTION_NODES
+    if t not in (cs.TS_ARROW_FUNCTION, cs.TS_METHOD_DEFINITION)
+)
 
 
 def _scope_qn_candidates(scope_qn: str) -> list[str]:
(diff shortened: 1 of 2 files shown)


vitali87:
@coderabbitai review

PR #977, opened 2026-07-29
feat: NestJS-aware dead-code roots for DI classes and external OptionsFactory implementers

### LINKED ISSUE #973: NestJS-aware dead-code roots: @Injectable/@Controller/@Module constructors, lifecycle hooks, and interface-factory methods
## Summary
NestJS runs a lot of code the static graph cannot see a call to, producing dead-code false positives for the DI/framework idiom. Add NestJS-aware reachability roots, mirroring cgr's existing Flask/FastAPI (`@route`) and C# (`[HttpGet]`) framework roots.

## Patterns (from dogfooding NestJS, ~24 findings in example apps)
- **DI-instantiated constructors** (9): `@Injectable()` / provider classes are instantiated by the DI container, never via a first-party `new`, so their constructor is reported dead.
- **Interface-factory methods** (7): `createTypeOrmOptions()` / `createGqlOptions()` etc. implement `...OptionsFactory` interfaces and are invoked by Nest.
- **Lifecycle hooks** (2): `onModuleInit`, `onApplicationBootstrap`, `onApplicationShutdown`, `beforeApplicationShutdown`, `onModuleDestroy` are invoked by name (like Python dunders).

## Fix direction (user-approved: NestJS-aware roots)
- Root class constructors decorated `@Injectable`/`@Controller`/`@Module` (decorator-name roots, like DEFAULT_ROOT_DECORATORS).
- Root the NestJS lifecycle-hook method names on `.ts` files (name-based, like PY_ENUM_HOOK_METHOD_NAMES / dunders).
- Root methods of a class that `implements` a known NestJS `...OptionsFactory` interface, or accept the `overrides_external` path if the interface resolves.

Note: `@Get`/`@Post` HTTP decorators already root correctly (`_norm_decorator` lowercases). This covers the rest.

vitali87:
## Summary
NestJS runs a lot of code the static graph sees no call to (the DI/framework
idiom), producing dead-code false positives. This adds NestJS-aware reachability
roots, mirroring cgr's existing Flask/FastAPI (`@route`) and C# (`[HttpGet]`)
framework roots:

- **DI-decorated classes** (`@Injectable`/`@Controller`/`@Module`/`@Catch`/
  `@Resolver`/`@WebSocketGateway`) are instantiated by the container and driven
  by Nest, so their **constructor** and **lifecycle/interface-contract methods**
  (`onModuleInit`, `onApplicationBootstrap`, `use`, `intercept`, `configure`,
  `canActivate`, `catch`, `transform`, ...) are roots. The framework method names
  root only on a NestJS component class, so a same-named ordinary method on a
  plain class is not force-rooted.
- **External-interface implementers**: a class with an `IMPLEMENTS` edge to an
  interface outside the project (e.g. `@nestjs.typeorm.TypeOrmOptionsFactory`)
  has its methods invoked by Nest, so they are roots (mirrors the existing
  `overrides_external` rule). This required adding `IMPLEMENTS` to the dead-code
  rel-fetch query, which previously omitted it.

Validated by re-running dead-code on a full NestJS index: **24 -> 9 findings**.
The remaining 9 carry no NestJS DI signal (a provider without `@Injectable`, a
factory without an `implements` clause, genuinely-dead example helpers), so they
are correctly out of scope for this decorator/interface-driven approach.

## Type of Change
- [x] Feature (dead-code precision; eliminates the NestJS DI false-positive class)

## Related Issues
Closes #973.

## Test Plan
- [x] `test_nest_dead_code_roots.py`: RED->GREEN for (a) `@Injectable` constructor
  + framework methods rooted while a plain private helper stays dead, (b) external
  `OptionsFactory` methods rooted, (c) a same-named `use` on a plain class stays
  dead (no over-rooting), (d) the rel query fetches `IMPLEMENTS`.
- [x] Existing dead-code root suites (C#, C/C++, collect) still pass.
- [x] Full unit suite: 5965 passed, 5 pre-existing env skips, 0 failed.
- [x] Real NestJS repo: 24 -> 9 dead-code findings, no over-rooting.
- [x] ruff / ty / pre-commit clean; adversarial diff review found no defect.

## Checklist
- [x] RED demonstrated before the change
- [x] Framework roots gated to JS/TS extensions
- [x] Known, intended tradeoff documented below

## Note (intended tradeoff)
The external-interface rule roots every method of a class implementing an
external interface (not only `create*Options`), consistent with how NestJS
contract interfaces work and with `overrides_external`. A private helper on such
a class is therefore treated as reachable; in practice these classes are small
factories/guards and their helpers are reached through normal CALLS edges anyway.

## Summary by CodeRabbit

- **New Features**
  - Enhanced dead-code analysis for NestJS apps by treating DI-instantiated classes, constructors, and framework-invoked lifecycle methods as reachable.
  - Expanded rooting to include external `...OptionsFactory` implementations detected via interface implementation relationships.

- **Tests**
  - Added/updated coverage to verify Nest-specific roots are retained (including when class rooting is enabled).
  - Added negative coverage to ensure unrelated third-party interface helpers remain flagged as dead.
  - Confirmed reachability logic now includes `IMPLEMENTS` relationships for external interface rooting.

### DIFF
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -210,6 +210,48 @@ TEST_PATH_PATTERNS: tuple[str, ...] = (
     "__tests__",
 )
 
+# NestJS component decorators that mark a CLASS as instantiated and driven by
+# the DI container / framework, never by a first-party `new` the graph can see:
+# `@Injectable` (providers/services), `@Controller`, `@Module`, `@Catch`
+# (exception filters), `@Resolver` (GraphQL), `@WebSocketGateway`. Such a class's
+# constructor is invoked by the container and its framework-contract methods by
+# Nest, so both are reachability roots (gated by a JS/TS extension). Names are the
+# lowercased, argument-stripped form _norm_decorator produces.
+NEST_ROOT_CLASS_DECORATORS: frozenset[str] = frozenset(
+    {
+        "injectable",
+        "controller",
+        "module",
+        "catch",
+        "resolver",
+        "websocketgateway",
+    }
+)
+
+# NestJS methods invoked by the framework through a lifecycle or interface
+# contract, never by a named call the graph can see: lifecycle hooks
+# (`onModuleInit`, `onApplicationBootstrap`, ...) and the single-method
+# interface contracts (`NestMiddleware.use`, `NestInterceptor.intercept`,
+# `NestModule.configure`, `CanActivate.canActivate`, `ExceptionFilter.catch`,
+# `PipeTransform.transform`). Rooted only on a method of a NestJS component
+# class (a NEST_ROOT_CLASS_DECORATORS decorator), so a same-named ordinary
+# method (`use`, `transform`) on a plain class is NOT force-rooted.
+NEST_FRAMEWORK_METHOD_NAMES: frozenset[str] = frozenset(
+    {
+        "onModuleInit",
+        "onModuleDestroy",
+        "onApplicationBootstrap",
+        "onApplicationShutdown",
+        "beforeApplicationShutdown",
+        "configure",
+        "use",
+        "intercept",
+        "canActivate",
+        "catch",
+        "transform",
+    }
+)
+
 # Python Enum protocol hooks: the enum machinery invokes these sunder
 # METHODS by NAME (_generate_next_value_ on auto(), _missing_ on a failed
 # value lookup), never through a call the graph can see -- runtime roots
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -228,6 +228,14 @@ NEST_ROOT_CLASS_DECORATORS: frozenset[str] = frozenset(
     }
 )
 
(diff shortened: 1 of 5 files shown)


vitali87:
Both addressed in 72728b39 (fixed, not rebutted):

- **Root component class nodes (dead_code.py):** With `--classes`, a non-exported `@Injectable`/`@Controller`/`@Module` class (and an external `OptionsFactory` implementer) now roots itself. A rooted constructor cannot revive the class node because `DEFINES_METHOD` is not a reachability edge, so `_is_nest_root` now returns true for the CLASS candidate directly. Added `test_component_class_node_is_rooted_with_include_classes` (RED→GREEN).
- **Restrict external-interface roots to Nest factory contracts:** the IMPLEMENTS-external rule now only fires when the interface leaf ends with `OptionsFactory` (the NestJS async-options convention), so an unrelated third-party interface implementer keeps its private helpers reportable. Added `test_non_nest_external_interface_helper_stays_dead` (RED→GREEN).

Re-validated on the full NestJS index: still 24 → 9, and the `create*Options` factory methods stay resolved (all their interfaces end in `OptionsFactory`). Full suite: 5967 passed, 5 env skips, 0 failed.

PR #979, opened 2026-07-29
feat: React class-component lifecycle methods are dead-code roots

### LINKED ISSUE #978: React class-component lifecycle methods are dead-code false positives
## Summary
A React class component's lifecycle methods (`render`, `componentDidMount`, `componentWillUnmount`, `componentDidUpdate`, the constructor, ...) are invoked by React at runtime, never by an explicit call the graph can see, so `cgr dead-code` reports them dead. Because the reachability walk starts from these (unreached) methods, every method they call via `this.` and every nested callback inside them is also reported dead, so an entire live component collapses into the dead list.

This is the same class of framework-invocation root as the NestJS lifecycle hooks (#973), Python dunders, and Java serialization hooks.

## Evidence (dogfooding excalidraw/excalidraw)
`cgr dead-code` reports **161** candidates; **123** of them are under three React class components:

- `packages/excalidraw/components/App.tsx` &mdash; `class App extends React.Component` (the main editor component): **89** dead items (`componentDidMount`, `componentDidUpdate`, `componentWillUnmount`, `constructor`, and the methods/handlers they reach).
- `excalidraw-app/collab/Collab.tsx` &mdash; `class Collab extends PureComponent`: **34** dead items.
- `excalidraw-app/components/TopErrorBoundary.tsx` &mdash; `class TopErrorBoundary extends React.Component`.

All three are exported and rendered via JSX, so their lifecycle methods genuinely run. The graph already records the `INHERITS` edge to `react.Component` / `react.PureComponent`.

## Fix direction
Add a React-aware reachability root: a method whose name is a React lifecycle method AND whose class `INHERITS` (transitively is fine to start with direct) from a React component base (`Component` / `PureComponent`) is a root. Gated to JS/TS, mirroring `_is_nest_root` (#973). Lifecycle set: `render`, `constructor`, `componentDidMount`, `componentDidUpdate`, `componentWillUnmount`, `shouldComponentUpdate`, `getSnapshotBeforeUpdate`, `componentDidCatch`, `getDerivedStateFromProps`, `getDerivedStateFromError`, and the legacy/UNSAFE_ variants.

vitali87:
## Summary
A React class component's lifecycle methods (`render`, `componentDidMount`,
`componentDidUpdate`, `componentWillUnmount`, the constructor, ...) are invoked by
React at runtime, never by an explicit call the graph can see, so `cgr dead-code`
reported them dead. Since the reachability walk starts from these (unreached)
methods, everything they call via `this.` and every nested callback collapsed
into the dead list too, so an entire live component was reported dead.

This adds a React-aware reachability root, mirroring the NestJS (#973), Python
dunder, and Java serialization roots: a method whose name is a React lifecycle
method AND whose class (transitively) extends a React component base is a root.

Validated by dogfooding excalidraw/excalidraw: dead-code candidates **161 → 78**.
Collab 34 → 1, App (the main editor component) 89 → 41.

## Type of Change
- [x] Feature (dead-code precision; eliminates the React class-component false-positive class)

## Related Issues
Closes #978.

## Test Plan
- [x] `test_react_dead_code_roots.py` (RED→GREEN): lifecycle methods + the helpers
  they reach via `this.` are rooted while an unreferenced private helper stays
  dead; a lifecycle-named method on a plain class stays dead; a non-React
  `Component`-named base (Glimmer) does NOT root; a transitive React base
  (`App extends BaseWidget extends react.Component`) does root.
- [x] Full unit suite: 5971 passed, 5 pre-existing env skips, 0 failed.
- [x] Real excalidraw index: 161 → 78, no wins lost after the precision/transitivity fixes.
- [x] ruff / ty / pre-commit clean; adversarial diff review (which surfaced the
  namespace-precision and transitive-inheritance points, both fixed) re-verified.

## Checklist
- [x] RED demonstrated before the change
- [x] React base match gated to the `react` namespace (not just the simple name)
- [x] Transitive inheritance to a React base covered

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dead-code detection for React class components.
  * React lifecycle methods, constructors, and reachable helper methods are now correctly recognized as active.
  * Added support for components inheriting from `React.Component` or `React.PureComponent`, including indirect inheritance.
  * Prevented similarly named components from unrelated frameworks from being incorrectly treated as React components.

### DIFF
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -260,6 +260,45 @@ NEST_FRAMEWORK_METHOD_NAMES: frozenset[str] = frozenset(
     }
 )
 
+# React component base classes: a class that `extends` one of these is a class
+# component whose lifecycle methods React drives at runtime. Matched on the base
+# interface/class leaf name (`React.Component`, `PureComponent`, and the rarely
+# spelled-out `React.PureComponent`), so the INHERITS target's last segment
+# identifies it regardless of the import alias.
+REACT_COMPONENT_BASE_NAMES: frozenset[str] = frozenset({"Component", "PureComponent"})
+
+# The module-namespace token that distinguishes React's `Component`/`PureComponent`
+# base (`react.Component`, `React.Component`) from an unrelated base sharing that
+# simple name (Ember/Glimmer's `@glimmer/component.Component`). Lowercased match.
+REACT_NAMESPACE_TOKEN = "react"
+
+# React class-component lifecycle methods the runtime invokes (mount/update/
+# unmount/render/error), plus the constructor React calls when it instantiates
+# the component. Never called by a first-party call the graph can see, so they
+# are reachability roots on a React component class (gated by INHERITS to a
+# REACT_COMPONENT_BASE_NAMES base and a JS/TS extension); the methods and
+# callbacks they reach via `this.` then expand from them.
+REACT_LIFECYCLE_METHOD_NAMES: frozenset[str] = frozenset(
+    {
+        "render",
+        "constructor",
+        "componentDidMount",
+        "componentDidUpdate",
+        "componentWillUnmount",
+        "shouldComponentUpdate",
+        "getSnapshotBeforeUpdate",
+        "componentDidCatch",
+        "getDerivedStateFromProps",
+        "getDerivedStateFromError",
+        "componentWillMount",
+        "componentWillReceiveProps",
+        "componentWillUpdate",
+        "UNSAFE_componentWillMount",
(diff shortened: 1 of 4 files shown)


vitali87:
Addressed in ae1f23da: `_is_react_base_qn` now requires the base namespace to equal `react` EXACTLY (lowercased), so `preact.Component` / `notreact.Component` are no longer taken for React. Added `test_react_lookalike_namespace_is_not_rooted` (RED→GREEN). Real excalidraw index unchanged at 78 (its bases are `react.PureComponent` / `React.Component`, namespace exactly `react`/`React`). Full suite: 5972 passed, 5 env skips, 0 failed.

PR #983, opened 2026-07-29
fix: link JS/TS fn.bind(this)() immediately-invoked calls

### LINKED ISSUE #982: TypeScript: function invoked as `fn.bind(this)()` loses its CALLS edge (false-positive dead code)
## Summary

A locally defined JS/TS function invoked immediately through `fn.bind(this)()` records no CALLS edge to `fn`. The plain `fn()` form works. As a result the function (and anything reachable only through it) is reported as unreachable by `cgr dead-code`.

## Where it was found

Dogfooding `typeorm/typeorm`. `src/query-builder/JoinAttribute.ts` defines a local `getValue` inside the `relation` getter and invokes it as:

```ts
this.relationCache = getValue.bind(this)()
```

The sibling `isSelected` getter defines a `getValue` too but calls it as plain `getValue()`, which links correctly. Only the bound-then-invoked form drops the edge.

## Root cause

`CallProcessor._get_call_target_name` derives the callee name from a call node's `function` field. It handles an `identifier`/`member_expression` callee, but for `fn.bind(ctx)()` the callee is a nested `call_expression` (`fn.bind(ctx)`), which matches no case, so no name is produced and no edge is emitted. The repo already has `_unwrap_bound_function` (covers `bind`/`call`/`apply`) used for the reference path; the call-name path simply never consults it.

## Reproduction

```ts
export class C {
  get direct() { const getValue = () => 1; return getValue(); }        // linked
  get bound()  { const getValue = () => 2; return getValue.bind(this)(); } // NOT linked
}
```

`direct` emits CALLS to `getValue`; `bound` emits nothing.

vitali87:
## Summary

- Resolve JS/TS calls of an immediately invoked bound function (`fn.bind(this)()`) to the underlying callable `fn`, so the CALLS edge is recorded instead of dropped.
- Peel to a fixpoint so chained binds and interleaved casts (`fn.bind(a).bind(b)()`) also resolve; a plain call-of-a-call (`getFactory()()`) stays unnamed.
- Fixes a dead-code false positive found dogfooding TypeORM's `JoinAttribute.relation` (`getValue.bind(this)()`).

## Type of Change

- [x] Bug fix

## Related Issues

Closes #982

## Test Plan

- [x] New tests added

New tests in `test_ts_bound_immediate_call_edge.py`: `bound` (`fn.bind(this)()` links), `chained` (fixpoint peel), `direct` (plain `fn()` regression guard), `call_of_call` (`getFactory()()` links only the inner call, no fabricated outer edge).

## Checklist

Scope note: idiomatic single `fn.call(x)` / `fn.apply(x)` are intentionally left unchanged; their member-expression callee is ambiguous with a real method named `call`/`apply`, so this PR does not touch them.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -2111,6 +2111,19 @@ class CallProcessor:
                             receiver = receiver_node.text.decode(cs.ENCODING_UTF8)
                             return f"{receiver}{cs.SEPARATOR_DOT}{method}"
                         return method
+                case cs.TS_CALL_EXPRESSION if language in _JS_TS_LANGUAGES:
+                    # A bound function that is itself invoked (`fn.bind(ctx)()`)
+                    # has a nested `fn.bind(ctx)` call_expression as its callee, so
+                    # no earlier case names it. Resolve to fn so the invocation
+                    # links to fn, not the Function.prototype builtin (issue:
+                    # JoinAttribute.relation's local getValue called as
+                    # getValue.bind(this)()). Guard on _unwrap_bound_function so a
+                    # plain call-of-a-call (`getFactory()()`) stays unnamed, then
+                    # peel to a fixpoint for chained binds / interleaved casts.
+                    if self._unwrap_bound_function(func_child) is not None:
+                        peeled = self._peel_bound_callable(func_child)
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile review

Issue #994, opened 2026-07-31
fix: generator function expressions bound to variables are not registered as function nodes (JS/TS)


vitali87:
## Summary

A generator function expression bound to a variable, `const gen = function* () { yield 1 }`, is not registered as a function node by the definition pass. The declaration form `function* gen () {}` registers correctly. Because the node does not exist, no reference or call edge can ever target it: any use of `gen` keeps nothing alive and the generator's body is invisible to dead-code analysis and retrieval.

## Repro

```js
function main () {
  const gen = function* () { yield 1 }
  gen.call(this)
  return 2
}
module.exports = { main }
```

No node with `gen` in its qualified name is created; a plain `const f = function () {}` in the same position registers as `main.f`. Found while hardening the #988 receiver resolution: the lexical binder recognises the generator value, but the span lookup correctly refuses an unregistered node, so the call emits nothing (pinned by `test_generator_function_expression_receiver_is_safe`).

## Fix direction

Wherever the JS/TS ingest accepts `function_expression` as a declarator value (binding-name registration, deferred anonymous registration), accept `generator_function` the same way.

vitali87:
PR #1003 fixes this, and grew to close the whole IIFE orphan family the registration exposed (parenthesised, direct, named, sequence, nested wrappers to fixpoint, branch-valued callees). Eight adversarial review rounds; corpus deltas: fastify unchanged, excalidraw three genuine false positives cleared.

vitali87:
Fixed by PR #1003 (merged): generator function expressions register, name and reference like function expressions, and the whole IIFE orphan family is closed (parenthesised, direct, named span-resolved, sequence, nested wrappers to fixpoint, branch-valued callee references, anonymous generator scope bubbling).

PR #995, opened 2026-07-31
fix: resolve statement-position fn.call and fn.apply receivers by unique relevant binding

### LINKED ISSUE #988: fix: identifier-receiver fn.call/fn.apply in statement position emits no call edge (JS/TS)
## Summary

A JavaScript or TypeScript call of the form `fn.call(this, ...)` or `fn.apply(this, ...)`, where `fn` is a plain identifier that resolves to a known function, emits no CALLS or REFERENCES edge when the call is an expression statement. Functions whose only invocations use this idiom are falsely reported dead.

## Witness (fastify)

`lib/server.js` invokes `multipleBindings.call(this, server, httpHandler, options, listenOptions, cb)` at two sites; `lib/plugin-utils.js` `registerPlugin` invokes `registerPluginName.call(this, fn)`, `checkPluginHealthiness.call(this, fn, pluginName)` and `checkVersion.call(this, fn)`; `lib/route.js` invokes `validateSchemaBodyOption.call(this, ...)` and `addNewRoute.call(this, ...)`; `lib/four-oh-four.js` invokes `_setNotFoundHandler.call(this, ...)`; `fastify.js` invokes `_addHook.call(this, name, fn)`. All twelve affected functions (including the four closures nested inside `multipleBindings`) appear in the dead-code report even though they run on every boot.

## Repro

```js
'use strict'
function helper (x) { return x + 1 }
function main () {
  helper.call(this, 1)
  return 2
}
module.exports = { main }
```

No edge into `helper` is emitted. Notably, when the same call appears in a value position (`return helper.call(this, 1)` or `const x = helper.call(this, 1)`), the value-reference pass peels the bound callable and emits a REFERENCES edge, which is why sibling functions such as `normalizeAndValidateMethod` (always invoked in assignment position) are not flagged. Statement position emits nothing.

## Root cause direction

PR #983 deliberately left `fn.call(...)` and `fn.apply(...)` unresolved because a member-expression callee named `call` is ambiguous with a real method of that name. That ambiguity does not exist when the receiver is a plain identifier that resolves in scope to a function node: `.call` on a function value is `Function.prototype.call`. For that receiver shape a CALLS edge to the resolved function is sound.

vitali87:
Fixes #988.

## Problem

A JS/TS invocation of the form `fn.call(this, ...)` or `fn.apply(this, ...)`, where `fn` names a first-party function, emitted no edge when the call was an expression statement. Functions whose only invocations use this idiom were falsely reported dead: twelve fastify candidates, including `multipleBindings` (with its four nested closures), the plugin-utils trio (`registerPluginName`, `checkPluginHealthiness`, `checkVersion`), `validateSchemaBodyOption`, `addNewRoute`, `_setNotFoundHandler` and `_addHook`.

## Root cause

PR #983 deliberately left `fn.call(...)`/`fn.apply(...)` unresolved because a member callee named `call` is ambiguous with a real method of that name. Value positions (`return fn.call(...)`, `x = fn.call(...)`) still linked via the bound-callable peel in the value-reference passes, which is why only statement-position invocations lost their targets.

## Fix: unique-relevant-binding resolution

For a call whose callee is `<receiver>.call` or `<receiver>.apply` (casts and parens peeled to a bare identifier), the receiver resolves through a whole-file binding index, never through name lookup in the resolver trie:

- `_js_collect_file_bindings` makes ONE cached pass per file collecting every INTRODUCTION of every identifier: callable parameters (binding side only, skipping default values and TS type annotations), hoisted function and generator declarations, function expression self-names, variable declarators (recording an inline arrow/function/generator value as the bound function; `var` containers widen to the enclosing callable body, `let`/`const` to the declaration's parent, peeling `export_statement` and widening switch-case slots to the switch body), for-in/for-of lefts, catch parameters, class names, import clause names, and reassignment targets including destructuring. A `with` block poisons the whole file.
- At a call site, introductions are filtered to those whose container span encloses the site (file-wide entries always qualify). The edge is emitted only when EXACTLY ONE relevant introduction remains and it is a function registered by the definition pass, resolved by its own source span via `_recorded_caller` with the recorded label.

Two or more relevant introductions, a non-function introduction, or an unbound name emit nothing. This makes every shadowing, `var` hoisting, switch/for/catch scoping and duplicate-declaration question collapse into "a second relevant introduction" with no scope simulation, while an introduction confined to a sibling scope cannot suppress a legitimate edge. Nothing else about the site changes: name resolution, builtin handling and argument references proceed exactly as before, so callable-parameter flow never sees the `this`-shifted argument list.

## Review history

Seven adversarial review rounds shaped this, each verified against the live pipeline; every finding is pinned as a test. Round 1 killed a `callee_info` route (positional flow shifted one slot by the `this` argument). Round 2 killed a resolver-name-lookup route (trie cross-file edges; parameter and declarator shadowing). Round 3 killed a hand-rolled scope walk (var hoisting, C-style for and switch scopes, method self-name phantom self-loops, hardcoded label). Round 4 killed its successor (for-of vars, block function declarations vs var redeclaration, static blocks, quadratic per-site scans). Round 5 refined the whole-file rule (export const containers, default values and `typeof` in parameter scans, sibling-scope suppression recovered via container relevance, static-block containment, destructuring reassignment, `with`). Round 6 widened hoisted declaration containers to the enclosing callable body (switch-case liveness and Annex B block declarations) and added TS enum, namespace and `import x = require(...)` bindings. Round 7 made the per-file index reset for reused updaters (watch mode re-parses; stale nodes resolved spans onto the wrong function), made namespace bodies their own containers, collected the namespace `import x = Other.thing` alias form, and taught the parameter scan the TS grammar's wrapperless defaults. The closing rounds hardened TS namespace declaration merging: exported direct members of a repeated namespace replicate one introduction per sibling block (never a container widening, which could grow relevance from zero and invent edges), merge groups key on the dotted namespace path from the nearest non-namespace container so `A { N }`, `A.N` and a top-level `N` group exactly as TS does, and all node bookkeeping uses tree-sitter's stable node ids so fresh Python wrappers cannot break identity checks.

## Validation

- 58 tests in `codebase_rag/tests/test_js_fn_call_apply_statement_edge.py`: positives (statement `.call`/`.apply`, nested arrows, casts, export const arrows, default-value and `typeof` non-suppression in both grammars, sibling-scope locals, switch case, namespace-internal calls, self-recursion, reused-updater freshness) and guards pinning every review finding.
- Real-repo verification on fastify after EVERY design iteration: dead-code candidates 36 to 24, exactly the twelve `.call`/`.apply` victims revived every time, no other candidate changed. The final reviewer sweep instrumented 28,897 candidate sites across fastify (27 resolved, all same-file, no false link, `route.call` in `lib/route.js` recovered) and confirmed zero resolutions and zero false edges across typeorm, zod, NestJS core and excalidraw sweeps.
- Discovered and filed separately: #994 (generator function expressions are not registered as function nodes).

## Known accepted approximations

- Flow-insensitive: `let f = () => 1; f = other; f.call()` counts two introductions and suppresses (conservative miss, never a wrong edge).
- The per-file binding cache stores tree-sitter nodes, keeping parsed trees alive for the duration of the call pass only; the pass already retains its AST cache, and the cache dies with the `CallProcessor`.

## Summary by CodeRabbit

* **New Features**
  * Improved JavaScript and TypeScript call tracking for `.call()` and `.apply()`.
  * More accurately resolves function references across imports, aliases, destructuring, namespaces, scopes, hoisting, generators, classes, and inline functions.

* **Bug Fixes**
  * Prevented stale or ambiguous bindings from creating incorrect call relationships.
  * Improved handling of reassigned, shadowed, and dynamically scoped functions.

* **Tests**
  * Added comprehensive coverage for call resolution, statement-position calls, and edge cases.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -2755,6 +2755,39 @@ class CallProcessor:
                 is_macro_target = callee_info[1] in self.macro_qns
                 if is_macro_target != (call_node.type == cs.TS_RS_MACRO_INVOCATION):
                     callee_info = None
+            if not callee_info and is_js_ts:
+                # `fn.call(this, ...)` / `fn.apply(this, ...)` on a plain
+                # identifier that resolves to a first-party function is
+                # Function.prototype.call, an invocation of fn. Value positions
+                # already link via the bound-callable peel; a STATEMENT
+                # invocation reached here nameless-in-effect and its target
+                # reported dead (fastify's `multipleBindings.call(this, ...)`
+                # and the plugin-utils trio, issue #988). Runs only after the
+                # primary resolution failed, so a genuine method named
+                # call/apply on a typed receiver keeps its own edge, and only
+                # for an identifier receiver: a member-expression receiver
+                # (`checks[instance].call(...)`, `router.route.call(...)`)
+                # stays ambiguous exactly as PR #983 left it.
+                receiver, sep, invoke = call_name.rpartition(cs.SEPARATOR_DOT)
+                if (
+                    sep
+                    and invoke in (cs.JS_METHOD_CALL, cs.JS_METHOD_APPLY)
+                    and receiver
+                    and cs.SEPARATOR_DOT not in receiver
+                    and (
+                        receiver_info := resolve_func(
+                            receiver,
+                            module_qn,
+                            call_var_types,
+                            class_context,
+                            caller_qn,
+                            language,
+                        )
+                    )
+                    is not None
+                    and receiver_info[0] == cs.NodeLabel.FUNCTION
+                ):
+                    callee_info = receiver_info
             if not callee_info and resolve_builtin is not None:
                 callee_info = resolve_builtin(call_name)
             if not callee_info and resolve_cpp_op is not None:
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -2762,18 +2762,28 @@ class CallProcessor:
                 # already link via the bound-callable peel; a STATEMENT
                 # invocation reached here nameless-in-effect and its target
                 # reported dead (fastify's `multipleBindings.call(this, ...)`
-                # and the plugin-utils trio, issue #988). Runs only after the
-                # primary resolution failed, so a genuine method named
-                # call/apply on a typed receiver keeps its own edge, and only
-                # for an identifier receiver: a member-expression receiver
+                # and the plugin-utils trio, issue #988). Only an identifier
+                # receiver qualifies: a member-expression receiver
                 # (`checks[instance].call(...)`, `router.route.call(...)`)
-                # stays ambiguous exactly as PR #983 left it.
+                # stays ambiguous exactly as PR #983 left it. Bare-name
+                # resolution guesses, so the guess is confined to bindings the
+                # name can actually mean here: never a caller PARAMETER, never
+                # a name the caller REBINDS locally (unless it resolved to that
+                # local function itself), and never outside this module (the
+                # project-wide trie would revive same-named functions in
+                # unrelated files). The edge is emitted directly and the site
+                # is finished: routing through callee_info would map the
+                # callable-param flow onto the RAW argument list, whose
+                # position 0 is the `this` value, binding parameters one slot
+                # off. Real function-valued arguments keep their reachability
+                # through the argument-reference pass instead.
                 receiver, sep, invoke = call_name.rpartition(cs.SEPARATOR_DOT)
                 if (
                     sep
                     and invoke in (cs.JS_METHOD_CALL, cs.JS_METHOD_APPLY)
                     and receiver
                     and cs.SEPARATOR_DOT not in receiver
+                    and receiver not in caller_params
                     and (
                         receiver_info := resolve_func(
                             receiver,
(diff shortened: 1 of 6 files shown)


codebase_rag/parsers/call_processor.py
```diff
@@ -6103,5 +6148,304 @@ def _is_unowned_js_scope(self, node: Node) -> bool:
...
+            or container.start_byte <= site.start_byte < container.end_byte
+        ]
+        if len(relevant) != 1:
+            return None
+        fn_node, _container = relevant[0]
+        if fn_node is None:
+            return None
+        return self._recorded_caller(fn_node, module_qn)
+
+    def _js_collect_file_bindings(
```
vitali87, line 6564: Applied the comprehension simplification in f6f7ad3d. On extracting the binding collector into its own module: deferring deliberately. The collector is coupled to the resolution path it serves (its output shape, container semantics and merge-group keys are consumed only by `_js_lexical_function_binding` three methods up), and the change set is already large; a mechanical extraction now would churn the review anchors for no behavioural gain. Happy to split it out in a follow-up refactor once #996 (ambient namespace members) lands, since that will touch the same walk.

vitali87:
@coderabbitai review

vitali87:
Applied the eviction suggestion in 8df3bb76: each module's receiver-binding entry is popped in a finally block when its file's call pass completes (errors and early returns included), so the map never retains more than the active file's tree.

vitali87:
@coderabbitai review

PR #997, opened 2026-07-31
fix: reference named operands of JS/TS logical default assignments

### LINKED ISSUE #990: fix: function referenced as a logical default operand in an assignment emits no reference edge (JS/TS)
## Summary

A function referenced as a fallback operand of a logical default expression in an assignment, `x = a || fn` or a chained `x = a || b || fn`, emits no REFERENCES edge. Default handlers wired this way are falsely reported dead.

## Witness (fastify)

`fastify.js` line 918: `options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler`. `lib/context.js` lines 69 to 72: `this.schemaErrorFormatter = schemaErrorFormatter || server[kSchemaErrorFormatter] || defaultSchemaErrorFormatter`. Both `defaultClientErrorHandler` and `defaultSchemaErrorFormatter` appear in the dead-code report even though they are the live defaults on every instance that does not override them.

## Repro

```js
'use strict'
function fallback () { return 1 }
function main (options) {
  options.handler = options.handler || fallback
  return options
}
module.exports = { main }
```

No edge into `fallback`.

## Design note

The assignment-value reference pass inspects the direct value and deliberately does not peel short-circuit expressions, a guard added because peeling `&&` inside JSX attributes falsely revived data props. In plain assignment position the semantics differ: the operands of `||`, `??` and the right operand of `&&` can each BE the assigned value, so a function identifier there genuinely flows as a value and must be referenced. The fix needs to stay gated away from the JSX attribute context that motivated the original guard.

vitali87:
Fixes #990.

## Problem

A function referenced as a fallback operand of a logical default in an assignment, `x = a || fn`, a chained `x = a || b || fn`, or `x = a ?? fn`, emitted no REFERENCES edge and was falsely reported dead. Fastify witnesses: `options.clientErrorHandler = options.clientErrorHandler || defaultClientErrorHandler` (fastify.js) and the chained `this.schemaErrorFormatter = schemaErrorFormatter || server[kSchemaErrorFormatter] || defaultSchemaErrorFormatter` (lib/context.js).

## Root cause

The assignment-reference pass scanned binary RHS operands for INLINE functions only (the express `done = done || function () {}` shape); a bare identifier operand was skipped, and nested logical chains were not flattened.

## Fix

For a binary RHS whose operator is `||`, `??` or `&&` (the operators that evaluate TO one of their operands, tested via the existing `cs.JS_SHORT_CIRCUIT_OPERATORS`), named operands in `_ASSIGNMENT_RHS_REF_TYPES` now emit REFERENCES like inline functions always did, with nested logical chains flattened through casts and parens. Non-logical operators keep the old inline-only behaviour, so `x === fn` stays silent.

The emission is gated to JS/TS via a `language` parameter threaded from all three call sites: Go reaches the same walk and spells boolean `||`/`&&` with the same `binary_expression` node, where a bool named like a method would fabricate a phantom edge (the adversarial review demonstrated exactly that with a bool parameter reviving an unrelated method; the repro is pinned RED to GREEN).

## Validation

- 7 tests in `codebase_rag/tests/test_js_logical_default_reference.py`: the two fastify shapes, `??`, `&&`, the comparison guard, the pre-existing inline-function shape, and the Go phantom-edge guard.
- Real-repo verification on fastify: dead-code candidates 36 to 34 relative to main; both OR-default victims revived, nothing else changed. Reviewer verification covered `.js`, `.ts` and `.tsx` plus Go and Python behaviour, confirming Go and Python are byte-identical to pre-change.

## Noted, pre-existing, not addressed here

- The left operand of the idiom (`o.bar = o.bar || fn`) resolves through the loose trailing-name member resolution shared with the ternary path, which can reference a same-named member on an unrelated class.
- A ternary nested inside a logical chain (`a || (c ? f1 : f2)`) emits nothing on either side's pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code reference tracking for JavaScript and TypeScript assignments using `||`, `??`, and `&&`.
  * Nested logical expressions now correctly identify referenced callable functions.
  * Reduced incorrect references from comparisons and Go boolean expressions.
  * Preserved support for inline fallback functions and language-aware scanning.

* **Tests**
  * Added regression coverage for logical defaults, chained expressions, inline functions, comparisons, and Go expressions.

### DIFF
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -212,6 +212,10 @@ FIELD_FINALIZER = "finalizer"
 FIELD_INCREMENT = "increment"
 
 # JS/TS module system node types
+# The operator field of a JS/TS binary_expression, and the short-circuit
+# operators that evaluate to one of their operands.
+JS_FIELD_OPERATOR = "operator"
+JS_LOGICAL_OPERATORS = frozenset({"||", "??", "&&"})
 TS_OBJECT_PATTERN = "object_pattern"
 TS_ARRAY_PATTERN = "array_pattern"
 TS_REST_PATTERN = "rest_pattern"
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -212,10 +212,6 @@ FIELD_FINALIZER = "finalizer"
 FIELD_INCREMENT = "increment"
 
 # JS/TS module system node types
-# The operator field of a JS/TS binary_expression, and the short-circuit
-# operators that evaluate to one of their operands.
-JS_FIELD_OPERATOR = "operator"
-JS_LOGICAL_OPERATORS = frozenset({"||", "??", "&&"})
 TS_OBJECT_PATTERN = "object_pattern"
 TS_ARRAY_PATTERN = "array_pattern"
 TS_REST_PATTERN = "rest_pattern"
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -197,6 +197,8 @@ TS_JS_TERNARY_EXPRESSION = "ternary_expression"
 # Short-circuit operators whose result IS one of the operands, so a
 # bind through them unions both operands' taints.
 JS_SHORT_CIRCUIT_OPERATORS: frozenset[str] = frozenset({"||", "??", "&&"})
+# `&&` alone among them cannot yield its LEFT operand as a function value.
+JS_OPERATOR_LOGICAL_AND = "&&"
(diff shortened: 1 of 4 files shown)


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
Both findings applied in 0b73fe4d: the operand flatten now threads value position (the left operand of && can never be the result, since a function value is always truthy; || and ?? flow both operands), pinned by a new test; and the test helper now asserts REFERENCES explicitly.

PR #999, opened 2026-07-31
fix: link whole-module commonjs function exports and load-time export wrappers

### LINKED ISSUE #991: fix: calling a module.exports = function expression through a required alias emits no call edge (CommonJS)
## Summary

When a module exports a function expression directly, `module.exports = function name (...) {...}`, a consumer that requires the module under a local alias and calls it, `const f = require('./mod'); f(x)`, produces no CALLS or REFERENCES edge into the exported function. The function and everything nested inside it are falsely reported dead.

## Witness (fastify)

`lib/error-serializer.js` (an autogenerated file) is exactly `module.exports = function anonymous (validator, serializer) {...}`. `lib/error-handler.js` consumes it with `const serializeError = require('./error-serializer')` and calls `serializeError({...})` at three sites. Both `anonymous` and its nested `anonymous0` appear in the dead-code report.

## Repro

```js
// ser.js
module.exports = function anonymous (v) { return String(v) }
// b.js
const serialize = require('./ser')
function main () { return serialize(1) }
module.exports = { main }
```

`main` gains no edge into `ser.anonymous`. Object-literal exports, `module.exports = { fn }`, link correctly; the direct function-expression export is the missing case.

vitali87:
Fixes #991.

## Problem

A CommonJS module whose entire export is one function was invisible to callers. Two shapes, both from fastify's generated `lib/error-serializer.js` family:

- `module.exports = function name (...) {...}`: a consumer's `const f = require('./mod'); f(x)` produced no edge into the exported function.
- `module.exports = function anonymous (...) {...}(args)` (the fast-json-stringify standalone output, with or without parentheses around the callee): the wrapper runs at module load and the export is its RETURN value; neither the wrapper nor its returned inner callable had any inbound edge, so both reported dead.

## Fix

A new tree-sitter query recognises `module.exports = <function|arrow|call>`. The definition pass COLLECTS the function nodes (skipping exports inside functions entirely, and peeling parenthesised IIFE callees), then finalises AFTER the deferred anonymous flush, resolving each node by its own SOURCE SPAN against the span registry:

- a direct export registers as an exported function (own name, or positional when anonymous), gains an `is_exported` merge, and records `module qn to function qn` in the new `ImportProcessor.commonjs_direct_exports` map; `_try_resolve_direct_import` falls back to that map when a whole-module require alias resolves to a module qn absent from the function registry, so the consumer call links;
- an immediately invoked wrapper gets a `Module CALLS wrapper` edge with the span-recorded label and qn (its returned callables stay alive through their internal references); the alias map is deliberately not populated for this shape, since the export is the return value, not the wrapper.

Span-first resolution means a name collision targets the wrapper's actual minted variant, a refused registration emits nothing, and a STALE span claim after a watch-mode removal is guarded by a registry membership check so nothing sparse is ever resurrected. The map is invalidated in `parse_imports` (beside the other per-module resets) and in `remove_file_from_state`.

## Review history

One adversarial review round over the live pipeline found six defects in the first design (all rooted in deriving qns from names: collision mistargeting, phantom edges for in-function exports, a wrong cross-module alias, an uninvalidated map, the parenthesised IIFE gap, and inconsistent export marking); the span-finalised redesign fixed all six, and the follow-up round confirmed them fixed while catching the stale-claim resurrection, closed by the registry guard. Every finding is pinned as a test.

## Validation

- 11 tests in `codebase_rag/tests/test_js_commonjs_direct_function_export_call.py`: the direct, anonymous-arrow and object-export forms, both IIFE spellings, the collision variant, in-function silence and no-alias, re-index map clearing, and the sparse-resurrection guard (RED against the watch-mode removal sequence).
- Real-repo verification on fastify: dead-code candidates 36 to 34 relative to main; both `error-serializer` candidates (`anonymous`, `anonymous0`) revived through the module-load CALLS edge and the internal reference.

## Summary by CodeRabbit

- **New Features**
  - Improved JavaScript/CommonJS support for modules that export a function directly.
  - Added more accurate linking when importing and calling these exported functions, including wrapped and immediately invoked functions.

- **Bug Fixes**
  - Prevented stale export references after files are reprocessed or removed.
  - Preserved correct behavior for object exports and avoided incorrect matches for nested functions or naming collisions.

- **Tests**
  - Added comprehensive coverage for direct exports, aliases, wrappers, reindexing, and related edge cases.

### DIFF
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -262,6 +262,17 @@ JS_COMMONJS_EXPORTS_FUNCTION_QUERY = """
 """
 
 # JS/TS CommonJS module.exports query
+# The DIRECT form `module.exports = function (...) {...}` / arrow: the whole
+# module IS one function (fastify's generated error-serializer). Two-level
+# member on the left, unlike the three-level named-property form below.
+JS_COMMONJS_DIRECT_EXPORT_QUERY = """
+(assignment_expression
+  left: (member_expression
+    object: (identifier) @module_obj
+    property: (property_identifier) @exports_prop)
+  right: [(function_expression) (arrow_function)] @export_function)
+"""
+
 JS_COMMONJS_MODULE_EXPORTS_QUERY = """
 (assignment_expression
   left: (member_expression
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -270,7 +270,7 @@ JS_COMMONJS_DIRECT_EXPORT_QUERY = """
   left: (member_expression
     object: (identifier) @module_obj
     property: (property_identifier) @exports_prop)
-  right: [(function_expression) (arrow_function)] @export_function)
+  right: [(function_expression) (arrow_function) (call_expression)] @export_function)
 """
 
 JS_COMMONJS_MODULE_EXPORTS_QUERY = """
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1300,6 +1300,9 @@ class GraphUpdater:
             else relative_path.with_suffix("").parts
         )
         module_qn_prefix = cs.SEPARATOR_DOT.join([self.project_name, *path_parts])
+        self.factory.import_processor.commonjs_direct_exports.pop(
+            module_qn_prefix, None
+        )
 
         qns_to_remove = set()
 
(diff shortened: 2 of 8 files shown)


vitali87:
Sonar's S3776 on the export collector is fixed in 2afb7e4a by extracting the per-capture entry builder. On CodeRabbit's docstring-coverage warning: deferring deliberately; this module follows the repo's comment-over-docstring convention for parser internals (the surrounding methods carry rationale comments, matching the sibling ingest passes), so adding docstrings here would diverge from the file's established idiom.

PR #1004, opened 2026-07-31
fix: resolve symbol-keyed member dispatch through a project-wide symbol index

### LINKED ISSUE #989: fix: method calls through a symbol-keyed member obj[kSym].method() emit no call edge across modules (JS/TS)
## Summary

A method call whose receiver is a symbol-keyed computed member, `obj[kSymbol].method(...)`, emits no CALLS edge when the symbol and the class live in different modules from the call site. Every method of a class installed this way is falsely reported dead.

## Witness (fastify)

`lib/symbols.js` defines `kLogController: Symbol('fastify.logController')`. `fastify.js` installs an instance with a computed object-literal key, `[kLogController]: logController`, where `logController` comes from the `createLogController` factory. Call sites across `lib/route.js`, `lib/reply.js`, `lib/four-oh-four.js` and `fastify.js` dispatch through the symbol, for example `context.server[kLogController].incomingRequest(request, reply)`. All nine `LogController` methods appear in the dead-code report. The same pattern flags eight `SchemaController` members via `this[kSchemaController].setValidatorCompiler(...)` and friends, seventeen candidates in total.

## Repro

Single-module resolution works (the call links), but the faithful cross-module shape fails:

```js
// symbols.js
module.exports = { kCtrl: Symbol('ctrl') }
// ctrl.js
class Ctrl { ping () { return 'pong' } }
function createCtrl (options) { return new Ctrl() }
module.exports = { Ctrl, createCtrl }
// main.js
const { kCtrl } = require('./symbols.js')
const { createCtrl } = require('./ctrl.js')
function build (options) {
  const ctrl = createCtrl(options)
  return { [kCtrl]: ctrl }
}
function use (server) {
  return server[kCtrl].ping()
}
module.exports = { build, use }
```

`use` gains no CALLS edge into `Ctrl.ping`.

## Root cause direction

A symbol constant is identity-unique across the project, so a sound project-wide map from the symbol's qualified name to the types assigned under it (via `[kSym]: value` object-literal keys and `obj[kSym] = value` assignments) would let the resolver type the receiver of `obj[kSym].method(...)` without falling back to name-global lookup, which is known to falsely revive same-named methods on unrelated classes.

vitali87:
## Summary

Fixes #989. A method call through a symbol-keyed computed member, `obj[kSymbol].method(...)`, emitted no CALLS edge when the symbol and the class lived in different modules from the call site, so every method of a class installed this way reported dead (seventeen fastify candidates across LogController and SchemaController).

## Design

A project-wide symbol index built in a pre-pass before call resolution:

1. **Symbol constants**: `kX = Symbol(...)` declarators and `kX: Symbol(...)` object pairs register the constant's qn (`Symbol(...)` and `Symbol.for(...)`).
2. **Installations**: `{ [kX]: value }` object keys, class fields (both grammars), and `obj[kX] = value` subscript assignments record the installed value's class, typed at sweep time by a depth-capped value chase: direct constructions, factory calls (bare, member, verified `Class.method = fn` static assignments), destructured-return object properties, and identifiers resolved to the INNERMOST introduction whose scope span encloses the read (parameters, catch and loop bindings and uninitialised declarators are opaque and refuse; the name-match-anywhere resolution that leaks was never used). A `new X(...)` target may be an old-style constructor FUNCTION only with per-member `X.prototype.<method> = ...` evidence, so nested private helpers sharing the `module.fn.name` namespace never masquerade as methods.
3. **Resolution**: `recv[kSym].method()` with exactly ONE installed class emits a precise CALLS; several installed classes emit REFERENCES to each candidate's matching method; a recognised symbol shape with NO typed installation degrades to REFERENCES over every registry function or method of that name, replacing the bare-name trie fallback that guessed one same-named method on an arbitrary class (a known method on the installed class set that fails the evidence gate emits nothing, matching the unknown-method rule). Divergent factory returns and ambiguous introductions always refuse rather than guess.

## Verification

Strict RED first for all sixteen pins (each confirmed failing before its fix; see commit history). Five adversarial review rounds in pinned worktrees against upstream main: nine findings (unscoped declarator chase in three forms, unrelated free-function fallback, divergent destructured returns, for-of `var` hoisting, edge-deleting suppression, and the Function-acceptance phantom in per-function and per-member forms), every one fixed and re-verified with its original repro.

Corpus results versus main: **fastify dead-code 18 to 0** with nothing added (this branch clears the seventeen symbol-dispatch candidates and their nested functions; the merged #1001 covers the rest), the four ContentTypeParser cross-checks land as precise CALLS, and excalidraw, zod and typeorm are delta-zero at both dead-code and (for excalidraw) full edge level. Full suite 6125 passed, 5 skipped, 0 failed.

With this PR merged, the fastify dogfood round that opened at 36 dead-code candidates closes at zero.

## Summary by CodeRabbit

* **New Features**
  * Improved JavaScript and TypeScript analysis for methods accessed through computed or symbol-based properties.
  * More accurately identifies class targets across constructors, factories, object properties, scopes, and modules.
  * Produces more precise call and reference relationships, including conservative handling of ambiguous cases.

* **Bug Fixes**
  * Prevents incorrect inferred calls when symbol bindings are unknown, shadowed, or associated with multiple classes.

* **Tests**
  * Added comprehensive coverage for symbol-based member dispatch scenarios.

### DIFF
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -104,6 +104,9 @@ TS_PROPERTY_IDENTIFIER = "property_identifier"
 TS_JS_COMPUTED_PROPERTY_NAME = "computed_property_name"
 # The `(a, b = 1, ...rest)` parameter list of a function.
 TS_JS_FORMAL_PARAMETERS = "formal_parameters"
+# A class field in the JS grammar (`[k] = v` in a class body); TS spells it
+# public_field_definition. Its key sits in the `property` field.
+TS_JS_FIELD_DEFINITION = "field_definition"
 # The Symbol global: `Symbol('x')` / `Symbol.for('x')` mint the unique keys
 # that symbol-keyed member dispatch (issue #989) indexes.
 JS_SYMBOL_GLOBAL = "Symbol"
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -100,6 +100,11 @@ TS_MODULE = "module"
 TS_CLASS_BODY = "class_body"
 
 TS_PROPERTY_IDENTIFIER = "property_identifier"
+# `[expr]` as an object-literal / class-member key.
+TS_JS_COMPUTED_PROPERTY_NAME = "computed_property_name"
+# The Symbol global: `Symbol('x')` / `Symbol.for('x')` mint the unique keys
+# that symbol-keyed member dispatch (issue #989) indexes.
+JS_SYMBOL_GLOBAL = "Symbol"
 
 # JS prototype property keywords
 JS_PROTOTYPE_KEYWORD = "prototype"
--- a/codebase_rag/constants/ast_js.py
+++ b/codebase_rag/constants/ast_js.py
@@ -102,6 +102,8 @@ TS_CLASS_BODY = "class_body"
 TS_PROPERTY_IDENTIFIER = "property_identifier"
 # `[expr]` as an object-literal / class-member key.
 TS_JS_COMPUTED_PROPERTY_NAME = "computed_property_name"
+# The `(a, b = 1, ...rest)` parameter list of a function.
+TS_JS_FORMAL_PARAMETERS = "formal_parameters"
 # The Symbol global: `Symbol('x')` / `Symbol.for('x')` mint the unique keys
 # that symbol-keyed member dispatch (issue #989) indexes.
 JS_SYMBOL_GLOBAL = "Symbol"
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1917,10 +1917,16 @@ class GraphUpdater:
                 self.queries,
                 func_class_captures_cache=captures_cache,
             )
+            self.factory.call_processor.collect_js_symbol_bindings(
+                file_path, root_node, language
+            )
         # Bindings are pending until every file's ctor metadata (param order,
         # param->attribute renames) is in: a construction site may be scanned
-        # before the file defining its class.
+        # before the file defining its class. Symbol-keyed installations wait
+        # the same way: a `[kX]: value` site may be swept before the file
+        # defining kX.
         self.factory.call_processor.finalize_callable_field_bindings()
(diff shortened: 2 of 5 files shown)


vitali87:
All three actionable comments and four of the eight nitpicks are taken in commit 1be341f4; the other four are rebutted below.

Taken:
1. The sweep walk is wrapped in the same broad exception guard as collect_callable_field_bindings, logging CALL_PROCESSING_FAILED, so one malformed file cannot abort Pass 3.
2. Symbol constants now register only from MODULE-scoped bindings (`_js_inside_callable` gates both the declarator and pair forms); a function-local `k: Symbol(...)` no longer hijacks same-spelled reads. RED first: test_function_local_symbol_pair_is_not_a_constant, confirmed failing before the fix.
3. The unresolved-symbol fan is restricted to FUNCTION and METHOD registry types, matching the file's variant guards.
4. (Nitpick) Symbol targets now expand through function_registry.variants like every sibling emission site, so duplicate-QN twins receive the edge.
5. (Nitpick) The prototype-evidence scan is memoized per (fn_qn, member) in a strings-only cache cleared with the rest of the per-run state.
6. (Nitpick) reset_js_receiver_bindings keeps its name for caller stability but its comment now states it clears ALL per-run JS call-pass state.
7. (Nitpick) `_walk_named` is deleted; its last caller went away in the introduction-resolution rewrite.

Rebutted:
- parametrize for shadow_first: the loop keeps one fixture body and both orders are asserted with a labelled failure message; a parametrize split adds ids but no coverage.
- positive controls in absence-only tests: the same file pins the positive direction of each mechanism (unique CALLS, REFERENCES candidate sets, module-binding and class-field installs), so a silent-pipeline false pass would fail those siblings first.
- extracting the Other-class fixture: each test reads standalone by design; the duplication is fixture text, not logic.
- bounding the fan: deliberately unbounded, since any cap trades liveness soundness for tidiness; the adversarial review measured the widest fastify site at 66 targets and we accept that until a dogfood round shows real retrieval pain.

Full suite 6145 passed, 5 skipped, 0 failed (plus exactly the one new pin).

PR #1006, opened 2026-08-01
fix(dead-code): treat JS/TS well-known symbol members as liveness roots (#993)

### LINKED ISSUE #993: fix: well-known symbol members like [Symbol.toStringTag] should be dead-code liveness roots (JS/TS)
## Summary

Class members keyed by well-known symbols, such as `get [Symbol.toStringTag] ()`, `[Symbol.iterator] ()` and `[Symbol.asyncIterator] ()`, are invoked implicitly by the JavaScript runtime and can never be called by name in user code. Dead-code analysis should treat them as liveness roots, the same way implicit protocol members are handled elsewhere.

## Witness (fastify)

`lib/content-type.js` defines `get [Symbol.toStringTag] () { return 'ContentType' }`, which the runtime reads whenever `Object.prototype.toString` is applied to an instance. It appears in the dead-code report; no source-level call site can ever exist for it.

## Repro

```js
class Box {
  get [Symbol.toStringTag] () { return 'Box' }
}
module.exports = Box
```

The member is registered as `Box.[Symbol.toStringTag]` and receives only a DEFINES_METHOD edge, so it reports dead.

## Fix direction

Members whose registered name matches a well-known symbol pattern, `[Symbol.<name>]`, should be excluded from dead-code candidates as runtime-invoked protocol members.

ChetanyaRathi:
Summary
Class members keyed by well-known symbols (get [Symbol.toStringTag](), [Symbol.iterator](), [Symbol.asyncIterator](), etc.) are invoked implicitly by the JS runtime and can never be called by name, but dead-code analysis was reporting them as dead. This treats them as liveness roots, the same way other implicit protocol members are handled.

Fix
Added is_well_known_symbol_member() in dead_code.py, matching the [Symbol.<name>] key form against a curated set of ES well-known symbols, and wired it into dead_code_from_graph so matching JS/TS members are added as roots.

Tests
Unit tests for the predicate (positive and negative cases, incl. [Symbol.for('x')] and local-symbol keys as negatives), plus a Docker-backed e2e test over box.js/iterables.ts fixtures asserting the symbol members are absent from the report while genuinely-dead members remain.

Closes #993.

## Summary by CodeRabbit

* **New Features**
  * Improved dead-code detection for JavaScript and TypeScript well-known `Symbol` properties, including iterator, matching, conversion, tagging, and resource-management hooks.
  * Built-in symbol members are now correctly recognized as reachable code.

* **Documentation**
  * Updated dead-code detection guidance to describe well-known symbol properties as reachable roots.

* **Tests**
  * Added coverage for symbol detection, end-to-end liveness analysis, and the dead-code command-line interface.


ChetanyaRathi:
The All Checks Pass failure here isn't a real test failure — the Unit Tests (macos-latest, py3.12) job was cancelled (runner timed out at ~35m), which trips the aggregate gate with exit code 1.

Every other job passed, including all the other Unit Tests matrix jobs (ubuntu + windows, py3.12 and py3.13) and the Docker-backed Integration Tests. The failure was isolated to that single cancelled macOS runner, not the test suite.

Could you re-run the cancelled job when you get a chance? I don't have permission to re-run workflow jobs on this run. Happy to push an empty commit to re-trigger CI instead if you'd prefer.

vitali87:
@greptile-apps review

vitali87:
Addressed the verified P1 in c0b3b1d4: the method-gated predicate no longer roots every `[Symbol.*]` member — the member must sit on the well-known allowlist, or be a `Symbol.for(...)` key that a RUNTIME itself invokes (nodejs.util.inspect.custom, nodejs.rejection). An application-defined `Symbol.for('app.tag')` member is reached only by code the graph can see, so it now correctly surfaces in dead-code output. New unit tests cover the previously untested method path — including the exact reproducer case, RED against the previous predicate — and both bracket-notation and quote variants.

vitali87:
@greptile-apps review

Issue #1010, opened 2026-08-01
fix: honour #[cfg(test)] on Rust mod declarations so gated modules like testutil count as test code


vitali87:
## Summary

A `#[cfg(test)]`-gated module DECLARATION that points at a separate file is invisible to dead-code test detection. ripgrep's `crates/searcher/src/lib.rs` declares:

```rust
#[cfg(test)]
mod testutil;
```

`testutil.rs` is compiled only for tests and is consumed exclusively by inline `mod tests` blocks across the crate, but nothing in the graph records the gate: the attribute sits on the `mod` declaration in lib.rs while the Module node is minted from testutil.rs itself. Ten `crates.searcher.src.testutil.*` methods (KitchenSink, SearcherTester, the test RegexMatcher) survive as dead-code candidates after #1008's attribute rooting, because they are trait-impl and helper methods the rooted `#[test]` functions reach only through dynamic dispatch.

This is the remaining subclass of #1008 (inline `mod tests` and `#[test]` attributes are handled there): the `#[cfg(test)]` attribute on a bodyless `mod NAME;` declaration should mark the TARGET module as test code, so its symbols root under `--include-tests` and are excluded under `--no-include-tests`.

## Minimal repro

`src/lib.rs`
```rust
#[cfg(test)]
mod testutil;

pub fn add(a: i32, b: i32) -> i32 { a + b }

#[cfg(test)]
mod tests {
    use super::*;
    #[test]
    fn t() { assert_eq!(add(crate::testutil::fixture(), 1), 8); }
}
```

`src/testutil.rs`
```rust
pub(crate) fn fixture() -> i32 { 7 }

pub(crate) struct Helper;

impl Helper {
    pub(crate) fn unused_by_name(&self) -> i32 { 9 }
}
```

`Helper.unused_by_name` (reachable only via test-side dispatch, or genuinely uncalled test infrastructure) is reported as dead production code; everything under a `#[cfg(test)]` module is test infrastructure and should never appear in the report.

## Direction

[...]

PR #1029, opened 2026-08-02
fix: root Rust inline #[test] functions and test-module symbols in dead-code analysis

### LINKED ISSUE #1008: fix: Rust inline #[test] functions are never rooted by dead-code, flagging tests and test-only helpers
## Summary

Rust puts unit tests **inside** source files (`#[cfg(test)] mod tests` with `#[test]` functions), so path-based test detection never sees them. `matches_test_path("crates/globset/src/glob.rs", ...)` is `False`, and the decorator rooting path mangles the attribute: `_norm_decorator("#[test]")` returns `"#[test"` (the leading `#[` survives the strip), and `test` is not in `DEFAULT_ROOT_DECORATORS` anyway.

The parser side is already correct: `#[test]` functions are ingested with `decorators: ['#[test]']` and their CALLS edges into production code are present. Only the dead-code rooting misses them.

Impact on ripgrep: **459 of 1811 candidates** are inline test functions (`.tests.` modules, e.g. `crates.cli.src.escape.tests.backslash`) or code reachable only from them, reported dead despite `--include-tests` defaulting to on.

## Minimal repro

`src/lib.rs`
```rust
pub fn add(a: i32, b: i32) -> i32 {
    a + b
}

fn helper_only_used_by_tests(x: i32) -> i32 {
    x * 2
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_add() {
        assert_eq!(add(1, 2), 3);
    }

    #[test]
    fn test_helper() {
        assert_eq!(helper_only_used_by_tests(2), 4);
    }
}
```

Graph (correct): `tests.test_add` and `tests.test_helper` are Function nodes with decorator `#[test]`, with CALLS edges to `add` and `helper_only_used_by_tests`.

Dead-code (wrong): `test_add`, `test_helper` and `helper_only_used_by_tests` are all reported as unreachable, because nothing roots the `#[test]` functions.

## Expected behaviour

With `--include-tests` (the default), a function carrying a `#[test]` family attribute (`#[test]`, `#[tokio::test]`, `#[bench]`, and scoped variants ending in `::test`) is a liveness root, exactly as test-path files are today, making both the tests and the production code they exercise live. With `--no-include-tests`, symbols inside a `#[cfg(test)]` context should be excluded from the report the same way test-path symbols are (they are not production dead code either way).

Verified on ripgrep at master with cgr current main.

vitali87:
Closes #1008.

## What

Rust unit tests live inside source files (`#[cfg(test)] mod tests` with `#[test]` functions), so path-based test detection never sees them: `matches_test_path("crates/globset/src/glob.rs")` is false, and `_norm_decorator("#[test]")` mangles the attribute to `#[test` (with `test` absent from the root-decorator set anyway). The tests and every production helper only they exercise were reported dead despite `--include-tests` defaulting to on: 459 of ripgrep's 1811 candidates.

The fix adds `_is_rust_test_symbol`, gated to `.rs` paths: a function carrying a `#[test]` family attribute (`#[test]`, `#[bench]`, or a scoped runner variant ending in `::test` such as `#[tokio::test]`), or any symbol inside a `tests` or `test` MODULE (the `#[cfg(test)] mod tests` convention in both spellings). The module check walks the qn's prefixes against real Module nodes rather than raw string segments, so a type or method named `tests`, a project directory named `tests`, and a dotted project name all stay ordinary reportable code instead of silently rooting their callee closures. Symbols lexically nested inside a test function (nested fns register flat under the module qn, closures as `anonymous_<line>_<col>`) are recognised by span containment and follow it on BOTH polarities: excluded with it when tests are off, rooted with it when they are on (a fn passed by value to `filter()` has no CALLS edge to revive it). Containment is strict on both ends because spans carry no columns, so a production fn packed onto a test fn's first or last physical line stays reportable. The module-name rule is deliberately wider than the path patterns (those match directory segments only, never a `src/test.rs` file or an inline mod): the name is a proxy for the `#[cfg(test)]` gate the graph does not yet record. A 972-crate corpus scan during review measured 4155 cfg-gated `mod test(s)` declarations against 41 ungated ones, of which exactly two ship as production API (aws-lc-rs `pub mod test`, alacritty's terminal test helpers), both test-support by nature; that residual silencing is accepted and documented until #1010 replaces the name proxy with cfg awareness. Wired into both sides of the existing test handling:

- with `--include-tests` (default), matching symbols are liveness roots, exactly as test-path files are today;
- with `--no-include-tests`, they are excluded from the candidate set, exactly as test-path symbols are (test infrastructure, not dead production code).

## Effect on ripgrep

Client side only (no re-index needed). On a fresh index of ripgrep at master, the pre-fix collector reports 1811 candidates (matching the issue's opening scan exactly, re-measured on the same graph) and the fixed collector reports 1157: the rooted `#[test]` functions revive their entire callee closures across every crate. With `--no-include-tests` the count is 1257, the span containment having removed the seventeen closures nested inside `#[test]`/`#[bench]` functions that a first review round showed leaking into the report. The residue includes ten `crates.searcher.src.testutil.*` methods gated by `#[cfg(test)]` on a cross-file `mod testutil;` declaration; that subclass is filed separately as #1010.

## Tests

TDD RED first: `test_rust_test_attribute_roots.py` holds fifteen repros (attribute roots plus callee revival with qns carrying NO `tests` segment so the decorator branch is load-bearing on its own, scoped `#[tokio::test(flavor = "multi_thread")]`/`#[bench]`, the module rule alone rooting an attribute-less helper, plain and non-test-attribute functions still reporting, `--no-include-tests` excluding both the decorator-decided and module-decided shapes, the non-Rust path guard, a production method named `tests` and a project directory named `tests` both staying reportable, the singular `mod test` matching the plural on both polarities with distinct real-shaped spans and no rescuing edge, the module rule keyed on real Module nodes with a lowercase type named `tests` staying reportable alongside its closure, a dotted project prefix not masking reporting, span-nested symbols excluded with their suppressed test fn AND rooted with an included one, and same-line-packed production fns staying reportable on both polarities), every behavioural case observed failing before its fix. Four adversarial review rounds drove the hardening: they confirmed the attribute matcher against real parses (raw `#[...]` decorator text, argument stripping, the `#[cfg(test)]` and `#[testing::...]` traps rejected), verified closure-revival transitivity, both polarities, class handling, and hostile span shapes on real parses, ran a fifteen-mutation vacuity matrix over the tests, measured the rule's width across a 972-crate corpus, and held the ripgrep candidate sets byte-stable at 1157/1257 across every refinement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rust dead-code analysis for test and benchmark functions, scoped test attributes, and symbols inside `test`/`tests` modules.
  * Test-related code is now correctly included or excluded according to analysis settings.
  * Reduced false positives and negatives involving symbol names, project paths, and source-code boundaries.

* **Tests**
  * Added comprehensive coverage for Rust test detection and dead-code reporting scenarios.

### DIFF
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -73,9 +73,13 @@ RUST_ROOT_FUNCTION_NAMES: frozenset[str] = frozenset({"main"})
 RUST_TEST_ATTRIBUTE_NAMES: frozenset[str] = frozenset({"test", "bench"})
 RUST_TEST_ATTRIBUTE_SUFFIX = "::test"
 
-# The `#[cfg(test)] mod tests` convention: any qn segment `tests` in a .rs
-# file marks inline test code (issue #1008).
-RUST_TESTS_MODULE_SEGMENT = "tests"
+# The `#[cfg(test)] mod tests` convention: a MODULE segment named `tests`
+# or `test` in a .rs file marks inline test code (issue #1008). Both
+# spellings, mirroring TEST_PATH_PATTERNS' /tests/ and /test/ directories.
+# Module segments only: the project prefix and the symbol's own name are
+# not modules, so a project directory or a method named `tests` stays
+# production code.
+RUST_TEST_MODULE_SEGMENTS: frozenset[str] = frozenset({"tests", "test"})
 
 # Rust trait-impl methods the language/std dispatches implicitly (Display::fmt
 # via format!, PartialEq::eq via ==, Iterator::next via for, operator traits,
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -67,6 +67,16 @@ GO_ROOT_FUNCTION_NAMES: frozenset[str] = frozenset({"init", "main"})
 # call site -- a reachability root (gated by .rs).
 RUST_ROOT_FUNCTION_NAMES: frozenset[str] = frozenset({"main"})
 
+# Rust test attributes: the harness invokes these functions with no call site.
+# Bare names match exactly; scoped runner variants (#[tokio::test],
+# #[async_std::test]) match by the ::test suffix. Gated by .rs (issue #1008).
+RUST_TEST_ATTRIBUTE_NAMES: frozenset[str] = frozenset({"test", "bench"})
+RUST_TEST_ATTRIBUTE_SUFFIX = "::test"
+
+# The `#[cfg(test)] mod tests` convention: any qn segment `tests` in a .rs
+# file marks inline test code (issue #1008).
+RUST_TESTS_MODULE_SEGMENT = "tests"
+
 # Rust trait-impl methods the language/std dispatches implicitly (Display::fmt
 # via format!, PartialEq::eq via ==, Iterator::next via for, operator traits,
 # Drop::drop, serde, ...), never through an explicit call the graph can see.
--- a/codebase_rag/constants/deadcode_roots.py
+++ b/codebase_rag/constants/deadcode_roots.py
@@ -73,12 +73,16 @@ RUST_ROOT_FUNCTION_NAMES: frozenset[str] = frozenset({"main"})
 RUST_TEST_ATTRIBUTE_NAMES: frozenset[str] = frozenset({"test", "bench"})
 RUST_TEST_ATTRIBUTE_SUFFIX = "::test"
 
-# The `#[cfg(test)] mod tests` convention: a MODULE segment named `tests`
-# or `test` in a .rs file marks inline test code (issue #1008). Both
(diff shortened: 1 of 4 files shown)


vitali87:
@greptileai

vitali87:
Both review findings addressed in the pushed commits:

- 1bddac0e fixes the whitespace-scoped attribute gap greptile verified: attribute paths are token streams, so the matcher now drops internal whitespace before comparing (`#[tokio :: test]` classifies identically to `#[tokio::test]`), with a regression fixture in test_scoped_test_attributes_root.
- 12f4b21a extracts CodeRabbit's suggested shared predicate: one _is_test_symbol covers both polarities so the exclusion and rooting branches cannot silently diverge.

@greptileai

Issue #1037, opened 2026-08-02
fix: nested fn inside a Rust impl method registers with a label its CALLS edge does not match


vitali87:
## Symptom

A nested `fn` declared inside a Rust impl method registers under the impl target's qn with a duplicate-variant marker, and the CALLS edge to it carries the FUNCTION label while the qn space belongs to METHOD nodes. The graph audit rejects the edge:

```
(Method 'p.src.foo.S.run')-[:CALLS]->(Function 'p.src.foo.S.helper@4') references a nonexistent node and would be dropped by the database
```

## Minimal repro

```rust
pub struct S;

impl S {
    fn helper(&self) -> u32 {
        1
    }
    pub fn run(&self) -> u32 {
        fn helper() -> u32 {
            2
        }
        helper()
    }
}
```

Any fixture with this shape hard-fails `_audit_recorded_graph`, so the edge is silently dropped on a real database and the nested function is unreachable in the graph.

## Direction

Nested fns inside impl methods should register with a label consistent with the qn their registration path produces (or register flat under the module qn as nested fns in free functions do), so the CALLS edge endpoints exist.

Found by the adversarial review of #1011's fix (round six, ripgrep).

vitali87:
Fixed in #1068. The body-local item was claiming the natural qn before the method pass ran; it now registers after the methods have taken their names, so the method keeps `S.helper` and the nested fn takes the `@<line>` variant as a Function node, which is what its CALLS edge already pointed at.

The pre-push review turned up the same theft with no impl involved, on main today: a fn nested in a free function is written before a module-level `pub fn` of the same name, takes its qn by source order, and leaves the public function with no incoming edge. Covered by the same fix.

PR #1038, opened 2026-08-02
fix: stop Rust bare calls binding same-named inherent methods

### LINKED ISSUE #1011: fix: bare calls inside Rust impl methods wrongly bind to same-named inherent methods
## Symptom

A bare function call inside a Rust `impl` method binds to a same-named inherent method of the surrounding type, fabricating a CALLS edge that Rust semantics forbid.

```rust
// src/foo.rs
pub fn helper() -> u32 {
    1
}

pub struct S;

impl S {
    fn helper(&self) -> u32 {
        2
    }
    pub fn run(&self) -> u32 {
        helper()
    }
}
```

cgr emits `CALLS src.foo.S.run -> src.foo.S.helper`. rustc disagrees: a bare path never resolves to a method, so `helper()` calls the free function `src.foo.helper`, and `cargo check` warns that `S::helper` is never used. The fabricated edge both misses the real call target and revives a dead inherent method in dead-code analysis.

## Root cause

`_resolve_enclosing_scope` in `codebase_rag/parsers/call_resolver.py` walks enclosing scopes generically and treats the surrounding class body as a scope whose members are bare-callable, which is true for some languages but not Rust: inherent methods are reachable only through `self.helper()`, `Self::helper()` or `S::helper()`. The Rust inline-module scope walk reaches the same wrong answer by a second route, so both paths need the language gate.

## Expected

For Rust callers, bare call resolution should skip the enclosing `impl`'s methods and resolve through module items and imports, so the example yields `CALLS src.foo.S.run -> src.foo.helper`.

Found while dogfooding dead-code analysis on ripgrep during the sixth adversarial review round of #1007; it predates that branch.

vitali87:
Closes #1011.

## What

A bare function call inside a Rust impl method bound to a same-named inherent method of the surrounding type, fabricating a CALLS edge Rust forbids: in the issue's repro, `run(&self)` calling `helper()` emitted `S.run CALLS S.helper`, while rustc resolves the bare path to the free function `helper` and warns that `S::helper` is never used. The fabricated edge both misses the real target and revives a dead inherent method in dead-code analysis.

Two resolution paths reached the wrong answer. The Rust inline-scope walk probes each enclosing qn segment for a local item, and an impl target's segment surfaces the inherent method; it now skips Method-labelled hits, since a bare Rust path can only name a method through `self.`, `Self::` or `Type::`, all of which resolve elsewhere. The generic enclosing-scope walk (Python LEGB) reaches the same method through the class segment; it now takes the language and applies the identical gate for Rust callers only, leaving every other language's behaviour untouched. With both gates in place the bare call falls through to same-module resolution and lands on the free function.

## Adversarial review

The single pre-push adversarial round confirmed the two gated walks regression-free across sixteen rustc-valid fixtures, and found the invariant still broken where no first-party binding exists: with `use std::cmp::max`, a prelude call, or a shadowing closure, the trie fallback still soaked the edge onto the inherent method. Fixed here: the trie excludes Method-labelled candidates for bare Rust names, and the duplicate-variant branch of the enclosing-scope walk carries the same gate. A related pre-existing defect it surfaced (a nested fn inside an impl method registering with a label its CALLS edge does not match) is filed as #1037.

## TDD

RED first: the issue's exact fixture, observed emitting `S.run -> S.helper` with the free-function edge missing, plus the review round's three trie-hole repros (external import, prelude, closure shadowing), each observed failing, and a `self.helper()` control pinning that receiver spellings keep their method edge. All green after the fix. Full suite: 6713 passed, 10 skipped, against a main baseline of 6708 with the same skips.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved call resolution across nested scopes while preserving the caller’s programming language context.
  - Corrected Rust bare-name lookups so they resolve to free functions, imports, prelude functions, or local closures instead of inherent methods.
  - Ensured receiver-qualified calls such as `self.method()` continue to resolve to the appropriate inherent method.

- **Tests**
  - Added coverage for Rust call resolution inside inherent implementations.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -383,7 +383,11 @@ class CallResolver:
                     scope[: len(scope) - len(last)] + last.split(cs.DUP_QN_MARKER, 1)[0]
                 )
                 natural_candidate = f"{natural_scope}{cs.SEPARATOR_DOT}{call_name}"
-                if natural_candidate in self.function_registry:
+                if natural_candidate in self.function_registry and (
+                    language != cs.SupportedLanguage.RUST
+                    or self.function_registry[natural_candidate]
+                    != cs.NodeLabel.METHOD.value
+                ):
                     return (
                         self.function_registry[natural_candidate],
                         natural_candidate,
@@ -946,7 +950,7 @@ class CallResolver:
                 self._simple_resolution_cache[cache_key] = result
             return result
 
-        result = self._try_resolve_via_trie(call_name, module_qn)
+        result = self._try_resolve_via_trie(call_name, module_qn, language)
         if use_cache:
             self._simple_resolution_cache[cache_key] = result
         return result
@@ -1696,7 +1700,10 @@ class CallResolver:
         return language
 
     def _try_resolve_via_trie(
-        self, call_name: str, module_qn: str
+        self,
+        call_name: str,
+        module_qn: str,
+        language: cs.SupportedLanguage | None = None,
     ) -> tuple[str, str] | None:
         search_name = _SEARCH_NAME_CACHE.get(call_name)
         if search_name is None:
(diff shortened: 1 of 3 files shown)


vitali87:
Bot reports addressed: SonarCloud's two new S3776 issues (the gated enclosing-scope and inline-scope walks) are resolved by extracting the shared method-bar predicate into _bare_call_allowed, behaviour unchanged (suite identical at 6713 passed / 10 skipped). CodeRabbit had no actionable comments and Greptile reported 5/5 with no findings.

vitali87:
Correction on the previous summary: the first extraction left both walks over the S3776 threshold (17 and 16). The scope-candidate probe and the duplicate-variant branch are now their own functions and both walks sit within the limit; behaviour is unchanged (suite identical at 6713 passed / 10 skipped across all three commits).

vitali87:
@coderabbitai review

Issue #1054, opened 2026-08-03
fix(rust): bind `use path::{self}` under the module name, not the `self` keyword


vitali87:
`use std::io::{self, IsTerminal};` records the imported module under the literal key `self` instead of the name it actually binds, `io`. In Rust a `self` entry in a brace list imports the base path itself under its LAST SEGMENT, so the module is in scope as `io` and every later `io::...` spelling in that file resolves through it.

`_process_use_tree` keys the entry on the keyword:

```python
case cs.KEYWORD_SELF:
    imports[cs.KEYWORD_SELF] = base_path or cs.KEYWORD_SELF
```

Nothing ever looks up `self`, so the binding is unreachable, while the name that was really introduced is absent from the map. Every consumer that asks "what does the head `io` mean here" gets nothing: qualified calls (`io::stdout()`), qualified types, and the external-head classification that decides whether a trait belongs to another crate.

## Evidence

ripgrep, after #1048:

```
crates/cli/src/wtr.rs:1      use std::io::{self, IsTerminal};
crates/cli/src/wtr.rs:67     impl io::Write for StandardStream { fn write, fn flush }
crates/cli/src/process.rs:1  use std::{io, process, ...};   (nested brace list)
crates/cli/src/process.rs:254 impl io::Read for CommandReader { fn read }
```

Import mapping recorded for both modules:

```
probe.crates.cli.src.wtr     -> {'self': 'std::io', 'IsTerminal': 'std::io::IsTerminal'}
probe.crates.cli.src.process -> {'self': 'std::io'}
```

The head `io` is unresolvable in both, so `StandardStream.write`, `StandardStream.flush` and `CommandReader.read` are still reported as dead code: their trait head cannot be proven to be the standard library. The same modules also lose every `io::`-qualified call edge.

The alias form is already correct (`use std::io::{self as sys}` goes through `_process_use_as_clause`); only the bare `self` form keys wrongly.

## Expected

A `self` entry binds the base path's last segment: `use std::io::{self, IsTerminal}` records `io -> std::io`. A bare `use self::thing` (no base path) keeps its current meaning.

## Out of scope

- Non-Rust import processing.
- Wildcard and alias forms, which already resolve.

vitali87:
PR #1056 implements this. Combined with #1048, ripgrep goes 189 to 161, 28 resolved and 0 newly listed; the six beyond #1048's own 22 are exactly the candidates that needed this head to resolve.

Issue #1047, opened 2026-08-02
fix: resolve Rust method calls on generic receivers through their trait bounds


vitali87:
A method call on a receiver whose type is a generic type parameter never resolves through the parameter's trait bound. Rust dispatches such calls to the bound trait's method, so cgr should emit CALLS to the trait method (which OVERRIDES then expands to the implementations).

## Repro

```rust
// src/lib.rs
pub trait Matcher {
    fn find(&self, hay: &str) -> bool;

    fn is_match(&self, hay: &str) -> bool {
        self.find(hay)
    }
}

pub fn search<M: Matcher>(m: M) -> bool {
    m.is_match("x")
}

pub struct Core<M> {
    matcher: M,
}

impl<M: Matcher> Core<M> {
    pub fn run(&self) -> bool {
        self.matcher.is_match("y")
    }
}

pub fn search_where<M>(m: M) -> bool
where
    M: Matcher,
{
    m.is_match("z")
}

pub struct Decoy;

impl Decoy {
    pub fn is_match(&self, _hay: &str) -> bool {
        false
    }
}

pub fn use_decoy(d: Decoy) -> bool {
    d.is_match("w")
}
```

Actual CALLS edges:

- `search` and `search_where` emit nothing for `m.is_match(...)`.
- `Core.run` emits `Decoy.is_match` for `self.matcher.is_match(...)`: the field type `M` never resolves, and the fallback binds the call to an inherent method on an unrelated type. Without the decoy the fallback happens to pick `Matcher.is_match`, so the field shape only works by uniqueness luck.

Expected: all three call sites emit `CALLS -> lib.Matcher.is_match`.

## Where it bites

ripgrep: `SearcherCore` stores `matcher: M` with `M: Matcher` and calls `self.matcher.is_match(line)` (crates/searcher/src/searcher/core.rs:121); the printer calls `matcher.new_captures()` and `matcher.capture_index(name)` through the same bound (crates/printer/src/util.rs:149, 96). None of these produce an edge, so all seventeen `Matcher` default methods sit in the dead code report while being the crate's core dispatch surface.

## Fix shape

[...]

Issue #1033, opened 2026-08-02
fix: resolve Rust workspace crate names in use paths so cross-crate calls bind precisely


vitali87:
## Summary

A Rust `use` whose head names another crate keeps its raw `::` path in the import mapping: only `crate::`/`super::`/`self::` paths are rewritten to project qns at parse time. Two consequences for `module::item` call resolution:

1. Workspace member crates are indexed first party, but a cross-crate call reaches its target only by trie luck. `use grep_searcher::sinks;` then `sinks::UTF8(...)` resolves through the simple-name trie, not through the mapping, because `grep_searcher::sinks` never becomes a qn even though the crate lives in the same workspace and its files are in the graph.

2. An external import shadowing a first-party sibling still yields a fabricated edge downstream. In `src/worker.rs`:

```rust
use std::fmt;
pub fn run() -> String { fmt::format(format_args!("hi")) }
```

with a sibling `src/fmt.rs` containing `pub fn format`, the call is `std::fmt::format`, yet the trie binds `src.fmt.format`. The module-qualified probe (#1009) stops undecided when the head is import-bound to an unindexed path, so it no longer binds the sibling itself, but it cannot decide a drop either: without workspace crate-name resolution, every cross-crate call would be severed with it.

## Direction

Map workspace member crate names (from the workspace manifest members and each member's `[package] name`, hyphens normalised to underscores) to their crate root qns. Rewrite `use` paths whose head is a member name to project qns at parse time, exactly as `crate::` paths are. Then an import-bound head that still resolves to no project qn is genuinely external, and the module-qualified probe can decide a drop instead of leaving the trie to fabricate a first-party edge.

## Out of scope

- Renamed dependencies (`[dependencies] alias = { package = ... }`)
- Path dependencies outside the workspace

Found on ripgrep (round six), from the adversarial review of #1009's fix.

PR #1055, opened 2026-08-03
fix(dead-code): root Rust methods implementing external traits

### LINKED ISSUE #1048: fix(dead-code): root Rust methods implementing external traits
Methods defined in an `impl <ExternalTrait> for Type` block are reported as dead code. They are invoked through the external trait's dispatch by code outside the graph: the standard library and third party crates call `io::Read::read`, `io::Write::write`, `fmt::Display::fmt`, `Hasher::write`, `FromIterator::from_iter`, serde calls `Visitor` methods from inside its deserialiser, and the `log` facade calls `Log::enabled`/`Log::log`. No first party call site can exist for most of them, so internal reachability can never prove them live.

The dead code pass should treat a method as a liveness root when its enclosing impl block implements a trait that is not first party: the trait path's head is a stdlib crate (`std`, `core`, `alloc`, `proc_macro`) or a manifest proven external dependency, or the trait name resolves to no project trait node. The head classification machinery already exists from #1033 (`rust_head_is_external_dep`, `RS_STDLIB_CRATES`). This mirrors the JS/TS well known symbol member roots from #993: members whose invocation contract belongs to the platform are roots, not candidates.

## Evidence

ripgrep residue after #1045 (221 candidates): 26 of them are external trait impl methods, none reachable by any first party edge, all invoked at runtime.

```
crates.cli.src.decompress.DecompressionReader.read      impl io::Read
crates.cli.src.process.CommandReader.read               impl io::Read
crates.cli.src.wtr.StandardStream.{write,flush}         impl io::Write
crates.cli.src.wtr.StandardStream.{supports_color,set_color,reset,
    is_synchronous,set_hyperlink,supports_hyperlinks}   impl termcolor::WriteColor
crates.core.logger.Logger.{enabled,log}                 impl log::Log
crates.globset.src.fnv.Hasher.write                     impl std::hash::Hasher
crates.globset.src.serde_impl.GlobSetVisitor.{expecting,visit_seq}  impl serde Visitor
crates.globset.src.serde_impl.GlobVisitor.{expecting,visit_str}     impl serde Visitor
crates.ignore.src.lib.Error.description                 impl std::error::Error
crates.pcre2.src.error.Error.description                impl std::error::Error
crates.printer.src.counter.CounterWriter.{write,flush}  impl io::Write
crates.printer.src.counter.CounterWriter.{is_synchronous,set_hyperlink,
    supports_hyperlinks}                                impl termcolor::WriteColor
crates.regex.src.literal.TSeq.from_iter                 impl FromIterator
```

## Out of scope

- First party trait impls (OVERRIDES liveness already covers them).
- Generic bound dispatch on first party traits (separate issue).
- Non-Rust languages.

vitali87:
Closes #1048.

## What

A method defined in an `impl <Trait> for Type` block whose trait belongs to another crate has no first-party caller and never can: the standard library calls `io::Read::read`, serde calls its `Visitor` methods from inside the deserialiser, the `log` facade calls `Log::log`, termcolor calls `WriteColor::set_color`. Internal reachability cannot prove such a method live, so all of them reported dead.

The fix flags the whole block's methods `overrides_external`, the property the dead-code surfaces already treat as a root (the same contract Python stdlib-base overrides and Dart framework overrides use). Each `impl Trait for Type` block records its ingested method qns alongside the deferred IMPLEMENTS entry, and `resolve_deferred_inherits` judges the trait once every first-party trait is registered.

## Only positive evidence roots

The decision is deliberately not "the trait resolves to no project node". A trait reached by a workspace sibling's crate name (`use grep_searcher::Sink`, re-exported from that crate's root) resolves nowhere here, and rooting on unresolvedness would have hidden ripgrep's entire `Sink` implementation surface, along with the genuinely dead code hanging off it, permanently. The first draft did exactly that: it rooted 66 candidates, of which 40 were first-party `Sink` impls and their call trees.

A trait counts as external only when the crate its written head speaks for is one of these:

- a toolchain crate (`std`, `core`, `alloc`, `proc_macro`);
- a head the importing package's own manifest binds outside the repo, a registry version or git dependency with no path (the `rust_head_is_external_dep` machinery from #1033, which reports `False` for a workspace path dependency);
- nothing at all, when the name is bare, unimported, and one of the prelude traits, the only way such a name can be in scope.

The written head is itself resolved through the module's imports first, because `use std::io;` then `impl io::Read` is Rust's usual spelling for a stdlib trait and the literal head `io` speaks for `std`. A trait that resolves to a registered first-party trait is skipped before any of this: OVERRIDES liveness already covers it.

## TDD

RED first: 5 of the 11 tests in `test_rust_external_trait_impl_flag.py` fail before the fix (scoped stdlib head, imported module head, bare imported external head, generic trait head, manifest registry dependency), with 3 guards passing throughout (same-file first-party trait, cross-file first-party trait, inherent impl on the same type). The 3 tests from adversarial review are RED against the reviewed commit and green after it.

The workspace-sibling guard was demonstrated RED against the loose first draft rather than against `main`: with the strictness stubbed out, `test_workspace_sibling_crate_trait_impl_is_not_flagged` fails, which is the ripgrep regression reproduced in a fixture. It only reproduces when the trait is declared in a submodule and re-exported from the crate root, exactly `grep_searcher::Sink`.

Full suite: 6793 passed, 10 skipped. The baseline on `main` is 6782 passed / 10 skipped, so the delta is exactly the 11 tests this branch adds, with no new skips, xfails or deselections.

Live ripgrep re-measure: 189 to 167, 22 resolved and 0 newly listed. The resolved set is the predicted surface: both `io::Read` and `io::Write` impls that spell their head as a scoped path, the whole termcolor `WriteColor` surface on `StandardStream` and `CounterWriter`, both `log::Log` methods, the serde `Visitor` methods, the `std::hash::Hasher` and `std::error::Error` impls, and `FromIterator::from_iter`.

## Adversarial review

One round, three findings, all fixed RED first, and re-measured afterwards: the 167 row set is byte for byte identical, so the extra strictness costs no recall.

1. The first-party guard accepted any registered node, not a trait. Parent resolution ends in a project-wide simple-name sweep, so a struct a project happens to call `Default` answered for the prelude trait and silenced the flag for every `impl Default` in the project (the same sweep also has the type "implement" a struct). Now gated on the resolved node being an INTERFACE, the same strictness #1047 needed for bound spellings.
2. `rust_head_is_external_dep` proves "this entry resolves to no repo directory", which is not the same as "this crate is outside the repo". A published workspace declaring a sibling by version alone, and a `package =` rename of a path dependency, both got their trait impls rooted, the exact direction the code's own comment calls unrecoverable. Fixed on both sides: a head naming a lib crate the repo holds is first-party whatever the manifest says about fetching it, and a dependency entry carrying a path binds that directory whether or not it also renames the package.
3. `Hash` is in no Rust prelude, so a bare `impl Hash` cannot compile without a `use`; the entry could only ever have fired on a first-party trait of that name. Removed.

Two conservative misses were reported and left alone, both recall gaps rather than wrong edges: a trait brought in by `use std::io::prelude::*;`, and an impl written inside a function body, whose scope-local `use` is not in the module import map.

## Residue

Three of the candidates listed in the issue survive: `StandardStream.write`, `StandardStream.flush` and `CommandReader.read`. Their modules spell the head through a brace-list `self` import (`use std::io::{self, IsTerminal};`), which the import processor records under the literal key `self` instead of `io`, so the head resolves to nothing and no positive evidence is available. That is a separate import-mapping defect affecting qualified calls in those modules too, filed as #1054 and fixed next; it is not worked around here.

## Summary by CodeRabbit

* **New Features**
  * Improved Rust trait analysis for external and first-party implementations.
  * Enhanced resolution of renamed Rust package dependencies and workspace crate imports.
  * Added support for fully qualified and generic trait paths.

* **Bug Fixes**
  * More accurately identifies methods associated with external traits while avoiding false positives for internal traits and similarly named structs.

* **Tests**
  * Added comprehensive coverage for scoped, imported, generic, registry-based, and workspace trait implementations.

### DIFF
--- a/codebase_rag/constants/ast_rust.py
+++ b/codebase_rag/constants/ast_rust.py
@@ -208,6 +208,39 @@ RS_MANIFEST_TARGET_TABLE_KEY = "target"
 # manifest needed to know a use head naming one is outside the project.
 RS_STDLIB_CRATES = frozenset({"std", "core", "alloc", "proc_macro"})
 
+# Traits the Rust prelude puts in scope with no `use`: an impl naming one
+# without importing it and without a same-named first-party declaration
+# implements the standard trait, whose dispatch lives outside the graph
+# (issue #1048). Marker traits with no methods are omitted as pointless.
+RS_PRELUDE_TRAITS = frozenset(
+    {
+        "AsMut",
+        "AsRef",
+        "Clone",
+        "Default",
+        "DoubleEndedIterator",
+        "Drop",
+        "ExactSizeIterator",
+        "Extend",
+        "Fn",
+        "FnMut",
+        "FnOnce",
+        "From",
+        "FromIterator",
+        "Hash",
+        "Into",
+        "IntoIterator",
+        "Iterator",
+        "Ord",
+        "PartialEq",
+        "PartialOrd",
+        "ToOwned",
+        "ToString",
+        "TryFrom",
+        "TryInto",
+    }
+)
+
 # Iterator-adaptor closure typing (issue #1045): a closure argument of one
 # of these adaptors receives the sequence's element (by value or
 # reference, indistinguishable for method binding), so its parameter can
--- a/codebase_rag/constants/ast_rust.py
+++ b/codebase_rag/constants/ast_rust.py
@@ -227,7 +227,6 @@ RS_PRELUDE_TRAITS = frozenset(
         "FnOnce",
         "From",
         "FromIterator",
-        "Hash",
         "Into",
         "IntoIterator",
         "Iterator",
--- a/codebase_rag/parsers/class_ingest/mixin.py
+++ b/codebase_rag/parsers/class_ingest/mixin.py
@@ -523,11 +523,22 @@ class ClassIngestMixin:
         # genuinely dead first-party trait impl for good.
         module_qn = impl.entry.module_qn
         resolved = self._resolve_deferred_parent_qn(impl.entry)
-        if resolved is not None and not resolved[1]:
-            # A registered first-party trait dispatches through OVERRIDES
-            # liveness, which already covers its impls.
+        if (
+            resolved is not None
+            and not resolved[1]
+            and self.function_registry.get(resolved[0]) == NodeType.INTERFACE
+        ):
+            # A registered first-party TRAIT dispatches through OVERRIDES
+            # liveness, which already covers its impls. The node kind has to
+            # be checked: parent resolution ends in a project-wide simple-name
+            # sweep, so a struct the project happens to call `Default` answers
+            # for the prelude trait and would silence every `impl Default`.
             return False
         head, scoped = self._rust_trait_head(impl.spelling, module_qn)
+        if self.import_processor.rust_head_is_repo_crate(head):
+            # The crate is indexed here whatever the manifest says about
+            # fetching it, so its traits are first-party.
+            return False
         if (
             head in cs.RS_STDLIB_CRATES
             or self.import_processor.rust_head_is_external_dep(head, module_qn)
--- a/codebase_rag/parsers/class_ingest/mixin.py
+++ b/codebase_rag/parsers/class_ingest/mixin.py
@@ -22,6 +22,7 @@ from ...types_defs import (
     FunctionSpanKey,
     NodeType,
     PropertyDict,
+    RustTraitImpl,
 )
 from ...utils.path_utils import cached_relative_path, cached_resolve_posix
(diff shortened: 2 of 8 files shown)


vitali87:
Addressed the `_prop` nitpick in f338f7f5: the helper now asserts a single match and names the collisions on failure. It is the right call for these fixtures specifically, since the decoy tests only mean what they say if the node selected is the intended one.

PR #1056, opened 2026-08-03
fix(rust): bind a brace-list self import under the module name

### LINKED ISSUE #1054: fix(rust): bind `use path::{self}` under the module name, not the `self` keyword

vitali87:
Closes #1054.

## What

`use std::io::{self, IsTerminal};` records the imported module under the literal key `self` instead of the name it actually binds, `io`. A `self` entry in a brace list imports the base path itself, in scope under its last segment, and every later `io::...` spelling in that file is written against that name. Nothing ever looks up `self`, so the binding was unreachable while the real name was absent from the map.

The damage is not a silent miss. With no entry for the head, a qualified call falls through to the trie, which picks any same-named function it can find:

```rust
// src/app.rs
use crate::io::{self, Reader};
pub fn run() -> i32 { io::open() }   // bound to src/decoy.rs::open, not src/io.rs::open
```

## Where the name comes from

The bound name is the base path's last segment, so `use std::io::{self, Write}` records `io -> std::io`, and the nested form `use crate::{io::{self}, util}` behaves the same once the enclosing lists have assembled the base.

A relative base is the exception: `use super::{self, parent_module}` ends in a keyword, which is no name at all, and the name the parent module is in scope under is knowable only from the resolved path, which this layer does not have. That shape binds nothing rather than something wrong. The alias form (`use crate::io::{self as sys}`) already resolved through a separate path and is untouched, with a regression guard added.

## Two changed expectations

Two existing tests asserted that `self` is a key in the import map, one on `use std::fs::{self, File, read_to_string}` and one on `use super::{self, parent_module}`. Both pinned the defect rather than a behaviour: no consumer reads that key (path HEADS spelled `self::x` are handled elsewhere, on the value side), so the entry was unreachable data. They now assert the reachable binding and its absence respectively.

## The binding is weak

Rust keeps types and values in separate namespaces, so a module `io` and a function `io` are both in scope in one file, legally, while the import map holds one slot per name. A `{self}` entry therefore never displaces a name another `use` in the same scope already bound: it is recorded under a marked key that the import processor drops when the slot is taken. Without that, adding the binding would have evicted the function and sent a bare `io()` call to the trie, which picks any same-named function it can find.

The reverse order, a later `use` overwriting the module binding, is the flat map's existing limit and is unchanged by this PR.

## TDD

RED first: 3 of the 4 tests in `test_rust_self_import_binding.py` fail before the fix, the direct and nested brace-list forms and the namespace collision, each showing the call bound to a decoy in an unrelated module. The fourth is the alias regression guard, green throughout by design.

Full suite: 6797 passed, 10 skipped, against a 6793 / 10 baseline on `main`, so the delta is exactly the 4 tests this branch adds, with no new skips, xfails or deselections.

## Adversarial review

One round, two findings, both acted on.

1. Confirmed regression: the new binding was written unconditionally, so a file importing both a module and a function of the same name lost the function, and its bare call bound to a decoy. Fixed by the weak binding above, pinned RED first.
2. The keyword branch's comment was factually wrong. `use super::{self, x}` is not a construct whose name is merely unknowable here; rustc rejects it outright ("imports need to be explicitly named"). The guard stays, since this parses whatever is on disk, but the comment now says why. One pre-existing fixture in `test_multilang_import_parsing.py` contains that uncompilable spelling, and its assertion now records that nothing is bound.

## Measurements

Live ripgrep: 189 to 187, 0 newly listed, on this change alone, since it mostly repairs call edges rather than roots.

On top of #1048, now merged, it takes ripgrep from 189 to 161, 28 resolved and 0 newly listed. The extra 6 over #1048's own 22 are exactly the candidates that needed this head to resolve: both `io::Write` methods on `StandardStream`, `CommandReader.read`, `DecompressionReader.read`, and a closure inside a revived method.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rust import handling for `{self}` module bindings.
  * Correctly resolves module-qualified calls and nested imports.
  * Prevents conflicting imports from overriding existing bindings.
  * Supports aliased `self` imports more reliably.

* **Tests**
  * Added regression coverage for Rust self-import resolution and collision scenarios.

### DIFF
--- a/codebase_rag/constants/ast_rust.py
+++ b/codebase_rag/constants/ast_rust.py
@@ -183,6 +183,12 @@ RS_USE_LIST_DELIMITERS = frozenset({"{", "}", ","})
 RS_ENCODING_UTF8 = "utf8"
 
 RS_WILDCARD_PREFIX = "*"
+# Marks a `use path::{self}` entry, whose name the base path supplies rather
+# than the source. Rust keeps types and values in separate namespaces while
+# the import map holds one slot per name, so such a binding is WEAK: it never
+# displaces a name another `use` in the same scope already claimed (#1054).
+# No Rust identifier can contain the marker character.
+RS_SELF_MODULE_PREFIX = "@"
 
 RS_FIELD_ARGUMENT = "argument"
 
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -1415,6 +1415,13 @@ class CallResolver:
             return self._resolve_rust_prefixed_path(
                 object_path, item, module_qn, caller_qn
             )
+        # A module the caller bound with `use path::{self}` is a module, full
+        # stop: decide through it before the type probe below, whose
+        # name-based resolution would otherwise answer with a same-named
+        # value and abandon the path (issue #1054).
+        if binding := self._rust_self_module_head(head, module_qn, caller_qn):
+            mapped, scope = binding
+            return self._decide_rust_module_item(mapped, object_path[1:], item, scope)
         # A registered type as the first segment is an associated-function
         # call, owned by the class-resolution paths.
         if self._resolve_class_name(head, module_qn):
@@ -1442,6 +1449,20 @@ class CallResolver:
             module_qn, list(object_path), module_qn
         )
         return self._decide_rust_base(base, item)
+
+    def _rust_self_module_head(
+        self, head: str, module_qn: str, caller_qn: str | None
+    ) -> tuple[str, str] | None:
+        # A `use path::{self}` binding names a MODULE outright, so it answers
+        # a qualified `head::item` even when the shared import map's one slot
+        # went to a value of the same name, which Rust's separate namespaces
+        # allow (issue #1054). Innermost scope first, like every other Rust
+        # name lookup here.
+        self_modules = self.import_processor.rust_self_module_imports
+        for scope in (*self._rust_enclosing_scopes(module_qn, caller_qn), module_qn):
+            if mapped := self_modules.get(scope, {}).get(head):
+                return mapped, scope
+        return None
 
     def _resolve_rust_prefixed_path(
         self,
--- a/codebase_rag/parsers/import_processor.py
+++ b/codebase_rag/parsers/import_processor.py
@@ -408,6 +408,7 @@ class ImportProcessor:
         "rust_fn_scope_imports",
         "rust_fn_scope_mod_imports",
         "rust_block_scope_imports",
+        "rust_self_module_imports",
         "_rust_fn_scope_keys",
         "_rust_pending_mod_scope_uses",
         "_rust_mod_scope_registry",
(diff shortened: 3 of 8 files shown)


vitali87:
@coderabbitai review

vitali87:
Good catch, and it was not just a missing assertion. I added the `open` decoy you asked for and the test went red: under the collision, `io::open()` was binding to `aaa.open` through the suffix trie, a fabricated edge my fixture had been hiding behind a single-candidate lookup.

Root cause: dropping the module binding to protect the value binding left the module namespace with nothing to answer from. Rust has two namespaces here and the import map has one slot, so 178958ec gives the module namespace its own map. A `use path::{self}` binding is recorded in `rust_self_module_imports` unconditionally, and qualified `head::item` resolution consults it innermost scope first.

That lookup also had to be hoisted above the "head names a registered type" guard: `_resolve_class_name` resolved `io` to the same-named *function* and abandoned the path before any module probe ran, so the guard was firing on something that is not a type at all.

Now `io()` reaches `helpers.io` and `io::open()` reaches `io.open`, with decoys for both, which is what rustc does. Suite 6797 passed, 10 skipped.

vitali87:
@coderabbitai review

Issue #1050, opened 2026-08-03
feat(flow): three-verdict flow queries so empty results distinguish no flow from unanalysed coverage


vitali87:
## Problem

A flow query that returns nothing is ambiguous. The caller cannot tell "no flow exists" apart from "the flow sits outside what the analysis covers". For assurance use cases (does this API key reach a log line or an outbound request?) an absent path must never be read as a PASS. This gap was raised independently by four commenters in the public discussion of the tool, so it clears the multiple-requesters bar.

## Current behaviour

`FLOWS_TO` edges are emitted only where the analysis resolves a flow. The known ceilings (docs/architecture/data-flow-edges.md, "Scope of the current phase") are:

- not path-sensitive: a kill on one branch drops taint conservatively
- no `Parameter` nodes, no SSA precision
- the source/sink registry covers Python, JavaScript, TypeScript, Go, Java, Rust, C, C++ and C#; any other language emits no I/O or flow edges at all

None of this coverage information is queryable. An empty result over an unregistered language looks identical to a verified absence of flow in a fully analysed Python file.

Note: the docs page also still lists single-pass return-taint ordering as a ceiling ("a callee whose body is processed after its caller is not yet known to return a tainted value"). That limit was removed by the fixpoint in #712 and the cross-file case verifiably works; the docs need updating as part of this work.

## Desired behaviour

1. Indexing records analysis coverage: per file (or per module), whether flow analysis ran, which capture groups were active, and whether the language is in the source/sink registry.
2. A flow reachability question answers with one of three verdicts:
   - `FOUND`: a path exists, return it
   - `NO_FLOW`: no path, and every node the query could touch was inside analysed coverage
   - `UNKNOWN`: no path found, but part of the reachable surface was outside coverage; name the gaps (files, languages, capture groups)
3. The data-flow edges docs state the current ceilings accurately, including the removal of the return-taint ordering limit.

## Acceptance criteria

- Coverage metadata is present in the graph after indexing and queryable in Cypher.
- The query surface (CLI/MCP) exposes the three-verdict answer for a source-to-sink reachability question.
- A query over a repo containing an unregistered language returns `UNKNOWN` with that language named, not an empty result.
- The same query over a fully covered repo with no path returns `NO_FLOW`.
- Integration tests pin all three verdicts.

## Out of scope

- Path-sensitive analysis
- SSA or `Parameter` nodes
- New language registries

vitali87:
[...]

PR #1057, opened 2026-08-03
docs(flow): correct the stale return-taint ordering ceiling

vitali87:
The "Scope of the current phase" section still described return-taint propagation as a single source-order pass in which a callee walked after its caller is unknown at the call site. That ceiling was removed by #712: `FlowProcessor.finalize()` resolves per-function summaries with a worklist fixpoint once every file has been walked, so forward and cross-file return taint is emitted, transitively.

Verified against current main with the exact scenario the docs claimed would fail: `reader.py` returns `os.getenv("API_KEY")`, `app.py` prints the result. Both the cross-file `RETURN` edge and the `ENV::API_KEY` to `STDOUT` resource flow are emitted. The three existing tests in `test_forward_return_taint_e2e.py` cover this and pass.

The rewritten section separates the three distinct ceilings that were previously conflated under "intra-procedural": the body walk (intra-procedural, not path-sensitive), return taint (transitive across functions and files), and argument hand-off (one level, no taint into the callee body since parameters seed only the shadow set). `graph-schema.md` carried the same conflation and is corrected to match.

Docs only, no behaviour change. Related: #1050, which covers making an empty flow result distinguish "no flow" from "outside coverage".

## Summary by CodeRabbit

* **Documentation**
  * Clarified that return-taint propagation works transitively across functions and files.
  * Documented that argument hand-off remains limited to one level.
  * Updated data-flow and graph-schema guidance while preserving existing resource-flow and assignment rules.

### DIFF
--- a/docs/architecture/data-flow-edges.md
+++ b/docs/architecture/data-flow-edges.md
@@ -351,17 +351,20 @@ module is re-exported under its own name. A project that does
 
 ## Scope of the current phase
 
-`FLOWS_TO` is intentionally conservative and intra-procedural in this phase:
+`FLOWS_TO` is intentionally conservative in this phase:
 
-- Value flow is tracked within a function body plus one level of argument/return
-  hand-off. A callee returning **different** sources on different branches carries
-  every origin to its callers. It is not path-sensitive: a kill on one branch of an
-  `if`/`else` drops taint conservatively.
-- Return-taint propagation is a single source-order pass, so a callee whose body is
-  processed **after** its caller (typically cross-file) is not yet known to return a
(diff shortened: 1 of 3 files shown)


vitali87:
@coderabbitai review

vitali87:
Both fixed in 9b6e615b. The arg hand-off is now described as a `FLOWS_TO` edge with `kind = arg`, matching `FlowKind.ARG = "arg"` in `codebase_rag/parsers/flow_access/constants.py`. The return-edge description now says the edge terminates at the calling function rather than at the assignment, matching `_emit_return_edge`, which targets `caller_spec`.

PR #1060, opened 2026-08-03
fix: bind a Rust block-local item over the enclosing function body use

### LINKED ISSUE #1026: fix: Rust function-body use outranks a function item declared in a nested block of the same body
## Summary

A `use` at the top of a function body outranks a function item declared in a nested block inside that same body. The call written inside the nested block binds the `use`, but rustc binds the block's own item.

## Minimal repro

`src/gamma.rs`:

```rust
pub const fn g() -> u32 {
    3
}
```

`src/a.rs`:

```rust
pub const fn f0() -> u32 {
    use crate::gamma::g;
    let z = {
        const fn g() -> u32 {
            21
        }
        g()
    };
    z
}
```

rustc evaluates `f0()` to 21: the inner block's `const fn g` shadows the function-body `use` for calls inside that block.

cgr emits `a.f0 -> gamma.g` and no edge to the block-local item.

## Cause

`_try_resolve_rust_inline_scope` consults `rust_fn_scope_imports[caller_qn]` (the function's body uses) before anything models the inner block's own items. This is the lookup-side dual of the storage-side rule fixed on the #1007 branch, where every nested plain block declaring direct function items is an item scope that opts an initializer-block `use` out for calls inside it: the same item-scope precedence needs to apply against the function's own body uses.

## Expected

The call inside the nested block binds the block's local item; calls after the block keep binding the `use`.

Found by rustc differential fuzzing during the #1007 adversarial review (five fuzz seeds, all this one class; behaviour is identical before and after the #1007 branch).

vitali87:
Closes #1026.

A `use` at the top of a Rust function body binds its name for the rest of that body, but a nested block declaring its own function item is a separate item scope: calls written inside the block bind the block's item, and only calls outside it keep binding the `use`.

```rust
pub const fn f0() -> u32 {
    use crate::gamma::g;
    let z = {
        const fn g() -> u32 {
            21
        }
        g()          // rustc: 21, the block's own item
    };
    z
}
```

`_try_resolve_rust_inline_scope` read the caller's body uses with no regard for where the call sits, so `g()` bound `crate::gamma::g` and the block-local item took no edge at all.

## Change

Each function-body `use` now records the spans of the nested blocks in that function that declare function items directly, reusing the item-scope walk the initializer-block machinery already runs. Those items resolve to the qn the registry actually handed them, through the same span lookup the function-scope key itself goes through, and a call sited inside a block binds its item directly.

Binding by span rather than by name is the part that matters. A block item and a same-named item at module level both register flat in one module, one holding the natural qn and the other a `@<line>` variant, so a name lookup answers with whichever of them registered first. The first cut of this branch stood the `use` down and left the name lookup to it, which turned a wrong edge into a different wrong edge; the adversarial pre-push review caught it and the test that missed it had no module-level twin in its fixture.

## Tests

Red first, five of them. `test_block_item_shadows_the_enclosing_function_body_use` bound `gamma.g` before the fix. Two pin the contest against a same-named twin, one at module level and one in an enclosing inline `mod`, where the wrong answer is the twin rather than the import. Two pin the boundaries: a call after the block closes still binds the `use`, and a block declaring some other name leaves the `use` alone.

## Out of scope

A block-local item still wins by name, not by span, for a function with no body `use` at all, so a call written after the block can reach it. Filed as #1061.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rust call resolution for nested block-scoped functions.
  * Local function declarations now correctly take precedence over imports and outer-scope functions.
  * Scope boundaries are respected so declarations do not incorrectly affect unrelated blocks.

* **Tests**
  * Added coverage for shadowing behavior across nested blocks, modules, and matching names.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -848,7 +848,7 @@ class CallResolver:
         # cached under the file-scoped key.
         if language == cs.SupportedLanguage.RUST and caller_qn:
             scoped = self._try_resolve_rust_inline_scope(
-                call_name, module_qn, caller_qn
+                call_name, module_qn, caller_qn, call_point
             )
             if scoped is not None:
                 return scoped[0]
@@ -1331,7 +1331,7 @@ class CallResolver:
         return None
 
     def _try_resolve_rust_inline_scope(
-        self, call_name: str, module_qn: str, caller_qn: str
+        self, call_name: str, module_qn: str, caller_qn: str, call_point: int | None
     ) -> tuple[tuple[str, str] | None] | None:
         """Resolve a bare call through the caller's enclosing Rust scopes.
 
@@ -1342,7 +1342,11 @@ class CallResolver:
         the file module, checking each scope's own items, the caller's
         weak entries (its enclosing mod's use fanned out per function, in
         case the mod's shared key was arbitrated to a twin), and the
-        scope's import map. Initializer-block uses are served earlier, by
+        scope's import map. A body use loses to an item the call's own
+        enclosing block declares, whose flat-registered node the later
+        probes find (issue #1026); the walk below needs no such gate,
+        since it consults each scope's own items first already.
+        Initializer-block uses are served earlier, by
         the site-gated probe in _resolve_function_call. Returns a 1-tuple
         so a deliberate drop (the scope imports the name from OUTSIDE the
         indexed first-party graph) is distinguishable from "no scope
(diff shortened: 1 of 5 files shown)


vitali87:
SonarCloud flagged cognitive complexity 16 on `_try_resolve_rust_inline_scope`. Addressed in cbe875b5 by extracting the enclosing-scope walk into its own function; behaviour unchanged, full suite green at 6503 passed / 5 skipped.

Issue #1007, opened 2026-08-01
fix: Rust crate::/super::/self:: use paths never resolve, so trait impls link to phantom externals and lose all OVERRIDES edges


vitali87:
## Summary

Rust `use crate::...` (and `super::`/`self::`) paths are stored **raw** in the import mapping and never rewritten to project qualified names. When an `impl Trait for Type` block names a trait imported this way, the deferred inheritance resolver sees a qn that does not start with the project prefix, declares it external, and emits `IMPLEMENTS` onto a phantom `ExternalModule` node such as `crate.flags.Flag`. Because the trait parent never resolves first party, the method override pass emits **no OVERRIDES edges** from the impl methods to the trait methods.

Dead-code impact is severe: calls through `&dyn Trait` receivers resolve (correctly) to the trait's method declarations, but without OVERRIDES edges the liveness walk cannot expand from a live trait method to its implementations. On ripgrep this single defect accounts for **938 of 1811 dead-code candidates** (926 impl methods in `crates/core/flags/defs.rs` plus 12 trait members): every one of the ~200 CLI flag structs implements `Flag` and is invoked solely through `pub(super) const FLAGS: &[&dyn Flag]`.

## Minimal repro

Three files, standard `src` layout (the ripgrep `crates/core` layout without a `src` directory fails identically):

`src/main.rs`
```rust
mod flags;

fn main() {
    for flag in crate::flags::defs::FLAGS {
        println!("{}", flag.name_long());
    }
}
```

`src/flags.rs`
```rust
pub(crate) mod defs;

pub trait Flag {
    fn name_long(&self) -> &'static str;
}
```

`src/flags/defs.rs`
```rust
use crate::flags::Flag;

pub(crate) struct AfterContext;

impl Flag for AfterContext {
    fn name_long(&self) -> &'static str {
        "after-context"
    }
}

pub(crate) const FLAGS: &[&dyn Flag] = &[&AfterContext];
```

Actual graph:

```
proj.src.flags.defs.AfterContext IMPLEMENTS crate.flags.Flag      <- phantom external
proj.src.main.main CALLS proj.src.flags.Flag.name_long
(no OVERRIDES edges at all)
```

Expected:

```
proj.src.flags.defs.AfterContext IMPLEMENTS proj.src.flags.Flag
proj.src.flags.defs.AfterContext.name_long OVERRIDES proj.src.flags.Flag.name_long
```

[...]

PR #1064, opened 2026-08-04
fix: scope Rust block-local items to the block that declares them

### LINKED ISSUE #1061: fix: a Rust block-local fn item binds calls written outside its block
## Summary

A function item declared inside a nested block is registered flat in the enclosing module, so bare-name resolution reaches it from anywhere in that scope. A call written outside the declaring block binds it, where rustc binds whatever is actually in scope there.

## Minimal repro

`src/gamma.rs`:

```rust
pub const fn g() -> u32 {
    3
}
```

`src/a.rs`:

```rust
pub mod inner {
    use crate::gamma::g;

    pub const fn f0() -> u32 {
        let z = {
            const fn g() -> u32 {
                21
            }
            0
        };
        z + g()
    }
}
```

rustc evaluates `f0()` to 3: the block's `const fn g` is out of scope by the time `g()` is written, so the mod-level `use` binds it.

cgr emits `inner.f0 -> inner.g`, the block item, and no edge to `gamma::g`.

## Cause

The block item registers under the enclosing scope's flat qn, and `_scope_candidate` (`call_resolver.py`) answers the scope walk from that registry by name alone. Nothing records where the item was declared, so every call in the scope sees it.

PR #1060 gave function-body uses a span-keyed view of the blocks in their own function (`rust_fn_scope_item_holes`) precisely so the contest could be decided by span. The same spans need to gate the scope walk, which means recording them for every function rather than only for functions that carry a body `use`.

## Expected

The call binds the mod-level `use`; only calls written inside the block bind the block's item.

Found by the adversarial review on PR #1060.

vitali87:
Closes #1061. Extends #1060, now merged.

A Rust `fn` item declared inside a block is in scope for that block alone, but it registers flat in the enclosing module beside the module's own items. Bare-name resolution therefore reached it from anywhere in that scope.

```rust
pub mod inner {
    use crate::gamma::g;

    pub const fn f0() -> u32 {
        let z = {
            const fn g() -> u32 {
                21
            }
            0
        };
        z + g()      // rustc: 3, through the mod-level use
    }
}
```

cgr bound the block's `g`, which is out of scope by the time that call is written.

## Change

Block item scopes move from the caller-keyed map #1060 introduced to one keyed by file, collected once per parse where the root node and the registered qns are both in hand. Keying by file is what fixes the other direction of the same question: a `fn` declared inside a block is inside the block, so its own body binds the block's items even though it is a caller in its own right.

The registered qns of those items also form a rejection set, consulted by the three probes that could otherwise reach one by name from outside its block: the same-module probe, the enclosing-scope walk, and the simple-name trie fallback. Only the span-gated probe can reach a block item now.

## Tests

Red first, all three. A call after the block closes binds the mod-level `use`; a sibling function binds the module's own item and not the block's; and a `fn` nested inside the block binds the block's item from its own body.

Worth recording from the sibling fixture: which of two same-named items holds the natural qn and which gets the `@<line>` suffix follows registration order, not nesting, so there the block item took the natural qn and the module-level item is the suffixed one. Nothing but the span distinguishes them.

## Three defects the pre-push review found in the first cut, each now covered

Making the item map file-wide **armed the file-keyed resolution cache** for exactly the files that must not use it. A file with no `use` anywhere left it on, so whichever call resolved first answered for every later call of that name and an in-block call silently lost its edge. The cache gate now stands down for any file holding block items.

The rejection was **unconditional, and the taint walk resolves callees with no call site position**, so it could not be placed against a block span and lost `FLOWS_TO` edges outright. Scoping now answers "outside" only where the site is known; an unplaceable site keeps the item reachable.

The **dedup bucket fanned a resolved callee onto its `@<line>` twin**, which walks straight across the boundary in whichever direction the natural qn happened to fall. Those twins are different functions sharing a name, not spellings of one callee, so a bucket mixing a block item with another scope's item no longer fans out.

A fourth was a test that asserted only membership and let a wrong edge through unchecked; it now asserts the absence too.

The suite then caught one more of my own: a module-level `static` initializer's caller qn **is** the module qn, so the block-item probe never ran there. That probe moved up beside the initializer-block probe, where every call site reaches it. The existing `#1007` test covering that shape had been passing on the dedup fanout alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rust call resolution for functions declared inside blocks.
  * Prevented references from crossing block boundaries or incorrectly linking to same-named module functions.
  * Preserved correct resolution for nested functions, declaration order, and calls outside local blocks.
  * Maintained flow-analysis paths to block-local functions and reliable results when resolution caching is used.

* **Tests**
  * Added coverage for Rust block scope, caching, nested functions, declaration order, and flow-edge traversal.

### DIFF
--- a/codebase_rag/parsers/call_processor.py
+++ b/codebase_rag/parsers/call_processor.py
@@ -3695,7 +3695,20 @@ class CallProcessor:
                 callee_type = cs.NodeLabel.METHOD
                 callee_qn = init_qn
 
-            for target_qn in resolver.function_registry.variants(callee_qn):
+            targets = resolver.function_registry.variants(callee_qn)
+            if len(
+                targets
+            ) > 1 and not resolver.import_processor.rust_block_item_qns.isdisjoint(
+                targets
+            ):
+                # The dedup bucket holds a Rust block-local item beside an
+                # item of another scope. Those are different functions
+                # sharing a natural name, not spellings of one callee, and
+                # the span-gated probe already chose between them, so the
+                # hedge that fans an ambiguous callee onto its twins would
+                # cross the block boundary here (issue #1061).
+                targets = [callee_qn]
+            for target_qn in targets:
                 # A duplicate-suffixed variant may be a DIFFERENT kind of
                 # node (a TS namespace merged onto a function registers as
                 # a class); only callable variants take a CALLS edge, and
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -802,6 +802,7 @@ class CallResolver:
             and caller_qn
             and (
                 module_qn in self.import_processor.rust_block_scope_imports
+                or module_qn in self.import_processor.rust_block_items
                 or (
                     caller_qn.startswith(f"{module_qn}{cs.SEPARATOR_DOT}")
                     and (
@@ -861,7 +862,7 @@ class CallResolver:
         # this point same-module wins. Elsewhere imports shadow module-level
         # definitions.
         if language == cs.SupportedLanguage.RUST and (
-            result := self._try_resolve_same_module(call_name, module_qn)
+            result := self._try_resolve_same_module(call_name, module_qn, call_point)
         ):
             if use_cache:
                 self._simple_resolution_cache[cache_key] = result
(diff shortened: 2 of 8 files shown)


vitali87:
@coderabbitai review

vitali87:
Addressed the docstring-coverage warning. `record_rust_block_items` now carries a docstring; the remaining functions the check counted are the test functions, which follow this repo's convention of a module docstring stating the behaviour under test plus a comment on each case explaining what would go wrong, rather than per-test docstrings. The other new functions (`rust_block_item_scopes`, `_rust_block_item_at`, `_rust_block_item_hidden`) already had docstrings.

PR #1067, opened 2026-08-04
chore: update retired Claude model ids in server.json and review workflow

kasparovabi:
## Summary

- `server.json` advertises `claude-sonnet-4-20250514` as the default for `ORCHESTRATOR_MODEL` and `CYPHER_MODEL`. That id was retired on 15 June 2026, so anyone who installs the MCP server and takes the published defaults starts from an id the API no longer serves.
- The commented model line in `.github/workflows/claude-code-review.yml` points at `claude-opus-4-1-20250805`, which retires on 5 August 2026. Uncommenting it after that date breaks the review job.

| File | Line | Before | After |
| --- | --- | --- | --- |
| `server.json` | 37 | `claude-sonnet-4-20250514` | `claude-sonnet-4-6` |
| `server.json` | 53 | `claude-sonnet-4-20250514` | `claude-sonnet-4-6` |
| `.github/workflows/claude-code-review.yml` | 41-42 | `claude-opus-4-1-20250805`, "Claude Opus 4.1" | `claude-opus-4-8`, "Claude Opus 4.8" |

Retirement dates come from the Anthropic deprecation page: https://platform.claude.com/docs/en/about-claude/model-deprecations

Both replacements already appear in `codebase_rag/constants/providers.py`, so the context-window table covers them without further edits.

## What I left alone

`codebase_rag/constants/providers.py` and the tests under `codebase_rag/tests/` already use current ids, and the entries for older models there are a capability table rather than a model choice, so removing them would drop real data.

The same workflow comment says the action "defaults to Claude Sonnet 4". I did not touch that clause because it describes `anthropics/claude-code-action` behaviour rather than anything configured here, and I have not verified what that default is today.

## Type of Change

- [x] CI/CD or tooling
- [x] Bug fix

## Test Plan

- [x] Manual testing: `server.json` still parses as JSON, and the workflow change is inside a comment, so the job is unaffected until someone uncomments it.

Nothing executable changed, so the Python suite is untouched by this diff.

## Checklist

---

I read this diff line by line before opening it, and I am happy to adjust or drop any part of it. What brought me here was a checker I maintain that flags retired Claude model ids, so I am saying that up front rather than leaving you to wonder. I have not run either id against the API to reproduce a failure, the claim rests on the published retirement dates. If you would rather handle this your own way, or would rather not get this kind of contribution at all, tell me and I will not send another.

## Summary by CodeRabbit

* **Chores**
  * Updated the default AI model configuration to use the latest Claude Sonnet version.
  * Updated the example Claude Opus model reference in the code review workflow documentation.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -38,8 +38,8 @@ jobs:
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
 
-          # Optional: Specify model (defaults to Claude Sonnet 4, uncomment for Claude Opus 4.1)
-          # model: "claude-opus-4-1-20250805"
+          # Optional: Specify model (defaults to Claude Sonnet 4, uncomment for Claude Opus 4.8)
+          # model: "claude-opus-4-8"
 
           # Direct prompt for automated review (no @claude mention needed)
           direct_prompt: |
--- a/server.json
+++ b/server.json
@@ -34,7 +34,7 @@
         {
           "name": "ORCHESTRATOR_MODEL",
           "description": "Model name for the orchestrator agent",
-          "default": "claude-sonnet-4-20250514"
+          "default": "claude-sonnet-4-6"
         },
         {
           "name": "ORCHESTRATOR_API_KEY",
@@ -50,7 +50,7 @@
         {
           "name": "CYPHER_MODEL",
           "description": "Model name for Cypher query generation",
-          "default": "claude-sonnet-4-20250514"
+          "default": "claude-sonnet-4-6"
         },
         {
           "name": "CYPHER_API_KEY",


kasparovabi:
Filling in the one check the bot couldn't complete.

Greptile's T-Rex run flagged `T-Rex validation blocked` because its Anthropic key came back 401, so the live Models API check never ran. I ran it from my side today (7 Aug 2026) against `GET /v1/models`:

- `claude-sonnet-4-6` — present in the response
- `claude-opus-4-8` — present in the response
- `claude-sonnet-4-20250514` (removed by this PR) — **absent**
- `claude-opus-4-1-20250805` (removed by this PR) — **absent**

That matches what the local routing check already concluded, so the two signals now agree.

One thing worth noting beyond this diff: the replacements are unversioned aliases rather than dated snapshots. A pinned id like `claude-sonnet-4-20250514` keeps working right up to its retirement date and then fails everywhere at once, with no warning in between. The aliases don't rot the same way, which is why the same edit shouldn't be needed again next cycle.

If it's useful, I wrote a small script for another repo that scans a codebase for Claude model ids and verifies each one against the Models API, exiting non-zero when a referenced id is no longer served — it makes this a check you can schedule rather than something you notice when CI breaks. Happy to port it here if you'd want it.

vitali87:
@coderabbitai full review

vitali87:
@greptile-apps review

PR #1077, opened 2026-08-04
fix: bind a Rust trait-impl method to the trait its own block names

### LINKED ISSUE #1076: fix: a dedup-variant method gets no OVERRIDES edge, so it is unreachable from every root
## Summary

When two Rust methods on one type share a name, the second registers as the dedup variant `natural@<line>`. The override pass splits that qualified name on the last dot, producing `method_name = "run@13"`, and then looks for a base method literally named `run@13`. No such method exists, so the variant never receives an `OVERRIDES` edge.

The variant is therefore unreachable from every root, and everything it calls is dead inside dead. This is what stops the caller-attribution fix behind #1014 from clearing the real-world false positives that motivated it.

## Minimal repro

```rust
// src/lib.rs
pub mod alpha;
pub mod foo;

// src/alpha.rs
pub(crate) fn other() -> u32 { 2 }

// src/foo.rs
pub trait Alpha { fn run(&self) -> u32; }
pub trait Beta { fn run(&self) -> u32; }

pub struct S;

impl Alpha for S {
    fn run(&self) -> u32 {
        other()
    }
}

impl Beta for S {
    fn run(&self) -> u32 {
        use crate::alpha::other;
        other()
    }
}

pub fn other() -> u32 { 1 }
```

## Expected

```
foo.S.run    OVERRIDES foo.Alpha.run
foo.S.run@13 OVERRIDES foo.Beta.run
```

## Actual

```
foo.S.run    OVERRIDES foo.Alpha.run
```

`foo.S.run@13` has no `OVERRIDES` edge. With `alpha::other` made `pub(crate)` so that export does not root it, it is reported dead despite its only caller being a live trait method.

## Analysis

`process_all_method_overrides` in `codebase_rag/parsers/class_ingest/method_override.py:34` does `method_qn.rsplit(SEPARATOR_DOT, 1)`, which keeps the `@<line>` suffix on the method name. `check_method_overrides` then looks up `foo.Beta.run@13` and misses, and `_find_override_by_arity` bails because `_signature_arity` returns `None` for a name with no parentheses.

The base-method lookup needs the natural name, while the edge itself must still be emitted from the variant qn. Every other consumer already splits on `DUP_QN_MARKER` and keeps the base.

## Out of scope

- The caller-attribution half of this shape, fixed and pinned by #1075.
- Non-Rust languages that mint the same variants.

vitali87:
Closes #1076.

The override pass rebuilt the relationship from the method's qualified name: split off the last segment, then walk the implemented traits in sorted order and take the first that declares a matching name. Two traits declaring one name broke that walk twice over.

## The two faces

A dedup variant keeps its `@<line>` suffix in the derived name, so the lookup for `Beta.run@13` misses and the second method gets no `OVERRIDES` edge at all, leaving it unreachable from every root.

And the surviving edge lands on whichever trait sorts first rather than the one written above the method:

```rust
impl Beta for S  { fn run(&self) -> u32 { 1 } }   // line 7,  registers as S.run
impl Alpha for S { fn run(&self) -> u32 { 2 } }   // line 13, registers as S.run@13
```

`S.run` is Beta's method, yet it got `OVERRIDES Alpha.run`. A wrong edge is worse than a missing one, because reachability flows through `OVERRIDES`: a live caller of `Alpha::run` marks the wrong method live and leaves the real one dead.

## The fix

`_rust_trait_impls` already pairs each impl block's trait with the method qns it ingested, and `resolve_deferred_inherits` runs before `process_all_method_overrides`. So the binding is recorded there for first-party traits and consumed as the authoritative parent, instead of being re-derived from strings. Where nothing is recorded, the ancestry walk runs exactly as before.

The dedup suffix is dropped only after an impl block has vouched for the method. Stripping it for every method instead caused two regressions, both caught in review and both now pinned by tests:

- an INHERENT `impl S { fn run }` sharing a trait method's name gained a false `OVERRIDES` edge, reviving a genuinely dead method
- a C# verbatim identifier `@event` collapsed to the empty string, silently losing its real override edge

## Scope

Impact is limited to first-party traits declared in the same crate. The `Display`/`Debug` shape common in real Rust already took the external-trait path, where both `Glob.fmt` and `Glob.fmt@12` are flagged `overrides_external` and neither was ever a false positive.

## Tests

Four new tests, each verified to fail without the change it pins. The two impl-order fixtures are exact mirrors, so an implementation reading the sorted trait list rather than the impl block passes one and fails the other.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved override relationship resolution for Rust trait implementations, including ambiguous and deduplicated method names.
  * Prevented false override relationships for inherent Rust methods.
  * Preserved correct C# overrides when methods use verbatim identifiers.

* **Tests**
  * Added regression coverage for Rust trait selection, method naming, and C# override handling.

### DIFF
--- a/codebase_rag/parsers/class_ingest/method_override.py
+++ b/codebase_rag/parsers/class_ingest/method_override.py
@@ -35,10 +35,6 @@ def process_all_method_overrides(
             parts = method_qn.rsplit(cs.SEPARATOR_DOT, 1)
             if len(parts) == 2:
                 class_qn, method_name = parts
-                # The dedup variant's name carries the line that made it
-                # unique; no base method is spelled that way, so a derived
-                # lookup keeping the suffix always misses (issue #1076).
-                method_name = method_name.split(cs.DUP_QN_MARKER, 1)[0]
                 if _emit_recorded_impl_override(
                     method_qn,
                     method_name,
@@ -179,6 +175,12 @@ def _emit_recorded_impl_override(
     trait_qn = (impl_method_traits or {}).get(method_qn)
     if trait_qn is None:
         return False
+    # Only now, with an impl block vouching for this method, is the dedup
+    # suffix safe to drop. Stripping it for every method instead would hand
+    # the ancestry walk a name it could not match before, inventing an
+    # override for an INHERENT method that merely shares a trait method's
+    # name, and would empty out a C# verbatim identifier like `@event`.
+    method_name = method_name.split(cs.DUP_QN_MARKER, 1)[0]
     parent_method_qn = f"{trait_qn}{cs.SEPARATOR_DOT}{method_name}"
     if function_registry.get(parent_method_qn) != NodeType.METHOD:
         # An inherent method, or one the trait declares nowhere: the walk has
--- a/codebase_rag/parsers/class_ingest/method_override.py
+++ b/codebase_rag/parsers/class_ingest/method_override.py
@@ -22,6 +22,7 @@ def process_all_method_overrides(
     interface_implementers: dict[str, set[str]] | None = None,
     csharp_methods: set[str] | None = None,
     csharp_override_methods: set[str] | None = None,
+    impl_method_traits: dict[str, str] | None = None,
 ) -> None:
     logger.info(logs.CLASS_PASS_4)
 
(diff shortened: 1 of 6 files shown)


vitali87:
Confirmed real, and confirmed pre-existing rather than introduced here. Running the inherent-first fixture on `origin/main` and on this branch:

```
main    S.run -> Alpha.run
branch  S.run -> Alpha.run
        S.run@10 -> Alpha.run
```

with

```rust
pub trait Alpha { fn run(&self) -> u32; }
pub struct S;
impl S { pub fn run(&self) -> u32 { 2 } }
impl Alpha for S { fn run(&self) -> u32 { 1 } }
```

`S.run` is the inherent method in both runs, and main already gives it the false edge. This branch adds the correct `S.run@10 -> Alpha.run` for the trait method beside it, so the false edge is neither created nor widened here.

The regression this PR's own test pins is the mirror case, where stripping the dedup suffix for every method invented an edge that main did not emit. That one was introduced by an earlier draft of this change and is fixed.

The remaining defect is the same root cause one level further: absence from `impl_method_traits` is read as no information, when for a Rust inherent impl block it is positive evidence that the method implements nothing. Rust has no inheritance, so an inherent method can never override. Fixing it means recording inherent impl-block methods too, which is a separate change against a separate fixture. Filed as a follow-up and taken next rather than folded into this one.

PR #1079, opened 2026-08-04
fix: never give a Rust inherent impl method an override edge

### LINKED ISSUE #1078: fix: a Rust inherent impl method gets a false OVERRIDES edge to a same-named trait method
## Summary

Rust has no inheritance, so a method can only override when it is written in an `impl Trait for Type` block. A method in an inherent `impl Type` block implements nothing and must have no `OVERRIDES` edge. The override pass does not know which kind of block a method came from, so when an inherent method shares a name with a trait the type also implements, the ancestry walk links the two.

## Minimal repro

```rust
// src/lib.rs
pub mod foo;

// src/foo.rs
pub trait Alpha { fn run(&self) -> u32; }

pub struct S;

impl S {
    pub fn run(&self) -> u32 { 2 }
}

impl Alpha for S {
    fn run(&self) -> u32 { 1 }
}
```

## Expected

```
foo.S.run@10 OVERRIDES foo.Alpha.run
```

Only the trait-impl method. The inherent `foo.S.run` implements nothing.

## Actual

```
foo.S.run    OVERRIDES foo.Alpha.run
foo.S.run@10 OVERRIDES foo.Alpha.run
```

The inherent method carries an edge to a trait it does not implement.

## Consequence

Reachability expands along `OVERRIDES`, so a live caller of `Alpha::run` marks the unused inherent method live too, suppressing a genuine dead-code candidate. Override-based dispatch also gains a target that Rust name resolution never picks.

## Analysis

#1077 records `impl_method_traits` for methods in first-party trait impl blocks and uses it as the authoritative parent. A method absent from that map falls through to the generic ancestry walk. For a Rust inherent impl block, that absence is not missing information: it is positive evidence that the method implements nothing.

The fix is to record inherent impl-block method qns alongside the trait ones, and skip the walk for any method known to be inherent. `_ingest_rust_impl` already distinguishes the two branches by whether `extract_impl_trait` returns a name; only the trait branch collects method qns today.

Pre-existing behaviour, present on main independently of #1077. Surfaced by a review bot on that PR.

## Out of scope

- Supertrait method resolution.
- Languages other than Rust, where a base class legitimately supplies the parent.

vitali87:
Closes #1078.

Rust has no inheritance, so a method can only override when it is written in an `impl Trait for Type` block. A method in an inherent `impl Type` block implements nothing. The override pass could not tell the two apart, so an inherent method sharing a name with a trait the type also implements was linked to it.

```rust
pub trait Alpha { fn run(&self) -> u32; }
pub struct S;
impl S { pub fn run(&self) -> u32 { 2 } }          // inherent
impl Alpha for S { fn run(&self) -> u32 { 1 } }    // trait impl
```

Before, both methods pointed at `Alpha.run`. After, only the trait-impl method does, whichever of the two happens to hold the deduplicated qualified name.

Reachability expands along `OVERRIDES`, so the false edge kept an unused inherent method alive and suppressed a genuine dead-code candidate.

## The fix

`_ingest_rust_impl_methods` already knows which block it is in. It now collects the block's method qns either way and records the inherent ones, so the override pass is told rather than left to infer from absence.

The classifier keys on `extract_impl_trait_path`, not `extract_impl_trait`. The latter reads no name off a generic wrapped around a scoped path, returning `None` for `impl std::ops::Add<u32> for S` exactly as it does for a genuinely inherent block:

```
trait=None  path=std::ops::Add       impl std::ops::Add<u32> for S
trait=None  path=crate::al::Alpha    impl<T> crate::al::Alpha<T> for S
trait=None  path=None                impl S
trait=None  path=None                impl<T> S<T>
```

An earlier draft classified by name and barred those real trait impls from ever overriding. That regression is pinned by `test_a_scoped_generic_trait_impl_is_not_mistaken_for_inherent`, verified to fail against it.

The recorded trait binding is also consulted before the inherent gate, so positive evidence outranks a classification drawn from absence if the two ever disagree.

## Tests

Two new tests, each verified to fail without the change it pins, plus the existing three still green. The inherent-first fixture asserts nothing at all comes from the natural qn rather than merely no edge to `Alpha`, so an over-broad gate that silenced another trait would not pass.

## Out of scope

`impl path::Trait<Args> for Type` records no entry in `interface_implementers`, because that path also reads the trait NAME. So such a block alone still yields no override edge; the test above reaches `Base` through a supertrait instead. Pre-existing and unchanged here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Rust code analysis to distinguish inherent implementation methods from trait implementations.
  - Prevented inherent methods from being incorrectly reported as overrides, regardless of implementation block order.
  - Improved detection of scoped and generic trait implementations.
  - Ensured trait override relationships use accurate, deduplicated method names.

- **Tests**
  - Added regression coverage for Rust inherent implementations and generic trait implementations.

### DIFF
--- a/codebase_rag/parsers/class_ingest/method_override.py
+++ b/codebase_rag/parsers/class_ingest/method_override.py
@@ -29,41 +29,41 @@ def process_all_method_overrides(
 
     implemented_interfaces = _invert_implementers(interface_implementers or {})
     for method_qn in function_registry.keys():
-        if (
-            function_registry[method_qn] == NodeType.METHOD
-            and cs.SEPARATOR_DOT in method_qn
+        if function_registry[method_qn] != NodeType.METHOD:
+            continue
+        # A dotless qn has no class to walk from; rpartition leaves class_qn
+        # empty for it, which is the same set the membership test used to skip.
+        class_qn, _, method_name = method_qn.rpartition(cs.SEPARATOR_DOT)
+        if not class_qn:
+            continue
+        # Positive evidence first: a recorded trait binding names the parent
+        # outright, so it outranks a classification drawn from absence,
+        # whichever way the two ever disagree.
+        if _emit_recorded_impl_override(
+            method_qn,
+            method_name,
+            function_registry,
+            ingestor,
+            impl_method_traits,
         ):
-            parts = method_qn.rsplit(cs.SEPARATOR_DOT, 1)
-            if len(parts) == 2:
-                class_qn, method_name = parts
-                # Positive evidence first: a recorded trait binding names the
-                # parent outright, so it outranks a classification drawn from
-                # absence, whichever way the two ever disagree.
-                if _emit_recorded_impl_override(
-                    method_qn,
-                    method_name,
-                    function_registry,
-                    ingestor,
-                    impl_method_traits,
-                ):
-                    continue
-                if inherent_impl_methods and method_qn in inherent_impl_methods:
-                    # Written in an inherent `impl Type` block, so it implements
-                    # nothing. Rust has no inheritance, and the walk below would
-                    # otherwise hand it the first trait the type implements that
-                    # happens to declare this name.
-                    continue
-                check_method_overrides(
-                    method_qn,
-                    method_name,
-                    class_qn,
-                    function_registry,
-                    class_inheritance,
-                    ingestor,
-                    implemented_interfaces,
-                    csharp_methods,
-                    csharp_override_methods,
-                )
+            continue
+        if inherent_impl_methods and method_qn in inherent_impl_methods:
+            # Written in an inherent `impl Type` block, so it implements
+            # nothing. Rust has no inheritance, and the walk below would
+            # otherwise hand it the first trait the type implements that
+            # happens to declare this name.
+            continue
+        check_method_overrides(
+            method_qn,
+            method_name,
+            class_qn,
+            function_registry,
+            class_inheritance,
+            ingestor,
+            implemented_interfaces,
+            csharp_methods,
+            csharp_override_methods,
+        )
     _process_mro_shadow_overrides(function_registry, class_inheritance, ingestor)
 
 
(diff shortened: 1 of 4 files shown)


vitali87:
Real gap, and it is the one this PR already lists as out of scope. The suggested change does not close it, though, which is why it stays separate.

I tried it directly:

```python
if trait_name := (
    rs_utils.extract_impl_trait(class_node)
    or rs_utils.extract_impl_trait_path(class_node)
):
```

against a fixture holding only `impl crate::b::Base<u32> for S`. Still no OVERRIDES edge, and no IMPLEMENTS edge either. The fallback yields `crate::b::Base`, and `_resolve_to_qn` resolves a SIMPLE name, so the result is `<module_qn>.crate::b::Base`, a phantom that `resolve_deferred_inherits` discards by design rather than emit a dangling endpoint.

Closing it needs the trait path resolved through the Rust crate-path machinery, which is its own change against its own fixtures. Filed as #1080.

What this PR does claim is narrower and holds: the inherent-versus-trait classification keys on `extract_impl_trait_path`, so a scoped-generic block is never mistaken for inherent and its methods are never barred from overriding. That is pinned by `test_a_scoped_generic_trait_impl_is_not_mistaken_for_inherent`, which reaches `Base` through a supertrait precisely because the direct route is the gap now tracked in #1080.

vitali87:
SonarCloud flagged one new issue, `python:S3776`, cognitive complexity 17 against a limit of 15 in `process_all_method_overrides`. The quality gate passed, but the call is fair: the two guards this PR adds sat inside a `for` wrapping an `if` wrapping an `if`, so each one cost more than it should.

Fixed in db645c2a by inverting the two outer conditions into early `continue`s, which removes both nesting levels without changing what is skipped. `rpartition` leaves `class_qn` empty exactly for the dotless names the `SEPARATOR_DOT in method_qn` membership test used to filter, so the same set is skipped as before.

Full unit suite green at 6537 passed, 5 skipped.

Issue #1100, opened 2026-08-07
fix: Rust declaration reads count excluded neighbour and entry files


vitali87:
## Summary

`ImportProcessor._rust_module_is_declared`, `_rust_module_decls`, and `_rust_entry_decls` read candidate declaring files straight from disk with no exclude or unignore filtering. The redirect sweep itself now walks with the indexer's predicate (#1088), but the per-file reads behind it still see files the user excluded:

- A `#[path]` target whose physical neighbour declares it is left unmoved, on the grounds that rustc compiles it into the neighbour's tree. When that neighbour is excluded by `--exclude`, the graph holds no such tree, yet the target still keeps the physical `super::` climb instead of following the redirect that IS indexed.
- Entry-file declaration maps (`_rust_entry_decls`) feed crate path and gate resolution; an excluded entry file's `mod` declarations still shape both.

## Fix direction

Gate each disk read on the same `exclude_paths`/`unignore_paths` predicate the indexer uses (`should_skip_rel_file`), so a declaration the graph never holds cannot decide where an indexed module sits.

## Out of scope

- The sweep's directory walk (#1088).

Issue #1099, opened 2026-08-07
fix: the C++ module qn map walks without --exclude and .cgrignore


vitali87:
## Summary

`cpp_frontend/qn.py::_eligible_rel_files` reproduces `GraphUpdater._collect_eligible_files`' walk order, but filters with `should_skip_rel_file(rel_path_str, dir_parts, suffix)` and no exclude or unignore sets, and prunes no directories at all. `build_module_qn_map` is also called from `ImportProcessor` for C++ includes with the same two gaps.

The indexer walks with `exclude_paths` (`--exclude`) and `unignore_paths` (`.cgrignore` rescues), so the two walks disagree on the file set:

- An excluded file still claims its base qn in the collision map, so an indexed file colliding on the basename gets the extension-suffixed qn (`proj.foo.h`) while the tree-sitter pass, which never saw the excluded file, keys it `proj.foo`. Every qn the libclang path synthesises for that file misses the graph.
- A file rescued by `.cgrignore` out of a built-in ignore is indexed but invisible to the map, so includes resolving to it get no qn.

## Fix direction

Thread the same `exclude_paths` and `unignore_paths` the updater holds into both call sites, prune directories with `path_utils.should_keep_dir`, and pass the sets to `should_skip_rel_file`, as the Rust redirect sweep now does (#1088).

PR #1102, opened 2026-08-07
fix: prune the Rust redirect sweep with the indexer's exclude and unignore predicate

### LINKED ISSUE #1088: fix: the Rust redirect sweep ignores --exclude and .cgrignore
## Summary

`ImportProcessor._rust_redirect_parent_map` sweeps the repository for `#[path]` declarations and prunes directories with `has_ignored_dir_part`, which covers the built-in ignore set and the `src/bin` carve-out.

`GraphUpdater._should_keep_dir` prunes by two more things the sweep never sees: `exclude_paths` (`--exclude`) and `unignore_paths` (`.cgrignore` rescues).

`ImportProcessor` receives neither, so:

- A declaration inside a user-excluded subtree can claim the qn of an indexed file, and `super::` inside that file then climbs to a module the user asked cgr not to index.
- A subtree rescued by `.cgrignore` out of a built-in ignore is walked by the indexer but not by the sweep, so a redirect declared there is missed and `super::` in its target keeps the physical climb.

## Reproduction sketch

```
src/lib.rs           pub mod engine;
src/engine.rs        pub mod sib;
src/engine/sib.rs    pub fn run() -> i32 { 1 }
tools/gen.rs         #[path = "../src/engine/sib.rs"] mod moved;
```

with `--exclude tools/`. The sweep reads `tools/gen.rs`, so `src.engine.sib` is recorded as moved and declared by `tools.gen`, a module the graph does not hold.

## Fix direction

Pass the same exclude and unignore sets the updater walks with into `ImportProcessor`, and prune the sweep with the shared predicate rather than the built-in half of it.

## Out of scope

- The built-in ignore set and the `src/bin` carve-out, which the sweep already honours (#1085).

vitali87:
## Summary

- The `#[path]` redirect sweep now prunes directories and skips files with the same `--exclude`/`.cgrignore` predicate the indexer walks with, so a declaration in an excluded subtree claims nothing and one in a rescued subtree is read (issue #1088).
- The indexer's `_should_keep_dir` body moves to `path_utils.should_keep_dir` and both walks call it, so the two cannot drift again.
- `ImportProcessor` receives `exclude_paths`/`unignore_paths` from the factory; the sweep's blanket dot-directory skip is gone, since the indexer has no such rule (dot directories are excluded through `IGNORE_PATTERNS`, which both sides read).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Closes #1088. The same-family gaps found while fixing this are filed as #1099 (C++ module qn map) and #1100 (Rust neighbour/entry declaration reads).

## Test Plan

- [x] New tests added
- [ ] Manual testing (describe below)

Three new tests, each red before the fix:

- a redirect declared in a `--exclude`d subtree leaves the parent map empty (the resolver's fallback happens to land the right physical edge today, so the map itself is asserted);
- a redirect excluded by a FILE pattern, where the decoy module stays indexed, no longer records the wrong CALLS edge into the excluded declarer's tree;
- a redirect in a built-in-ignored directory rescued by `.cgrignore` moves its target, so `super::` in the target resolves through the declarer instead of the physical shadow sibling.

## Checklist

- [ ] No new comments or docstrings (code should be self-documenting): the new `should_keep_dir` carries a docstring and the sweep carries a why-comment, matching the density of the surrounding predicates in `path_utils.py`; they record the invariant (sweep and indexer walk the same set) that the code alone cannot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository indexing consistency for excluded, ignored, and unignored directories and files.
  * Rust redirect mappings now reflect the same filtering rules as indexed source files.
  * Preserved discovery of eligible files in Cargo `src/bin` directories.
  * Ensured restored paths are correctly included in indexing and redirect mapping.
* **Tests**
  * Added coverage for excluded directories, file patterns, and paths restored by unignore rules.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -81,10 +81,9 @@ from .utils.dependencies import has_semantic_dependencies
 from .utils.fqn_resolver import find_function_source_by_fqn
 from .utils.path_utils import (
     cached_relative_path,
-    matches_ignore_patterns,
+    should_keep_dir,
     should_skip_path,
     should_skip_rel_file,
-    unignore_could_match_within,
 )
 from .utils.source_extraction import extract_source_with_fallback
 
@@ -1480,28 +1479,8 @@ class GraphUpdater:
         return None, None
 
     def _should_keep_dir(self, dirname: str, dir_prefix: str) -> bool:
-        rel_dir = f"{dir_prefix}{dirname}"
-        # an explicit exclude can never be rescued by unignore (excludes win
-        # at the file level too), so prune the subtree outright.
-        if self.exclude_paths and matches_ignore_patterns(
-            f"{rel_dir}/", self.exclude_paths
-        ):
-            return False
-        if dirname not in cs.IGNORE_PATTERNS:
-            return True
-        # Cargo's src/bin/ holds first-party binaries, not build output;
-        # mirrors has_ignored_dir_part.
-        if (
-            dirname == cs.DIR_BIN
-            and dir_prefix.rstrip(cs.SEPARATOR_SLASH).rsplit(cs.SEPARATOR_SLASH, 1)[-1]
-            == cs.DIR_SRC
-        ):
-            return True
-        return bool(
-            self.unignore_paths
-            and any(
-                unignore_could_match_within(u, rel_dir) for u in self.unignore_paths
-            )
+        return should_keep_dir(
+            dirname, dir_prefix, self.exclude_paths, self.unignore_paths
         )
 
     def _drop_cache_if_graph_lost(self) -> None:
--- a/codebase_rag/parsers/factory.py
+++ b/codebase_rag/parsers/factory.py
@@ -79,6 +79,8 @@ class ProcessorFactory:
                 project_name=self.project_name,
                 ingestor=self.ingestor,
                 function_registry=self.function_registry,
+                exclude_paths=self.exclude_paths,
+                unignore_paths=self.unignore_paths,
             )
         return self._import_processor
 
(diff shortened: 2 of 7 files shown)


vitali87:
Bot report disposition:

**SonarCloud S3776** (cognitive complexity 17 > 15 in `_rust_redirect_parent_map`): fixed in 1e7e64c7 by extracting the per-directory file filter into `_swept_rust_files`; both functions now sit well under the threshold and the behaviour is unchanged (139 Rust sweep and redirect tests pass).

**Greptile P1** (dot directories no longer pruned by the sweep): rebutted. The indexer has never had a blanket dot-directory rule; it prunes only the named entries in `IGNORE_PATTERNS` (`.git`, `.venv`, and friends), which both walks still honour. A live probe confirms an arbitrary hidden directory such as `.gen/decl.rs` IS indexed and gets a `Module` node, so the old sweep-only dot-skip was itself a parity bug: a redirect declared in an indexed hidden directory was invisible to the sweep, exactly the divergence class this PR removes. Aligning the sweep on `should_keep_dir` is therefore the fix, not a regression. If a blanket hidden-directory policy is ever wanted it now has a single home in `should_keep_dir`.

**CodeRabbit docstring coverage warning** (33.33% < 80%): rebutted. The repository's PR checklist mandates self-documenting code and bans new docstrings as the default; the one docstring added (`should_keep_dir`) records the cross-walk invariant that the code alone cannot. The warning gate conflicts with repo policy and is non-blocking.

PR #1108, opened 2026-08-07
fix: walk the C++ module qn map with --exclude and .cgrignore

### LINKED ISSUE #1099: fix: the C++ module qn map walks without --exclude and .cgrignore

vitali87:
Closes #1099.

## Changes

`cpp_frontend/qn.py::_eligible_rel_files` now mirrors `GraphUpdater._collect_eligible_files` completely:

- prunes directories with `should_keep_dir(...)` (it pruned nothing at all before)
- passes `exclude_paths`/`unignore_paths` to `should_skip_rel_file`
- skips `cs.CGR_STATE_FILENAMES`, which the indexer skips and this walk did not

`build_module_qn_map` and `CppQnResolver` take the two sets, threaded from `GraphUpdater` through `run_cpp_frontend` / `run_cpp_frontend_hybrid`, and from `ImportProcessor` (which already held both) at the include-resolution call site.

## Tests

`codebase_rag/tests/test_cpp_qn_map_exclude_unignore.py`:

- an excluded `foo.cpp` no longer claims `proj.foo`, so the indexed `foo.h` keeps it instead of being pushed onto `proj.foo.h`
- a `.cgrignore`-rescued file under `node_modules/` gets a qn instead of being invisible
- an excluded directory is pruned
- **parity**: the qn map's key set equals `GraphUpdater._collect_eligible_files`' file set, under exclude, under unignore, with no filters, and with a state file present

6 of the 9 fail without the fix. Full C++ suite: 462 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * C++ code indexing now consistently respects excluded and explicitly included paths across standard and hybrid processing.
  * Improved module-file discovery by skipping filtered directories, state files, and ineligible files.
  * Prevented incorrect qualified-name mappings caused by excluded files or filename collisions.

* **Tests**
  * Added coverage confirming filtering behavior and consistency between C++ indexing and repository file discovery.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -360,6 +360,8 @@ class GraphUpdater:
                 structural_elements=(
                     self.factory.structure_processor.structural_elements
                 ),
+                exclude_paths=self.exclude_paths,
+                unignore_paths=self.unignore_paths,
             )
             logger.info(
                 ls.CPP_FRONTEND_HYBRID_PENDING.format(
@@ -376,6 +378,8 @@ class GraphUpdater:
             function_registry=self.function_registry,
             simple_name_lookup=self.simple_name_lookup,
             structural_elements=self.factory.structure_processor.structural_elements,
+            exclude_paths=self.exclude_paths,
+            unignore_paths=self.unignore_paths,
         )
         logger.info(
             ls.CPP_FRONTEND_COVERED.format(count=len(self._cpp_frontend_covered))
--- a/codebase_rag/parsers/cpp_frontend/frontend.py
+++ b/codebase_rag/parsers/cpp_frontend/frontend.py
@@ -803,6 +803,8 @@ def run_cpp_frontend(
     function_registry: FunctionRegistryTrieProtocol | None = None,
     simple_name_lookup: SimpleNameLookup | None = None,
     structural_elements: dict[Path, str | None] | None = None,
+    exclude_paths: frozenset[str] | None = None,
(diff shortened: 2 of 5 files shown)


vitali87:
@coderabbitai full review

@greptile-apps review

vitali87:
@coderabbitai full review

Issue #1092, opened 2026-08-05
ci: unit tests download the embedding model from HuggingFace at run time


vitali87:
## Summary

Three tests in the unit suite download `microsoft/unixcoder-base` from HuggingFace while running:

- `test_embedder.py::test_embed_code_integration`
- `test_embedder.py::test_similar_code_has_similar_embeddings`
- `test_semantic_search_eval.py::test_cgr_semantic_search_retrieves_expected_function`

They are gated on `_has_semantic_deps()`, which checks that torch and transformers are importable. Nothing checks that the weights are present or that the hub is reachable, so the unit job depends on a live third-party service.

## Observed

Two failures on one day, on different runners and for different reasons:

```
Unit Tests (ubuntu-latest, py3.12)
  429 Too Many Requests for url: https://huggingface.co/microsoft/unixcoder-base/resolve/main/config.json

Unit Tests (windows-latest, py3.13)
  OSError: We couldn't connect to 'https://huggingface.co' to load the files,
  and couldn't find them in the cached files.
```

Both were unrelated to the changes under test and both cleared on a rerun, so the cost is a red PR and a manual rerun each time.

## Fix direction

Cache the hub directory in CI so the download happens once per cache key rather than once per job, keeping the tests a hard failure when they do run. Skipping on a network error is the wrong direction: it turns an outage into a silent pass and hides a real embedder break.

For a developer machine with neither cache nor network, a skip whose reason names the missing weights is honest, since the alternative is an unrunnable suite.

vitali87:
Third occurrence today, this time taking down the coverage job rather than a unit matrix entry:

```
SonarCloud Analysis / Run tests with coverage
  429 Too Many Requests for url: https://huggingface.co/microsoft/unixcoder-base/resolve/main/config.json
```

So the blast radius is every job that runs the unit suite, including the one producing the coverage number the quality gate reads. Each occurrence costs a red PR and a manual rerun.

PR #1113, opened 2026-08-07
fix: drop a module's function_locations before it re-parses

### LINKED ISSUE #1019: fix: stale function_locations survive on a reused GraphUpdater and mis-key Rust method-body uses
## Summary

`function_locations` on the DefinitionProcessor is never cleared between runs of a reused GraphUpdater, and its key is `(module_qn, start_line, start_col)`. If a Rust function is renamed in place (same start position), a second run on the same updater reads the stale record and attributes by the dead qn.

Latent today: every CLI and MCP path constructs a fresh GraphUpdater per invocation, and the incremental tests build a new updater for the second run. But the codebase defends exactly this invariant elsewhere (`reset_rust_path_caches`, `reset_js_receiver_bindings` exist so a reused updater's second run stays correct), and the #1007 branch added a new consumer of the map (`finalise_rust_function_scope_uses` files method-body `use` imports under the located qn), so a future updater-reuse path would file imports under dead qns.

## Repro (reused updater, two runs)

```rust
// run 1: src/foo.rs
impl S {
    pub fn run(&self) -> u32 {
        use crate::alpha::helper;
        helper()
    }
}
// run 2: the method is renamed to `go` in place (same start position)
```

Second run on the SAME updater instance: `finalise_rust_function_scope_uses` looks up the method node's span, receives the stale `FunctionLocation(qualified_name="foo.S.run")`, and files the body use under `foo.S.run`; the live `foo.S.go` then resolves through the file map, emitting `foo.S.go CALLS beta.helper` instead of `alpha.helper`.

## Suggested direction

Purge (or overwrite-verify) a module's `function_locations` entries when that module re-parses, tracked per module the way `_rust_inline_scope_keys` and `_rust_fn_scope_keys` already are, so the cost stays proportional to the re-parsed file rather than the whole map.

Surfaced in round thirteen of the #1007 adversarial reviews (reproducer executes the two-run sequence directly against the processor).

vitali87:
Closes #1019.

Branches off `main` — independent of the other Rust PRs in flight.

## Change

`function_locations` is keyed by `(module_qn, start_line, start_col)` and was never cleared between runs of a reused GraphUpdater. A function renamed in place keeps its key, so the stale record survives, `_claim_function_span`'s first-claim guard blocks the live registration, and `finalise_rust_function_scope_uses` files the method's body `use` under the dead qn.

The issue suggested tracking the keys per module. I found **nine** write sites across the definition and class passes, so instead of wrapping each one — where the next new write site would silently miss the index — the map maintains the index itself: `FunctionLocations` (a `dict` subclass in `types_defs.py`) records `key[0]` on every `__setitem__` and exposes `drop_module(module_qn)`. `update` is overridden too, since `dict.update` bypasses `__setitem__` on a subclass.

`DefinitionProcessor.process_file` calls `drop_module` before the module's ingestion passes run. Cost stays proportional to the re-parsed file, as the issue asked.

## Tests

`codebase_rag/tests/test_function_locations_reuse.py` executes the two-run sequence against one updater instance:

- the renamed method resolves through its own body `use` (`alpha`), not the file-level one (`beta`)
- `function_locations` holds the live qn and no longer the dead one
- **control**: a module whose functions kept their names still resolves after a re-parse

The first two fail without the fix, reproducing the issue's predicted symptom exactly — `foo.S.go CALLS beta.helper` — and the map showing `foo.S.run` with no `foo.S.go` at all.

## Note on severity

Still latent in production: every CLI and MCP path builds a fresh updater. This pins the invariant the codebase already defends with `reset_rust_path_caches` and `reset_js_receiver_bindings`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale function mappings when reprocessing modules with reused parsing sessions.
  - Updated method renames to resolve to their current locations instead of outdated entries.
  - Preserved correct resolution for unchanged methods after module re-parsing.

- **Tests**
  - Added regression coverage for refreshed and reused parsing workflows.

### DIFF
--- a/codebase_rag/parsers/definition_processor.py
+++ b/codebase_rag/parsers/definition_processor.py
@@ -15,7 +15,7 @@ from ..types_defs import (
     CppDefinitionSpan,
     DeferredCppInherit,
     DeferredInherit,
-    FunctionLocation,
+    FunctionLocations,
     FunctionRegistryTrieProtocol,
     FunctionSpanKey,
     RustTraitImpl,
@@ -207,7 +207,7 @@ class DefinitionProcessor(
         # method node Pass 2 registered, so Pass-3 caller attribution reuses
         # the registered label/qn instead of re-deriving them structurally
         # (the walks diverge on preprocessor-distorted class bodies).
-        self.function_locations: dict[FunctionSpanKey, FunctionLocation] = {}
+        self.function_locations: FunctionLocations = FunctionLocations()
         # {rel path: [full line spans]} of every C/C++ function/method the
         # tree-sitter pass ingested; the hybrid C++ frontend's macro-use
         # CALLS resolve against these after Pass 2 (see CppDefinitionSpan).
@@ -369,6 +369,13 @@ class DefinitionProcessor(
                         cache_entry[key] = combined_captures[key]
                 if cache_entry:
                     self._func_class_captures_cache[file_path] = cache_entry
+
+            # A reused updater's second run re-parses this module into a map
+            # still holding the first run's spans. The key is (module_qn,
+            # line, col), so a function renamed in place keeps its key and the
+            # first-claim guard would block the live registration, filing body
+            # `use` imports under the dead qn (issue #1019).
+            self.function_locations.drop_module(module_qn)
 
             self.import_processor.parse_imports(
                 root_node,
--- a/codebase_rag/parsers/function_ingest.py
+++ b/codebase_rag/parsers/function_ingest.py
@@ -17,6 +17,7 @@ from ..types_defs import (
     DeferredCppInherit,
     DeferredParentLink,
     FunctionLocation,
+    FunctionLocations,
     FunctionRegistryTrieProtocol,
     FunctionSpanKey,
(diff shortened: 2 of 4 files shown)


vitali87:
Heads-up on the red SonarCloud check here: it is not caused by this PR. The `sonarcloud.yml` job runs the unit suite, which downloads the embedding model, and it hit HuggingFace 429s:

```
FAILED test_embedder.py::test_embed_code_integration - OSError: We couldn't connect to 'https://huggingface.co'
FAILED test_embedder.py::test_similar_code_has_similar_embeddings
FAILED test_semantic_search_eval.py::test_cgr_semantic_search_retrieves_expected_function
3 failed, 6567 passed
```

That is issue #1092, and the fix is #1118 — which I have now extended to cover `sonarcloud.yml` as well as `ci.yml` (it originally only patched the latter, which is why this job still downloads). **Merging #1118 first should clear this check**; this branch needs no change of its own.

vitali87:
@coderabbitai full review

@greptile-apps review

PR #1114, opened 2026-08-07
fix: name a Rust inline mod under class scope by one qn rule

### LINKED ISSUE #1018: fix: inline mod inside a trait or impl body gets inconsistent Module node and DEFINES qns
## Summary

An inline `mod` declared inside a trait or impl body (for example in a const default's initializer block, or in a trait default method) produces inconsistent Module bookkeeping: the Module node and the DEFINES edge disagree on the module's qn, so the graph audit reports both an orphan Module node and a dangling DEFINES edge.

## Repro

```rust
// src/foo.rs
pub trait T {
    const C: u32 = {
        mod inner {
            pub const fn g() -> u32 { 1 }
        }
        inner::g()
    };
}
```

Indexing this file and running `graph_audit.collect_violations` over the recorded batches yields:

```
Module 'foo.T.inner' has no relationships
(Module 'foo.T')-[:DEFINES]->(Module 'foo.T.inner') references a nonexistent node and would be dropped by the database
```

The same class of failure appears for a mod inside a trait default method (round twelve of the #1007 reviews): the inline-mod Module node is named by mod segments only (`foo.inner`) while its functions register with the class segment (`foo.T.inner.g`), so the DEFINES chain dangles.

## Analysis

The module ingestion pass and the function fqn walk use different scope rules for inline mods under class scopes (mods-only versus class-segment-keeping), so the node, its DEFINES edge, and the contained functions end up under divergent qns. Reproduces identically on main and on the #1007 branch (verified by execution in rounds twelve and thirteen of the #1007 adversarial reviews), so it is independent of the crate-path work.

Related: #1017 (module qn collisions) shares the underlying theme that Rust module scopes below non-mod items lack a coherent single qn scheme.

vitali87:
Closes #1018.

> **Stacked on #1113 (issue #1019)** only because it was branched after it; the two changes are independent. Merge #1113 first and this retargets cleanly.

## What was diverging

Three different scope rules named the same thing. For `impl S { fn run() { mod inner { fn g() {} } } }`:

| | qn produced |
|---|---|
| Module node (`build_nested_qualified_name_for_class`) | `foo.inner` — mods + types, impl target dropped |
| functions inside it (FQN specs walk) | `foo.S.inner.g` — impl target kept |
| the containment DEFINES source (`build_module_path` default) | `foo.inner` — mods only |

The audit reported exactly what the issue described:

```
Module 'foo.T.inner' has no relationships
(Module 'foo.T')-[:DEFINES]->(Module 'foo.T.inner') references a nonexistent node
(Module 'foo.inner')-[:DEFINES]->(Function 'foo.T.inner.g') references a nonexistent node
```

## Change

1. **One rule.** New `_rust_inline_module_qn` names the inline module with impl targets *and* type segments included — the rule its own functions already use. The containment walk in `function_ingest` uses the same rule, so the DEFINES source names the module that exists.
2. **The parent is not always a Module.** `mod` in a trait const initializer or an impl method body hangs off the TYPE, whose node carries a Class/Interface/Enum label. The lexical-parent edge now resolves the parent's registered label via the function registry (the pattern already used for Dart `implements` and Java anonymous overrides) instead of hard-coding `Module`.
3. `RELATIONSHIP_SCHEMAS` documents Interface/Enum/Type/Union as DEFINES sources, which the audit then accepts.

The containment walk keeps using the **mods-only** walk to decide *whether* an inline mod encloses the item at all, and only then spells it with the full rule. Without that guard the branch fires for every method in an impl and 15 Rust tests fail on a DEFINES from a Module named after the type — caught and fixed before commit.

## Tests

`codebase_rag/tests/test_rust_inline_mod_under_class_scope.py` covers all three shapes from the issue and its linked review rounds — trait const initializer, trait default method, impl method — asserting a single `.inner` Module qn that the contained functions actually nest under. `create_and_run_updater` runs the graph audit, so each test also fails on any orphan or dangling edge.

All three fail without the fix. Rust suite: 873 passed.

### DIFF
--- a/codebase_rag/parsers/class_ingest/mixin.py
+++ b/codebase_rag/parsers/class_ingest/mixin.py
@@ -1661,10 +1661,9 @@ class ClassIngestMixin:
                 continue
 
             module_name = safe_decode_text(module_name_node)
-            nested_qn = id_.build_nested_qualified_name_for_class(
+            inline_module_qn = self._rust_inline_module_qn(
                 module_node, module_qn, module_name or "", lang_config
             )
-            inline_module_qn = nested_qn or f"{module_qn}.{module_name}"
 
             module_props: PropertyDict = {
                 cs.KEY_QUALIFIED_NAME: inline_module_qn,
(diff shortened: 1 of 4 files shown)


vitali87:
@coderabbitai full review

@greptile-apps review

PR #1155, opened 2026-08-10
fix: key File nodes on the link path so a deleted symlink is prunable (#1154)

### LINKED ISSUE #1154: Realtime & batch File deletion lose symlink identity (resolve() can't follow a deleted link)
Split out from Greptile's review of #1153.

## Problem
`File.absolute_path` is stored at ingest time via `cached_resolve_posix()` = `Path.resolve().as_posix()`, which **follows symlinks** — so a symlinked file's node is keyed on the *target's* canonical path.

At deletion time the link is already removed from disk, so `path.resolve()` can no longer follow it and returns the symlink's own pathname. `CYPHER_DELETE_FILE` (`MATCH (f:File {absolute_path: $path})`) then receives an identity that never matches the indexed node, leaving a stale `File` node behind.

## Affected code paths (both share the limitation)
- `realtime_updater.py` `_process_change_locked` — the realtime delete (this is what #1153 changes to `path.resolve().as_posix()`).
- `codebase_rag/graph_updater.py:1860` — the batch delete, already `deleted_path.resolve().as_posix()` (issue #897).

## Scope note
This is **pre-existing** and **not a regression from #1153** — #1153 correctly fixes the common regular-file case and brings realtime into parity with the batch path. The symlink case needs a different identity strategy (e.g. resolving against a parent that still exists, or matching on `path`/`name` for links) and is tracked here separately.

## Repro sketch
1. Add a symlink `link.py -> ../real/target.py` inside a watched repo; ingest.
2. Delete `link.py`.
3. The `File {absolute_path: <target canonical>}` node remains in the graph.

vitali87:
Closes #1154.

## Problem
`File.absolute_path` was computed at ingest via `cached_resolve_posix()` = `Path.resolve().as_posix()`, which **dereferences a leaf symlink to its target**. Two consequences for a symlinked file:

1. **Deletion leaks the node.** Once the link is removed from disk, `resolve()` can no longer follow it, so the delete key (`CYPHER_DELETE_FILE` matches `absolute_path`) never matches the stored target path — the `File` node is left behind (both the realtime path and `graph_updater.py:1860`).
2. **Prune skips it too.** The batch prune only considers nodes whose `absolute_path` is inside `repo_path`. A link pointing outside the repo was stored under the *external* target path, so it failed the containment gate and was never swept.

## Fix
Add `cached_file_identity_posix()` — resolves only the **parent** (which outlives the leaf) and keeps the leaf name, so ingest and deletion agree on the key, and a link stays inside the repo it lives in (consistent with the GHSA-85gg containment stance). Applied to the sole `File`-node creator (`structure_processor.process_generic_file`, node + `CONTAINS_FILE` edge) and to the `contract_linking` `EXPOSES` edge / index lookup that key on the same `File` identity.

**Zero migration risk for regular files:** when the leaf is not a symlink, `parent.resolve()/name` == `resolve()`, so identities are byte-identical. Only symlink leaves change — from broken to correct.

## Scope
- Module/Function/Class nodes are unaffected: they are deleted by relative path + project scope (`CYPHER_DELETE_MODULE`), which is already symlink-safe.
- Does **not** touch `realtime_updater.py` — that delete line is owned by the in-flight #1153. This change is complementary: #1153 fixes relative→absolute at delete time; this fixes the ingest identity so the two agree for symlinks.

## Test
`test_symlinked_file_keyed_on_link_not_target` (RED-verified): ingests a symlinked file, asserts the `File` node is keyed on the link’s own in-repo path (not the target), and that `link.resolve()` after `unlink()` recovers exactly that key — proving ingest/delete agreement. Reverting the one-line ingest change fails the test.

## Summary by CodeRabbit

- **Bug Fixes**
  - File indexing now preserves the original symlink path instead of replacing it with the target file’s path.
  - Stored file paths remain stable and continue to identify the link correctly even after the linked file is removed.
  - Contract and source references now consistently retain the indexed file’s original path identity.
- **Tests**
  - Added coverage to verify symlink path preservation and stability.


vitali87:
@greptileai Thanks — the mechanism is real, but I think "not merge-safe" over-scopes it. Breaking it down:

**Full rebuild (`--clean`, the standard upgrade path):** `clean_database()` wipes the graph before re-ingest (cli.py:300), so no legacy record can survive. Fully safe.

**Incremental (`--update-graph`) over a legacy graph:**
- *In-root* symlinks self-heal: the real file MERGEs its own key, the link gets its new `/repo/link` key, both paths exist on disk so prune keeps both. No stale/duplicate node.
- The genuine residual is an *external* symlink recorded in a *pre-GHSA-85gg* graph — its stored `absolute_path` is outside the repo, so the prune containment gate skips it. But post-0.0.589 those are skipped at ingest and never created, and the old code could neither delete nor prune them either.

So this change is **strictly better than the status quo, not a regression** — it fixes the common case and makes new symlink nodes prunable. The legacy-migration cleanup for that narrow incremental-over-old-graph case is a separate concern, tracked in #1156 with a concrete approach.

Please re-review with the `--clean` vs `--update-graph` distinction in mind; I don't think a strictly-improving correctness fix should be gated on retrofitting a migration for pre-existing legacy data.

Issue #897, opened 2026-07-22
fix: Folder/File nodes keyed by relative path merge across projects, delete-project wipes sibling subtrees


vitali87:
## Summary

Folder and File nodes use the bare relative path as their unique key (\`_NODE_LABEL_UNIQUE_KEYS\` maps \`FOLDER\` and \`FILE\` to \`UniqueKeyType.PATH\`). In the shared graph, two projects that contain the same relative path (for example both have an \`app/\` directory) therefore merge onto a single Folder node. Every other label is project scoped (Project name or project prefixed qualified name); these two are not.

## Impact: cross-project data loss

\`CYPHER_DELETE_PROJECT\` walks containment from the Project node:

```cypher
MATCH (p:Project {name: $project_name})
OPTIONAL MATCH (p)-[:CONTAINS_PACKAGE|CONTAINS_FOLDER|CONTAINS_FILE|CONTAINS_MODULE*]->(container)
OPTIONAL MATCH (container)-[:DEFINES|DEFINES_METHOD*]->(defined)
DETACH DELETE p, container, defined
```

When the walk reaches a shared Folder node it continues into the sibling project's subtree and detach deletes it. Reproduced live: \`cgr delete-project\` on a \`py-client\` fixture (which has \`app/client.py\`) silently deleted every Module and Function of an unrelated \`catalog-service\` project (which has \`app/main.py\`). The sibling Project node survives but its subtree is empty, so the project looks indexed while holding nothing.

The incremental orphan prune has the same exposure: \`CYPHER_DELETE_FILE\` and \`CYPHER_DELETE_FOLDER\` delete by \`{path: $path}\` with no project bound (the Module branch of the prune already special cases exactly this hazard, with the comment "a sibling project's module can share the relative path").

## Repro

1. Two fixture projects, each with an \`app/\` directory containing one source file.
2. Index both into the shared graph.
3. \`MATCH (o)-[:CONTAINS_FOLDER]->(n {path: 'app'}) RETURN o.name, id(n)\` shows both Projects containing the SAME node id.
4. \`cgr delete-project --name <project-a>\`.
5. Project B's functions are gone: \`MATCH (f:Function) WHERE f.qualified_name STARTS WITH '<project-b>' RETURN count(f)\` returns 0.

## Proposed fix

Key Folder and File nodes by \`absolute_path\` (both labels already store it):

[...]

PR #1158, opened 2026-08-10
fix(graph): prune legacy target-resolved File identities

### LINKED ISSUE #1156: Incremental --update-graph does not reconcile legacy target-resolved File identities
Follow-up from Greptile's review of #1155.

## Context
#1155 keys `File` nodes on the link path (leaf not dereferenced) so a deleted symlink is prunable. That is correct for all newly written nodes.

## Residual gap (narrow)
On an **incremental** `--update-graph` run over a graph originally built by an **older** version, a legacy `File` node that was keyed on a symlink's *dereferenced target* may persist:
- In-root symlinks self-heal: the real file MERGEs its own key, the link gets its new key, both paths exist so prune keeps both. No stale data.
- The genuine residual is an **external** symlink recorded in a **pre-GHSA-85gg** graph: its stored `absolute_path` is outside the repo, so the prune containment gate skips it and it is never swept.

## Why this does not block #1155
- The standard upgrade path is `--clean` (full rebuild via `clean_database()`, cli.py:300), which wipes the graph — no legacy data survives.
- Post-0.0.589, external symlinks are skipped at ingest, so new graphs never create these nodes.
- The old code could not delete or prune these nodes either, so #1155 is strictly better, not a regression.

## Possible fix
During `_prune_orphan_nodes`, also drop a `File` node whose stored `absolute_path` disagrees with the identity derivable from its relative `path` (`cached_file_identity_posix(repo_path / path)`), i.e. a legacy target-resolved record. Needs careful handling of symlinked-ancestor edge cases and its own RED-verified test.

prondubuisi:
## Summary

- Incremental `--update-graph` now sweeps pre-GHSA-85gg `File` nodes that were keyed on a leaf-dereferenced **external** symlink target. Those nodes sit outside the repo, so the prune containment gate skipped them and they leaked forever.
- A still-present external symlink is recognized when `cached_resolve_posix(repo_path / path)` equals the stored `absolute_path` and that path disagrees with `cached_file_identity_posix(...)`.
- A deleted external symlink is swept only when this project owns the node (`CONTAINS_PACKAGE|CONTAINS_FOLDER|CONTAINS_FILE*` from `Project`). Sibling-project files that share a relative path stay untouched.
- In-repo symlink records that share the real file's key are left alone (the self-heal case from #1155). Symlinked *ancestors* are compared via parent-resolve identity so they do not false-positive.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Fixes #1156

## Test Plan

- [x] Unit tests pass (`uv run pytest codebase_rag/tests/test_graph_updater_pruning.py` — 21 passed, including the 16 existing prune tests)
- [x] New tests added (`TestPruneLegacyTargetResolvedFileIdentity`)
- [ ] Integration tests pass (`make test-integration`, requires Docker — not run locally; unit tests cover the prune decision table)

RED verification (tests against unpatched `_prune_orphan_nodes`):

- `test_prune_removes_live_legacy_external_symlink_file` failed (`0` deletes)
- `test_prune_removes_deleted_legacy_external_symlink_when_project_owned` failed (`0` deletes)
- keep-safe tests still passed on the unpatched code

GREEN after the fix: 21/21 in `test_graph_updater_pruning.py`. Also 48 passed across prune + incremental updater tests.

Keep-safe coverage:

- unowned outside `File` with a missing relative path (sibling project) is not deleted
- in-repo symlink sharing the target's key is not deleted
- file under a symlinked ancestor is not deleted (identity uses parent-resolve, same as ingest)

Pre-commit on the changed files: ruff, ruff format, ty, README generation, bandit.

## Checklist

- [x] All pre-commit checks pass (`pre-commit run --files` on the three changed files)
- [x] No new comments or docstrings except the issue-#1156 invariant next to the existing #897 containment comments

### AI assistance

AI assistance was used to locate the prune/identity code, implement the focused fix, and write the RED-verified regression tests. The final diff was reviewed against `CONTRIBUTING.md` (constants, no `Any`, conventional commit, tests required) and run through the pre-commit hooks listed above.

## Why this is safe

`File` nodes are unique on `absolute_path`. Deleting an **outside** key cannot collapse an in-repo file. Deleting an in-repo identity mismatch *would* be unsafe (it is the real file's MERGE key), so that case is explicitly not pruned. Ownership is consulted only for missing-on-disk outside keys, so `test_prune_skips_other_projects` still holds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale file references, including symlinked and externally located files.
  * Preserved valid in-repository symlinks and files under symlinked directories.
  * Avoided removing external files unless they are confirmed to belong to the project.
  * Added more accurate project file path tracking for graph-based operations.
  * Improved handling of unresolved file locations during project cleanup to reduce unintended deletions.

### DIFF
--- a/codebase_rag/constants/graph.py
+++ b/codebase_rag/constants/graph.py
@@ -29,6 +29,7 @@ KEY_RESPONSE = "response"
 KEY_START_LINE = "start_line"
 KEY_END_LINE = "end_line"
 KEY_PATH = "path"
+KEY_PATHS = "paths"
 KEY_ABSOLUTE_PATH = "absolute_path"
 # Whether flow analysis covered a Module: its language is in the source/sink
 # registry AND the FLOWS_TO capture group was enabled at indexing. Read by
@@ -391,6 +392,16 @@ CYPHER_COUNT_PROJECT_MODULES = (
 # Queries for orphan pruning: return all paths stored in the graph
 CYPHER_ALL_FILE_PATHS = (
     "MATCH (f:File) RETURN f.path AS path, f.absolute_path AS absolute_path"
+)
+# Pre-GHSA-85gg File nodes keyed on a leaf-dereferenced external target
+# fail the prune containment gate. Ownership via this project's CONTAINS_*
+# path lets a deleted symlink's leftover node be swept without touching a
+# sibling project's file that happens to share the relative path (#1156).
+CYPHER_PROJECT_OWNED_FILE_ABSOLUTE_PATHS = (
+    "MATCH (p:Project {name: $project_name})"
+    "-[:CONTAINS_PACKAGE|CONTAINS_FOLDER|CONTAINS_FILE*]->(f:File) "
+    "WHERE f.absolute_path IN $paths "
+    "RETURN DISTINCT f.absolute_path AS absolute_path"
 )
 CYPHER_ALL_MODULE_PATHS_INTERNAL = (
     "MATCH (m:Module) RETURN m.path AS path, m.qualified_name AS qualified_name"
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -80,7 +80,9 @@ from .types_defs import (
 from .utils.dependencies import has_semantic_dependencies
 from .utils.fqn_resolver import find_function_source_by_fqn
 from .utils.path_utils import (
+    cached_file_identity_posix,
     cached_relative_path,
+    cached_resolve_posix,
     should_keep_dir,
     should_skip_path,
     should_skip_rel_file,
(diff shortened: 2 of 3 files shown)


vitali87:
Thanks for picking up #1156 so quickly — the batched ownership verification (`_owned_file_absolute_paths` via `CONTAINS_*`) is the right direction and addresses CodeRabbit's Major (a node only a *sibling* reaches is now excluded).

Maintainer note on the residual **Greptile P1** (shared File identity), so you can decide how far to take it:

`NodeLabel.FILE` has a **global** unique constraint on `absolute_path` (`NODE_UNIQUE_CONSTRAINTS`, one constraint per label — not per-project). So if two projects in one shared graph both symlink to the *same* external target, they MERGE onto a **single** File node that both containment chains reach. `_owned_file_absolute_paths` returns that node for *either* project (both "own" it), so this project's prune still deletes the node the sibling depends on — the ownership check confirms *an* owner, not the *sole* owner.

Two ways to close it, your call:
1. **Sole-ownership guard** — only sweep when this project is the *only* Project reaching that `absolute_path` (e.g. count distinct `Project` ancestors and require exactly one). Fully correct, a bit more query work.
2. **Surgical delete** — instead of `CYPHER_DELETE_FILE` by bare `absolute_path`, detach only *this* project's `CONTAINS_FILE` edge and delete the node only if it becomes an orphan. Also correct, more invasive.
3. **Document as accepted limitation** — the trigger (two projects, same external target, both pre-GHSA-85gg, incremental update) is astronomically narrow, so a comment scoping it may be defensible.

Given how narrow this is, I lean 1 or 3. Whichever you choose, please add a regression test with two projects sharing one external-target File and assert the sibling's node survives. Nice work — this is close.

PR #1237, opened 2026-08-13
feat: build orchestrator prompt sections conditionally on registered tools

### LINKED ISSUE #1201: Build orchestrator prompt sections conditionally on registered tools (semantic_search-less MCP config)
Follow-up to #1199 / PR #1200, prompted by review feedback there.

## Gap

#1200 makes the orchestrator system prompt reference the *registered* short tool names, closing the rename drift. But the prompt text itself is static: when a canonical tool is not registered, the prompt still instructs the model to use it (with a runtime warning logged). One legitimate configuration hits this today: the MCP path (`codebase_rag/mcp/tools.py`) omits `semantic_search` when the vector backend is unavailable, yet the built prompt still contains the whole "WHEN TO USE SEMANTIC SEARCH FIRST" strategy section (`prompts.py`, the `t.semantic_search` sites around lines 156-183) telling the model to start with a tool that is not in its schema.

For prose-following models this is the same failure mode #1199 documented — a call to an unregistered name is dropped silently and the run burns retries on empty turns — just narrower: it needs a config where the tool is genuinely absent, rather than affecting every install.

## Work

Make the prompt sections conditional on tool availability rather than interpolation-only:

- `extract_tool_names` (or a sibling) should expose which canonical tools are actually registered, not just their names — e.g. return `(ToolNames, frozenset[AgenticToolName])` or make the availability set a field.
- `build_rag_orchestrator_prompt` builds the semantic-search strategy subsection (2a/2c workflows) only when `semantic_search` is registered; the graph-first guidance in 2b becomes the unconditional default in its absence.
- Same treatment for any other tool that can legitimately be absent (`structural_search`/`structural_replace` are also conditional in the MCP path, though they are not in `ToolNames` today).
- The runtime warning from #1200 stays — it then indicates a genuinely unexpected registration gap rather than a known configuration.

## Tests

Extend `test_prompt_tool_names.py`: build the prompt from a tool list without `semantic_search` and assert the prompt neither backtick-references it nor instructs starting with semantic search; with the full list, assert the strategy section is present (guards against the section silently disappearing).

alooshxl:
## Summary

- `build_rag_orchestrator_prompt` now builds its strategy sections conditionally on which canonical tools are registered: when `semantic_search` is absent (the MCP path without a vector backend), the SEMANTIC FIRST strategy section, the semantic investigation-cycle step, and the semantic token-management bullet are replaced by graph-first guidance, so the model is never instructed to start with a tool that is not in its schema.
- New `extract_registered_tools` sibling to `extract_tool_names` exposes the registered canonical-tool set. The #1200 drift warning stays but now fires only for tools the prompt references unconditionally, so it indicates a genuinely unexpected registration gap rather than a known configuration.
- With the full toolset the built prompt is byte-identical to the previous output (verified by diffing the old and new builder output), so existing configurations see no change.

## Type of Change

- [x] New feature

## Related Issues

Closes #1201

## Test Plan

- [x] New tests added

Extended `codebase_rag/tests/test_prompt_tool_names.py` in both directions per the issue: built from a tool list without `semantic_search`, the prompt neither backtick-references it nor instructs starting with semantic search (asserts `"semantic" not in prompt.lower()`), and graph-first guidance is present; built from the full list, the SEMANTIC FIRST / HYBRID APPROACH strategy section is still there, guarding against the section silently disappearing. Also covers `extract_registered_tools` and the warning split (missing `semantic_search` no longer warns; missing required tools still do). The new tests were RED-verified against the previous implementation.

`structural_search`/`structural_replace` are also conditionally registered in the MCP path but are not referenced anywhere in the orchestrator prompt today, so there is no section to gate for them; `extract_registered_tools` reports the full registered set, so gating them later needs no further API change.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved search guidance based on which search tools are available.
  * Added support for semantic-search and graph-search workflows, including candidate discovery and token-management instructions.
  * Improved recognition of project entry points during investigations.

* **Bug Fixes**
  * Added warnings and safer handling when required search tools are unavailable.

* **Tests**
  * Expanded coverage for tool detection, conditional guidance, and missing-tool scenarios.

### DIFF
--- a/codebase_rag/prompts.py
+++ b/codebase_rag/prompts.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING
 
 from loguru import logger
@@ -26,11 +28,19 @@ if TYPE_CHECKING:
     from pydantic_ai import Tool
 
 
-def extract_tool_names(tools: list["Tool"]) -> ToolNames:
+PROMPT_OPTIONAL_TOOLS = frozenset({AgenticToolName.SEMANTIC_SEARCH})
+
+
+def extract_registered_tools(tools: list[Tool]) -> frozenset[AgenticToolName]:
+    registered = {t.name for t in tools}
+    return frozenset(name for name in AgenticToolName if name in registered)
+
+
+def extract_tool_names(tools: list[Tool]) -> ToolNames:
     registered = {t.name for t in tools}
 
     def resolve_tool_name(canonical: AgenticToolName) -> str:
-        if canonical not in registered:
+        if canonical not in registered and canonical not in PROMPT_OPTIONAL_TOOLS:
             logger.warning(
                 f"Tool '{canonical}' is not registered on the agent; "
                 "the orchestrator prompt references it anyway"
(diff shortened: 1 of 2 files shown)


vitali87 (changes requested):
address bots comments please

Issue #1245, opened 2026-08-13
Dynamic capture foundation: edge provenance schema, trace interchange format, ingestion pipeline


vitali87:
Part of #1244. This must land before any per-language tracer.

## Scope

The language-agnostic half of dynamic call-graph capture: the graph schema for runtime-observed edges, the trace interchange format every tracer emits, and the ingestion pipeline that resolves trace records to existing graph nodes.

## Graph schema

- Extend `CALLS` edges (or add a parallel relationship) with provenance properties:
  - `source`: `static`, `dynamic`, or both (a dynamic observation on an existing static edge sets both, it does not duplicate the edge)
  - `call_count`: total observed invocations
  - `workloads`: which test/entrypoint produced the observation
  - `observed_receiver_types` where the runtime exposes them
- Decide how staleness is represented: dynamic edges are pinned to the commit they were traced at; after edits they must be re-traced or marked possibly stale.

## Trace interchange format

A single JSONL (or similar) format that every language tracer emits, so ingestion is written once. Each record needs at minimum: caller `(path, qualified name, line)`, callee `(path, qualified name, line)`, count, workload id, and optional receiver-type info. Versioned, documented, and covered by schema validation.

## Resolution layer

Mapping runtime frames to graph node IDs is the hard part and has a long tail:

- decorators and wrappers reporting the wrapper's location
- lambdas, closures, comprehensions, and other synthetic names
- generated code with no source on disk
- native/extension frames with no corresponding node

Unresolvable records must be counted and reported, not silently dropped, so per-language resolution quality is measurable.

## Ingestion pipeline

- CLI entry point (e.g. `cgr trace ingest <file>`) that reads the interchange format, resolves records, and writes/updates edges in Memgraph
- Idempotent re-ingestion for the same commit and workload
- Summary output: records read, edges created, edges confirming static, edges static missed, unresolved

## Acceptance criteria

- [ ] Edge schema documented in `docs/`
- [ ] Interchange format documented and validated
- [ ] Ingestion CLI with idempotent behavior and resolution-quality reporting
- [ ] Query surface: retrieval can distinguish and filter by edge provenance

vitali87:
Delivered by #1258 (Python pilot carrying the language-agnostic core) and proven language-agnostic by the nine tracers that landed on top of it (#1268, #1261–#1267) with zero changes to the ingestion contract.

Against the acceptance criteria:
[...]

Issue #1244, opened 2026-08-13
Epic: dynamic runtime call-graph capture (hybrid static/dynamic analysis)


vitali87:
## Motivation

Static, AST-based analysis is complete but imprecise: it cannot resolve dynamic dispatch, reflection, `getattr`-style access, dependency injection, monkey-patching, plugin registries, or callbacks routed through frameworks. Dynamic analysis has the opposite profile: every edge observed at runtime is ground truth (that call really happened, with that receiver type), but you only see the paths the workload exercised.

This epic adds a dynamic capture tier: run the target codebase's test suite (or another instrumented workload) under a language-appropriate tracer, capture actual caller/callee relationships, resolve them to existing graph nodes, and merge them into the graph alongside the static edges.

## Design principles

- **Never discard either source.** Edges carry provenance: `STATIC`, `DYNAMIC`, or both. A static edge confirmed at runtime and a runtime-only edge that static analysis missed are both high-value signals.
- **Coverage honesty.** Absence of a dynamic edge must never imply dead code; the dynamic graph only reflects the traced workload.
- **Scoped execution.** We trace instrumented test runs in a controlled environment. We do not promise to safely run arbitrary codebases; environment setup, sandboxing, and side effects are the integrator's responsibility, documented per language.
- **Rich edge metadata.** Call counts, observed receiver/argument types where the runtime exposes them, and workload provenance (which test or entrypoint produced the observation).

## Structure

The shared schema and ingestion pipeline land first, then per-runtime tracers. Python is the pilot: it is our implementation language and the worst offender for static-analysis blind spots.

### Foundation
- [x] #1245

### Per-runtime tracers
- [x] #1246 Python (`sys.monitoring`, PEP 669) — pilot
- [x] #1247 JavaScript / TypeScript / TSX (Node.js, V8)
- [x] #1248 Java + Scala (JVM agent)
- [x] #1249 C# (.NET EventPipe / profiler API)
- [x] #1250 Go
- [x] #1251 Rust
- [x] #1252 C / C++
- [x] #1253 PHP (Xdebug)
- [x] #1254 Lua (`debug.sethook`)
- [x] #1255 Dart (VM Service)

## Rollout order

1. Foundation (schema, resolution layer, ingestion)
2. Python end to end, proving the schema and the frame-to-node resolution approach
[...]

PR #1260, opened 2026-08-14
feat: add JVM runtime call tracing agent with dynamic CALLS edge provenance

### LINKED ISSUE #1248: Dynamic call-graph tracer for Java and Scala (JVM agent)

Covers Java and, by extension, Scala: both compile to JVM bytecode, so one agent serves both. Scala adds name-mangling on top (objects, traits, default parameters), handled in the resolution layer.

## Approach

A `java.lang.instrument` agent (likely built on ByteBuddy or ASM) that instruments method entry to record caller/callee pairs, aggregated in memory and flushed to the interchange format on JVM exit. This is the highest-payoff runtime for dynamic capture: interface dispatch, dependency injection frameworks (Spring, Guice), reflection, and proxies are pervasive and mostly opaque to static analysis.

Alternatives to evaluate: JFR `MethodTrace`/`MethodTiming` events (low overhead, JDK 25+), async-profiler wall-clock mode (sampled, approximate edges only).

## Scope

- Agent attachable via `-javaagent:` with include/exclude package filters (instrumenting everything including the JDK is not viable)
- Wrapper for Maven Surefire / Gradle test runs so the standard workflow is one flag or plugin
- Capture the concrete receiver class on virtual/interface calls: this is the core payoff (which implementation actually handled the call)
- Resolution handling: lambdas (`lambda$` synthetic methods), inner/anonymous classes, generated proxies (JDK dynamic proxies, CGLIB), Scala name mangling
- Emit the shared interchange format from #1245

## Acceptance criteria

- [ ] Traced test run of a sample Java project produces dynamic edges with concrete receiver types
- [ ] Demonstrated runtime-only edge through an interface or DI container that static analysis could not resolve
- [ ] Scala sample resolves through name mangling
- [ ] Overhead measured and documented with recommended filters

vitali87:
Closes #1248. Part of #1244. **Stacked on #1258** (base branch is `feat/1246-python-dynamic-tracing`); only the last commit is new.

## What

A dynamic call-graph tracer for Java and Scala: a `java.lang.instrument` agent records real caller/callee pairs at runtime and emits the shared trace interchange format from #1245, which `cgr trace ingest` resolves onto the static graph with the same provenance properties the Python pilot introduced.

### The agent (`codebase_rag/trace/jvm_agent/`, ~350 lines of Java, zero dependencies)

- Built with the `java.lang.classfile` API (JDK 24+), so there is no ByteBuddy/ASM/Maven anywhere: `make jvm-agent` needs only `javac` and `jar`.
- Injects a recorder call at method entry for classes under `include=` package prefixes; the caller is recovered with `StackWalker`, walking through JDK internals, lambda-metafactory hidden classes, and generated proxies to the nearest project frame. An edge like `list.forEach(this::handle)` or a call through a DI proxy is attributed to the code that initiated it, which is exactly what static analysis cannot see.
- Samples the concrete receiver class on instance methods (bounded per pair), so interface dispatch records which implementation actually ran.
- Aggregates in memory (`ConcurrentHashMap` + `LongAdder`-style counters) and writes the JSONL trace on JVM exit. Workload labels via agent arg or `TraceRecorder.setWorkload(...)`.

### Resolution (`JvmFrameResolver`)

JVM runtime names and the graph's path-derived qualified names disagree in every interesting way; the resolver bridges: package-derived paths join node paths by suffix (handles Maven/Gradle source roots), `Outer$Inner` becomes dotted nesting, `<init>` maps to the class-named constructor node, raw-source parameter signatures (which JVM descriptors cannot reproduce) are stripped for matching with overloads disambiguated by line span, lambda bodies and anonymous-class methods resolve to their enclosing method's node by innermost span, and `<clinit>`/anonymous constructors are counted as synthetic rather than guessed. Scala's `Util$`/`$anonfun$` mangling is normalized.

`ingest_trace` now dispatches per the trace header's `language`; the Python path is untouched.

## Verification

Live end-to-end against Memgraph: sample Maven-layout Java project indexed with `cgr start --update-graph`, traced with the agent, ingested:

- Interface dispatch resolved to concrete implementations with receiver types (`describe(Animal) -> Dog.sound()` / `Cat.sound()`, both `static_missed` since the static tier cannot resolve the dispatch)
- Registry dispatch through a `HashMap` + method reference captured as a runtime-only edge (`handle(String) -> greet()`, `static_missed: true`)
- Anonymous-class `run()` resolved to its span-threaded `Function` node; its `<init>` correctly dropped as synthetic instead of fabricating a self-edge
- Implicit constructors counted `no_match` rather than misattributed

Overhead measured and documented: ~1.2µs per instrumented call (6M-call JIT-friendly loop: 3ms untraced, 7s traced), fine for test suites, with `include=` scoping recommended.

Scala: name-mangling normalization is unit-tested; live Scala validation is deferred until Scala static parsing leaves development status (its label/qn tests are currently skipped in-repo).

18 new unit tests (JVM resolution, ingest dispatch, anonymous/lambda/constructor edge cases), all red/green. Agent sources live under `codebase_rag/` per the Go/Roslyn frontend precedent and are Sonar-analyzed with coverage waived like the other non-Python tool sources.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help all install dev test test-parallel test-integration test-all test-parallel-all clean python build-grammars watch readme lint format typecheck check pre-commit release
+.PHONY: help all install dev test test-parallel test-integration test-all test-parallel-all clean python build-grammars watch readme lint format typecheck check pre-commit release jvm-agent
 
 PYTHON := uv run
 
@@ -79,6 +79,13 @@ check: lint typecheck test ## Run all checks: lint, typecheck, test
 
 release: ## Build, verify, and publish the current pyproject version to PyPI, then tag and create a GitHub Release
 	./scripts/release.sh
+
+jvm-agent: ## Build the JVM runtime tracing agent (requires JDK 24+)
+	rm -rf build/jvm-agent
+	mkdir -p build/jvm-agent
+	javac --release 24 -d build/jvm-agent codebase_rag/trace/jvm_agent/src/cgr/trace/*.java
+	jar cfm build/cgr-jvm-agent.jar codebase_rag/trace/jvm_agent/MANIFEST.MF -C build/jvm-agent .
+	@echo "Agent built: build/cgr-jvm-agent.jar"
 
 pre-commit: ## Run all pre-commit checks locally (comprehensive test before commit)
 	@echo "Running pre-commit checks..."
--- a/codebase_rag/constants/trace.py
+++ b/codebase_rag/constants/trace.py
@@ -22,13 +22,22 @@ TRACE_KEY_WORKLOADS = "workloads"
 TRACE_KEY_RECEIVER_TYPES = "receiver_types"
 
 TRACE_LANGUAGE_PYTHON = "python"
+TRACE_LANGUAGE_JVM = "jvm"
 TRACE_TOOL_NAME = "cgr-trace"
+TRACE_TOOL_NAME_JVM = "cgr-trace-jvm"
 TRACE_DEFAULT_OUTPUT = "cgr-trace.jsonl"
 
 # Python runtime qualname markers.
 TRACE_QUALNAME_LOCALS = "<locals>"
 TRACE_QUALNAME_MODULE = "<module>"
 TRACE_SYNTHETIC_PREFIX = "<"
+
+# JVM runtime name markers.
+TRACE_JVM_CONSTRUCTOR = "<init>"
+TRACE_JVM_STATIC_INITIALIZER = "<clinit>"
+TRACE_JVM_LAMBDA_PREFIX = "lambda$"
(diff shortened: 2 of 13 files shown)


vitali87:
Both findings fixed in the follow-up commit:

- **Package prefix overreach**: both `included` checks (transformer on internal names, recorder on binary names) now match only at name boundaries: exact match, `prefix.` / `prefix/` package members, or `prefix$` nested classes. Verified live: running the sample with the truncated prefix `com.example.zo` now instruments nothing (0 records), while `com.example.zoo` still traces fully.
- **Workload state crosses threads**: `currentWorkload` replaced with an `InheritableThreadLocal` seeded from the agent arg on the premain thread. `setWorkload` now labels the calling thread and threads it spawns afterwards, so concurrent runners cannot clobber each other's provenance. Verified live: run-level labelling still lands on every record.

Issue #1255, opened 2026-08-13
Dynamic call-graph tracer for Dart (VM Service)


vitali87:
Part of #1244. Depends on #1245 and the Python pilot (#1246) validating the pipeline.

## Approach

Dart's VM Service protocol is the sanctioned instrumentation surface. Candidates:

1. **VM Service CPU sampling (`getCpuSamples`)**: samples carry full stacks, so caller/callee pairs are recoverable, but edges are approximate (sampled); cheap to build since `dart run --observe` and `package:vm_service` do the heavy lifting
2. **Coverage service (`package:coverage`)**: executed-region data, not edges; supplementary only
3. **Kernel-level instrumentation**: transform the kernel AST (`.dill`) to inject entry hooks before running; complete edges, more engineering

Start with route 1 to get sampled edges flowing through the pipeline, and evaluate whether route 3 is warranted by resolution quality. Sampled edges must be flagged as approximate in the interchange format (a field #1245 should accommodate).

The dynamic payoff in Dart: calls through function values and callbacks (pervasive in Flutter), `dynamic`-typed dispatch, and `noSuchMethod` forwarding.

## Scope

- Wrapper for `dart test` runs with the VM service enabled, collecting samples per test process
- Resolution handling: closures and anonymous functions, async function frames (resolve to the source async function), mixin applications, extension methods
- Emit the shared interchange format from #1245, with the sampled/approximate flag set where applicable

## Acceptance criteria

- [ ] Traced `dart test` run of a sample project produces dynamic edges
- [ ] Demonstrated runtime-only edge through a function value or `dynamic` dispatch
- [ ] Sampled edges clearly distinguishable from exact edges in the graph
- [ ] Docs page covering usage and caveats

PR #1273, opened 2026-08-15
feat: flag sampled-profiler trace edges as approximate in the interchange format

vitali87:
## What

Adds a `sampled` boolean to the trace interchange header and a `dynamic_sampled` property on the CALLS edges written by ingestion, so an edge captured by a sampling profiler is honestly distinguishable from one observed by an exact tracer.

Sampling profilers (Go pprof, Node.js/V8 cpuprofile, .NET `dotnet-trace`, Dart CPU samples) infer parent/child adjacency and counts from stack samples, so a missing edge is not evidence of dead code and `dynamic_call_count` is a relative weight. Exact tracers (Python `sys.monitoring`, the JVM agent, Xdebug, the C shim, the Lua hook) observe every call. Until now both looked identical in the graph.

## Why

This is the interchange-format field the runtime-tracing epic (#1244) called for and that #1255 (Dart, VM Service CPU samples) needs for its acceptance criterion "sampled edges clearly distinguishable from exact edges in the graph". The gap affected every sampled-profile language, not just Dart.

## Changes

- `TraceHeader.sampled` (defaults to `False`); written and parsed in the JSONL header. Pre-flag trace files omit the key and read back as exact; a present value must be a real boolean.
- Each converter tags its header: pprof/cpuprofile/speedscope and the Dart collector set `sampled=true`; the instrumented (C/C++), Xdebug (PHP), and Python tracers set `sampled=false`. The JVM and Lua agents are exact and rely on the default.
- Ingestion stamps `dynamic_sampled` on every edge from the header flag.
- Docs: new row in the edge-property table.

## Tests

- Round-trip of both flag values; backward-compat (missing key reads as exact); non-boolean rejection.
- Ingestion stamps `dynamic_sampled` true for a sampled trace and false for an exact one.
- Converter-level assertions (pprof sampled, instrumented exact).

Refs #1244, #1255.

## Summary by CodeRabbit

* **New Features**
  * Trace data now records whether it originated from sampling-based profiling.
  * Dynamic call-graph edges indicate when their data is approximate due to sampling.
  * Sampled and fully instrumented traces are clearly distinguished across supported formats.
  * Trace formats remain compatible with older records lacking sampling metadata.

* **Documentation**
  * Updated dynamic tracing documentation to explain sampled data, approximate counts, and trace accuracy.

### DIFF
--- a/codebase_rag/constants/trace.py
+++ b/codebase_rag/constants/trace.py
@@ -20,6 +20,12 @@ TRACE_KEY_LINE = "line"
 TRACE_KEY_COUNT = "count"
 TRACE_KEY_WORKLOADS = "workloads"
 TRACE_KEY_RECEIVER_TYPES = "receiver_types"
+# Header flag: True when the tracer sampled the call stack (pprof, V8 cpuprofile,
+# dotnet-trace, Dart CPU samples) so parent/child adjacency and counts are
+# approximate, False when it instrumented every call (Python sys.monitoring, the
+# JVM agent, Xdebug, the C shim, the Lua hook) so edges and counts are exact.
+# Absent in pre-flag trace files, which read back as exact.
+TRACE_KEY_SAMPLED = "sampled"
 
 TRACE_LANGUAGE_PYTHON = "python"
 TRACE_LANGUAGE_JVM = "jvm"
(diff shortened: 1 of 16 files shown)


vitali87:
@greptile-apps Good catch, fixed in 5db9b24e. `convert_speedscope` handles both `sampled` and `evented` speedscope profiles, and evented profiles carry explicit frame open/close events, so their edges are exact. `_accumulate_profiles` now returns whether any accumulated profile was `sampled`, and the header flag is set from that (a file mixing both is treated as approximate). Added a regression assertion: the evented-profile test asserts `header.sampled is False`, the sampled-profile tests assert `True`.

PR #1274, opened 2026-08-15
feat: add Rust dynamic call-graph tracer (dyn dispatch capture)

### LINKED ISSUE #1251: Dynamic call-graph tracer for Rust (dyn dispatch capture)

## Approach

Rust, like Go, has no runtime instrumentation hook, and static analysis already resolves most Rust calls (monomorphized generics are static). The dynamic payoff is narrower but real: `dyn Trait` dispatch, function pointers/closures passed across boundaries, and calls routed through trait objects in plugin-style architectures. Candidates:

1. **`-Z instrument-mcount` / `-C instrument-coverage`-adjacent compile-time instrumentation**: coverage gives executed regions, not edges; mcount-style entry hooks give callee plus return address, from which the caller is recoverable via symbolization
2. **eBPF uprobes on the test binary (Linux)**: no rebuild, needs debug symbols, inlining hides callees
3. **A proc-macro/attribute instrumentation crate** applied to the workspace for traced builds: explicit and portable, but intrusive

Given the narrower payoff, prioritize this issue below the managed runtimes and pick the cheapest mechanism that demonstrates `dyn` dispatch capture.

## Scope

- Wrapper for `cargo test` runs of a traced build profile
- Inlining: traced profile builds with reduced inlining, overhead documented
- Resolution handling: symbol demangling (v0 mangling), closures, monomorphized instances collapsing to their generic source definition
- Capture the concrete type behind `dyn Trait` calls where the mechanism allows
- Emit the shared interchange format from #1245

## Acceptance criteria

- [ ] Traced `cargo test` run of a sample project produces dynamic edges
- [ ] Demonstrated runtime-only edge through a `dyn Trait` object
- [ ] Monomorphized instances resolve to the generic definition node

vitali87:
## What

Adds a dynamic call-graph tracer for **Rust**, the last runtime in the epic (#1244) with no tracer. `pprof-rs` / `cargo flamegraph` write gzipped protobuf CPU profiles in the same wire format as Go, so this reuses the Go pprof decoder and accumulation and adds only Rust-specific symbol demangling and scoping.

## Why

Static analysis already resolves monomorphised Rust calls, so the dynamic payoff is narrower but real: `dyn Trait` dispatch, function pointers, and closures routed across boundaries. This closes #1251.

## Approach

- `codebase_rag/trace/rust_pprof.py`: `convert_rust_pprof` reuses `_decompress` / `_decode_profile` / `_accumulate_edges` from `pprof.py`; a Rust `_bare_name` strips the legacy `::h<hash>`, splits paths at angle/paren depth zero so `<Dog as Animal>::speak` yields `speak`, drops trailing turbofish (`fn::<u32>` -> `fn`), collapses generic instantiations, and marks `{{closure}}` anonymous. Frames outside the repo and under `target/` are seen through to the nearest project frame. The header is tagged `sampled=True` (pprof-rs samples).
- Resolution reuses `JsFrameResolver` (span-first, bare name + declaration line), so monomorphised instances resolve to their single generic source definition; `ingest.py` adds `rust` to that dispatch tuple.
- CLI: Rust pprof is gzipped protobuf indistinguishable from Go's by magic bytes, so `cgr trace convert --language rust` selects the Rust demangler. Unknown `--language` values are rejected.
- Docs: supported-runtimes list plus a "Recording a Rust trace" section (pprof-rs recipe and caveats).

## Tests

- Converter: sample-weighted edges, the `dyn Trait` dispatch edge captured through an intervening std frame, `target/`+stdlib frames out of scope, closures/generics normalised, `_bare_name` demangling table, workload labelling, malformed-profile rejection.
- Ingest: a Rust trace resolves the dyn-dispatch edge to the graph node and flags it `dynamic_sampled` + `static_missed`.
- CLI: `--language rust` routes to the Rust converter (and needs `--repo-path`); unknown language rejected. Plus the `TRACE_LANGUAGE_RUST` constant assertion.

Stacked on #1273 (the `sampled` interchange flag it depends on). Closes #1251. Refs #1244.


vitali87:
@greptile-apps Both fixed in eeac19db (docs-only):

- **P1 (Message trait import):** added `use pprof::protos::Message;` to the recording snippet so `write_to_writer` resolves.
- **P2 (debug vs inlining):** reworded the guidance. The recipe now traces the dev profile (`cargo test`, `opt-level = 0`) so callees are not inlined away, and explicitly states `debug = true` only preserves symbols/line tables and does not reduce inlining in an optimized `--release` build.

Also switched the snippet to write `cpu.pb` (uncompressed protobuf, which is what `write_to_writer` produces) and noted the converter reads the profile gzipped or not.

vitali87:
Superseded by #1275. This PR was stacked on #1273's branch and GitHub auto-closed it when that base branch was deleted on merge. #1275 is the same commits rebased onto main (the `sampled` flag it depends on is now merged via #1273).

Issue #1252, opened 2026-08-13
Dynamic call-graph tracer for C/C++ (-finstrument-functions)


vitali87:
Part of #1244. Depends on #1245 and the Python pilot (#1246) validating the pipeline.

Covers C and C++ in one issue: they share compiler and tooling infrastructure. The dynamic payoff in C++ is virtual dispatch and calls through function pointers; in C it is function pointers (vtable-style structs, callback registries), which static analysis essentially cannot resolve.

## Approach

Candidates, to be evaluated:

1. **`-finstrument-functions` (GCC/Clang)**: entry/exit hooks receive callee and call-site addresses; symbolize offline against the binary's debug info; complete edges, requires rebuilding the project with the flag
2. **Valgrind callgrind**: no rebuild, exact call graph, but 20-50x slowdown; viable for small test suites only
3. **eBPF uprobes / DTrace**: no rebuild, platform-specific, inlining hides callees

Route 1 is the likely primary, with callgrind as a no-rebuild fallback.

## Scope

- A small runtime shim (the `__cyg_profile_func_enter` implementation) that aggregates address pairs and flushes on exit, plus an offline symbolizer that turns address pairs into `(file, function, line)` using DWARF info
- Integration recipe for CMake-based projects (traced build type), the same mechanism the repo already assumes for parsing
- Inlining: traced builds at `-O0`/`-Og` with the tradeoff documented
- Resolution handling: C++ name demangling, templates collapsing to their source definition, lambdas, static functions with identical names across TUs
- Capture the concrete class behind virtual calls where recoverable
- Emit the shared interchange format from #1245

## Acceptance criteria

- [ ] Traced test run of a sample CMake project produces dynamic edges
- [ ] Demonstrated runtime-only edge through a function pointer (C) and a virtual call (C++)
- [ ] Symbolization handles templates and demangling; unresolved addresses reported
- [ ] Docs page covering usage and caveats

vitali87:
All acceptance criteria are met and merged (PR #1288, plus the foundation shipped earlier in the epic):

- [x] **Traced test run of a sample CMake project produces dynamic edges** — `test_live_cmake_project_produces_dynamic_edges` configures/builds/runs a real CMake project (shim in the target sources, `-finstrument-functions -g -O0`, pthreads linked) and converts its trace to edges.
[...]

PR #1276, opened 2026-08-15
feat: demonstrate C++ virtual-dispatch tracing and report unresolved addresses

vitali87:
## What

Closes the two dominant gaps that kept #1252 open: the C++ virtual-call demonstration and the silent dropping of unresolvable addresses.

## Why

The instrumented C/C++ tracer already worked for C (function-pointer dispatch, live-tested), but #1252's acceptance criteria also require a demonstrated **C++ virtual-call** edge and that **unresolved addresses be reported** rather than silently dropped. Neither existed.

## Changes

- **C++ virtual-dispatch live test** (`test_live_cpp_virtual_dispatch_produces_virtual_edge`, `@pytest.mark.slow`, skipped without a C/C++ toolchain): compiles an `Animal`/`Dog` program where `dispatch(Animal*)` calls `a->speak()` through the abstract interface, runs the instrumented binary, and asserts the runtime-only `dispatch -> speak` edge (count 5) that static analysis cannot resolve through the trait object. The test encodes the correct build recipe — the shim is C, so it is compiled with the C compiler (`cc -c`) and linked with the C++ objects by the C++ driver; only the C++ translation unit carries `-finstrument-functions`. Compiling the `.c` shim with a C++ driver fails, which is why this was never demonstrated before.
- **Unresolved-address reporting**: `convert_instrumented` now distinguishes an address that symbolised to no source position at all (empty name or `??` from addr2line/atos — stripped symbols or missing debug info) from one that resolved to glue outside the repo, and logs a warning with the count (`TRACE_MSG_ADDRS_UNRESOLVED`) instead of dropping it silently. Unit-tested.
- **Docs**: the C/C++ section now gives separate C and C++ build recipes (including the CMake note) and documents the unresolved-address reporting.

## Tests

- `test_live_cpp_virtual_dispatch_produces_virtual_edge` — real compile + run, asserts the virtual edge (verified locally with gcc/g++/addr2line).
- `test_unresolved_addresses_are_reported` — a `??` symbol is dropped and the warning fires.

Refs #1252, #1244.

## Summary by CodeRabbit

* **Bug Fixes**
  * Dynamic tracing now reports unresolved addresses instead of silently dropping them.
  * Unresolved trace edges are excluded from results while their counts remain visible in warnings.
  * C++ virtual dispatch traces now correctly capture runtime call relationships and invocation counts.

* **Documentation**
  * Updated C and C++ dynamic tracing setup instructions, including CMake guidance.
  * Clarified behavior for function pointers, virtual dispatch, and unsymbolized addresses.

### DIFF
--- a/codebase_rag/constants/trace.py
+++ b/codebase_rag/constants/trace.py
@@ -73,6 +73,10 @@ TRACE_ERR_ADDRS_DROPPED = (
     "{path} overflowed the shim's edge table: call edges were dropped, so the "
     "trace is incomplete and cannot honour exact invocation counts."
 )
+TRACE_MSG_ADDRS_UNRESOLVED = (
+    "{count} of {total} instrumented addresses did not symbolise to a source "
+    "position (missing debug info or stripped symbols); their edges were dropped."
+)
 TRACE_ERR_BAD_XDEBUG = (
     "{path} is not an Xdebug computerized trace (expected 'File format: 4')."
 )
--- a/codebase_rag/tests/test_dynamic_trace_instrumented.py
+++ b/codebase_rag/tests/test_dynamic_trace_instrumented.py
@@ -13,6 +13,7 @@ import textwrap
 from pathlib import Path
 
 import pytest
+from loguru import logger
 
 from codebase_rag import constants as cs
 from codebase_rag.trace.instrumented import convert_instrumented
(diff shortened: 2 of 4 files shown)


vitali87:
@coderabbitai Clarified in the latest commit: the docs now state that an edge whose caller or callee does not resolve to a project frame is excluded from the converted trace, and that the unresolved-symbolisation count is reported in addition, so the gap is visible rather than silent.

vitali87:
@greptile-apps Good catch, fixed in af6d93b7. `_frame` now classifies an address with a valid name but no usable source position (empty/`??` path or line <= 0 — stripped symbols / missing debug info) as unresolved, so it is counted and reported instead of taking the silent out-of-repo drop path. Only an address that resolves to a real position genuinely outside the repo is dropped as glue. Regression test extended: a `(stripped_fn, "", 0)` symbol is now included and the warning asserts `2 of 4` addresses unresolved.

PR #1277, opened 2026-08-15
test: demonstrate Lua metatable __index dispatch capture

### LINKED ISSUE #1254: Dynamic call-graph tracer for Lua (debug.sethook)

## Approach

Lua ships an in-language hook that makes this one of the simplest tracers: `debug.sethook(fn, "c")` fires on every call, and `debug.getinfo` supplies source, line, and name for callee and caller frames. High payoff relative to effort: Lua dispatch is dynamic by construction (tables of functions, metatables, `__index` chains), so static analysis misses a large share of real edges.

For LuaJIT, the hook disables JIT compilation on traced paths; acceptable for test workloads, documented.

## Scope

- A pure-Lua tracer module loadable via `-l` or `require`, aggregating caller/callee pairs and flushing to the interchange format from #1245 at exit
- Integration recipe for busted test runs (and a generic entrypoint-wrapping recipe, since many Lua codebases are embedded and have no standard test runner)
- Resolution handling: anonymous functions (Lua functions frequently have no name; resolve by `(source, linedefined)` which matches how the static tier identifies them), C functions (record as unresolved-native), methods called via `:` sugar vs `.`
- Capture the receiver table's identity where nameable (metatable-based classes)

## Acceptance criteria

- [ ] Traced busted run of a sample project produces dynamic edges
- [ ] Demonstrated runtime-only edge through a metatable `__index` dispatch
- [ ] Anonymous functions resolve by definition position; unresolved frames reported
- [ ] Works on PUC Lua 5.1-5.4; LuaJIT behavior documented

vitali87:
Closes #1254's headline acceptance criterion: a demonstrated runtime-only edge through metatable `__index` dispatch.

## What

Adds a live Lua test (`test_agent_records_metatable_index_dispatch`, skipped without a `lua` interpreter) exercising OOP-via-metatables: `Dog` inherits `speak()` from `Animal` through `__index`, so `d:speak()` on a Dog instance dispatches to `Animal.speak`. The receiver's class and the resolved method are invisible to static analysis (the method lives on a different table reached only through the inheritance chain); the `debug.sethook` agent observes the actual `run -> speak` edge (count 4) and resolves it by span to the single `Animal.speak` definition.

The existing Lua tests cover table-of-functions dispatch and C-glue see-through; this adds the metatable `__index` case the issue calls out as the high-payoff showcase. Verified locally against the real `lua` interpreter.

Refs #1254, #1244.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for Lua method dispatch through metatables.
  * Verified that repeated method calls record the correct method, caller, source location, and invocation count.
  * Expanded tracing test support to run multiple Lua sample programs.

### DIFF
--- a/codebase_rag/tests/test_dynamic_trace_lua.py
+++ b/codebase_rag/tests/test_dynamic_trace_lua.py
@@ -56,13 +56,44 @@ _SAMPLE = """
     require("cgr_trace").write()
 """
 
+_INDEX_SAMPLE = """
+    -- OOP via metatables: Dog inherits speak() from Animal through __index,
+    -- so a call on a Dog instance dispatches to Animal.speak. The receiver's
+    -- class and the resolved method are invisible to static analysis.
+    local Animal = {}
+    Animal.__index = Animal
+
+    function Animal:speak()
+        return "woof"
+    end
+
+    local Dog = setmetatable({}, { __index = Animal })
+    Dog.__index = Dog
+
+    function Dog.new()
+        return setmetatable({}, Dog)
+    end
+
+    local function run()
+        local d = Dog.new()
+        local out = ""
+        for _ = 1, 4 do
+            out = out .. d:speak()
+        end
+        return out
+    end
+
+    print(run())
+    require("cgr_trace").write()
+"""
+
 lua = shutil.which("lua")
 pytestmark = pytest.mark.skipif(lua is None, reason="lua interpreter not available")
 
 
-def _run_traced_lua(tmp_path: Path) -> tuple:
+def _run_traced_lua(tmp_path: Path, sample: str = _SAMPLE) -> tuple:
     script = tmp_path / "main.lua"
-    script.write_text(textwrap.dedent(_SAMPLE))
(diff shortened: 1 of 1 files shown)


vitali87:
@coderabbitai Hardened in the latest commit. The test now collects all `run -> speak` records and asserts exactly one, then compares `callee.line` against the `function Animal:speak()` line computed from `_INDEX_SAMPLE` (via `textwrap.dedent`), proving the tracer resolved the inherited Animal.speak definition rather than anything on Dog. No more dict key that could overwrite matching records, and no more `line > 0` weak check.

vitali87:
@greptile-apps Already addressed in 58b5bf3f (pushed just before this review landed). The test no longer uses `line > 0`: it collects the `run -> speak` records, asserts exactly one, and asserts `dispatch.callee.line` equals the `function Animal:speak()` line computed from `_INDEX_SAMPLE` via `textwrap.dedent`. A regression that kept the edge/count but reported a wrong definition site (e.g. Dog.new) now fails.

Issue #1253, opened 2026-08-13
Dynamic call-graph tracer for PHP (Xdebug function traces)


vitali87:
Part of #1244. Depends on #1245 and the Python pilot (#1246) validating the pipeline.

## Approach

PHP is high-payoff for dynamic capture: variable functions (`$fn()`), variable method calls (`$obj->$method()`), `call_user_func`, magic methods (`__call`, `__get`), and framework containers (Laravel, Symfony DI) are pervasive and opaque to static analysis. Candidates:

1. **Xdebug function trace** (`xdebug.mode=trace`, machine-readable format): complete call records with file/line, no custom extension needed, well understood; overhead is significant but acceptable for test runs
2. **A custom observer-API extension (PHP 8+)**: lower overhead and tailored output, but a C extension to build and maintain per platform
3. **phpspy / sampling**: approximate edges only, supplementary

Route 1 first: parse Xdebug's computerized trace format into the interchange format without writing any C.

## Scope

- Wrapper for PHPUnit runs with Xdebug tracing enabled, one trace per process
- Parser from Xdebug trace format to the shared interchange format from #1245
- Workload provenance per test where feasible (PHPUnit event hooks)
- Capture receiver classes on method calls, including through `__call`
- Resolution handling: closures (`{closure}` frames), generated proxy classes from DI containers, traits (method reported in using class vs trait definition)

## Acceptance criteria

- [ ] Traced PHPUnit run of a sample project produces dynamic edges
- [ ] Demonstrated runtime-only edge through a variable method call or container resolution
- [ ] Trait and closure frames resolve correctly; unresolved frames reported
- [ ] Overhead measured and documented
- [ ] Docs page covering usage and caveats

vitali87:
All acceptance criteria are met and merged (PR #1289, on top of the Xdebug parser/resolver shipped earlier in the epic):

- [x] **Traced PHPUnit run of a sample project produces dynamic edges** — `test_live_phpunit_run_captures_variable_call_closure_and_trait` runs a real `phpunit` suite under `xdebug.mode=trace`, converts the trace, and asserts project edges.
[...]

PR #1279, opened 2026-08-15
feat: capture the concrete receiver class on PHP instance-call trace edges

vitali87:
## What

Two of #1253's gaps: no real traced-run demonstration, and `receiver_types` hardcoded empty despite Xdebug naming instance calls with the concrete runtime class.

## Changes

- **Receiver capture**: `convert_xdebug_trace` now populates `receiver_types` with the concrete receiver class for instance calls (`Class->method` -> `Class`, namespace stripped). A dispatch through an interface or base type therefore records which implementation actually ran; static calls (`Class::method`) are not dynamic dispatch and carry none. Metadata only, no effect on resolution.
- **Live E2E test** (`test_live_xdebug_trace_captures_polymorphic_dispatch`, `@pytest.mark.slow`, skipped unless php+Xdebug are present): traces a real PHP program where `dispatch(Animal)` calls `$a->sound()` polymorphically, and asserts the runtime-only `dispatch -> Dog->sound` (x2) / `Cat->sound` edges with `receiver_types` of `Dog`/`Cat`, plus a trait method (`Service->greet`) captured under the using class. Verified locally against PHP 8.3 + Xdebug 3.
- **Unit test** for receiver capture against the synthetic sample; **docs** note the receiver-class capture and the leaf-trait-method resolution limitation.

## Scope note

The leaf-trait-method *resolution* gap (a trait method that makes no calls, so it falls to the `Owner.method` name tail which cannot match the graph's `TraitName.method` node) is documented rather than forced: a safe fix needs trait-usage information the frame resolver does not currently load, and guessing by bare method name would mis-resolve unrelated same-named methods. Non-leaf trait methods already resolve correctly by span.

Refs #1253, #1244.

## Summary by CodeRabbit

* **New Features**
  * Dynamic PHP tracing now records the concrete runtime class handling instance method calls.
  * Trace results distinguish instance dispatch from static calls and support trait-based method resolution.

* **Bug Fixes**
  * Improved attribution for polymorphic calls and magic `__call` handling.
  * Unresolved trait methods are represented accurately in trace results.

* **Documentation**
  * Updated dynamic tracing guidance to explain receiver types, trait methods, static calls, and unresolved methods.
  * Clarified how magic method calls are attributed in tracing output.

### DIFF
--- a/codebase_rag/tests/test_dynamic_trace_xdebug.py
+++ b/codebase_rag/tests/test_dynamic_trace_xdebug.py
@@ -6,11 +6,30 @@
 
 from __future__ import annotations
 
+import shutil
+import subprocess
+import textwrap
+
 import pytest
 
 from codebase_rag import constants as cs
 from codebase_rag.trace.records import read_trace_file
 from codebase_rag.trace.xdebug import convert_xdebug_trace
+
+
+def _php_with_xdebug() -> str | None:
+    """The php interpreter path when it can load Xdebug, else None."""
+    php = shutil.which("php")
+    if php is None:
+        return None
+    probe = subprocess.run(
+        [php, "-r", 'echo extension_loaded("xdebug") ? "yes" : "no";'],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    return php if probe.stdout.strip() == "yes" else None
+
 
 _HEADER = "Version: 3.5.3\nFile format: 4\nTRACE START [2026-08-14 02:07:42]\n"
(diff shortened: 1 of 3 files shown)


vitali87:
@greptile-apps Fixed in 8936e3f7. The import-time `_php_with_xdebug()` probe now runs with `timeout=15` and treats `subprocess.TimeoutExpired`/`OSError` as 'xdebug unavailable', so a stalled interpreter degrades to skipping the live test instead of hanging pytest collection.

Issue #1247, opened 2026-08-13
Dynamic call-graph tracer for JavaScript/TypeScript/TSX (Node.js/V8)


vitali87:
Part of #1244. Depends on #1245 and the Python pilot (#1246) validating the pipeline.

Covers JavaScript, TypeScript, and TSX in one issue: they share the Node.js/V8 runtime, and TS/TSX trace back to JS via source maps.

## Approach

Candidate mechanisms, to be evaluated in this order:

1. **V8 Inspector Protocol precise coverage + instrumentation**: attach via `node --inspect` or `inspector` module; `Profiler.startPreciseCoverage` gives executed functions but not edges, so it complements rather than replaces call capture
2. **Instrumentation via module hooks**: a Node loader (`module.register`) that wraps exported functions to record caller/callee pairs; robust for CommonJS/ESM app code, blind to intra-module private calls
3. **V8 CPU profiling with high sampling**: cheap but sampled, edges are approximate; likely acceptable only as a supplementary signal

The chosen mechanism must attribute TS/TSX source positions through source maps so records resolve to the nodes the static tier created from the original sources.

## Scope

- Tracer packaged so `npm test` / `vitest` / `jest` runs can be wrapped with minimal config
- Workload provenance from the test runner where feasible (jest/vitest reporters)
- Source-map translation from transpiled JS back to TS/TSX positions
- Resolution handling: anonymous/arrow functions, class methods vs prototype assignment, re-exported bindings, bundled or transpiled output
- Emit the shared interchange format from #1245

## Acceptance criteria

- [ ] Traced test run of a sample TS project produces dynamic edges resolved to TS source nodes
- [ ] At least one demonstrated runtime-only edge (e.g. dispatch through an event emitter or dynamic `import()`)
- [ ] Resolution rate reported; source-map failures categorized
- [ ] Docs page covering usage and caveats

vitali87:
All acceptance criteria are met and merged (PR #1292, on top of the V8 cpuprofile converter and Source Map v3 remapping shipped earlier in the epic):

- [x] **Traced test run of a sample TS project produces dynamic edges resolved to TS source nodes** — `test_live_typescript_trace_resolves_to_source` compiles a TS project with `tsc --sourceMap`, runs it under `node --cpu-prof`, converts, and asserts project frames relocate off `dist/*.js` onto their `src/*.ts` source.
[...]

PR #1280, opened 2026-08-15
feat: remap transpiled-JS trace frames to TypeScript sources via source maps

vitali87:
## What

Closes the central gap that kept #1247 open: transpiled-JS trace frames now resolve back to their TypeScript sources via source maps. A project indexed as `src/*.ts` but run from `dist/*.js` previously produced `unresolved[unknown_path]` frames; now they land on the indexed `.ts` nodes.

## How

- **`codebase_rag/trace/sourcemap.py`** (new): a dependency-free Source Map v3 reader — Base64-VLQ decode, `mappings` parse (delta-decoded across the payload, generated column reset per line), and nearest-preceding-segment lookup. `SourceMapIndex` lazily loads and caches each generated file's map, following its `sourceMappingURL` (external `.map` only; inline `data:` maps are left to the on-disk file) or falling back to `<file>.map`.
- **`cpuprofile.py`**: `_project_frame` now reads the frame's `columnNumber` and, when a map covers the generated `(line, column)`, relocates the frame to the original `(source, line)` before the repo-scope and excluded-dir checks. Frames with no mapping (runtime glue, the occasional module wrapper) keep their generated position.
- **Docs**: the "Transpiled output" caveat now describes the source-map support instead of calling it a planned follow-up.

## Tests

- `test_dynamic_trace_sourcemap.py`: VLQ/mapping/lookup against a **real `tsc --sourceMap` map** (positions verified: `greet`->app.ts:4, `handle`->:12, `run`->:16); nearest-segment fallback; unmapped/malformed -> None; a converter test proving the registry-dispatch edge `handle -> greet` relocates both endpoints from `dist/app.js` to `src/app.ts`.
- **Live E2E** (`@pytest.mark.slow`, skipped without `node`+`tsc`): really transpiles a TS file, profiles it with `node --cpu-prof`, converts, and asserts a genuine function->function edge was relocated to `.ts` and no named-function frame stays on the transpiled `.js`. Verified locally against node 18 + TypeScript.

Refs #1247, #1244.

## Summary by CodeRabbit

* **New Features**
  * Trace records now map generated JavaScript and CPU-profile frames to their original TypeScript/JavaScript files and positions.
  * Supports external, adjacent, and indexed source maps.
  * Unmapped frames retain their generated locations.

* **Bug Fixes**
  * Improved handling of invalid, missing, malformed, and unmapped source-map data.

* **Documentation**
  * Updated dynamic tracing guidance with source-map behavior and setup requirements.

### DIFF
--- a/codebase_rag/tests/test_dynamic_trace_sourcemap.py
+++ b/codebase_rag/tests/test_dynamic_trace_sourcemap.py
@@ -13,7 +13,6 @@ from pathlib import Path
 
 import pytest
 
-from codebase_rag import constants as cs
 from codebase_rag.trace.cpuprofile import convert_cpuprofile
 from codebase_rag.trace.records import read_trace_file
 from codebase_rag.trace.sourcemap import SourceMapIndex, load_source_map
@@ -119,10 +118,13 @@ def test_index_map_sections_are_flattened(tmp_path):
     assert source_map is not None
     # First section: unshifted, identical to the flat map.
     first = source_map.original_position(10, 12)
-    assert first is not None and first[0].endswith("src/app.ts") and first[1] == 16
+    assert first is not None
+    assert first[0].endswith("src/app.ts")
+    assert first[1] == 16
     # Second section at line 50: the same inner `run` frame is now at line 60.
     shifted = source_map.original_position(60, 12)
-    assert shifted is not None and shifted[0].endswith("src/app.ts")
+    assert shifted is not None
+    assert shifted[0].endswith("src/app.ts")
     assert shifted[1] == 16
 
 
@@ -211,7 +213,7 @@ def test_live_typescript_trace_resolves_to_source(tmp_path):
         "function run(): void {\n"
         "    register('greet', greet);\n"
         "    let out = '';\n"
-        "    for (let i = 0; i < 100000; i++) { out = handle('greet'); }\n"
+        "    for (let i = 0; i < 3000000; i++) { out = handle('greet'); }\n"
         "    if (out.length < 0) console.log(out);\n"
         "}\n"
         "run();\n"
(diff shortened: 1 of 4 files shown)


vitali87:
@coderabbitai All three fixed in the latest commit:

- **Truncated VLQ:** `_decode_vlq` now raises `ValueError` when the final digit sets the continuation bit with no digit following (e.g. `AAAAg`); `_parse_mappings` propagates it and `load_source_map` returns None, so a malformed map is rejected rather than mis-decoded. Regression test added.
- **Per-frame allocation:** `original_position` no longer rebuilds a column list per lookup; it bisects the sorted segment tuples directly with `key=lambda seg: seg[0]`, so a minified line's many segments cost no per-lookup allocation.
- **Docs:** the caveat now says to enable the source-map option explicitly (tsc `--sourceMap`, the bundler equivalent whose default varies) and that without a map, or for an unmapped position, the frame keeps its generated `dist/*.js` location.

vitali87:
@greptile-apps Both fixed in da1ee7c6, each with a regression test:

- **Index (sectioned) maps:** `_from_document` now detects a top-level `sections` list and flattens it via `_from_sections` — each section's inner map is parsed, its segments shifted by the section's generated `offset` (column only on the section's first line), and its sources absolutised, so a bundler's index map resolves exactly like a flat map. Test wraps the real tsc map in two sections and asserts the offset section relocates `run` from line 10 to line 60 -> the same src/app.ts:16.
- **Encoded map URLs:** `_sniff_map_reference` now runs the `sourceMappingURL` through `urlsplit` (dropping `?query`/`#fragment`) and `unquote` (percent-decoding) before resolving the file, so `app%20map.js.map?v=123` finds `app map.js.map`. Test covers it.

vitali87:
@coderabbitai Fixed: `_sniff_map_reference` now wraps the `urlsplit`/`unquote` call in `try/except ValueError`, so a malformed sourceMappingURL (e.g. `http://[`) returns None and the frame keeps its generated location rather than aborting conversion. Regression test added.

PR #1282, opened 2026-08-15
test: demonstrate Dart dynamic-dispatch capture via the VM Service collector

vitali87:
Closes a #1255 acceptance gap: the collector was only tested against function-value/registry dispatch, not `dynamic`-typed dispatch, which the docs already claim to capture.

## What

Adds a live test (`test_collector_captures_dynamic_dispatch`, `@pytest.mark.slow`, skipped without the Dart toolchain) exercising `int dispatchDynamic(dynamic a) => a.speak();`. Because the receiver is `dynamic`, the concrete class is unknowable statically; the VM Service CPU samples resolve the call to the concrete `Dog.speak` / `Cat.speak` implementations that actually ran (each at its own definition line). Verified locally against Dart 3.13 — a single dynamic call site produces runtime-only `dispatchDynamic -> speak` edges to both concrete methods.

The collector runner is refactored into `_run_collector(tmp_path, source)` shared by the existing and new fixtures.

Refs #1255, #1244.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for Dart dynamic dispatch through abstract interfaces and concrete implementations.
  * Verified collected VM samples resolve calls to the correct concrete methods and valid source locations.
  * Improved test execution reuse while retaining existing registry-dispatch coverage.
  * Added a reusable workload fixture to support reliable dynamic-dispatch trace validation across test runs.

### DIFF
--- a/codebase_rag/tests/test_dynamic_trace_dart.py
+++ b/codebase_rag/tests/test_dynamic_trace_dart.py
@@ -154,11 +154,21 @@ def test_collector_captures_dynamic_dispatch(dynamic_dispatch_trace):
         if (r.caller.qualname, r.callee.qualname) == ("dispatchDynamic", "speak")
     ]
     assert dispatched, sorted({(r.caller.qualname, r.callee.qualname) for r in records})
-    # Each concrete implementation resolves to its own definition line, so a
-    # single dynamic call site maps to the real methods that ran.
     for record in dispatched:
         assert record.callee.path.endswith("main.dart")
-        assert record.callee.line > 0
+    # Both Dog.speak and Cat.speak run through the one dynamic call site, and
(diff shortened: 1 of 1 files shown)


vitali87:
@coderabbitai @greptile-apps Good catch, strengthened. The test now computes both `int speak() {` definition lines from the sample and asserts `speak_defs <= observed` — i.e. **both** `Dog.speak` and `Cat.speak` must appear as distinct `dispatchDynamic -> speak` callee lines. A regression that dropped one concrete receiver, or collapsed both onto one line, now fails. Verified stable across repeated runs (each concrete method does 6M iterations over 10 alternating rounds, so both are reliably sampled).

Issue #1249, opened 2026-08-13
Dynamic call-graph tracer for C# (.NET profiler / EventPipe)


vitali87:
Part of #1244. Depends on #1245 and the Python pilot (#1246) validating the pipeline.

## Approach

Candidate mechanisms for .NET, to be evaluated in this order:

1. **CLR Profiling API (`ICorProfiler`)** with IL rewriting at method entry: complete edge capture, but a native profiler component per platform
2. **`System.Diagnostics.Metrics` / EventPipe with a startup hook**: managed-only, simpler to ship, but needs explicit instrumentation of call sites via IL weaving (e.g. Fody/Mono.Cecil at build time) rather than attach-time rewriting
3. **`dotnet-trace` sampled CPU profiles**: cheap but sampled; supplementary signal only

Like the JVM, .NET is high-payoff: interface dispatch, DI (`IServiceCollection`), reflection, and source-generated or emitted code are common and static analysis cannot resolve them.

## Scope

- Tracer runnable against `dotnet test` with minimal configuration, plus include/exclude namespace filters
- Capture concrete receiver types on virtual/interface dispatch
- Resolution handling: async state machines (`MoveNext` frames must map back to the original async method), iterators, lambdas/local functions (compiler-generated display classes), generated code
- Workload provenance from the test framework (xUnit/NUnit/MSTest) where feasible
- Emit the shared interchange format from #1245

## Acceptance criteria

- [ ] Traced `dotnet test` run of a sample project produces dynamic edges with receiver types
- [ ] Async methods resolve to their source declarations, not state-machine internals
- [ ] Demonstrated runtime-only edge through DI or reflection
- [ ] Overhead measured and documented
- [ ] Docs page covering usage and caveats

vitali87:
All acceptance criteria are met and merged (PR #1291, on top of the dotnet-trace speedscope converter and CLR demangling shipped earlier in the epic):

[...]

PR #1283, opened 2026-08-15
feat: demangle C# local-function CLR frames to their nested source node

vitali87:
Closes a #1249 resolution gap: C# local-function frames from dotnet-trace were dropped as synthetic.

## What

A C# local function compiles to the CLR name `<EnclosingMethod>g__LocalName|N_M` (verified via reflection on a real `dotnet` build: `int Local()` inside `Run` -> `<Run>g__Local|0_0`). `_demangle_clr_name` had no rule for `g__`, so the frame hit the `"<" in method` guard and was dropped as `unresolved[synthetic]`.

Unlike a lambda (anonymous, resolved to its enclosing method), a local function keeps a source name, and the static C# tier nests it as a `Method.Local` node under the hosting method (confirmed by `test_local_function_*` in test_csharp_definitions.py). The new `_DOTNET_LOCAL_FUNCTION` rule demangles `<Run>g__Local|0_0` -> `Run.Local`, so `N.Sample.<Run>g__Local|0_0` resolves to the `N.Sample.Run.Local` node instead of being dropped.

## Tests

- `test_resolves_local_function_to_nested_node` (unit, synthetic CLR frame matching the reflection-verified format) asserts `MyApp.Worker.<Run>g__Local|0_0` -> `...MyApp.Worker.Run.Local`.
- Docs: added local functions to the CLR-demangling list.

Note: the dominant #1249 gap (concrete-receiver capture) still needs an instrumented CLR profiler and is out of scope here; this closes the local-function resolution hole the audit flagged.

Refs #1249, #1244.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved .NET trace resolution for C# local functions, mapping runtime names to their corresponding nested methods.
  * Preserved correct resolution for async methods, lambdas, constructors, and property accessors.

* **Documentation**
  * Updated dynamic tracing guidance with local-function and nested-class resolution examples.

### DIFF
--- a/codebase_rag/tests/test_dynamic_trace_resolution_dotnet.py
+++ b/codebase_rag/tests/test_dynamic_trace_resolution_dotnet.py
@@ -64,6 +64,13 @@ _NODES = [
         f"{_P}.Worker.MyApp.Worker.Run.Local",
         path="Worker.cs",
     ),
+    # An accessor-hosted local nests at class level (the accessor body is not
+    # its own scope in the static tier).
+    _node(
+        cs.NodeLabel.METHOD,
+        f"{_P}.Worker.MyApp.Worker.Local",
+        path="Worker.cs",
+    ),
 ]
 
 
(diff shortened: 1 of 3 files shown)


vitali87:
@coderabbitai Handled the accessor case; scoped out the ctor case with a reason.

- **Accessor-hosted locals** (`<get_Size>g__Local|0_0`): the static C# tier does not treat an accessor body as its own scope, so it nests the local directly under the class (`N.Sample.Local`, verified by running the real frontend on a fixture). `_local_function_target` now drops a `get_`/`set_` host and resolves to that class-level node. Regression test added.
- **Constructor-hosted locals** (`<.ctor>g__Local|0_0`): the `.ctor` token carries a dot that `_demangle_clr_name`'s `rpartition('.')` consumes before the local-function pattern is reached, so these can't be cleanly demangled without bracket-aware name splitting. They're rare, and rather than mis-resolve them I leave them `unresolved[synthetic]` (no regression) and documented the limitation in the pattern comment.

Issue #1287, opened 2026-08-16
feat: ingest eBPF-based continuous-profiler pprof profiles (production dynamic call graphs)


vitali87:
Sibling of #1244. That epic tracks per-runtime collectors for test/dev workloads; this issue tracks a **production-grade collection path**: ingesting pprof profiles produced by eBPF-based continuous profilers (Parca, Pyroscope, the OpenTelemetry eBPF profiler, `perf`-derived pprof) into the same hybrid graph.

## Why

All current collectors live inside the target runtime (`sys.monitoring`, the JVM agent, V8's profiler, the `-finstrument-functions` shim) and are aimed at test suites — the docs explicitly say the C shim is "not for production." eBPF profilers sit outside the process at the kernel level: zero instrumentation of the target, ~1% overhead, fleet-wide, continuous. Their output is the pprof wire format we already decode, so production call chains can fuse through the existing `cgr trace ingest` door with provenance-tagged `CALLS` edges (`dynamic_sampled: true`, `static_missed`, workload attribution) — no ingest changes.

## What already exists (reused as-is)

- `convert_pprof_profile()` in `codebase_rag/trace/pprof.py` — the dependency-free protobuf decoder, leaf-first stack walk, see-through-glue frame collapsing, and inline-frame (`Line`) expansion.
- The JSONL interchange format (`trace/records.py`) — `sampled: true` header, aggregated `CallRecord`s. No schema change for v1.
- Ingest (`trace/ingest.py`) — untouched.

## v1 scope

**1. New front door: `cgr trace convert profile.pb.gz --format ebpf`** (sibling of the Go/Rust converters, which differ only in `build_frame` and header tags). Three structural differences from `go test -cpuprofile` output the decoder currently ignores:

- **Mappings (pprof field 3, currently skipped).** Fleet profiles contain stacks from many binaries (target service, libc, kernel). Parse the `Mapping` table to filter frames to the target binary's build-id, and count unsymbolized locations per mapping instead of silently dropping them (same symbolisation-gap honesty the C shim documents).
- **Sample labels (inside field 2, currently skipped).** Parca/OTel attach `pid`, `thread`, `container`, and service labels per sample. Uses: filter to the target service; map a chosen label (e.g. `endpoint`, `service_name`) to `workloads` — production's analogue of "which test exercised this edge." The schema already takes arbitrary workload strings.
[...]

PR #1297, opened 2026-08-17
feat: resolve interpreted-runtime frames in eBPF pprof profiles (--language python)

vitali87:
Closes the last open follow-up on #1287: interpreted-runtime frames.

## What

`cgr trace convert --format ebpf --language python` now resolves the source-level frames an eBPF profiler (OpenTelemetry, Parca, Pyroscope) unwinds from the interpreter stack in the kernel. Those frames arrive as names against the real `.py` file (`<module>`, a bare `leaf_compute`, or `Service.handle_request`), not as native binary addresses, so the native Go/Rust/C++ path could not handle them.

## How

The native path assumes each frame is an address in the target's own binary: it demangles the symbol and filters frames to a `--build-id` mapping. Neither holds for interpreted frames, so `--language python` takes a different route, all automatic:

- **Names pass through untouched**, except for stripping the `py::` prefix the perf-map symbolisation route (which some profilers reuse) prepends. No native demangler runs.
- **No `--build-id` filtering.** Interpreted frames live in the interpreter's mapping, not the target binary, so a native build-id would drop every one. They are scoped to the project by source-path containment instead (the same containment/`--path-map` re-anchoring the native path already applies).
- **Emitted as `language=python`**, so ingestion routes the records to the existing Python `FrameResolver` — the same resolver the in-process `sys.monitoring` tracer feeds. No ingest changes.
- Frames from outside the repository (stdlib, site-packages) are counted and reported like any other unmapped path, never silently dropped.

## Verification

The two frame shapes the tests are built against were captured from **real profiles** of a Python 3.12 workload:

- perf-map symbolisation: `py::Service.handle_request` (Class.method, `py::`-prefixed, definition line in `Function.start_line`)
- py-spy-style: `leaf_compute` (bare name)

4 new tests cover `py::` stripping + Class.method preservation, `<module>` binding, seen-through/counted out-of-repo stdlib frames, the build-id bypass, and a full convert -> ingest -> resolve proof that the frames bind to Module/Method/Function graph nodes with zero unresolved. Full `dynamic_trace` suite: 223 passed.

## Scope

Python only, since that is the shape I could verify against real captures. The machinery is language-parameterised (`_INTERPRETED_NAME_FNS`), so JVM/Ruby drop in as small follow-ups once a real profile of each is available to design the frame/label shape against. Docs and `--language` help updated.

## Summary by CodeRabbit

* **New Features**
  * Added support for converting and ingesting Python eBPF profiles.
  * Python traces resolve interpreted frames, preserve metadata, and record sampled call relationships.
  * Added support for common Python profiler formats and repository source-path filtering.
  * Added the `--language python` option for tracing Python runtimes.
  * Python frames remain available when native build-ID filters are applied.

* **Bug Fixes**
  * Improved cross-platform repository-relative source-path handling for interpreted-language trace frames.

* **Documentation**
  * Added guidance for converting, filtering, and ingesting interpreted Python eBPF profiles.

### DIFF
--- a/codebase_rag/cli_help.py
+++ b/codebase_rag/cli_help.py
@@ -122,8 +122,9 @@ HELP_TRACE_INCLUDE = (
 HELP_TRACE_LANGUAGE = (
     "Override the auto-detected source language. Rust pprof profiles share Go's "
     "gzipped-protobuf format, so pass 'rust' to demangle them as Rust. For an "
-    "eBPF profile of an interpreted runtime, pass 'python' to resolve the "
-    "in-kernel-unwound source frames against the Python graph."
+    "eBPF profile of an interpreted runtime, pass '--format ebpf --language "
+    "python' together to resolve the in-kernel-unwound source frames against "
+    "the Python graph ('python' is an eBPF-only override)."
 )
 HELP_TRACE_FORMAT = (
     "Force a profile format instead of auto-detecting it. Use 'ebpf' for pprof "
--- a/codebase_rag/cli_help.py
+++ b/codebase_rag/cli_help.py
@@ -121,7 +121,9 @@ HELP_TRACE_INCLUDE = (
 )
 HELP_TRACE_LANGUAGE = (
     "Override the auto-detected source language. Rust pprof profiles share Go's "
-    "gzipped-protobuf format, so pass 'rust' to demangle them as Rust."
+    "gzipped-protobuf format, so pass 'rust' to demangle them as Rust. For an "
+    "eBPF profile of an interpreted runtime, pass 'python' to resolve the "
+    "in-kernel-unwound source frames against the Python graph."
 )
 HELP_TRACE_FORMAT = (
     "Force a profile format instead of auto-detecting it. Use 'ebpf' for pprof "
--- a/codebase_rag/tests/test_dynamic_trace_ebpf.py
+++ b/codebase_rag/tests/test_dynamic_trace_ebpf.py
@@ -577,3 +577,189 @@ def test_pull_strips_auth_on_redirect(tmp_path):
     # Neither the standard nor the custom credential header crossed the redirect.
     assert received["auth"] is None
     assert received["apikey"] is None
+
+
+# --- Interpreted-language frames (issue #1287 follow-up) ------------------
+#
+# The OTel/Parca/Pyroscope eBPF profilers unwind the interpreter stack in the
+# kernel and report Python frames as source-level names against the real ``.py``
+# file, not as native binary addresses. The shapes below are the two observed in
+# real captures on a Python 3.12 workload:
+#
+#   * perf-map symbolisation:  ``py::Service.handle_request``  (Class.method,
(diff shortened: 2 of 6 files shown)


vitali87:
@coderabbitai Addressed in b82a4e16. Added a sentence to the interpreted-runtimes docs section stating Python eBPF profiles are sampled: ingested edges carry `dynamic_sampled: true` and `dynamic_call_count` is an approximate sample count, explicitly contrasted with the exact per-call totals the in-process `sys.monitoring` tracer records.

vitali87:
@greptile-apps Thanks, both reviewed:

1. **cli_help.py:126 (P2, valid)** — fixed in b82a4e16. The `--language` help now shows `--format ebpf --language python` together and notes `python` is an eBPF-only override, since the sniffed (non-ebpf) path only accepts `rust`. Good catch on the misleading standalone guidance.

2. **test:636 `_py_convert` naming (P2)** — keeping the name: it deliberately mirrors the existing `_convert` sibling helper directly above it in the same file (same build-profile / convert / read-back / return-tuple shape), so the parallel name is the clearer signal here. Renaming only one of the pair would obscure that they are the native vs Python variants of the same helper.

vitali87:
Added 792c1e0c: the new convert -> ingest end-to-end test surfaced a pre-existing cross-platform bug on the Windows CI runner. The pprof/eBPF (and V8/Lua/Dart/PHP) converters emit POSIX frame paths, but the frame resolvers gated repo-containment on `str(repo_root) + os.sep`, which is backslash-separated on Windows, so every pprof-sourced frame read as OUTSIDE_REPO and resolved to nothing. Normalized the containment check to POSIX (`_repo_relative`) across FrameResolver / JsFrameResolver / PhpFrameResolver; the change is idempotent on POSIX hosts (52 local trace-resolution/ingest tests green) and the interpreted-frames end-to-end test now doubles as the Windows regression guard.

vitali87:
@coderabbitai Partially adopted in eadcb560.

**Taken:** `..` traversal. `_repo_relative` now runs `posixpath.normpath` before the containment check, so an in-repo frame that dips through a parent still resolves and one that walks out stays out. Added a regression test (`test_repo_relative_collapses_parent_traversal_and_gates_escapes`) covering in-and-back, escape, and unrelated paths.

**Declined (with reasoning):** the full `Path.resolve()` canonicalization for case-insensitivity and symlink aliases.
- Both were pre-existing behaviors, not regressions this PR introduced: the original code used lexical `relative_to` + case-sensitive `str.startswith` too. This PR's goal was the separator fix.
- `resolve()` hits the filesystem per frame on the hot ingest path (thousands of frames), and it would canonicalize symlinks/case that the **indexer never resolves** when it keys nodes by lexical relative paths, so resolving only the frame side would introduce *new* mismatches, not fix them. Containment and node identity must use the same normalization, and the graph's is lexical.
- There is no traversal risk to harden against: the returned relative path only keys `_callables_by_path` (known in-repo nodes), so a stray fragment fails to match rather than escaping. Documented this in the helper's docstring.

PR #1310, opened 2026-08-18
fix: reject invalid source location ranges

Whxuan0701:
## Summary

- Reject negative and reversed source ranges during location validation.
- Keep invalid locations from reaching the source extraction path.
- Add focused regression coverage for each invalid boundary.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

No existing issue found.

## Test Plan

- [x] New tests added

Focused validation:

- `uv run pytest -q codebase_rag/tests/test_source_extraction.py` (37 passed)
- `uv run ruff format --check codebase_rag/utils/source_extraction.py codebase_rag/tests/test_source_extraction.py`
- `uv run ruff check codebase_rag/utils/source_extraction.py codebase_rag/tests/test_source_extraction.py`
- `uv run ty check codebase_rag/utils/source_extraction.py`
- `git diff --check`

Repository-wide local unit tests were also attempted on the unchanged base and reached 7,161 passed / 142 skipped, with two unrelated failures and one collection error caused by optional dependencies and the Intel macOS environment. CI remains the authoritative full-suite result.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source location validation to reject missing paths, unspecified line values, zero or negative line numbers, and invalid ranges where the ending line precedes the starting line.
  * Prevents malformed source references from being accepted, resulting in more reliable source extraction and clearer handling of invalid input.

### DIFF
--- a/codebase_rag/tests/test_source_extraction.py
+++ b/codebase_rag/tests/test_source_extraction.py
@@ -239,6 +239,18 @@ class TestExtractSourceWithFallback:
 
 
 class TestValidateSourceLocation:
+    def test_rejects_zero_start_line(self) -> None:
+        valid, path = validate_source_location("/path/to/file.py", 0, 10)
+
+        assert valid is False
+        assert path is None
+
+    def test_rejects_zero_end_line(self) -> None:
+        valid, path = validate_source_location("/path/to/file.py", 1, 0)
+
+        assert valid is False
+        assert path is None
+
     def test_rejects_negative_start_line(self) -> None:
         valid, path = validate_source_location("/path/to/file.py", -1, 10)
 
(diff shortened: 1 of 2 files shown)


vitali87 (commented):
Good catch on both fronts: the old truthiness check rejected zero by accident and let negative and reversed ranges straight through. One addition requested (echoing CodeRabbit's inline comment): add explicit zero-boundary cases (start_line=0 and end_line=0) to the parametrised tests, since zero rejection is now intentional behaviour rather than a falsy accident, and a future refactor to explicit comparisons should not silently change it. With that, this is merge-ready.

Whxuan0701:
Addressed in commit 643092e. I consolidated the invalid range cases into a parametrized test, including explicit start_line=0 and end_line=0 cases. The focused tests and Ruff checks pass.

vitali87:
@Whxuan0701 thank you again.

PR #1312, opened 2026-08-18
chore(ruff): add PLR0917 to ignore list alongside PLR0913

aoright:
### Summary
- Add `PLR0917` (too many positional arguments) to Ruff's ignore list in `pyproject.toml` alongside `PLR0913`.
- Ensures `ruff check .` passes cleanly on latest Ruff versions where `PLR0917` is enabled under the `PL` ruleset.

## Summary by CodeRabbit

* **Chores**
  * Updated code quality checks to allow functions with many positional arguments without raising a lint warning.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,6 +147,7 @@ ignore = [
     "PLR2004",  # magic value comparison (too noisy)
     "PLC0415",  # import not at top-level (intentional lazy imports)
     "PLR0913",  # too many arguments (often unavoidable)
+    "PLR0917",  # too many positional arguments (often unavoidable)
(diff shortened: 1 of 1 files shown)


vitali87 (commented):
Verified locally against the pinned ruff (0.14.14): ruff check passes clean today because PLR0917 is preview-gated, and forcing --preview --select PLR0917 shows 141 latent hits, so this ignore is exactly the future-proofing the description claims and is coherent with the adjacent PLR0913 suppression (sibling rules; suppressing one without the other would break CI the day the rule stabilises). Will merge once checks are green.

Issue #1179, opened 2026-08-10
Go semantic frontend via go/packages + go/types


vitali87:
## Why Go first

Go has the best payoff-to-friction ratio of any language without a semantic layer:

- `parsers/go/` is the thinnest real package (~480 LOC: `type_inference.py` 204, `module_paths.py` 153, `utils.py` 105), with receiver-method binding deferred into `graph_updater.py`.
- **Interface-method dispatch is the single worst heuristic failure mode in Go** — the name trie cannot know which concrete types satisfy an interface, so calls through interfaces either fan out wrong or miss.
- Friction is minimal: where there is Go code there is a Go toolchain and a `go.mod`, and the type-checker is fast.
- A working seed exists: `evals/oracles/go_ast.go` already proves the bundled-Go-tool pattern.

## Plan

A small Go binary, bundled and built once like the Roslyn tool (`csharp_frontend`'s `~/.cgr/` build-lock pattern), loading the module via `golang.org/x/tools/go/packages` (`NeedSyntax | NeedTypes | NeedTypesInfo`) and emitting a JSON fact bundle per the `LanguageFrontend` protocol:

- **resolved_call_sites**: exact callee per call expression from `types.Info.Uses` — method values, embedded-struct promotion, shadowing handled by the real scope rules
- **interface dispatch**: for calls through interface types, the set of in-module concrete implementations (`types.Implements` / method sets) — emitted as candidate CALLS with a property distinguishing static from dispatch edges
- **external_sites**: callees resolving outside the module → suppress fabricated first-party edges
- **base_kinds**: interface satisfaction as IMPLEMENTS facts (Go's implicit interfaces are exactly the thing tree-sitter can never derive)
- **expression_types**: declared/inferred types feeding `parsers/go/type_inference.py`

Later (separate issue once this lands): `go/ssa` def-use chains to upgrade Go FLOWS_TO from lexical lean-walk to compiler-verified flow.

## Enablement and degradation

- `GO_FRONTEND: auto | treesitter | gotypes` — auto activates iff `go.mod` present and `go` binary found.
- Load errors (missing deps, build tags) degrade per-package to heuristics; unresolved imports become external proofs.
- Register with `parser_fingerprint.py`.

## Tests

Fixture Go repo exercising: interfaces with multiple implementors, embedded structs with promoted methods, shadowed names across scopes, method values passed as arguments, vendored/external calls. Compare against `run_go_oracle` where applicable.

[...]

Issue #1229, opened 2026-08-12
Semantic frontends: incremental/watch staleness + rehydration alias (C# and Go)


vitali87:
Follow-up from PR #1228 review (Greptile 2×P1 + CodeRabbit Major). A **shared, pre-existing** limitation of the location-keyed semantic frontends — it affects the merged C# Roslyn frontend identically, so it should be fixed uniformly for both, not divergently in the Go PR.

## The limitation

`_run_csharp_frontend` and `_run_go_frontend` are invoked **only** in the full `run()` path (`graph_updater.py`), never in the realtime/incremental update path. Pass 3 recomputes CALLS edges for `self._parsed_files` only. Two consequences when semantic facts are location-keyed:

1. **Stale edges on incremental / watch.** If file X changes the compiler binding of a call in unchanged file Y (a new `B.M` shadowing `A.M`, a promoted-method target moving), Y is not reprocessed and its old fact-driven edge survives until a full re-index. The frontend itself is not re-run on the realtime path, so Pass 3 consults facts from the last full build.

2. **Rehydration drops the Go name-token alias (Go-specific sub-case).** A fresh process doing incremental work rehydrates `function_locations` from the persisted graph at the **func-node** span; PR2's Go **name-token** alias (registered only during a live Pass-2 parse) is not reconstructed, so `_declared_location` misses and the call degrades to the tree-sitter heuristic. C# is unaffected because Roslyn's target position coincides with the func-node span the graph stores.

## Safety today

Both degrade to the **tree-sitter baseline** on a miss — never worse than not having the frontend, and a full re-index is always correct. This is the same posture the C# frontend shipped with.

## Fix (uniform, both frontends)

- Re-run the applicable frontend on the realtime path, or invalidate + reprocess the affected callers when semantic facts can change (reprocess all callers in the language when any file in a fact-carrying module changes).
- For the Go rehydration sub-case: either reconstruct the name alias on rehydration (needs the target file re-parsed), or have the `gotypes` tool emit the callee's **declaration start** (func keyword) instead of `fn.Pos()` (the name ident) so the target joins the standard `function_span_key` with no alias — which also removes the rehydration gap entirely. The latter is the cleaner long-term shape.
- Add incremental/watch regression coverage for a cross-file binding change.

vitali87:
[...]

PR #1345, opened 2026-08-20
feat: bind Java calls through the javac facts and suppress proven-external edges

### LINKED ISSUE #1181: Java semantic frontend via the javac Compiler Tree API
## Why

Java's current resolution is the largest pure-heuristic stack in the repo — `parsers/java/method_resolver.py` (745 LOC), `type_resolver.py` (304), `variable_analyzer.py` (560) — matching by name and arity where the language demands overload resolution *by argument type*, generic instantiation, and inner/anonymous-class scoping. The JDK ships the authoritative answer: the Compiler Tree API (`com.sun.source.*`), and `evals/oracles/java_oracle/Oracle.java` already uses it.

## Plan

A bundled Java tool (built once, same pattern as the Roslyn tool) that runs `JavacTask` parse + attribute over repo sources and emits facts per the `LanguageFrontend` protocol:

- **resolved_call_sites**: the attributed `ExecutableElement` per invocation — overloads by argument type, varargs, boxing, static imports, lambdas and method references (`String::valueOf`), anonymous and inner classes
- **external_sites**: symbols attributed to classpath/JDK targets → suppress fabricated first-party CALLS
- **base_kinds**: `extends` vs `implements` by resolved symbol, including generics (`class A extends B<C>`)
- **expression_types**: attributed types feeding `java/type_inference.py`, replacing heuristic generic-erasure guessing

## The classpath question (main friction)

Stage it:

1. **No-dependency attribute** first: `javac` parse+attribute *without* the project classpath is still exact for intra-repo binding; every unresolvable symbol becomes an external proof. This alone covers the graph's first-party precision goal.
2. **Classpath capture** later, only if stage 1's external-proof rate on real projects shows it matters: either lightweight Maven/Gradle interrogation, or adopting `scip-java` (which solves build-tool integration and emits a SCIP index — sharing the SCIP→facts adapter built for the Rust issue).

## Synergy

Annotation processing is how #1140 (Lombok/Dagger generated code) becomes tractable: running attribute with processors on and exposing generated members as `invisible_constructs` facts. Keep it out of scope here but design the fact schema so generated members fit.

## Enablement, degradation, tests

- Auto iff `pom.xml`/`build.gradle`/`.java` sources and a JDK are found; per-file fallback to heuristics on attribute errors; `parser_fingerprint.py` registration.
- Fixture Java repo exercising: overload sets differing only by parameter type, generic methods, inner + anonymous classes, static imports, method references, a call into an external library (must become an external proof, not a first-party CALLS).

Depends on the `LanguageFrontend` protocol issue.

vitali87:
Closes #1181. PR2 of two: #1341 shipped the javac fact provider, this makes the resolver consume it. Same two-step rollout the Go frontend used in #1179.

## The join had to be fixed first

The facts key a call target at the callee NAME token, but `function_span_key` keys a Java method at the DECLARATION start, which is the modifiers: `    public String handle(` records col 4 while the tool reports the col of `handle`. Same line, different column, so every single fact would have missed. That is the failure mode worth naming, because it is silent: the heuristics answer instead, the tests look green, and the frontend appears to work while contributing nothing.

`_register_java_name_alias` registers a SECOND `function_locations` entry at the name token, mirroring `_register_go_name_alias`, which exists for exactly the same asymmetry.

## A latent bug in FunctionLocations

The alias writes through `setdefault` (the span record must win if two methods share a line). `FunctionLocations` overrides `__setitem__` and `update` to maintain its module index, and its docstring argues the index belongs to the map rather than to each of nine write sites, but `setdefault` was not overridden, and `dict.setdefault` bypasses `__setitem__` exactly like `dict.update` does. A record written through it would have been invisible to `drop_module` and survived an incremental re-parse: the stale-record bug (#1019) this class exists to prevent. Now overridden, with tests.

## Consumption

`resolve_java_call_site` on the Java engine: a HIT resolves the target through the shared `declared_location` join, and a proven-external site returns the sentinel so no first-party edge is fabricated. Any miss returns None and the existing Java resolver answers, so the frontend can only add precision, never remove a resolution it has no opinion on. The facts load after Pass 2 (the alias index has to exist first) and the maps are cleared at the start of each run, so a reused updater in watch mode does not keep applying stale facts after the setting flips.

Note the hook placement: the Java branch in `call_processor` sits EARLIER in the dispatch chain than the Go and Python ones, so the semantic path had to go inside it. My first attempt appended a branch in the Go/Python position and it was simply unreachable.

## Tests

Every assertion is paired with the same repo indexed with the frontend OFF, so a pass cannot come from the fallback:

- **Overload by argument type**: `handle(Object)` vs `handle(CharSequence)` called with a `String`. Java picks the most specific applicable overload; the argument type matches NEITHER parameter name, so name-and-arity matching takes the first declaration. Frontend on binds `CharSequence`, frontend off binds `Object`.
- **External proof**: a static import of `java.lang.String.valueOf` with a same-named local static in the file. The call is unqualified, so the module-wide name scan finds the local one; javac knows it resolved to the JDK. Frontend on emits no edge, frontend off fabricates `Decoy.valueOf(int)`.

I dropped a first pair of external tests that used a qualified `String.valueOf(42)`: the heuristics already emit nothing there, so the test would have passed without proving anything.

Batteries: 700 passed across java, function_locations, frontends and fingerprint.

## Summary by CodeRabbit

* **New Features**
  * Improved Java call resolution using compiler-derived type information.
  * Added accurate handling for overloaded methods and external static calls.
  * Java processing now falls back to heuristic resolution when compiler information is unavailable.
  * Real-time updates now reprocess changed Java files automatically.

* **Bug Fixes**
  * Prevented incorrect repository links for external Java methods.
  * Improved function location tracking and cleanup.

* **Tests**
  * Added coverage for Java resolution, external calls, real-time updates, and function locations.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -548,6 +548,35 @@ class GraphUpdater:
             )
         )
 
+    def _run_java_frontend(self) -> None:
+        # The javac frontend (issue #1181) runs AFTER Pass 2, like the Jedi one:
+        # its call facts join against the function_locations Pass 2 just filled,
+        # including the name-token alias the Java method registration adds.
+        # Reset first so a reused updater (watch mode) that previously ran the
+        # frontend does not keep applying stale facts on a later run with it
+        # off; the maps are mutated in place because the type-inference engine
+        # holds a reference.
+        dp = self.factory.definition_processor
+        dp.java_call_sites.clear()
+        dp.java_external_sites.clear()
+        if settings.JAVA_FRONTEND == cs.JavaFrontend.HEURISTIC:
+            return
+        frontend = FRONTENDS.get(cs.SupportedLanguage.JAVA)
+        if frontend is None or not frontend.applies(self.repo_path):
+            return
+        if not frontend.available():
+            logger.warning(ls.JAVA_FRONTEND_UNAVAILABLE)
+            return
+        facts = frontend.run(self.repo_path, ())
+        dp.java_call_sites.update(facts.resolved_call_sites)
+        dp.java_external_sites.update(facts.external_sites)
+        logger.info(
+            ls.JAVA_FRONTEND_FACTS.format(
+                calls=len(facts.resolved_call_sites),
+                externals=len(facts.external_sites),
+            )
+        )
+
     def _run_python_frontend(self) -> None:
         # In-process Jedi facts (issue #1183): exact first-party callees
         # through re-exports/decorators and external proofs for calls leaving
@@ -883,6 +912,10 @@ class GraphUpdater:
         # calls against the function_locations Pass 2 just filled) and needs
         # the parsed-file list, which Pass 2 produced (issue #1183).
         self._run_python_frontend()
+
+        # Same posture for Java (issue #1181): the facts resolve against the
+        # method name-token locations Pass 2 registered.
+        self._run_java_frontend()
 
         # HYBRID must run after Pass 2: an incremental run deletes each
         # changed file's Module subtree before re-parsing it, so macro
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -934,6 +934,9 @@ JAVA_FRONTEND_RUN_FAILED = "javac frontend did not finish ({error}); using tree-
 JAVA_FRONTEND_PARSE_FAILED = (
     "javac frontend emitted unreadable output (stdout={stdout}, stderr={stderr})"
 )
+JAVA_FRONTEND_UNAVAILABLE = (
+    "JAVA_FRONTEND is set to javac but no working JDK was found; using tree-sitter"
+)
 JAVA_FRONTEND_FACTS = (
     "javac facts: {calls} resolved call sites, {externals} external sites"
 )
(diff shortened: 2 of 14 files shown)


vitali87:
@greptileai Confirmed and fixed in a529329d. The watch path re-runs the applicable semantic frontend per changed language (issue #1229 phase 3) and Java was not wired in, so a `.java` edit recomputed CALLS against the previous run's facts. Since those facts are keyed by (file, line, byte col), an edit that shifted a call by a single byte would keep binding it through a stale position, and a stale external proof would keep suppressing an edge that had become first-party.

`realtime_updater` now has a JAVA arm alongside Go, C# and Python: `_run_java_frontend()` (which clears its own maps first) then `_rehydrate_function_locations()`, because a watch process that did not perform the full build has no in-memory locations for unchanged files. `test_java_change_reruns_the_javac_frontend` covers it and was verified red.

On the blocked T-Rex validation: `javac`/`java` are genuinely absent in your sandbox, and every end-to-end test in this PR is gated on `java_frontend_available()`, so they skip rather than fail there. I ran them locally against JDK 17: 9 passed in the Java semantic and function-locations files, 700 across the java, function_locations, frontends and fingerprint batteries.

PR #1349, opened 2026-08-20
fix: type a nested Java call argument from its selected overload

### LINKED ISSUE #1348: Java: a nested call argument is typed by name, so an overloaded inner call picks the first declaration
## Summary

`_infer_java_method_return_type` types a nested call by NAME only (`_resolve_java_method_return_type` builds a `receiver.method` string), so an overloaded nested call is typed from whichever declaration comes first. `take(make("x"))` can be typed from `make(int)` and bind the wrong `take`.

Raised by CodeRabbit on PR #1346. It is the same defect Greptile found on the chained-receiver path in PR #1343, which was fixed there by resolving the inner call and reading the return type back off the SELECTED signature (`_declared_return_type_of`). The argument path still has it.

## Reproduction

```java
class Wrong { }
class Right { }
class Factory {
    public Wrong make(int n) { return new Wrong(); }
    public Right make(String s) { return new Right(); }
    public void take(Wrong wrong) { }
    public void take(Right right) { }
}
public class Main {
    public static void main(String[] args) {
        Factory factory = new Factory();
        factory.take(factory.make("x"));
    }
}
```

Actual: `take(Wrong)`. Expected: `take(Right)`. The inner `make(String)` edge itself is already correct, so only the argument TYPE is wrong.

## Why it did not land with #1346

The obvious fix is to route the nested call through `_do_resolve_java_method_call` and reuse `_declared_return_type_of`, exactly as the chained-receiver path does. I tried that and it does not work from where the code sits: `_infer_java_type_from_expression` and `_infer_java_method_return_type` do not receive `local_var_types`, and passing an empty map means the RECEIVER cannot be typed (`factory` is a local), so the inner resolution returns None and nothing improves.

Doing it properly means threading `local_var_types` through the expression-typing path, which touches every caller of `_infer_java_type_from_expression` across the variable analyzer. That is a wider change than the literal-typing fix it was found in, so it belongs on its own.

## Scope

- Thread `local_var_types` into `_infer_java_type_from_expression` / `_infer_java_method_return_type` (or give the engine a call-scoped current-scope map).
- Resolve the nested call through `_do_resolve_java_method_call`, take the return type from the resolved signature via `_declared_return_type_of`, keep the name-based lookup as the fallback.
- Watch for recursion: argument typing already calls into resolution, and resolution calls back into argument typing. `_find_inherited_method` uses `@recursion_guard`; the same treatment probably applies here.
- Regression: the reproduction above, plus a nested call whose receiver is a field rather than a local.

vitali87:
Closes #1348. Raised by CodeRabbit on #1346, filed rather than bolted on, fixed here.

## The bug

`_infer_java_method_return_type` typed a nested call by NAME (`_resolve_java_method_return_type` builds a `receiver.method` string), so an overloaded inner call contributed the FIRST declaration's return type. `take(make("x"))` was typed from `make(int)` and bound `take(Wrong)`, even though the inner `make(String)` edge itself was already correct.

Same defect the chained-receiver path had in #1343, fixed the same way: resolve the inner call, then read the return type back off the SELECTED signature via `_declared_return_type_of`.

## Why it needed its own PR

It could not be done from where the code sat. `_infer_java_type_from_expression` and `_infer_java_method_return_type` never received `local_var_types`, so the receiver of an instance call (`factory`, a local) could not be typed, the inner resolution returned None, and nothing improved. This threads the caller's variable types through both, which is the change that touches every caller and is why it did not ride along with the literal-typing fix.

`_resolve_nested_java_call` carries a `@recursion_guard`. To be precise about what that is for: argument typing calls resolution and resolution types arguments, so the descent is mutual. I did not observe an infinite recursion (each step moves to a child node), so the guard is defensive rather than a fix for an observed hang; `test_a_deeply_nested_argument_chain_terminates` exercises three levels.

## A mistake worth recording in the diff

My first attempt declared `_do_resolve_java_method_call` and `_declared_return_type_of` as `@abstractmethod` stubs on `JavaVariableAnalyzerMixin`, matching how `JavaMethodResolverMixin` declares ITS cross-mixin dependencies. That broke 72 tests: the MRO is (TypeResolver, VariableAnalyzer, MethodResolver), so a stub in VariableAnalyzer PRECEDES the real implementation and wins, resolving every Java call to None. The direction matters -- only a mixin that comes later may declare abstracts for methods implemented earlier. They are annotation-only declarations now, with a comment saying why, since an annotation creates no runtime attribute.

## Tests

`test_java_nested_call_arguments.py`, verified red first: the issue's `make(int)`/`make(String)` reproduction, its mirror with the int argument selecting the other target (so the pick follows the argument, not the declaration order of `make`), the same shape with a FIELD receiver rather than a local, and the three-level chain.

Java battery: 689 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Java method resolution for nested calls and overloaded methods.
  * Nested call arguments now use local-variable and field types to select the correct overload.
  * Improved return-type inference across nested method calls and field-access chains.
  * Prevented unbounded recursion when resolving deeply nested method-call chains.

* **Tests**
  * Added coverage for nested overloads, field-based receivers, deeply nested calls, and locally declared variables.

### DIFF
--- a/codebase_rag/constants/ast_python.py
+++ b/codebase_rag/constants/ast_python.py
@@ -151,3 +151,4 @@ TYPE_INFERENCE_BASE_MODEL = "BaseModel"
 
 ATTR_TYPE_INFERENCE_IN_PROGRESS = "_type_inference_in_progress"
 GUARD_INHERITED_METHOD = "_inherited_method_guard"
+GUARD_NESTED_JAVA_CALL = "_nested_java_call_guard"
--- a/codebase_rag/parsers/java/method_resolver.py
+++ b/codebase_rag/parsers/java/method_resolver.py
@@ -204,7 +204,10 @@ class JavaMethodResolverMixin:
 
     @abstractmethod
     def _infer_java_type_from_expression(
-        self, expr_node: ASTNode, module_qn: str
+        self,
+        expr_node: ASTNode,
+        module_qn: str,
+        local_var_types: dict[str, str] | None = None,
     ) -> str | None: ...
 
     @abstractmethod
@@ -748,7 +751,9 @@ class JavaMethodResolverMixin:
                 # or a call carries it in the expression: the shared inference
                 # types all of them (issue #1344).
                 arg_types.append(
-                    self._infer_java_type_from_expression(child, module_qn)
+                    self._infer_java_type_from_expression(
+                        child, module_qn, local_var_types
+                    )
                 )
                 continue
             name = safe_decode_text(child)
--- a/codebase_rag/parsers/java/variable_analyzer.py
+++ b/codebase_rag/parsers/java/variable_analyzer.py
@@ -7,6 +7,7 @@ from loguru import logger
 
 from ... import constants as cs
 from ... import logs as ls
+from ...decorators import recursion_guard
 from ...types_defs import ASTNode
 from ..utils import safe_decode_text
 from .utils import (
@@ -159,7 +160,7 @@ class JavaVariableAnalyzerMixin:
 
         if value_node := declarator_node.child_by_field_name(cs.FIELD_VALUE):
             if inferred_type := self._infer_java_type_from_expression(
-                value_node, module_qn
+                value_node, module_qn, local_var_types
             ):
                 resolved_type = self._resolve_java_type_name(inferred_type, module_qn)
                 local_var_types[var_name] = resolved_type
@@ -229,7 +230,7 @@ class JavaVariableAnalyzerMixin:
             return
 
         if inferred_type := self._infer_java_type_from_expression(
-            right_node, module_qn
+            right_node, module_qn, local_var_types
         ):
             resolved_type = self._resolve_java_type_name(inferred_type, module_qn)
             local_var_types[var_name] = resolved_type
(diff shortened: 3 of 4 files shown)


vitali87:
@greptileai Confirmed and fixed in ec6f36a1, and thank you for running it rather than just reading it.

`_infer_java_type_from_expression` now forwards the caller-local map into `_infer_java_field_access_type` and on into `_resolve_field_access_base_type`, which prefers it over the module-wide lookup. I extended the same treatment to the bare `TS_IDENTIFIER` case in the same switch, which had the identical flaw: `_lookup_variable_type` is name-keyed across every method in the module, so it cannot tell two methods' same-named variables apart.

Worth recording how the reproduction behaves, because my first attempt at your scenario PASSED and I nearly wrote the finding off. Two methods declaring `holder` with different types is not sufficient on its own: with the conflicting method declared FIRST, the module-wide map ends up holding the right type anyway and the bug hides. It only surfaces when the other method comes LATER and its declaration is the last writer. The committed test puts `run()` before `other()` for exactly that reason, and it is red without the change.

Java battery: 690 passed.

Issue #1344, opened 2026-08-20
Java: literal arguments are untyped, so all-literal calls pick overloads by arity alone


vitali87:
## Summary

Java overload selection ignores literal arguments. `_infer_arg_types` types identifiers whose local variable type is known, but a literal argument contributes no type, so a call whose arguments are all literals falls back to arity-only matching and binds to whichever same-arity overload the class declares first.

Found while fixing #1334 (PR #1343): the first draft of the overload regression test passed the argument as a literal and bound to the wrong overload for this reason, so the committed test hoists it into a typed local instead.

## Reproduction

```java
class Wrong {
    public void selectedOnly() { }
}
class Selected {
    public void selectedOnly() { }
}
class Factory {
    public Wrong create(Integer wrongArg) { return new Wrong(); }
    public Selected create(String selectedArg) { return new Selected(); }
}
public class Main {
    public static void main(String[] args) {
        Factory factory = new Factory();
        factory.create("pick me").selectedOnly();
    }
}
```

Actual: `create(Integer)` (first same-arity declaration), and the chain then types from `Wrong`, so the edge is `Wrong.selectedOnly()`.

Expected: `create(String)` and `Selected.selectedOnly()`.

Hoisting the literal into `String selectedArg = "pick me";` resolves both correctly today, which is what isolates this to literal typing rather than to the overload pick or the chain typing.

## Scope

The Java grammar gives each literal its own node type, so the mapping is direct and needs no inference: `string_literal` -> String, `character_literal` -> char, `true`/`false` -> boolean, `null_literal` -> the null type (matches any reference parameter, so it must NOT be allowed to exclude a candidate), and the numeric literals carry their type in the suffix (`1` -> int, `1L` -> long, `1.0` -> double, `1.0f` -> float).

Two details worth respecting:

- Boxing and widening mean an exact simple-name comparison is too strict: `create(Integer)` is a valid target for an `int` literal, and `create(long)` is a valid target for an `int` literal. The existing `_overload_matches_arg_types` compares simple names, so it needs a small widening/boxing equivalence table rather than raw equality, otherwise adding literal types would REMOVE correct edges that arity-only matching finds today.
- A `null` literal argument must stay untyped (`None`) rather than typed as a class, since it is compatible with every reference parameter.

Because of the first point this is not a pure addition: it should land with tests that cover a literal binding to a boxed parameter and to a widened one, not just the exact-match case.

Issue #1334, opened 2026-08-19
Java: chained method-call receivers resolve by module-wide name scan, not return-type propagation


vitali87:
## Summary

Java chained method calls (fluent/builder style) resolve their CALLS edges by the unqualified-call fallback (a module-wide name scan), not by propagating the receiver's return type through the chain. The exact example from discussion #199 happens to produce correct edges, but only by coincidence: as soon as another class in scope declares a method with the same name, the chain mis-binds.

Follow-up to #199. The other two points raised there (chained field access like `obj.address.zipCode` and mixed chains like `obj.engine.start()`) were fixed in 7774b542 and follow-ups, and are covered by `codebase_rag/tests/test_java_field_access_chains.py`. This is the remaining gap.

## Reproduction

```java
class Decoy {
    public Decoy where(String c) { return this; }
    public void execute() { }
}
class QueryBuilder {
    public QueryBuilder from(String table) { return this; }
    public QueryBuilder where(String condition) { return this; }
    public void execute() { }
}
public class Main {
    public static void main(String[] args) {
        QueryBuilder obj = new QueryBuilder();
        obj.from("users").where("id = 1").execute();
    }
}
```

Actual CALLS edges produced by `GraphUpdater`:

```
proj.src.Main.Decoy.execute()
proj.src.Main.Decoy.where(String)
proj.src.Main.QueryBuilder.from(String)
```

Expected:

```
proj.src.Main.QueryBuilder.from(String)
proj.src.Main.QueryBuilder.where(String)
proj.src.Main.QueryBuilder.execute()
```

`from("users")` resolves correctly because its receiver is the identifier `obj`, whose type is known. `where(...)` and `execute()` have a `method_invocation` node as their receiver, and that case is never typed, so they fall through to the module-wide scan and bind to whichever matching method the scan finds first (`Decoy` here).

## Root cause

`extract_method_call_info` in `codebase_rag/parsers/java/utils.py` matches the receiver (`object_node.type`) against `identifier`, `field_access`, `this`, `super`, `parenthesized_expression`, and `cast_expression`, but not `method_invocation`. For a chained call it returns `object=None`, so `_do_resolve_java_method_call` in `codebase_rag/parsers/java/method_resolver.py` treats the call as unqualified and ends up in `_resolve_static_or_local_method`, which scans the module by name.

[...]

PR #1350, opened 2026-08-20
feat: index SQL routines and resolve calls that name their target in a string

raulgonzalezdev:
## The gap

In codebases that keep business logic in the database, the call graph stops at the database boundary.

`.sql` files are ingested as `File` nodes, but nothing inside them is parsed, so a stored routine never becomes a callable node. And the call sites that reach those routines do not name them syntactically — they pass the name as a **string** to a dispatcher:

```ts
const rows = await callSp('usp_invoice_list', params)   // TypeScript -> PL/pgSQL
```
```python
cursor.callproc('usp_invoice_list')                     # Python DB-API
```
```java
conn.prepareCall("{call usp_invoice_list(?)}")          // JDBC
```

No parser resolves those as calls. The result is that the routine looks unreachable: dead-code reports it, and "what breaks if I change this procedure" has no answer.

Measured on a real ERP monorepo (4.217 tracked files, ~3.700 stored procedures): **1.015 `.sql` files in the graph, 0 symbols parsed inside them, 0 edges from application code to any routine.**

## What this adds

**1. SQL (PostgreSQL) as an indexed language** — through the existing `LanguageSpec` machinery, using `tree-sitter-sql`.

The one non-obvious piece: `create_function` names its routine through an `object_reference` child rather than a `name` field, so the generic extractor returns `None` for every routine. Schema-qualified names (`app.usp_x`) reduce to the last identifier, which is what a caller writes.

Marked **In Development** deliberately: the upstream grammar parses plain SQL statement bodies only (its own TODO notes PL/pgSQL bodies are not modelled), so routine bodies are not walked in depth. Measured coverage on the same monorepo: names extracted for **345 of 373** stored procedures (92.5%); the remaining 28 fail to parse in the grammar itself.

**2. String-call resolution** — a new `parsers/string_call.py`.

The dispatcher is project-specific, so it is **declared, never guessed**, in a `.cgr.toml` at the repository root:

```toml
[[string_calls]]
callee = "callSp"      # matched on the last name segment: db.callSp works too
arg_index = 0          # which argument carries the name
```

Without a declaration this does nothing at all. Only string **literals** resolve — a variable or an interpolated template names its target at runtime, and guessing there would invent edges. The dispatcher's own edge still emits; this only **adds** the edge the syntax hides. `CGR_CONFIG_PATH` points the config elsewhere for read-only or third-party checkouts.

## Result on the same repo

| | Before | After |
|---|---|---|
| Stored routines as callable nodes | 0 | **2.288** |
| `CALLS` edges from application code to routines | 0 | **230** |

## Two bugs worth calling out

Both are shaped like traps for anyone adding a language, so the fixes carry regression tests:

- **A `LanguageSpec` alone is not enough.** `_language_imports()` is an explicit list; a language missing from it has its files walked but never parsed, with no error anywhere. The first full ingestion produced zero SQL nodes and reported success.
- **Naming a node type the grammar does not define fails the whole language at load time** (`Invalid node type ...: create_procedure`) rather than degrading. The grammar's `main` branch has `create_procedure`; the published package (0.3.11) does not. A test now asserts every declared node type exists in the installed grammar.

## Tests

25 new tests, `ruff check` clean over `codebase_rag/`:

- `test_sql_language.py` — registration in both spec maps, name extraction (plain, `OR REPLACE`, schema-qualified), and the guard that every declared node type exists in the installed grammar.
- `test_string_call.py` — dispatcher forms that occur in real code (awaited, through a receiver, schema-qualified, non-default `arg_index`) and, just as important, what must **not** resolve: a variable, an interpolated template, an undeclared dispatcher, a missing argument, malformed config.

## Notes

No issue number: this started as an evaluation of the tool on a PL/pgSQL-heavy codebase and the gap turned out to be general rather than specific to us. Happy to open one and split this into two PRs (language, then resolution) if you prefer that shape — or to adjust the config surface, which is the part most likely to want a different design.

## Summary by CodeRabbit

- **New Features**
  - Added in-development PostgreSQL/SQL language support, including `.sql` file recognition and routine name extraction.
  - Added configurable detection of string-based routine dispatch calls through project configuration.
  - Resolved direct, awaited, schema-qualified, and argument-indexed string call targets.

- **Bug Fixes**
  - Safely disables string-call detection for missing or invalid configuration.

- **Tests**
  - Added coverage for SQL parsing and configurable string-call resolution.


vitali87:
@raulgonzalezdev thanks for the PR. Please address ALL review comments by the bots.

vitali87:
Thanks for this, and for the way it is written up: measured numbers on a real repo, an explicit statement of what deliberately does NOT resolve, and two incidental bug findings carried with regression tests. That is more rigour than most feature PRs arrive with.

I am reviewing rather than merging, so treat everything below as input for you to weigh, not instructions.

## The part I would keep exactly as it is

Literal-only resolution, and nothing at all without a declaration. A variable or an interpolated template names its target at runtime and guessing there would invent edges. This repo has been bitten repeatedly by resolution that is right by coincidence -- a module-wide name scan that binds to whichever same-named thing it finds first (#1334), an argument matcher that picks the first same-arity overload (#1344). A fabricated edge is far worse than a missing one, because nothing downstream can tell it apart from a real one. Your design gets that right.

The two bugs you found are real and I would take them even if the rest of the PR went nowhere:

- `_language_imports()` being an explicit list, so a language with a `LanguageSpec` gets its files walked and never parsed, silently reporting success. Silent-success is the worst failure mode this codebase has; I have hit the same class three times this week in the semantic frontends, where a location join missed and the heuristics answered instead, leaving every test green while the frontend contributed nothing.
- The node-type validation test. Naming a node type the installed grammar does not define failing the WHOLE language at load time is exactly the kind of thing that should be asserted once rather than rediscovered.

## The design question worth settling before this lands: the config surface

`.cgr.toml` introduces a THIRD configuration mechanism. Today the repo has pydantic settings via `.env`/environment (`codebase_rag/config.py`) and `.cgrignore` for path patterns. A new root-level TOML is not wrong, but it is a decision with a long tail: precedence when both a config file and env exist, what happens in a monorepo with nested projects, and whether `CGR_CONFIG_PATH` should instead be a CLI flag for consistency with how `--exclude` complements `.cgrignore`.

Since you flagged this yourself as the part most likely to want a different design, my suggestion is to get @vitali87's call on the mechanism BEFORE polishing it, rather than after. It is the one piece where a maintainer preference would force a rewrite of otherwise finished code.

## On splitting the PR

Yes, I would split it, and your instinct about the seam is right. The repo has a precedent that matches almost exactly: the Go semantic frontend shipped as #1179 (fact provider) then a separate resolver-consumer PR, and Java just did the same across #1341 and #1345. Concretely:

- **PR1: SQL as an indexed language.** Self-contained, valuable alone (routines become searchable nodes), and its correctness argument is entirely "does the grammar name things properly". The `_language_imports()` fix and the node-type guard belong here.
- **PR2: string-call resolution.** Depends on PR1 for anything to point at, and it is where the config-surface discussion lives, so it does not block PR1 from landing.

That also makes the "In Development" marking easier to reason about: PR1 can say plainly that bodies are not walked because the upstream grammar does not model PL/pgSQL bodies, without that caveat clouding the resolution work.

## Smaller things

- 92.5% name extraction with 28 files failing IN THE GRAMMAR is worth an explicit note in the code, not only the PR description -- ideally a log line when a `create_function` yields no name, so a user with a low hit rate can tell it is the grammar rather than their schema.
- CodeRabbit and Greptile have six findings between them on the current diff; worth clearing those before the split so the work is not done twice.

An issue would help even though the code came first: it gives the config-surface discussion somewhere to live that is not a diff thread. Happy to keep reviewing as this evolves.

vitali87:
Adpoting this due to slow pace of development!

vitali87:
Adopted in #1394: rebased onto main with your commits and authorship preserved, and all six bot findings fixed on top (schema-qualified names now stay qualified end to end, JDBC {call ...} unwrapping, the macOS arm64 bundle collecting the SQL grammar, CREATE PROCEDURE remnants, edge labels, boolean arg_index). Thanks @raulgonzalezdev for the excellent groundwork — this PR will be closed once #1394 lands.

vitali87:
Landed via #1394 with all review findings addressed — thanks again @raulgonzalezdev, your commits and authorship are preserved in main.

Issue #1353, opened 2026-08-21
Follow-up to #1351: complete the review that the merge skipped (ref write-back)


vitali87:
Follow-up to #1351 (slice 2 of #1187), which was merged before its review was actually complete. Filing rather than reopening, since the PR is merged.

## What went wrong at merge time

The merge gate checked "0 pending checks", which is not the same question as "the reviewers have reviewed THIS commit". Measured after the fact:

| bot | last commit reviewed | PR head at merge |
|---|---|---|
| greptile-apps | `ee14ed4a` (commit 1 of 6, 22:47Z) | `7e3e10ad` |
| coderabbitai | `7e3e10ad`, but that entry is a REPLY 23s after mine, not a fresh pass | `7e3e10ad` |

So Greptile never reviewed five of the six commits, including the whole of `RefParameterIsWritten` -- which exists BECAUSE of Greptile's own P1 finding on commit 1. The fix for its finding was never seen by it.

CodeRabbit produced a new, genuine defect on every single push (six findings across five rounds). Merging eight minutes after resolving the fifth, with the sixth-round base rate that high, was not justified by a green checkmark.

Three threads were also resolved by me while they carried an open question I had asked the reviewer ("say the word and I will add the test", "tell me if you would rather use `IArgumentOperation`", "say so if the workspace threading belongs here"). Resolving your own question and merging is not consensus.

## Unfinished work

1. **Local-function `ref` callees are untested.** `CallableBody` handles `LocalFunctionStatementSyntax`, block- and expression-bodied, but no test exercises it, so the path is unproven. CodeRabbit asked for this test and I declined it as unrealistic; that judgement should be re-made by someone else, or the branch removed if the shape really is not worth supporting.

2. **Cross-project callee bodies silently produce no fact.** A `ref` parameter written in a REFERENCED project cannot be proven: `FactCollector` holds only a per-document `SemanticModel`, and asking this compilation for a model of a tree it does not own throws. It is guarded with `ContainsSyntaxTree` and the callee is treated as read-only. That is the conservative direction (a missed edge, not a fabricated one), but it means multi-project solutions -- the common case in real C# -- lose `ref` write-back across project boundaries entirely. Fixing it needs a workspace or compilation map threaded into the collector.

[...]

PR #1355, opened 2026-08-21
feat: prove ref writes for callee bodies in referenced projects

vitali87:
Part of #1353, item 2. Closes the cross-project gap that #1351 left guarded rather than solved.

## The gap

A `ref` parameter written by a callee in a REFERENCED project could not be proven. `RefParameterIsWritten` needs a semantic model for the callee body, that body belongs to another compilation, and asking the caller's compilation for a model of a tree it does not own THROWS. #1351 guarded it with `ContainsSyntaxTree` and treated such a callee as read-only -- safe, but it means multi-project solutions, which is most real C#, lost `ref` write-back across every project boundary.

## The change

`Main` now materializes every project's `Compilation` first and indexes each syntax tree by its owner, then hands that index to `FactCollector`. `ResolveOwner` returns the caller's own compilation when it contains the tree and the owning one otherwise, so a cross-project callee is provable. A tree from outside the solution has no owner and stays unprovable, which keeps the conservative answer for genuinely external callees.

On memory: the workspace already caches compilations behind its solution snapshot, so holding the list adds little over building them one at a time. Worth a second opinion if you disagree -- it is the one judgement call in this diff.

## Verified in both directions

A two-project fixture (`App` with a `ProjectReference` to `Lib`, the `ref` callee in `Lib`):

- with the old `ContainsSyntaxTree` guard: `out_writes` is `{}` -- the fact is dropped
- with the owner index: `{('App/Program.cs', 11, 15, 'Fill'): {1: 'sink'}}`

`test_ref_write_back_across_a_project_reference` builds that fixture and asserts the ENV -> STDOUT edge end to end. 18 passed in `test_csharp_semantic_flow.py`.

## Remaining on #1353

Item 3, `IArgumentOperation` vs syntax-level argument binding, is a reviewer preference rather than a correctness question and I am deliberately not deciding it alone. Item 4 is a full adversarial pass over the helper, which commits 2-6 of #1351 never received.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved C# analysis across referenced projects, allowing `ref` parameter write-back flows to be tracked correctly.
  * Fixed cases where data flow between project boundaries was incorrectly omitted when the relevant method was defined in another project.

* **Tests**
  * Added coverage for cross-project C# data flow, including flows from environment values to standard output.

### DIFF
--- a/codebase_rag/parsers/csharp_frontend/roslyn/Frontend.cs
+++ b/codebase_rag/parsers/csharp_frontend/roslyn/Frontend.cs
@@ -56,17 +56,36 @@ public static class Frontend
             .Select(p => p.AssemblyName)
             .Where(n => !string.IsNullOrEmpty(n))
             .ToHashSet(StringComparer.OrdinalIgnoreCase);
-        var collector = new FactCollector(rootFull, IgnoredDirs(), firstPartyAssemblies);
+        // Materialize every compilation FIRST and index each syntax tree by its
+        // owner. A callee body can live in a REFERENCED project, and asking the
+        // caller's compilation for a model of a tree it does not own throws, so
+        // without this index a cross-project `ref` write cannot be proven and
+        // the fact is silently dropped (issue #1353). The workspace already
+        // caches these compilations behind its solution snapshot, so holding
+        // the list adds little over building them one at a time.
+        // GetCompilationAsync runs source generators, so symbols that resolve
+        // only through generated members still bind; the generated trees
+        // themselves have no first-party path and never become facts.
+        var compilations = new List<Compilation>();
         foreach (var project in projects)
         {
-            // GetCompilationAsync runs source generators, so symbols that resolve
-            // only through generated members still bind; the generated trees
-            // themselves have no first-party path and never become facts.
-            var compilation = await project.GetCompilationAsync();
-            if (compilation is null)
+            if (await project.GetCompilationAsync() is { } built)
             {
-                continue;
+                compilations.Add(built);
             }
+        }
+        var treeOwners = new Dictionary<SyntaxTree, Compilation>();
+        foreach (var built in compilations)
+        {
+            foreach (var tree in built.SyntaxTrees)
+            {
+                treeOwners.TryAdd(tree, built);
+            }
+        }
+        var collector = new FactCollector(
+            rootFull, IgnoredDirs(), firstPartyAssemblies, treeOwners);
+        foreach (var compilation in compilations)
+        {
             foreach (var tree in compilation.SyntaxTrees)
             {
                 var path = tree.FilePath;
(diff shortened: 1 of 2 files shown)


vitali87:
@coderabbitai took your non-blocking suggestion rather than leaving it as a follow-up, since "would improve test precision" tends not to come back around once a PR is merged.

`test_cross_project_ref_emits_the_write_fact_itself` (c525dbae) asserts the fact the frontend actually produces -- argument index 1 of `Fill` writes `sink` -- alongside the existing end-to-end edge assertion. The distinction matters here more than usual: an end-to-end edge can in principle survive for a reason unrelated to the change under test, and this PR has already been through one round of exactly that failure mode on #1351, where a test passed without the fix it was supposed to be proving.

@greptileai on the unavailable tooling: I ran the suite locally against .NET 8.0.130 and Python 3.12 -- 19 passed in `test_csharp_semantic_flow.py`, including the cross-project fixture you flagged. I also verified the change in both directions before opening this: with the previous `ContainsSyntaxTree` guard the fixture yields `out_writes: {}`, and with the owner index it yields `{("App/Program.cs", 11, 15, "Fill"): {1: "sink"}}`. So the capability gain is measured rather than argued.

Note the head has moved to c525dbae, so this needs a fresh pass rather than carrying forward the verdict on cdc4be36.

vitali87:
@coderabbitai review

Requesting an explicit pass on `c525dbae`. Greptile has a check-run attached to that SHA (started 03:07:20Z, after the 03:06:37Z push, concluded success), but I cannot find one for CodeRabbit on the new head -- the most recent CodeRabbit activity here is a reply 18 seconds after my comment, which is an acknowledgement rather than a review of the diff.

Asking rather than assuming, because that exact ambiguity is what went wrong on #1351: I read "no pending checks" as "the reviewers reviewed this commit", merged, and only afterwards found Greptile had never seen five of the six commits -- including the code written in response to its own finding. The new commit here is small (a direct assertion of the `out_writes` fact, per your suggestion), but "small" is not a reason to skip the confirmation.

vitali87:
@coderabbitai review
@greptileai review

Requesting an explicit pass on `69644745`. Neither of you has a check-run attached to that SHA (26 CI check-runs are, all green), and it has been ~20 minutes, versus the ~6 minutes Greptile took on the previous head. The pass currently shown in the checks column is a PR-level status carried over from `c525dbae`.

The new commit is a pure test refactor -- manual save/restore of a setting replaced with the `monkeypatch` fixture, which is what Sonar S8997 asked for -- so I expect nothing here. I am asking anyway because "it is only a test change" is not the criterion I want to be using: on #1351 I merged on a green checks column while Greptile was five commits behind, and the whole point of the follow-up work in #1353 is to not repeat that.

If your position is that the head is already covered, say so and name the SHA -- that is a confirmation I can act on. Silence is not.

Issue #1363, opened 2026-08-21
Lean-walk languages lose pass-through return taint (param-to-return summaries missing)


vitali87:
Found while testing the C# type-test pattern fix (#1362), and confirmed NOT to be a pattern problem.

## Reproduction

```csharp
public class Leaky
{
    private string Forward(string v) { return v; }

    public void Run()
    {
        var secret = Environment.GetEnvironmentVariable("K");
        var t = Forward(secret);
        Console.WriteLine(t);        // no ENV -> STDOUT edge
    }
}
```

`Forward` returns its parameter unchanged, so `t` holds the tainted value, but no flow edge is emitted. Verified against `main`.

## Why this is filed separately

I hit it while writing a regression test for #1362 and initially assumed the pattern binding was at fault. Probing the plain declaration form showed the same result, so patterns are not implicated -- I removed the assertion from that PR rather than let a test claim ownership of a gap it does not own, and left a comment there pointing here.

## Scope

#1168 added parameter-to-return taint for pass-through helpers, so the machinery exists; this shape is not reaching it. Worth checking first:

- whether the summary is computed for a method whose body is a bare `return v;` with no intervening statement
- whether the CALLER is walked before the callee's summary is available (an ordering dependency would explain why some pass-through shapes work and this one does not)
- whether it is C#-specific: the same shape should be probed in Python and Go, since #1168 was implemented across the lean walk generally

I have not investigated beyond establishing that patterns are innocent, and I would rather it be scoped by someone who knows why #1168's summaries fire where they do than guess at it.

vitali87:
Scope correction, measured. This is **not C#-specific and not a bug in a C# path** -- it is a structural gap in the LEAN walk, and the title I gave this issue is wrong.

## Same shape, three languages

| language | walk | ENV -> STDOUT |
|---|---|---|
| Python | deep | **emitted** |
| C# | lean | missing |
| JavaScript | lean | missing |

Python is the only language on the deep walk, and it is the only one that gets this. So every lean-walk language is affected: C#, JS, TS, Java, Go, Rust, C, C++, Lua, PHP, Dart, Scala.

## The mechanism, isolated

The lean walk DOES record return summaries (`_summaries[caller_qn] = _acc_return_taint`, consumed in `_resolve_summaries`). The distinction is WHICH returns it can see:

| callee | summary recorded? |
|---|---|
[...]

Issue #1195, opened 2026-08-11
Lean-walk parameter-name extractors for Java, C#, Rust, and C (unlock positional taint composition)


vitali87:
Follow-up to #1169. The lean-walk parameter seeding lands with a name extractor for only three grammar families, and says so itself (`_lean_parameter_names`, `flow_access/processor.py`):

> Java/C#/Rust have no extractor yet -- positional composition stays inert for them until one is added.

C is in the same position (falls through to `return []`).

## Gap

The #1142/#1168/#1169 summary machinery is language-agnostic — it keys on `(function qn, parameter name)` and composes at finalize. But composition needs the callee's positional parameter-name list, and `_lean_parameter_names` only extracts names for Go (`go_parameter_names`), JS/TS/TSX (`js_ts_parameter_names`), and C++ (`cpp_parameter_names`). For **Java, C#, Rust, and C**, the extractor returns `[]`, so:

- parameters are never seeded as pseudo-origins in `_process_lean_flow`;
- no parameter-to-sink or parameter-to-return summaries are ever recorded;
- a tainted argument passed to a logging wrapper in these languages still dies at the `kind=arg` edge and never reaches the sink inside the callee body — the exact false negative #1142 fixed for Python and #1169 fixes for Go/JS/TS/C++.

## Work

Add one extractor per grammar, mirroring the existing three:

- **Java** — `formal_parameters` → `formal_parameter` / `spread_parameter`, `name` field; receiver parameters skipped.
- **C#** — `parameter_list` → `parameter`, `name` field; `this`/`ref`/`out` modifiers don't change the name node.
- **Rust** — `parameters` → `parameter`, whose `pattern` field is usually an `identifier` but can be `mut_pattern`/`reference_pattern`/tuple patterns — unwrap like the JS destructuring path; `self_parameter` contributes nothing positional.
- **C** — `parameter_list` → `parameter_declaration`, name lives inside a (possibly nested) `declarator` chain (`pointer_declarator` → `identifier`); abstract declarators (prototypes without names) yield nothing.

Each extractor slots into `_lean_parameter_names`; no changes to the composition machinery are needed. While here, thread `caller_qn` into `_emit_taint_to_sink` so the Rust `println!` macro and C++ `cout` stream sink paths also record parameter-sink summaries (noted in #1169's approach as an open item).

## Tests

[...]

Issue #1168, opened 2026-08-10
feat: forward parameter-to-return taint summaries (pass-through helpers)


vitali87:
Follow-up to #1142 (PR #1167), which delivered the parameter-to-**sink** direction.

## Gap
Taint does not yet compose through a callee's **return** value when the returned value came from a parameter:

```python
def redact(v):
    return v            # pass-through: parameter flows to the return value

def caller():
    secret = os.getenv('K')
    y = redact(secret)
    print(y)             # NO_FLOW today: redact has no internal source, so its
                         # return summary is empty and y is treated as clean
```

The return-taint fixpoint (issue #712) only marks a function's return tainted when a **source inside its body** reaches the return. A pure pass-through has no internal source, so `y` never lights up.

## Approach (from the #1142 plan)
Compute a parameter-to-return closure (which parameters reach the return, transitively through other pass-throughs — the `Taint.params` field added in #1167 already flows into `_return_taint_source`, so the signal is captured). Then at each call site where a tainted argument enters a return-parameter, fold the argument's resolved origins/pending into the callee's entry in the return fixpoint (`_resolve_summaries`) so any caller holding `pending={callee}` resolves through it. This is the conservative MAY over-approximation, consistent with the existing model.

## Tests
- `y = redact(secret); print(y)` — three-function pass-through chain connects ENV -> STDOUT.
- Cross-file and forward-defined pass-through.
- Negative control: a helper that returns a **fresh** value (not its parameter) produces no flow.

Issue #1227, opened 2026-08-12
Go/C# frontend robustness: nested-module loading + crash-safe build lock


vitali87:
Follow-up from PR #1226 review (Greptile, two P1s) — both are **coverage/robustness** gaps that degrade safely (never a wrong fact) and are **shared with the existing C# Roslyn frontend**, so they should be fixed uniformly across both frontends rather than diverging in the Go PR.

### 1. Nested / multi-module loading
`run_go_frontend` invokes the bundled `gotypes` tool once from the repo root, loading `./...`. Go treats a nested `go.mod` as its own module boundary, so `./...` from the root does **not** descend into independently-nested modules — their local calls are absent from the emitted facts (those calls fall back to tree-sitter; never wrong, just missing).

**Fix:** enumerate module roots via `discover_go_module_paths`, run the tool per module root (or add Go-workspace support), and re-relativize emitted paths to the repo root before merging. The C# frontend has the analogous single-project-root limitation (`find_csharp_project` picks one), so consider the shared shape.

### 2. Crash-safe build lock
`_acquire_build_lock` uses an mkdir-lock with no staleness reclamation (ported verbatim from `csharp_frontend`). A holder SIGKILLed between `mkdir` and the `finally: rmdir` leaves `.build-lock` behind; later runs then wait the full retry budget (~5 min) and return empty facts until the stale dir is removed by hand.

**Fix:** reclaim a lock whose owner is demonstrably dead/expired (PID liveness or an mtime-age threshold), applied to the shared lock discipline in **both** `go_frontend` and `csharp_frontend`. Add coverage for a terminated lock holder.

Both are safe to defer: the worst case is degradation to the pure tree-sitter graph, the standing frontend invariant.

PR #1364, opened 2026-08-21
fix: lean-walk pass-through helpers keep their argument's taint (#1363)

vitali87:
Addresses #1363 for the direct pass-through case.

## The defect

A helper that returns its parameter unchanged laundered taint in every lean-walk language:

```js
function forward(v) { return v; }
const t = process.env.SECRET;
console.log(forward(t));      // no ENV -> STDOUT edge
```

Python emitted this edge; C#, JS, Java and Go did not. I filed the issue as C#-specific and that was wrong: it affects all twelve lean-walk languages.

## Cause

#1168 built the composition generically and `finalize()` is language-agnostic. The lean walk already had every part of it:

- parameters seeded as `Taint(params={name})`
- `_param_call_sites` recorded WITH a per-call token
- `_resolve_return_params` closing over the callee summaries

The one break was the join. `_js_expr_taint` returned a call's result as `Taint(pending={callee_qn})` while the Python path returns `{callee_qn, token}`. The origins injected under `token` in `finalize()` were therefore unreachable from any caller, so the composition silently resolved to nothing. The fix adds the token to the lean call result.

## Why this does not fabricate edges

The token is the precision mechanism, not a shortcut, and the tests pin each way it could go wrong:

- callee discards its parameter (`return "constant"`) gives no edge
- callee returns its SECOND parameter: secret passed first gives no edge, passed second gives the edge
- two calls to the same helper, one tainted and one clean: the clean call's result stays clean. A callee-keyed summary would fail this; the per-call token is what makes it pass.

A token for a non-pass-through callee resolves to nothing, so calls that were not pass-throughs are unaffected.

## Scope

Direct pass-through (`return v`) now composes in every language. Chaining transitively through a further call (`return other(p)`) still needs the `_return_param_edges` recording that only the Python walk does; that stays Python-only and #1363 stays open for it rather than widening this change.

## Verification

- 7 new tests in `codebase_rag/tests/test_flow_lean_passthrough_return.py`, parameterised over JavaScript, Java, C# and Go, verified RED before the fix and GREEN after.
- Full flow/taint suite: 433 passed, no regressions.
- Docs corrected: `data-flow-edges.md` claimed the return composition was Python-only, and separately still claimed Java and C# have no parameter-name extractor (#1195 added those).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data-flow tracking for pass-through return values across supported languages, including Rust.
  * Recognized Rust explicit returns and trailing expressions as function return values.
  * Prevented taint from leaking between separate call sites.
  * Ensured discarded and non-returned arguments remain clean.
  * Preserved correct positional parameter selection.

* **Documentation**
  * Clarified language-specific return propagation behavior and parser limitations.

* **Tests**
  * Added regression coverage for JavaScript, Java, C#, Go, Dart, and Rust.

### DIFF
--- a/codebase_rag/constants/ast_rust.py
+++ b/codebase_rag/constants/ast_rust.py
@@ -351,3 +351,6 @@ RS_ITER_MAP = "map"
 RS_ITER_COLLECT = "collect"
 TS_RS_ARRAY_TYPE = "array_type"
 RS_FIELD_ELEMENT = "element"
+# Rust spells a `return x;` node return_expression, NOT return_statement:
+# the shared TS_RETURN_STATEMENT never matches it (issue #1365).
+TS_RS_RETURN_EXPRESSION = "return_expression"
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -2065,7 +2065,14 @@ class FlowProcessor:
                     self._return_edge_candidates.append(
                         (callee[0], callee[1], jc.flow.caller_spec)
                     )
-                    return Taint(frozenset(), frozenset({callee[1]})), frozenset()
+                    # Dart resolves a call's value here rather than through
+                    # _js_expr_taint, so it needs the same per-call pass-through
+                    # token to compose a helper's returned parameter (#1363).
+                    token = _passthrough_result_token(jc.flow.caller_qn, call_sel)
+                    return (
+                        Taint(frozenset(), frozenset({callee[1], token})),
+                        frozenset(),
+                    )
             return None, frozenset()
         if len(rhs) == 1 and rhs[0].type == cs.TS_DART_IDENTIFIER and rhs[0].text:
             alias = rhs[0].text.decode(cs.ENCODING_UTF8)
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -196,6 +196,25 @@ def _lean_parameter_slots(
     return [], None
 
 
+# A Rust block's final child is the function's value only when it is an
+# expression: a statement or declaration yields `()` and returns nothing.
+_RUST_NON_VALUE_TAIL = frozenset({cs.TS_EXPRESSION_STATEMENT, cs.TS_RS_LET_DECLARATION})
+
+
+def _rust_tail_expression(func_node: Node) -> Node | None:
+    # The value a Rust body yields with no `return` keyword and no trailing
+    # semicolon -- the idiomatic form, which carries no return node at all and
+    # so reaches none of the return branches in the walk (issue #1365).
+    body = func_node.child_by_field_name(cs.FIELD_BODY)
+    if body is None or body.type != cs.TS_RS_BLOCK:
+        return None
+    children = [c for c in body.named_children if c.type != cs.TS_COMMENT]
+    if not children:
+        return None
+    tail = children[-1]
+    return None if tail.type in _RUST_NON_VALUE_TAIL else tail
+
+
 class Taint(NamedTuple):
     # What is tainting a variable: resolved resource origins plus the set of
     # callee QNs whose (possibly not-yet-processed) return value it carries. A
@@ -799,6 +818,15 @@ class FlowProcessor:
             # not leak its shadow past the join.
             for node in statements:
                 state = self._walk_flat_stmt(node, state, jc)
+        if ctx.language == cs.SupportedLanguage.RUST:
+            tail = _rust_tail_expression(caller_node)
+            if tail is not None:
+                returned = self._js_expr_taint(tail, state.taint, jc)
+                if returned is not None:
+                    self._acc_returns_taint = True
+                    self._acc_return_taint = _merge_taint(
+                        self._acc_return_taint, returned
+                    )
         if self._acc_returns_taint:
             self._summaries[ctx.caller_qn] = self._acc_return_taint
 
@@ -1322,7 +1350,9 @@ class FlowProcessor:
             self._flow_stream(node, tainted, handles, jc)
         elif d.keyword_stdout_write_types and node_type in d.keyword_stdout_write_types:
             self._flow_keyword_write(node, tainted, jc)
-        elif node_type == cs.TS_RETURN_STATEMENT:
+        elif node_type in (cs.TS_RETURN_STATEMENT, cs.TS_RS_RETURN_EXPRESSION):
+            # Rust names this node return_expression, so matching only
+            # return_statement dropped every Rust return summary (issue #1365).
             returned = self._js_return_taint(node, tainted, jc)
             if returned is not None:
                 self._acc_returns_taint = True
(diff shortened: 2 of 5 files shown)


vitali87:
@greptileai @coderabbitai Both findings confirmed and both fixed or scoped. I verified each independently rather than taking the reports at face value, and the measurement turned up a third gap neither of you flagged.

## Method

Every language was run twice: a direct `ENV -> STDOUT` flow as a baseline (proving the source/sink idioms are recognised, so a negative means the composition failed rather than the fixture) and the same flow through `forward(v) { return v; }`.

| language | baseline | pass-through (before) | pass-through (now) |
|---|---|---|---|
| JS, TS, Java, C#, Go, PHP, C, C++ | yes | yes | yes |
| Dart | yes | **no** | **yes (fixed)** |
| Lua | yes | **no** | no (tracked) |
| Scala | yes | **no** | no (tracked) |
| Rust | yes | **no** | no (tracked) |

## CodeRabbit: Dart

Correct, and it was the same defect in a second place. Dart resolves a call's value through `_dart_rhs` rather than `_js_expr_taint`, and that path returned `Taint(pending={callee_qn})` with no token. Fixed in 856bf91f; Dart now composes in both the block form (`{ return v; }`) and the arrow form (`=> v`), and is in the parameterised test.

## Greptile: Lua and Scala

Correct, and the cause is upstream of this PR. `_lean_parameter_slots` (`processor.py:167`) dispatches per language and falls through to `return [], None` for Lua and Scala, so no parameter is ever seeded as `Taint(params={name})` and the token has nothing to compose against. That needs new slot extractors, not a change to this PR's mechanism.

Worth noting for whoever takes it: this means the parameter-to-SINK direction from #1169 is also inert for those two, not just the return direction.

## The third gap

Rust fails too, and neither review caught it. It is the interesting case because `rust_positional_parameter_slots` IS registered, so the Lua/Scala explanation does not apply. I confirmed it fails on both the tail-expression form (`{ v }`) and the explicit `return v;`, which rules out the implicit-return shape. I have not diagnosed it further and did not want to guess in a PR description.

## What changed here

The documentation claim was the real problem you both pointed at, and you were right to block on it: I wrote "every language" having tested four. It now names the ten languages that compose and states plainly that Lua, Scala and Rust do not, with the reason. Lua, Scala and Rust are tracked in #1365 with the measurements and a suggested order.

The transitive case (`return other(p)`) remains Python-only and #1363 stays open for it.

vitali87:
@greptileai Thanks, and one correction so #1365 does not start from a false premise.

Your hypothesis 3 (parameter-name mismatch) is already ruled out: calling `rust_positional_parameter_slots` directly on the fixture returns `(["v"], None)`, the bare name the walker matches against.

Hypothesis 2 is the whole story, and the reason it looked insufficient is a wrong premise about the grammar. Rust `return v;` does NOT produce a `return_statement`. It produces **`return_expression`**, which `cs.TS_RETURN_STATEMENT` never matches, and the tail form produces no return node at all. Two forms, two reasons, one root cause. Full detail with the measurements is on #1365.

That also makes the impact wider than pass-through: no Rust return is ever recorded by the lean walk, so an own-read helper (`fn fetch() -> String { return std::env::var("SECRET").unwrap(); }`) loses its flow too, with no parameter involved.

Separately: your review check has not attached to the current head 856bf91f, which carries the Dart fix and the narrowed documentation. Could you re-review that SHA so the verdict reflects what would actually merge?

vitali87:
@greptileai Confirmed, reproduced independently, and now fixed in this PR at 8fa16cf3. I said I would keep Rust out of scope; that was the wrong call once this became the only thing standing between a correct ten-language fix and a merge, so I have folded it in rather than leaving the PR blocked on a defect I had already diagnosed.

Your finding matches my measurements exactly, including the own-read case with no parameter involved, which is the part that shows the loss is wider than pass-through composition.

## What was wrong

Two independent reasons, one per return form, which is why both failed at once and why no single fault site explained it:

1. `return v;` parses to `return_expression`. The walk tested `TS_RETURN_STATEMENT`, so it never matched, and NO Rust return summary was ever recorded, for any function.
2. `fn forward(v: String) -> String { v }` has no return node at all. The value is the block trailing expression.

## The fix

- `TS_RS_RETURN_EXPRESSION` added and matched alongside `TS_RETURN_STATEMENT`.
- A Rust block trailing expression is read as the return value, distinguished from a statement: a child that is an `expression_statement` or `let_declaration` yields `()` and returns nothing.

## Precision

The trailing-expression rule is the risky half, since "the block mentions a parameter" is not the same as "the block returns it". Two tests pin it:

- `fn size(v: String) -> usize { v.len() }` yields a LENGTH, not the secret. No edge.
- `fn discard(v: String) { let _x = v; }` ends in a statement, writes nowhere, and returns `()`. No edge. A bogus return summary is the only thing that could produce an edge in that fixture, so it pins the statement-versus-expression distinction itself rather than passing for an unrelated reason.

Plus the idiomatic tail form and the own-read form as positives. 13 tests in the file now, all verified RED before the fix.

Documentation updated: Rust joins the composing set, explicitly including the trailing-expression form. #1365 now covers only Lua and Scala, which genuinely need new parameter-slot extractors.

vitali87:
@greptileai Good catch, and it was my error: the sentence was a fragment left behind when I edited the docstring after the Rust fix landed. "Rust has an extractor, so no parameter is ever seeded" was both false and a non-sequitur, and "Both" was left pointing at three things.

Fixed in 64c925e0 along the lines you suggested. The Lua/Scala paragraph now says only what is true of Lua and Scala, and the Rust paragraph that follows already carried the accurate description of the two faults.

Worth noting the class of mistake, since it is the second time in this PR I have shipped a stale claim in prose while the code underneath was correct: both times the code was verified by execution and the prose was not verified by anything. The tests caught nothing because a docstring cannot fail. Thanks for reading it as carefully as the diff.

vitali87:
CI note: `Unit Tests (ubuntu-latest, py3.12)` went red on 64c925e0 with `test_eval_java_inheritance.py::test_no_overrides_row_is_emitted_for_java`. It is a flake, not a regression, and I established that before re-running rather than assuming it:

- 64c925e0 differs from 8fa16cf3 by **a two-line docstring in a test module**, nothing else. 8fa16cf3 passed this exact job.
- On 64c925e0 itself, `Unit Tests (ubuntu-latest, py3.13)` passed with identical code; only the py3.12 job failed.
- The job passed on all three earlier commits of this PR.

Root cause found and filed as #1366: `java_oracle.py::_ensure_compiled` compiles `Oracle.java` into the shared in-repo oracle directory with no lock, and the traceback shows a pytest-xdist worker (`popen-gw3`). The mtime guard is a check-then-act, and `javac` writes the class in place, so one worker can launch the JVM against a class file another is still writing. It is the same failure mode already solved for the javac frontend in #1341 and the Go/C# frontends in #1227, and the fix is to reuse `build_lock.py` plus a staging-and-rename publish.

The job has been re-enqueued and is running.

vitali87:
@greptileai Could you review the current head 64c925e0? Your last check-run is against 8fa16cf3, and the only change since is the two-line docstring correction you asked for, which is now applied and the thread resolved.

The flaky `Unit Tests (ubuntu-latest, py3.12)` failure has been re-run and passes, confirming it was the Java oracle compile race rather than a regression (filed as #1366).

Everything else on the gate is green: all checks passing, Sonar quality gate OK with 0 issues, no unresolved threads. Yours is the only outstanding verdict, and I would rather have it against the exact commit that would merge than infer it from the previous one.

PR #1369, opened 2026-08-21
fix: lock the Java oracle build so parallel test workers cannot race (#1366)

### LINKED ISSUE #1366: Java oracle compiles into a shared directory with no lock, so parallel test workers race
Flaky CI failure observed on PR #1364, run [32524121790](https://github.com/vitali87/code-graph-rag/actions/runs/32524121790):

```
FAILED codebase_rag/tests/test_eval_java_inheritance.py::test_no_overrides_row_is_emitted_for_java
subprocess.CalledProcessError: Command
  [/usr/bin/java, -cp, .../evals/oracles/java_oracle, Oracle, /tmp/.../popen-gw3/.../proj]
returned non-zero exit status 1
```

## It is a flake, established before filing

- The commit that failed differs from its parent by **a two-line docstring in a test module** and nothing else. The parent passed this exact job.
- On the failing commit, `Unit Tests (ubuntu-latest, py3.13)` passed with identical code. Only the py3.12 job failed.
- The same job passed on all three earlier commits of the PR.

So the failure is not code-dependent.

## Mechanism

`evals/oracles/java_oracle.py::_ensure_compiled` compiles `Oracle.java` into the **shared, in-repo** `evals/oracles/java_oracle/` directory with no lock:

```python
def _ensure_compiled() -> None:
    if _CLASS.is_file() and _CLASS.stat().st_mtime >= _SOURCE.stat().st_mtime:
        return
    subprocess.run([javac, str(_SOURCE)], cwd=str(_ORACLE_DIR), check=True)
```

The traceback shows worker `popen-gw3`, so the suite runs under pytest-xdist. Several workers can enter this concurrently. The guard is a check-then-act on mtime with no atomicity, and `javac` writes `Oracle.class` (plus any nested-class files) in place and non-atomically. A worker can therefore launch `java -cp <dir> Oracle` against a class file another worker is midway through writing, and the JVM exits 1.

The mtime check makes it rare, since it only opens the window when the class is missing or stale, which on a fresh CI checkout is exactly the first run: precisely when all workers arrive at once.

## Fix

This is the same failure mode already solved for the bundled javac frontend in #1341 and for the Go/C# frontends in #1227. Reuse the existing machinery rather than inventing a second one:

- `codebase_rag/parsers/build_lock.py` provides `acquire_build_lock` / `release_build_lock`.
- Compile into a staging directory and publish by atomic rename, so a partially written class is never visible under the path the JVM loads.

The same audit should cover the other oracles that compile or build on demand: whichever of the Go, C#, and Rust oracle helpers share this compile-into-a-shared-path shape have the same race.

## Note

Not urgent for correctness of shipped code, since it only affects the eval oracle test harness, but it will keep costing CI reruns until fixed.

vitali87:
Closes #1366.

## The defect

`java_oracle.py::_ensure_compiled` compiled `Oracle.java` into the shared in-repo oracle directory with no lock. The freshness test is a check-then-act, and `javac` writes the class in place, so under `pytest -n auto` one worker could exec the JVM against a class file another worker was still writing. It surfaced on PR #1364 as:

```
FAILED test_eval_java_inheritance.py::test_no_overrides_row_is_emitted_for_java
subprocess.CalledProcessError: [/usr/bin/java, -cp, .../java_oracle, Oracle, .../popen-gw3/...] exit 1
```

The mtime guard is why it is rare rather than constant: the window only opens when the class is missing or stale, which on a fresh CI checkout is the first run, precisely when every worker arrives at once.

The real oracle directory shows `Oracle.class` plus `Oracle$1.class`, `Oracle$2.class`, `Oracle$3.class`, so a compile publishes four files in place, none of them atomically.

## Audit: Java was the only one unprotected

I checked every oracle that builds on demand before writing code, since a second mechanism would have been the wrong fix:

| oracle | protection |
|---|---|
| Rust | `FileLock`, added for this exact flake ("one worker exec'd the binary while another rewrote it (ETXTBSY) ... under `pytest -n auto`") |
| C# | mkdir lock with polling |
| TypeScript / Lua / PHP (node deps) | mkdir lock with polling in `_common.py` |
| **Java** | **none** |

So this reuses the Rust oracle's pattern rather than introducing anything new: take the lock, re-test freshness inside it, compile. Every path that runs the JVM goes through `_ensure_compiled`, so a caller that waits on the lock only proceeds once the class is complete.

## Verification

The race is reproduced deterministically rather than argued. `test_concurrent_callers_compile_the_oracle_once` points the module at a scratch directory, substitutes a compiler that takes measurable time before publishing the class, and runs six threads:

- **before the fix: 6 compiles** (`assert 6 == 1`), each able to publish over another
- **after: 1**

Two further tests keep the freshness short-circuit honest in both directions, since a lock that always compiles or never recompiles would also pass the race test: a class newer than its source is left alone, and a class older than its source is rebuilt.

- Oracle/eval suite: 235 passed, 12 skipped.
- The previously flaky Java oracle tests pass under `-n 4`.
- `.build-lock` added to `.gitignore` (the Rust oracle's lock is already ignored; this one was not).

## Scope

Test-harness only. No shipped code path changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Java oracle compilation reliability during concurrent builds.
  * Prevented partially written class files from being used.
  * Skipped recompilation for current classes and rebuilt outdated classes automatically.
  * Added temporary build-lock handling to isolate compilation artifacts.

* **Tests**
  * Added coverage for concurrent compilation, temporary-path isolation, and class freshness behavior.


vitali87:
@greptileai Confirmed and fixed in c4311a48. The finding is exactly right, and it is the failure mode the test existed to prevent, which makes it worth stating plainly: a regression test that can pass while its subject is broken is worse than no test, because it reports safety it has not established.

`threading.Thread` swallows whatever its target raises and `join(timeout=...)` returns nothing when it expires, so "exactly one compile happened" held equally well if five workers had died and the sixth did the work.

Now the workers record their exceptions, and the test asserts every thread finished and none raised before it believes the count.

I also re-verified the test still detects what it was written for, rather than assuming a strengthened test is still a working one. With the lock neutralised and nothing else changed, it fails at the compile-count assertion; with the lock restored it passes. So the added assertions did not turn it into a test that passes regardless.

vitali87:
@coderabbitai Confirmed and fixed in 093d0f25. You are right, and the version I wrote was weaker than it looked.

Post-dating the SOURCE meant the class my fake compiler then wrote was itself still older than the source, so the artifact stayed stale after the "rebuild". The test only passed because it called `_ensure_compiled()` once: a real run in that state would recompile on every single call and never settle.

Now the test ages the CLASS instead, calls `_ensure_compiled()` twice, and asserts exactly one compile. That checks what the original was reaching for (a stale class IS rebuilt) plus the part it silently skipped (the rebuild actually resolves staleness).

To be clear about impact: this was a defect in the test, not the product. Real javac writes the class after reading the source, so the class is genuinely newer and the guard settles. The pathology existed only in my artificial fixture.

@greptileai Your P1 on the timed joins is addressed in c4311a48 (workers record exceptions, and the test asserts every thread finished and none raised). I also confirmed the strengthened test still fails with the lock neutralised, so the extra assertions did not turn it into a test that passes regardless. Resolving that thread now.

vitali87:
@greptileai Could you review the current head 093d0f25? Your check-run is against 8d53401a, two commits back, and both of your and CodeRabbit's findings have been addressed since:

- c4311a48: workers capture exceptions and the test asserts every thread finished, fixing your timed-join P1
- 093d0f25: the stale-class test ages the class rather than post-dating the source, so the rebuild converges and a second call does not recompile

Everything else on the gate is green: all checks passing, Sonar quality gate OK with 0 issues, no unresolved threads. Yours is the only outstanding verdict and I would rather have it against the commit that would actually merge.

Issue #1169, opened 2026-08-10
feat: forward parameter taint for the lean-walk languages (Go/JS/TS/Java/C#)


vitali87:
Follow-up to #1142 (PR #1167), which delivered forward parameter-to-sink taint for the **Python** walk only.

## Gap
The lean (non-Python) flow walk (`_process_lean_flow` / `_js_call`) does not yet seed parameters as pseudo-origins, so a tainted argument passed to a logging wrapper in Go/JS/TS/Java/C# still stops at the `kind=arg` edge and never reaches the sink inside the callee body.

## Approach
The summary/closure/composition machinery from #1167 is language-agnostic (it keys on function qn + parameter name and composes at finalize). What's missing on the lean side:
- Seed each parameter into the `tainted` map in `_process_lean_flow` (mirror the Python seeding), using each language's parameter-name extraction.
- Record parameter-to-sink in the `_js_call` write-sink branch and the callee-arg branch (mirror the Python `_apply_call` additions).
- `_flow_param_names` already covers Go/JS-TS/C++; positional composition works for those. Java/C# lack recorded parameter names, so positional mapping yields nothing there until that table is extended (matches the existing arg-edge limits) — keyword/direct cases still compose.
- Consider the `_emit_taint_to_sink` path (Rust `println!` macro, C++ `cout` stream) — needs `caller_qn` threaded to record parameter-sink there.

## Tests
Mirror the Python cases from `test_flow_edges.py` in at least one lean-walk language (Go or Java): logging-wrapper positional, two-hop, cross-file, and a negative control.

vitali87:
Follow-up filed: #1195 adds the missing parameter-name extractors (Java/C#/Rust/C) so the positional composition this issue lands stops being inert there. Two adjacent gaps found in the same audit are tracked separately: #1196 (nested-definition headers skipped by the lean walks) and #1197 (closure-capture taint, all walks).

PR #1375, opened 2026-08-22
feat: Lua and Scala parameter slots complete the lean taint composition (#1365)

### LINKED ISSUE #1365: Lean walk: Lua and Scala lack parameter-slot extractors, so no taint composition runs
Follow-up to #1363 / PR #1364. That PR made a direct pass-through helper (`return v`) carry its argument's taint on the lean walk, and it works in JavaScript, TypeScript, Java, C#, Go, PHP, C, C++ and Dart. Three languages still do not compose, each verified against a baseline so the fixture itself is known good.

## Measured

Each language was run twice: a direct `ENV -> STDOUT` flow (baseline, proves the source and sink idioms are recognised) and the same flow through `forward(v) { return v; }`.

| language | baseline | pass-through |
|---|---|---|
| PHP, C, C++, TypeScript, JS, Java, C#, Go, Dart | yes | yes |
| Lua | yes | **no** |
| Scala | yes | **no** |
| Rust | yes | **no** |

## Lua and Scala: no parameter slots

`_lean_parameter_slots` (`codebase_rag/parsers/flow_access/processor.py:167`) dispatches per language and falls through to `return [], None` for both. With no slots, no parameter is seeded as `Taint(params={name})`, so `_resolve_return_params` has nothing to match and the per-call token composes against nothing. These need `lua_positional_parameter_slots` / `scala_positional_parameter_slots` in the style of the extractors #1195 added, plus the seeding registration.

Note this also means the parameter-to-SINK direction (#1169) is inert for Lua and Scala, not just the return direction, so fixing the extractor should light up both. Worth confirming when the work is done.

## Rust: extractor present, still no composition

`rust_positional_parameter_slots` is registered, and Rust fails on BOTH return forms:

- tail expression: `fn fwd(v: String) -> String { v }`
- explicit statement: `fn fwd(v: String) -> String { return v; }`

Since both fail, it is not the implicit-tail-return shape. I have not diagnosed further. Whoever picks this up should check whether the parameter names the Rust extractor returns actually match the identifier text the walk sees (a `mut v` or pattern-bound parameter would not), and whether `let s = fwd(t);` reaches the same call-result path that carries the token.

## Suggested order

Rust first (extractor exists, so likely a narrow mismatch), then Lua and Scala together since they need the same new extractor work. Each needs the positional-precision tests the existing pass-through file uses: a discarded parameter, a wrong-slot argument, and two call sites where only one is tainted.

vitali87:
Closes #1365.

## What was missing

`_lean_parameter_slots` dispatched per language and fell through to `return [], None` for Lua and Scala. With no slots, no parameter was ever seeded as `Taint(params={name})`, so BOTH composition directions were inert for those two languages, not just the return one:

- argument into a callee sink (#1169)
- pass-through return (#1363)

Every other lean language got its extractor in #1195/#1169. These two were simply never added.

## Two pieces

**Slot extractors** (`utils.py`), following the #1195 precedent:

- Lua: `parameters` holds bare `identifier` children plus a trailing `...` (`vararg_expression`) which takes the variadic slot and binds no name.
- Scala: `parameters` -> `parameter` with a `name` field; a repeated parameter (`xs: String*`) is identified by its TYPE node being `repeated_parameter_type`. Curried definitions have several sibling `parameters` lists and only the first maps to `arg:<index>`, so the rest are deliberately left alone.

**Scala body-as-value.** `def forward(v: String): String = v` has no return node at all: a def's BODY is its value. This is the same shape as Rust's trailing block expression, so the existing hook is generalised to cover both. A block body resolves to its final expression; a block ending in a `val`/`var` definition yields Unit and returns nothing.

## Precision

The body-as-value rule is the risky half, so it is pinned in both directions:

- `def size(v: String): Int = v.length` yields a LENGTH, not the secret. No edge.
- `def hold(v: String): Unit = { val kept = v }` ends in a definition, writes nowhere. No edge.
- Positional mapping for both languages: the secret as the second argument reaches the sink only when the callee uses its second parameter.

## A correction to my own test

I first wrote a Lua test asserting that `...` prevents a positional shift. Measuring instead of assuming showed the premise was wrong: **Lua requires `...` to be last**, so it can never displace a named parameter and that shift is unreachable. The real, narrower limitation is that taint landing in `...` binds no name and stops there. The test now asserts that measured behaviour, and the documentation states it.

## Verification

- 10 tests, all verified RED before the change.
- Flow/taint/Lua/Scala suite: 729 passed, 5 skipped, no regressions.
- `data-flow-edges.md` updated: direct pass-through now composes in all thirteen covered languages, with the two keyword-less return forms and the Lua vararg limitation called out explicitly.

## Summary by CodeRabbit

* **New Features**
  * Improved data-flow analysis for Lua and Scala function parameters, including variadic and repeated parameters.
  * Added support for tracking Scala expression-, block-, and indented-body return values.
  * Improved propagation of values through Lua and Scala function calls.

* **Bug Fixes**
  * Prevented incorrect propagation through Lua varargs and Scala methods returning `Unit`, including qualified `Unit` types.
  * Preserved correct handling of user-defined types ending in `Unit`.

* **Documentation**
  * Updated data-flow documentation with Scala return handling and Lua vararg behavior.

### DIFF
--- a/codebase_rag/constants/ast_lua.py
+++ b/codebase_rag/constants/ast_lua.py
@@ -33,3 +33,7 @@ TS_LUA_FUNCTION_DEFINITION = "function_definition"
 IMPORT_REQUIRE = "require"
 IMPORT_PCALL = "pcall"
 IMPORT_IMPORT = "import"
+
+# Lua `...` in a parameter list: occupies the trailing variadic slot and
+# binds no simple name (issue #1365).
+TS_LUA_VARARG_EXPRESSION = "vararg_expression"
--- a/codebase_rag/constants/ast_scala.py
+++ b/codebase_rag/constants/ast_scala.py
@@ -32,3 +32,6 @@ TS_SCALA_REPEATED_PARAMETER_TYPE = "repeated_parameter_type"
 TS_SCALA_INDENTED_BLOCK = "indented_block"
 # A def declared `Unit` discards its body's value, so no return summary.
 SCALA_UNIT_TYPE = "Unit"
+# Every spelling of scala.Unit. A user-defined type merely ENDING in Unit
+# (`example.Unit`) is a real type whose return value must still compose.
+SCALA_UNIT_TYPES = frozenset({"Unit", "scala.Unit", "_root_.scala.Unit"})
--- a/codebase_rag/constants/ast_scala.py
+++ b/codebase_rag/constants/ast_scala.py
@@ -29,3 +29,6 @@ TS_SCALA_TYPE_IDENTIFIER = "type_identifier"
 # parameter (`xs: String*`) is spelled as the parameter's TYPE node.
 TS_SCALA_PARAMETER = "parameter"
 TS_SCALA_REPEATED_PARAMETER_TYPE = "repeated_parameter_type"
+TS_SCALA_INDENTED_BLOCK = "indented_block"
+# A def declared `Unit` discards its body's value, so no return summary.
+SCALA_UNIT_TYPE = "Unit"
--- a/codebase_rag/constants/ast_scala.py
+++ b/codebase_rag/constants/ast_scala.py
@@ -24,3 +24,8 @@ TS_SCALA_LAMBDA_EXPRESSION = "lambda_expression"
 TS_SCALA_IDENTIFIER = "identifier"
 TS_SCALA_INSTANCE_EXPRESSION = "instance_expression"
 TS_SCALA_TYPE_IDENTIFIER = "type_identifier"
+
+# Scala parameter shapes for lean slot extraction (issue #1365). A repeated
+# parameter (`xs: String*`) is spelled as the parameter's TYPE node.
+TS_SCALA_PARAMETER = "parameter"
+TS_SCALA_REPEATED_PARAMETER_TYPE = "repeated_parameter_type"
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -214,16 +214,14 @@ _SCALA_NON_VALUE_TAIL = frozenset(
 
 
 def _scala_returns_unit(func_node: Node) -> bool:
-    # `Unit` and the qualified `scala.Unit` (and `_root_.scala.Unit`) name the
-    # same type, so compare the LAST dotted segment rather than the whole
-    # annotation. A user-defined type that happens to end in `Unit` would also
-    # match; that direction only SUPPRESSES a summary, so the cost is a missed
-    # flow rather than a fabricated one, which is the safer way to be wrong.
+    # The spellings of scala.Unit are enumerable, so match them exactly rather
+    # than by trailing segment: a user-defined `example.Unit` is an ordinary
+    # type whose returned value DOES matter, and suppressing it would drop a
+    # real flow.
     return_type = func_node.child_by_field_name(cs.FIELD_RETURN_TYPE)
     if return_type is None or return_type.text is None:
         return False
-    spelled = return_type.text.decode(cs.ENCODING_UTF8).strip()
-    return spelled.rpartition(cs.SEPARATOR_DOT)[2] == cs.SCALA_UNIT_TYPE
+    return return_type.text.decode(cs.ENCODING_UTF8).strip() in cs.SCALA_UNIT_TYPES
 
 
 def _scala_body_value(func_node: Node) -> Node | None:
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -72,11 +72,13 @@ from ..utils import (
     go_positional_parameter_slots,
     java_positional_parameter_slots,
     js_ts_positional_parameter_slots,
+    lua_positional_parameter_slots,
     php_positional_parameter_slots,
     python_free_variable_names,
     python_parameter_names,
     rust_positional_parameter_slots,
     safe_decode_text,
+    scala_positional_parameter_slots,
 )
 from .constants import (
     KEY_KIND,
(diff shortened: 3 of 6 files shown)


vitali87:
@greptileai @coderabbitai All three findings confirmed by measurement and fixed in 59443880.

## Greptile: Unit methods leaked a return value (P1)

Correct, and it is the exact false-positive direction I flagged as the risk of this rule in the PR description, so you found the instance I failed to look for.

Reproduced before fixing:

| case | before | after |
|---|---|---|
| `def hold(): Unit = { System.getenv("SECRET") }` then `println(hold())` | **edge emitted (wrong)** | no edge |
| same body under `: String` (control) | edge | edge |

`Unit` discards the body value, so summarising the trailing expression invented a return the caller can never observe. `_scala_body_value` now returns nothing when the def is DECLARED `Unit`. Only an explicit annotation is trusted: an inferred type is unknown to the walk, so those are left alone rather than guessed at.

The control matters as much as the fix. An exclusion like this can silently swallow real flows, so the same body under a `String` return type is pinned as a test to prove the rule is narrow.

## CodeRabbit: Scala 3 indented bodies

Also correct, and I had not considered Scala 3 at all. `object M:` / `def f(v: String) =` with significant indentation spells the body `indented_block`, not `block`. I confirmed it against the grammar: matching only the braced form meant every Scala 3 helper silently missed its return. Both block forms are handled now, with a test using indentation syntax throughout.

This one was a false NEGATIVE rather than a false positive, which is why my own tests could not have caught it: they were all written in Scala 2 braces.

## CodeRabbit: the parameter-sink language list

Correct, and staler than this PR. Line 364 listed Go, JavaScript, TypeScript/TSX and C++, but Java, C#, Rust and C were added in #1195 and PHP and Dart later still. I checked the actual dispatch table in `_lean_parameter_slots` rather than patching in just my two, and the list now names all twelve.

## Verification

13 tests in the file, each verified RED first. Full flow/taint/Lua/Scala regression is running against this commit and I will report the result before merging.

vitali87:
@greptileai Could you review the current head 59443880? Your check-run is against 64600e3a, before the fixes.

Both findings are addressed and pinned by tests:

- Your Unit P1: `_scala_body_value` returns nothing for a def DECLARED `Unit`, so `println(hold())` no longer invents a flow. The `: String` control is tested too, so the exclusion cannot silently swallow real returns.
- CodeRabbit's Scala 3 gap: `indented_block` bodies are handled alongside braced ones.

Local flow/taint/Lua/Scala regression against this commit: 732 passed, 5 skipped, no failures. All 26 CI checks green, Sonar quality gate OK with 0 issues, no unresolved threads. Yours is the only outstanding verdict.

vitali87:
@greptileai @coderabbitai Both of you landed on the same gap independently, and it is a real one. Fixed in 9d89f745.

`scala.Unit` names the same type as `Unit`, and my exact-string check on the annotation let the qualified spelling through, re-opening exactly the leak the previous commit closed:

| return type | before 9d89f745 | after |
|---|---|---|
| `Unit` | no edge | no edge |
| `scala.Unit` | **edge emitted (wrong)** | no edge |
| `String` (control) | edge | edge |

The check now compares the LAST dotted segment, which also covers `_root_.scala.Unit`. A user-defined type ending in `Unit` would match too; I chose that deliberately because this direction only SUPPRESSES a summary, so being wrong costs a missed flow rather than a fabricated one, and for a taint rule that is the safer way to be wrong. The tradeoff is written into the code comment rather than left implicit.

@greptileai on your earlier question: I measured both shapes you named rather than reasoning about them.

| case | result |
|---|---|
| `def leak(v: String): Unit = { v.toUpperCase }` (trailing non-Unit, discarded) | no edge |
| `def wrap(v: String): Unit = { ...; println(x) }` (trailing Unit call) | no edge |
| Scala 3 indented `Unit` with trailing non-Unit expression | no edge |
| `def show(v: String): Unit = { println(v) }` (param reaches a real sink) | **edge, correctly** |

That last row is the one worth calling out: the Unit exclusion must not suppress the parameter-to-SINK direction, since a Unit method that writes its argument is a genuine flow no matter what it returns. It is now a test, so the exclusion cannot quietly widen into swallowing real edges.

16 tests in the file. Full regression running against this commit; I will post the numbers before merging.

vitali87:
@coderabbitai You are right and I have taken the correction, in 4acac31b.

I had argued that trailing-segment matching was the safer way to be wrong because over-suppression only costs a missed flow. That reasoning holds when the set cannot be enumerated. Here it can: `Unit`, `scala.Unit` and `_root_.scala.Unit` are the spellings, so accepting collateral damage to `example.Unit` bought nothing. A user-defined type ending in `Unit` is an ordinary type whose returned value is observable, and dropping it is a real miss, not a conservative default.

The match is now exact against that set, with a regression test asserting `def fetch(): example.Unit` still composes its return.

@greptileai this also supersedes the last-segment approach I described in my previous comment: same defect closed, narrower rule.

## Regression, as promised

Local flow/taint/Lua/Scala suite against 9d89f745: **735 passed, 5 skipped, no failures**. 17 tests now in the pass-through file, each verified RED first. I am re-running the suite against 4acac31b and will confirm before merging.

Current gate: 26 CI checks green, Sonar quality gate OK with 0 issues.

vitali87:
@greptileai Could you review the current head 4acac31b? Your check-run is against 64600e3a, four commits back.

Everything raised across the three review rounds is now fixed and pinned by tests:

| finding | by | fixed in |
|---|---|---|
| `Unit` def leaked its body value | Greptile | 59443880 |
| Scala 3 `indented_block` bodies never composed | CodeRabbit | 59443880 |
| parameter-sink language list stale since #1195 | CodeRabbit | 59443880 |
| qualified `scala.Unit` still leaked | both | 9d89f745 |
| trailing-segment match over-suppressed `example.Unit` | CodeRabbit | 4acac31b |

**Regression against this exact commit: 736 passed, 5 skipped, no failures.** 17 tests in the pass-through file, each verified RED before its fix. 26 CI checks green, Sonar quality gate OK with 0 issues, no unresolved threads.

Yours is the only outstanding verdict.

PR #1377, opened 2026-08-22
feat: lean-walk transitive pass-through chaining (closes #1363)

### LINKED ISSUE #1363: Lean-walk languages lose pass-through return taint (param-to-return summaries missing)

vitali87:
Closes #1363, the half deliberately left out of #1364.

## What was missing

#1364 made a DIRECT `return v` carry its argument's taint in every language. A callee returning the result of a FURTHER call still ended the chain:

```js
function inner(v) { return v; }
function outer(p) { return inner(p); }      // chain stopped here
console.log(outer(process.env.SECRET));     // no ENV -> STDOUT edge
```

`_resolve_return_params` already closes over `_return_param_edges` transitively, and the Python walk has recorded that hand-off since #1168. The lean walk never did, so its closure had only direct returns and every chain ended one hop in.

## The change

When a lean return returns a CALL, record which of the enclosing function's parameters that call forwards, tagged with the argument's `via` slot. The existing finalize closure does the rest, at any depth.

Params-only, and never touches the deferred-candidate lists: the VALUE side of the returned call is already handled by `_js_expr_taint` on the same node, so this adds the hand-off without double-counting the return itself.

## Precision

The two negative tests are the point of this PR, not decoration. Both passed BEFORE the change simply because nothing propagated at all, so their whole value is in still passing now:

- `outer(p) { return inner("clean"); }` -- the parameter never reaches the return, so the caller stays clean. A chain that recorded the hand-off unconditionally would fail this.
- `inner(a, b) { return b; }` with `outer` forwarding its parameter into the FIRST slot -- the chain is positional at every hop, not just the last one.

Plus a three-deep chain (`c -> b -> a`) to confirm this is the transitive closure rather than a one-hop special case.

## Verification

- 6 tests, 4 verified RED before the change, across JavaScript, Java and Go.
- Direct pass-through, Lua/Scala and lean-params suites: 37 passed, no regressions.
- Full flow/taint suite: **462 passed, 0 failures**.
- `data-flow-edges.md` updated in the same commit: the transitive case is no longer Python-only, and the positional constraint is stated.

## Summary by CodeRabbit

* **New Features**
  * Improved data-flow analysis to track values through nested wrapper functions across supported languages, including awaited and parenthesized calls.
  * Supports multi-level parameter-to-return propagation when functions forward the appropriate argument.

* **Bug Fixes**
  * Prevents incorrect propagation when wrappers return constants or forward a different argument.

* **Documentation**
  * Updated data-flow documentation to describe cross-language transitive return-flow behavior.

### DIFF
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -2947,10 +2947,45 @@ class FlowProcessor:
         # `return a, b` several); union the taint over every returned value.
         result: Taint | None = None
         for expr in self._lean_return_values(node):
+            self._record_lean_return_handoff(expr, tainted, jc)
             taint = self._js_expr_taint(expr, tainted, jc)
             if taint is not None:
                 result = taint if result is None else _merge_taint(result, taint)
         return result
+
+    def _record_lean_return_handoff(
+        self, expr: Node, tainted: _TaintMap, jc: _JsCtx
+    ) -> None:
+        # `return other(p)`: record that each of THIS function's parameters
+        # appearing in the returned call reaches its return through that callee,
+        # so a wrapper of a wrapper resolves (issue #1363). The Python walk has
+        # recorded this hand-off since #1168; without it the lean closure had
+        # only direct `return v` and every chain ended one hop in.
+        #
+        # Params-only, never the deferred-candidate lists: the VALUE side of the
+        # returned call is already handled by _js_expr_taint on the same node.
+        if expr.type != jc.descriptor.call_type:
+            return
+        raw = call_name(expr)
+        if raw is None:
+            return
+        callee = self._resolve(
+            raw,
+            jc.flow.module_qn,
+            jc.flow.class_context,
+            jc.flow.caller_qn,
+            jc.flow.language,
+            jc.flow.local_var_types,
+        )
+        if callee is None:
+            return
+        for via, taint in self._js_arg_taints(expr, tainted, jc):
+            if taint is None:
+                continue
+            for pname in taint.params:
+                self._return_param_edges.append(
+                    (jc.flow.caller_qn, pname, callee[1], via)
+                )
 
     @staticmethod
     def _lean_return_values(node: Node) -> list[Node]:
--- a/codebase_rag/parsers/flow_access/processor.py
+++ b/codebase_rag/parsers/flow_access/processor.py
@@ -873,6 +873,10 @@ class FlowProcessor:
                 else _scala_body_value(caller_node)
             )
             if tail is not None:
+                # A keyword-less return is still a return: the chain hand-off
+                # has to be recorded here as well, or Scala and Rust wrappers
+                # stop one hop in (issue #1363).
+                self._record_lean_return_handoff(tail, state.taint, jc)
                 returned = self._js_expr_taint(tail, state.taint, jc)
                 if returned is not None:
                     self._acc_returns_taint = True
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai @coderabbitai Both findings confirmed by measurement and fixed in 317b3492. CodeRabbit's was the broader one and it caught an overclaim in my own documentation.

## Greptile: wrapped calls

Correct. `_record_lean_return_handoff` tested the node type before unwrapping, so a chain stopped at a pair of parentheses:

| form | before | after |
|---|---|---|
| `return inner(p)` | edge | edge |
| `return (inner(p))` | **no edge** | edge |
| `return await inner(p)` | **no edge** | edge |

Now unwraps exactly what `_js_expr_taint` unwraps, before the guard.

## CodeRabbit: the other return paths

Also correct, and this is the part that mattered most. The helper was only reachable from `_js_return_taint`, so three return paths never saw it:

- Rust and Scala keyword-less returns (the body IS the value, no return node)
- Dart, which resolves calls as a selector chain rather than the shared call node

Measured before fixing: Scala chain **false**, Dart chain **false**. Rust happened to pass through another route, but it was not recorded and I have wired it explicitly rather than leave it to coincidence. Dart needed its own recorder using `dart_call_name` / `_dart_arg_taints`, since the generic one cannot recognise a selector call.

## The overclaim

My documentation said the transitive case "composes in those same languages", listing all thirteen, when I had tested three. That was wrong when I wrote it, and it is the second time in this session I have generalised a coverage claim from a small sample.

So I measured every language, both forms, before touching the wording again:

| language | direct | chain |
|---|---|---|
| Python, JavaScript, TypeScript, Java, C#, Go, PHP, C, C++, Rust, Dart, Lua, Scala | yes | yes |

Thirteen of thirteen, each with a direct-pass-through baseline alongside so a positive cannot come from a mis-built fixture. The claim is now true and measured rather than assumed.

## Precision held

The unwrap could easily have become "any returned call forwards taint", so `return (inner("clean"))` is now tested to stay clean: wrapped or not, the hand-off is still positional and still only records parameters the call actually forwards.

12 tests. Full flow/taint/dart/scala/rust regression running against this commit; numbers to follow before merge.

vitali87:
@greptileai Could you review the current head 317b3492? Your check-run is against 21523a23, before the fixes.

Both findings are addressed and pinned:

- your wrapped-call P1: parenthesized and awaited returned calls now chain, verified false before and true after
- CodeRabbit's broader one: Rust and Scala keyword-less returns and Dart selector calls now record the hand-off on their own paths, with Dart getting a dedicated recorder

**Local regression against this commit: 1291 passed, 7 skipped, no failures.** 12 tests in the file. All 26 CI checks green, Sonar quality gate OK with 0 issues, no unresolved threads.

The coverage claim in the documentation is now backed by a thirteen-language measurement with a direct-pass-through baseline per language, rather than the three-language sample I generalised from originally.

Issue #1012, opened 2026-08-01
cgr daemon publishes Memgraph, Lab, and Qdrant ports on all interfaces by default


vitali87:
## Problem

`codebase_rag/docker-compose.yaml` maps the stack ports with bare `host:container` entries (`7687`, `7444`, `3000`, `6333`, `6334`). Docker's default host address for such mappings is `0.0.0.0`, so `cgr daemon up` exposes Memgraph (unauthenticated Bolt), Memgraph Lab, and Qdrant to every host on the local network. Anyone on the same network can read the full code graph.

This is the same exposure class as #808, which we fixed for the MCP HTTP server by defaulting to a loopback bind.

## Proposal

Prefix each published port with `127.0.0.1:` by default, with an env override (for example `CGR_STACK_BIND_HOST`) for users who intentionally want the stack reachable from other hosts.

## Notes

Found while writing the security model documentation for the OpenSSF Best Practices silver badge work. The documentation discloses the current behaviour until this is fixed.

PR #1379, opened 2026-08-22
docs: security guide describes the loopback bind that shipped with #1012 (#1372)

### LINKED ISSUE #1372: docs: security.md still claims Memgraph/Qdrant ports are network-reachable, but the compose file binds 127.0.0.1 (#1012 is fixed)
## Problem

Reported in an external field-test review of v0.0.593 ([article, Chinese](https://index.zshipu.com/ai002/post/20260817/Code-Graph-RAG-0.0.593-%E5%AE%9E%E6%B5%8B%E7%BB%99-Coding-Agent-%E5%8A%A0%E7%9F%A5%E8%AF%86%E5%9B%BE%E8%B0%B1%E5%80%BC%E4%B8%8D%E5%80%BC%E5%BE%97%E5%A4%9A%E5%85%BB%E4%B8%A4%E4%B8%AA%E6%95%B0%E6%8D%AE%E5%BA%93/)) — and still present on `main` today.

`docs/architecture/security.md` (line 25) states:

> **The graph and vector stores are local processes, but their ports are currently network-reachable.** … The compose file presently publishes their ports (7687, 7444, 3000, 6333, 6334) without a host-address restriction, which Docker binds to all interfaces … issue #1012 tracks changing the default to a loopback bind.

But #1012 was closed (commit `fd34afe`) and `codebase_rag/docker-compose.yaml` now binds every published port to loopback by default:

```yaml
- "${CGR_STACK_BIND_HOST:-127.0.0.1}:${MEMGRAPH_PORT:-7687}:7687"
```

The security guide is describing a threat that no longer exists in the default configuration, and pointing at a tracking issue that is already resolved. The reviewer flagged this documentation drift as a maintenance risk — a reader can't tell which parts of the security posture description are current.

## Suggested fix

Rewrite that paragraph to reflect reality: ports bind `127.0.0.1` by default, `CGR_STACK_BIND_HOST` can widen the bind (with the caveat that Memgraph/Qdrant are unauthenticated, as the doc already notes around line 45), and reference #1012 as resolved rather than tracking.

vitali87:
Closes #1372.

## The drift

`docs/architecture/security.md` still said:

> **The graph and vector stores are local processes, but their ports are currently network-reachable.** ... The compose file presently publishes their ports ... without a host-address restriction ... issue #1012 tracks changing the default to a loopback bind.

Every claim in that paragraph is false today. Verified each before rewriting:

| claim | reality |
|---|---|
| ports published without a host address | `"${CGR_STACK_BIND_HOST:-127.0.0.1}:${MEMGRAPH_PORT:-7687}:7687"` and the same for 7444, 3000, 6333, 6334 |
| Docker binds them to all interfaces | binds `127.0.0.1` unless overridden |
| #1012 "tracks" the fix | #1012 CLOSED 2026-08-08 |

A security document that OVERSTATES exposure is not harmlessly out of date. A reader cannot tell which parts are still true, which is exactly what the external reviewer flagged.

## The rewrite

States the current posture, and adds something the docs did not cover anywhere: `CGR_STACK_BIND_HOST` appeared only in the compose file, `stack/constants.py` and a test, so the one operation that RE-CREATES the exposure was undocumented. The new text says widening the bind is a deliberate risk and that Memgraph Bolt, Memgraph Lab and Qdrant are all unauthenticated, so a wider bind puts the whole graph on the network with no credential in front of it.

## Guard against recurrence

The issue is really about drift, so a prose fix alone would leave the same failure mode in place. `test_the_security_guide_describes_the_loopback_default` fails if the guide reverts to describing the pre-#1012 exposure, or stops naming the variable that widens the bind.

Verified the guard actually works rather than assuming: with the doc reverted to its old text the test FAILS, with the fix in place it PASSES. (My first attempt at that check stashed the test along with the doc and proved nothing, which is worth mentioning because a guard that cannot fail is worse than none.)

- `test_stack_loopback_bind.py`: 18 passed.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that newly generated Docker Compose files bind Memgraph and Qdrant ports to loopback by default.
  * Documented the `CGR_STACK_BIND_HOST` setting and risks of exposing unauthenticated services.
  * Added guidance to stop the stack before deleting and regenerating Compose files.
  * Clarified that deleting a file while containers are running does not change active port bindings.
  * Documented manual loopback edits followed by a restart as an alternative remediation.

* **Tests**
  * Expanded regression coverage for defaults, overrides, existing files, and remediation ordering.


vitali87:
@greptileai @coderabbitai Both findings confirmed and fixed in 9dcb2ff0. Greptile's P1 is the important one and I want to be direct about what I did wrong.

## Greptile: preserved compose files (P1)

Correct, and I made the dangerous version of this mistake. The issue was that the guide OVERSTATED exposure; my fix replaced it with a claim that UNDERSTATES exposure, which is the direction that actually gets someone hurt. I wrote "other hosts on the network cannot reach them" as an unconditional property without checking the upgrade path.

The code says so plainly, in `StackManager._warn_if_ports_are_public`:

> The file is rendered once and never overwritten, so an existing install keeps publishing the unauthenticated Memgraph and Qdrant endpoints on every interface.

So a user who installed before #1012 still has bare `host:container` mappings, is only warned, and is never migrated. The guide now scopes the loopback claim to a NEWLY rendered compose file, states that `~/.cgr/docker-compose.yaml` is never overwritten, and gives the remedy the warning itself names: delete and re-render, or add the `127.0.0.1:` prefix by hand.

## Both bots: the guard was blind

Also correct. Rejecting one obsolete phrase would have passed a guide claiming the services bind everywhere and are authenticated. It now asserts the positive claims: the `127.0.0.1` default, `CGR_STACK_BIND_HOST`, that the services are UNAUTHENTICATED, and that a pre-fix file is never overwritten.

I did not want to take "18 passed" as evidence that a guard made of substring checks actually guards anything, so I mutation-tested each assertion by breaking one claim at a time in the document:

| mutated claim | guard |
|---|---|
| `127.0.0.1` changed to `0.0.0.0` | **failed** |
| `CGR_STACK_BIND_HOST` renamed | **failed** |
| "UNAUTHENTICATED" softened to "protected" | **failed** |
| "never overwritten" changed to "always regenerated" | **failed** |

Four for four, so every assertion is load-bearing rather than decorative.

18 passed with the document restored.

vitali87:
@greptileai Could you review the current head 9dcb2ff0? Your check-run is against cad60ee9, before the fixes.

Both of your findings are addressed:

- **P1, preserved compose files:** the loopback claim is now scoped to a NEWLY rendered file, and the guide states that `~/.cgr/docker-compose.yaml` is never overwritten, that `cgr daemon up` only warns, and how to remediate. You were right that my rewrite understated exposure, which is the worse direction for a security document.
- **P2, blind guard:** the test now asserts the positive claims and I mutation-tested all four by breaking one claim at a time in the document; each mutation fails the guard, so none of the assertions are decorative.

All 26 CI checks green, Sonar quality gate OK with 0 issues, no unresolved threads.

vitali87:
@greptileai Third finding confirmed and fixed in 77b702d3, and it exposed a product defect beyond the documentation.

## Your finding

Correct. I copied the remediation out of `WARN_COMPOSE_PORTS_PUBLIC` without checking that it works. Traced it:

```
cgr daemon up -> stack/cli.py::up_cmd -> ensure_running()
    if current.state == RUNNING: return      # up() never runs
    self.up() -> ensure_compose_file() -> re-render + warn
```

So deleting `~/.cgr/docker-compose.yaml` while the stack is up does nothing at all: the containers keep their old bindings and `daemon up` returns before it would re-render. Someone following my instructions would have believed they had fixed an exposure they still had, which is a worse outcome than the stale paragraph I set out to correct.

Fixed in three places, since the wrong advice was in the product too, not only the guide:

- `docs/architecture/security.md`: run `cgr daemon down` FIRST, then delete and re-run, and it says explicitly why deleting while up achieves nothing.
- `WARN_COMPOSE_PORTS_PUBLIC`: the warning users actually see carried the same ineffective instruction.
- A sixth guard assertion pinning `cgr daemon down` in the remediation.

## The defect underneath, filed as #1380

Following the same early return one step further: `_warn_if_ports_are_public` is reachable ONLY through `ensure_compose_file`, which `ensure_running` skips on a healthy stack. So a pre-#1012 install that has been running for weeks is never warned at all. The warning is silent for exactly the population it exists to protect, since a long-running install is the profile of a legacy compose file.

That is a behaviour change rather than a docs fix, so it is #1380 rather than more scope here.

## Verification

- `test_stack_loopback_bind.py`: 18 passed, guard now at six assertions.
- Stack/compose/daemon suites: 66 passed.

Worth saying plainly: this is the third round on a documentation PR, and each round found something real that I had asserted without checking. First the stale claim, then my replacement understating exposure, now a remediation that does not work. Security prose gets no benefit of the doubt from me after this.

vitali87:
@coderabbitai Confirmed and fixed in 5bc6718b. You are right that presence is not the property worth guarding: "delete the file, then run `cgr daemon down`" contains the same words and leaves the reader exposed.

I did not take your literal-string suggestion, and want to say why rather than silently diverge. `assert "'cgr daemon down' FIRST, then delete the file" in doc` pins the guard to one sentence, punctuation and capitalisation included, so any rewording of the guide breaks CI for a reason that has nothing to do with security. The property is ORDER, so the test now compares positions:

```python
stop_at = doc.find("cgr daemon down")
delete_at = doc.find("delete the file")
assert stop_at < delete_at
```

Rewording stays free; reversing the steps fails. Mutation-tested it exactly that way: rewrote the paragraph to "delete the file first and then run `cgr daemon down`", keeping both phrases present, and the guard **failed** as intended. Restored, 18 passed.

If you would rather have the literal assertion for stricter pinning, say so and I will switch, but I think order-by-position is the more durable form of the same check.

@greptileai your remedy P1 thread is resolved by 77b702d3, and the deeper defect it uncovered (the warning never firing for an already-running legacy stack) is filed as #1380.

vitali87:
@coderabbitai Both findings confirmed. Fixed in 6d91632f, and fixing them convinced me the approach itself was wrong, so I changed it rather than patch again.

## What you found

1. **Scope the position check.** Right, and worse than described: my "fix" for a first-match bug HAD a first-match bug. `find("cgr daemon up")` matched an earlier sentence in the same paragraph, the one explaining that the command merely warns. The guard was measuring the wrong occurrence.
2. **Assert the full flow.** Also right: dropping the re-render step or the manual alternative both passed.

## Why I stopped patching

I fixed those and mutation-tested, and two mutants still passed, both for the same reason: the paragraph names `cgr daemon up` and `127.0.0.1:` several times for different purposes, so ANY position-based prose matching stays fragile. That was the third layer of patch on a substring search, each layer with its own occurrence bug. The tool was wrong for the job.

The guide now states the remediation as numbered steps, and the guard parses THEM:

```
1. `cgr daemon down` to stop the stack
2. delete the file `~/.cgr/docker-compose.yaml`
3. `cgr daemon up` to re-render it with the loopback bind
```

Unambiguous for a reader, and structural for the test. The parse is anchored on "remediate in this order:" because the guide contains other numbered lists (the security invariants) that a document-wide match picked up instead, which was the same class of bug one more time.

## Mutation results, all five caught

| mutation | guard |
|---|---|
| steps reversed (delete before stop) | failed |
| re-render step removed | failed |
| stop weakened to "restart at some point" | failed |
| manual `127.0.0.1:` alternative removed | failed |
| ordered list replaced with prose | failed |

The manual-alternative check needed its own scoping too: `127.0.0.1:` also appears where the guide says what to look for, so a document-wide search stayed green with the remedy deleted.

18 passed in the file, 66 across stack/compose/daemon suites.

vitali87:
@coderabbitai Both confirmed and fixed in 7a3ec4c1. The second one I initially thought was theoretical, so I tested it before answering, and you were right.

## Step 2 target

Fixed: step 2 must name the compose file, not just say "delete the file". `delete the container volumes` and a bare `delete the file` both fail the guard now.

## The manual remedy anchor

My first attempt required any line carrying "by hand" plus `127.0.0.1:`. I built a mutant to check whether your concern was reachable in practice: remedy REMOVED, replaced with "You can tweak things by hand; the `127.0.0.1:` value is shown above." It **passed** -- the decoy satisfied both markers while the actual instruction was gone. Blind, exactly as you said.

I did not anchor on `To keep local edits instead`, because pinning the sentence is what sent this guard through three rounds of rewording breakage already. Instead the line must name WHAT it applies to:

```python
"by hand" in line and f"{LOOPBACK}:" in line and "published port" in line
```

That is the instruction's meaning rather than its phrasing. Rewording stays free; losing the instruction does not.

## Mutation results

| mutation | guard |
|---|---|
| step 2 loses the filename | failed |
| step 2 deletes the wrong thing | failed |
| remedy replaced by a decoy carrying both markers | failed |
| remedy removed entirely | failed |
| steps reversed | failed |
| decoy ADDED while the real remedy stays | passed, correctly |

That last row matters as much as the others: a guard that fails on a harmless addition would be noise, so I checked that direction too rather than only checking that bad things fail.

18 passed in the file. This is the fifth round on this PR and every one of your findings has been real; the guard is materially better for it.

vitali87:
Sonar flagged `python:S9073` on my step-2 check: a composite `assert a and b` reports one failure for two distinct problems, so the message cannot say which half broke. Split in 1aff96b0.

Re-ran the mutants for that step to confirm the split kept both cases covered rather than merely satisfying the linter:

| mutation | guard |
|---|---|
| step 2 loses the filename | failed |
| step 2 deletes the wrong thing | failed |

Each now fails with its own message. 18 passed.

Worth noting for the record: my four-part merge gate reads Sonar issues as a number, and it moved from 0 to 1 on my previous push. Had I trusted the earlier reading, this would have merged with a new code smell in the file whose whole purpose is to keep a security document honest.

vitali87:
@greptileai Could you review the current head 1aff96b0? Your last check-run is against cad60ee9, five commits back, and the PR has changed shape since:

- **77b702d3** fixed your P1: the remediation now stops the stack before deleting, in the guide AND in `WARN_COMPOSE_PORTS_PUBLIC`, which carried the same ineffective instruction. The deeper defect it exposed (a running pre-#1012 stack never reaching the warning at all) is filed as #1380.
- **6d91632f** restructured the remediation into numbered steps, because three rounds of substring/position matching each had their own first-match bug. The guard parses the ordered list now.
- **7a3ec4c1 / 1aff96b0** tightened the assertions to name the compose file and the published-port prefix, then split a composite assertion Sonar flagged.

Every assertion is mutation-tested: reversing the steps, dropping the re-render, weakening the stop, deleting the manual remedy, replacing it with a decoy, and removing the ordered list all fail the guard, while a harmless addition still passes.

Gate: all checks green, Sonar quality gate OK with 0 issues, no unresolved threads. Yours is the only outstanding verdict.

vitali87:
@greptileai Confirmed and fixed in 173d8c93. This one is the best catch of the six, because the instruction LOOKED complete and did nothing.

Docker fixes a container's published ports when the container is CREATED. So editing `~/.cgr/docker-compose.yaml` to add a `127.0.0.1:` prefix leaves the running Memgraph, Memgraph Lab and Qdrant containers bound exactly as before. A reader following that path would have finished the edit, seen the corrected file, and remained fully exposed with no signal that anything was wrong.

My wording made it worse than yours describes: I had written "and restart the stack", which reads naturally as `docker restart` and does NOT rebind ports. Both places now say to run `cgr daemon down` then `cgr daemon up` to RECREATE, and say why, including that `docker restart` is insufficient.

Fixed in the guide and in `WARN_COMPOSE_PORTS_PUBLIC`, since the product warning had the same edit-only advice. A seventh guard clause requires the recreate step in the manual remedy, mutation-tested: reverting it to "restart the stack" fails, and an unrelated rewording still passes.

66 passed across stack/compose/daemon suites.

## Pattern I am taking from this PR

Six rounds, and every finding shares one shape: an instruction that is *stated* versus an instruction that *works*. The original guide named a resolved issue as tracking; my replacement claimed a property the upgrade path breaks; the remedy skipped the stop; the manual remedy skipped the recreate. Each time the words looked right and the mechanism said otherwise. On the whole PR my tests never caught any of it: prose has no execution to fail, so the only real check is tracing the mechanism the sentence describes. Thanks for doing that six times.

vitali87:
@greptileai Could you review head 173d8c93? Your check-run is against 1aff96b0, before the container-recreation fix.

That commit addresses your P1: the manual `127.0.0.1:` edit now says to run `cgr daemon down` then `cgr daemon up` to RECREATE the containers, in the guide and in `WARN_COMPOSE_PORTS_PUBLIC`, with the reason stated (Docker fixes published ports at creation, `docker restart` does not rebind).

Gate: all checks green, Sonar 0 issues, no unresolved threads. Yours is the only outstanding verdict.

Issue #1373, opened 2026-08-21
release: tag/Release/PyPI drift confuses users — v0.0.710 tagged but latest GitHub Release with binaries is v0.0.670


vitali87:
## Problem

Reported in an external field-test review of v0.0.593 ([article, Chinese](https://index.zshipu.com/ai002/post/20260817/Code-Graph-RAG-0.0.593-%E5%AE%9E%E6%B5%8B%E7%BB%99-Coding-Agent-%E5%8A%A0%E7%9F%A5%E8%AF%86%E5%9B%BE%E8%B0%B1%E5%80%BC%E4%B8%8D%E5%80%BC%E5%BE%97%E5%A4%9A%E5%85%BB%E4%B8%A4%E4%B8%AA%E6%95%B0%E6%8D%AE%E5%BA%93/)).

The reviewer found that "tag versions" and "Release versions with binary attachments" have to be tracked separately: at the time, `v0.0.593` existed as a tag while the latest full GitHub Release was `v0.0.589`. The same gap exists today — `main` is at `v0.0.710` (tag) while the latest Release is `v0.0.670`. Users landing on the repo can't easily tell which version line to trust: git tag, GitHub Release binaries, or PyPI.

This is the user-facing side of the cadence problem analyzed in #1149 (interim tags never trigger `publish.yml`; only every-50 release tags publish). #1149 focused on the supply-chain consequence for security fixes; this issue is about the everyday confusion.

## Suggested fix (either or both)

- Document the release cadence prominently (README installation section and/or `docs/getting-started/installation.md`): interim versions are tags only; binaries + PyPI releases land on the every-50 cadence; how to get the newest code (`uv tool install` from git vs PyPI).
- Or narrow the gap itself, per the options laid out in #1149 (publish every version, or publish on a version-changed push to main).

vitali87:
The documentation half shipped in PR #1381 (merged as `07ecd09a`). **Leaving this open** for the cadence question itself, which is your call, not a docs fix.

## What shipped

- README and PYPI_README now explain that three version lines exist and intentionally differ: git tags move per merge, GitHub Releases and PyPI move every 50 versions, plus immediately for any security fix. Includes the git-install command for running newer code than the latest release.
[...]

PR #1381, opened 2026-08-22
docs: explain why the newest tag runs ahead of the newest release (#1373)

vitali87:
Closes the documentation half of #1373. The cadence-change half (publish every version, or on version-changed push) is left alone: that is a release-policy decision, not a docs fix.

## Verified the mechanism before writing it

The issue says interim tags never trigger `publish.yml`. `publish.yml` triggers on `push: tags: ["v*"]`, which looks like it should fire for every tag, so I traced why it does not.

`version-bump.yml` chooses the CREDENTIAL based on whether the version is a release:

```yaml
git tag "v${NEW}"
if [ release == "true" ]; then
  git push origin "v${NEW}"                                    # SSH deploy key
else
  git push "https://x-access-token:${GITHUB_TOKEN}@..." "v${NEW}"
fi
```

GitHub does not trigger workflows for events pushed with `GITHUB_TOKEN`. So interim tags land on the remote and publish nothing, while every-50 release tags go out over the SSH key and do trigger the publish. The every-50 cadence is implemented by that credential choice.

Confirmed against reality: tags `v0.0.709` through `v0.0.716` all exist on the remote, and `publish.yml` last ran on 2026-08-18 for `v0.0.670`, which is also what PyPI serves. The behaviour is deliberate, not a stuck pipeline.

## What the docs now say

A short section in README and PYPI_README: three version lines exist, git tags move per merge while Releases and PyPI move every 50, the newest tag therefore runs ahead by design, and here is the git install command for running newer code.

## One thing I removed from my own draft

My first version included a "today" column with `v0.0.716` and `v0.0.670` in it. That is drift by construction: hardcoded version numbers in a README are exactly the staleness this issue is about, and they would be wrong within a day. The table now describes only the RELATIONSHIP, which stays true. The live PyPI version is already on the badge at the top of the README.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how Git tags, GitHub Releases, and PyPI versions differ, including when they may be published together.
  * Documented that security fixes may be released immediately outside the regular 50-version cadence.
  * Clarified which version package installers provide.
  * Simplified instructions for installing the latest code directly from the `main` branch.
  * Added guidance on release timing, publishing triggers, release cadence, and expected interim-tag behavior.

### DIFF
--- a/PYPI_README.md
+++ b/PYPI_README.md
@@ -49,6 +49,19 @@ instead of locally, set `CGR_EMBEDDING_PROVIDER=openai` with
 `OPENAI_EMBEDDING_BASE_URL` and `OPENAI_EMBEDDING_MODEL`; torch and
 transformers are then not required locally.
 
+### Which version am I getting?
+
+`pip install code-graph-rag` gives you the newest PyPI release, which is deliberately
+not the newest commit. Git tags exist for every version (one per merge), while PyPI
+uploads and GitHub Release binaries land on an every-50 cadence, so the tag on `main`
+runs ahead of the published version.
+
+To run code newer than the latest release, install from git:
+
+```bash
+pip install "git+https://github.com/vitali87/code-graph-rag@main#egg=code-graph-rag[treesitter-full]"
+```
+
 ### Prerequisites
 
 - Python 3.12+
--- a/PYPI_README.md
+++ b/PYPI_README.md
@@ -54,12 +54,13 @@ transformers are then not required locally.
 `pip install code-graph-rag` gives you the newest PyPI release, which is deliberately
 not the newest commit. Git tags exist for every version (one per merge), while PyPI
 uploads and GitHub Release binaries land on an every-50 cadence, so the tag on `main`
-runs ahead of the published version.
+usually runs ahead of the published version. A security fix is the exception: it
+ships immediately rather than waiting for the cadence.
 
 To run code newer than the latest release, install from git:
 
 ```bash
-pip install "git+https://github.com/vitali87/code-graph-rag@main#egg=code-graph-rag[treesitter-full]"
+pip install "code-graph-rag[treesitter-full] @ git+https://github.com/vitali87/code-graph-rag@main"
 ```
 
 ### Prerequisites
--- a/README.md
+++ b/README.md
@@ -108,20 +108,22 @@ Three version lines exist and they intentionally differ:
 | where | what it tracks |
 |---|---|
 | git tags | every version, one per merge |
-| GitHub Releases (binaries, signatures) | every 50th version |
-| PyPI | every 50th version |
+| GitHub Releases (binaries, signatures) | every 50th version, plus any security fix |
+| PyPI | every 50th version, plus any security fix |
 
-So the newest tag on `main` always runs ahead of the newest release, often by
-tens of patch versions. Nothing is stuck; the cadences differ by design.
+So the newest tag on `main` usually runs ahead of the newest release, often by
+tens of patch versions; they coincide only just after a release. Nothing is
+stuck, the cadences differ by design. A security fix does NOT wait for the
+cadence: it ships a release and a PyPI upload immediately.
 
 `uv tool install` and `pipx install` give you the newest PyPI version, which is the
-newest RELEASE, not the newest tag. That is deliberate: interim tags exist so every
-merge is addressable, while binaries and PyPI uploads land on the every-50 cadence.
+newest RELEASE, not the newest tag. Interim tags exist so every merge is
+addressable; binaries and PyPI uploads follow the cadence above.
 
 To run code newer than the latest release, install from git:
 
 ```bash
-uv tool install "git+https://github.com/vitali87/code-graph-rag@main#egg=code-graph-rag[treesitter-full,semantic]"
+uv tool install "code-graph-rag[treesitter-full,semantic] @ git+https://github.com/vitali87/code-graph-rag@main"
 ```
 
 You also need Docker (for Memgraph), `cmake`, and `ripgrep`. Full prerequisites, source installs, and environment setup are in the [Installation](docs/getting-started/installation.md) guide.
--- a/README.md
+++ b/README.md
@@ -101,6 +101,29 @@ uv tool install "code-graph-rag[treesitter-full,semantic]"
 pipx install "code-graph-rag[treesitter-full,semantic]"
 ```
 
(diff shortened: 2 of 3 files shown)


vitali87:
@coderabbitai @greptileai Three findings, two adopted as stated and one where the measurement disagrees with the diagnosis. Fixed in 8eb2fd10.

## Security-release exception (both of you)

Correct, and I verified it in `version-bump.yml` rather than taking it on trust. The `decide` step sets `RELEASE=true` immediately when the commit carries a GHSA id, an explicit `[security]` marker, or an associated PR labelled `security`:

```bash
if [ "$SECURITY" = "true" ]; then
  RELEASE=true
elif [ -n "$LATEST" ]; then
  ... every-50 cadence ...
```

This omission mattered more than a wording nit: a reader waiting on a security fix would have concluded they might wait up to 50 versions for it. Both files now state that a security fix ships a release and a PyPI upload immediately.

## "Always runs ahead" (CodeRabbit)

Correct, the absolute is false at a release boundary where the newest tag IS the release tag. Now "usually runs ahead ... they coincide only just after a release".

## The uv command (Greptile P1)

Adopted, but the stated reason does not reproduce. The finding says the `#egg=` form "cannot be parsed as the intended uv VCS target, preventing installation". I tested both forms:

| check | result |
|---|---|
| `uv tool install` with `#egg=` form | parses, proceeds to `git fetch` (failed only on a deliberately fake ref) |
| `uv tool install` with direct-reference form | same, parses and proceeds to `git fetch` |
| extras honoured via `#egg=`? | yes: 120 packages resolved without extras, 132 with `[treesitter-full]` |

So the old command would have worked. I still switched to `code-graph-rag[extras] @ git+...`, because that is the PEP 508 direct-reference spelling uv documents, while `#egg=` is legacy pip syntax supported for compatibility. Better documentation, not a broken command.

Flagging the discrepancy rather than quietly accepting it: a docs PR that changes a command on the grounds it is broken should be right about whether it is broken.

vitali87:
@coderabbitai Adopted the substance, placed differently, in 54304924.

You are right that the trigger mechanism is the missing piece: without it "tags move faster than releases" reads as an unexplained assertion, and the next person to see a tag with no PyPI upload will file it as a bug. That is worth writing down.

But not in README and PYPI_README. Those are read by someone deciding how to install the tool; which credential pushes a git tag is release engineering they cannot act on, and `PYPI_README.md` is the PyPI project description, where CI internals are noise. It would also rot faster than the rest of the section, since it is pinned to the current workflow implementation.

So it went to `docs/contributing.md`, next to the CI Pipeline section, where a maintainer debugging exactly this will be looking:

- the credential branch in `version-bump.yml`, quoted
- why `publish.yml` looks like it should fire for every tag and does not (GitHub does not run workflows for `GITHUB_TOKEN` pushes)
- the every-50 cadence plus the immediate security path
- an explicit note that a tag not publishing is intended, not a failure

The user-facing files keep only what a user can act on: which version they get, and the command for running newer code.

If you would still rather have a one-line pointer in README, say so and I will add a link, but I do not think the mechanism itself belongs on the PyPI page.

vitali87:
CI note: `Unit Tests (macos-latest, py3.13)` went red on 54304924. It is a flake, established before re-running:

- the commit changes ONE documentation file (`docs/contributing.md`) and nothing else, so it cannot affect Rust attribute scanning
- the same job passed on both earlier commits of this PR
- the failure is a wall-clock assertion overshooting by 19%: `assert 0.0594 < 0.05` on a shared macOS runner

Re-enqueued and it is running.

Filed as #1382, because the test has a design problem worth fixing rather than just re-running. Its ratio check (`large < small * 10`) genuinely tests linear-versus-quadratic behaviour, but the `max(..., 0.05)` floor takes over whenever the small case is fast, and 50ms of absolute wall clock on a noisy runner says nothing about complexity. The branch that fires is the one carrying the least signal. I listed three options there rather than picking one, since the choice changes what the test guarantees.

vitali87:
@coderabbitai All three fixed in af460ad8, and the first one was a genuine defect in what I wrote, not a style point.

## The snippet was broken shell

I "simplified" the workflow when quoting it and produced `if [ release == "true" ]`, which compares the literal string `release` and is ALWAYS false. Your executable check demonstrating `literal_result=false` is exactly right. A maintainer reading my version would have concluded the release branch never runs, which is the opposite of the truth, in a document whose entire job is explaining that branch.

Now quoted verbatim from `version-bump.yml`, and I verified the fidelity mechanically rather than by eye:

```
OK   if [ "${{ steps.decide.outputs.release }}" = "true" ]; then
OK   git push origin "v${{ steps.bump_version.outputs.new }}"
```

Both lines are present in the workflow AND in the doc. Paraphrasing code into prose was the error; the reason to quote is that a paraphrase can be wrong in ways the original cannot.

## The "intentional" reassurance was too broad

Your Major finding is the important one. I wrote that a tag not publishing is intended behaviour. That holds for an interim tag and is dangerously wrong for a release tag, which goes out over the SSH deploy key and MUST trigger `publish.yml`. My sentence would have told a maintainer to shrug at a genuinely failed release.

Now split explicitly: interim tag means nothing to fix, release tag means a real failure, investigate the workflow run rather than assuming the cadence explains it.

That is the second time in two PRs I have written an over-broad reassurance in a place where the failure direction is "reader stops looking". Worth naming.

## Cadence wording

Taken as suggested: "A release happens every 50 versions".

CI is green again after the macOS timing flake re-ran (diagnosed above, filed as #1382).

vitali87:
@greptileai Could you review head af460ad8? Your check-run is against b58c928a, three commits back.

Since then:

- **8eb2fd10**: your P2 (security-release exception) and the uv command form. On the uv P1 I measured both forms before switching, and the `#egg=` variant does parse and does honour extras (120 packages without, 132 with `[treesitter-full]`), so I adopted your suggestion as the documented PEP 508 spelling rather than as a broken-command fix. Details in the comment above, in case that changes your read.
- **54304924**: moved the trigger mechanism into `docs/contributing.md` instead of the README, since which credential pushes a tag is maintainer detail rather than something a user installing the tool can act on.
- **af460ad8**: CodeRabbit caught that my quoted snippet was broken shell, and that "a tag not publishing is intended" is only true for interim tags. Both fixed; the snippet is now verbatim from the workflow and verified line-by-line against it.

Gate: all checks green (the macOS failure was a wall-clock flake, re-run passes, filed as #1382), Sonar 0 issues, no unresolved threads.

Issue #1089, opened 2026-08-05
perf: the attribute-run group in the Rust declaration scan is still superlinear


vitali87:
## Summary

#1087 made the LEADING whitespace of the three Rust declaration patterns line-local, which removes the quadratic rescan on blank source. The whitespace inside the repeated attribute group still crosses newlines:

```
(?:#\[[^\]\n]*\]\s*)*
```

A long unbroken run of attribute lines with no item between them is therefore still quadratic. Measured at n=1000/2000/4000 attribute lines, identical before and after #1087, so this is a residual rather than a regression.

## Why it is filed rather than fixed

No plausible Rust file was constructed that reaches it: real attribute runs are short and always terminate in an item, and the run has to be unbroken by any declaration to accumulate. Generated code is the one place worth watching, since bindgen and prost emit long attribute sequences.

## Fix direction

The group's `\s*` has to keep crossing newlines, because an attribute on its own line above its declaration is idiomatic. Bounding the repetition, or anchoring the group so it cannot restart at every line, would remove the accumulation without changing what matches.

Verified by differential fuzzing in the #1087 review: 500k random inputs across three alphabets plus 3992 real `.rs` files, zero match differences between the old and new patterns.

PR #1385, opened 2026-08-22
test: make the Rust scan perf guard immune to runner load (#1382)

### LINKED ISSUE #1382: test: wall-clock perf assertion flakes on loaded macOS runners
Observed on PR #1381, a commit that changed ONE documentation file and nothing else:

```
FAILED test_rust_attr_run_scan_perf.py::test_unbroken_attribute_run_scans_linearly
AssertionError: (0.0015252499999860447, 0.059420374999945125)
assert 0.059420374999945125 < 0.05
```

## Why it is a flake

- The commit touched only `docs/contributing.md`, which cannot affect Rust attribute scanning.
- The same job passed on both earlier commits of the same PR.
- The assertion is wall-clock: `assert large < max(small * 10, 0.05)`. It overshot by 19% (0.0594 vs 0.05) on a shared macOS runner.

Re-running it passes, so this costs a CI cycle rather than blocking anything permanently.

## The shape of the problem

The test wants to prove the scan is LINEAR rather than quadratic, which is a good thing to protect. The ratio check (`small * 10`) expresses that well and is scale-free. The trouble is the `max(..., 0.05)` floor: when `small` is very fast (1.5ms here), the floor becomes the operative bound, and 50ms of absolute wall clock on a noisy runner is not a statement about complexity at all. A quadratic regression would blow through 10x the small time; a busy runner blows through 50ms without any regression at all.

So the failing branch is the one that carries the least signal.

## Options

1. **Raise the floor** to something a loaded runner will not hit (0.25s or so). Cheapest, keeps the test, weakens the absolute bound that was not doing much anyway.
2. **Drop the floor and keep the ratio**, using a larger `small` input so the ratio is measured against a stable baseline rather than 1.5ms of noise.
3. **Replace timing with an operation count**: count regex scan steps or input characters consumed rather than seconds. Deterministic, immune to runner load, and a more direct statement of "linear in input size".

Option 3 is the most robust and the most work; option 1 unblocks CI immediately. I have not implemented any of them since the choice affects what the test actually guarantees, and the same reasoning applies to any sibling timing assertions in this file.

Related: #1366 was a different CI flake (a compile race) fixed by locking; this one is a measurement-design issue rather than a race.

vitali87:
Closes #1382. The flake cost a CI cycle on PR #1381, where a commit touching one documentation file failed `Unit Tests (macos-latest, py3.13)`.

## Why it flaked

```python
assert large < max(small * 10, 0.05)
```

Two problems, and the second is the one that fired:

1. Each size was measured ONCE. Scheduler noise on a shared runner inflated a single sample roughly 6x (0.0594s against a ~0.009s local median).
2. The `max(..., 0.05)` floor became the operative bound whenever the small case was fast, and 50ms of absolute wall clock on a loaded runner is not a statement about complexity at all.

Worth noting the ratio branch would ALSO have failed that run (0.0594 > 10 x 0.0015). Single-sample timing was the root cause, not just the floor.

## The fix

Ratio only, from best-of-N samples:

```python
small = _best_scan_time(_attr_run(2000))
large = _best_scan_time(_attr_run(8000))
assert large < small * 10
```

Best-of-N because scheduler noise only ever ADDS time, so the minimum is the closest estimate of real cost, and one clean run among five is far likelier than one clean sample. The ratio is scale-free, so machine speed cancels, which an absolute ceiling can never do.

## Measured, not assumed

| check | result |
|---|---|
| linear scaling, 4x input | 4.2x time |
| ratio across 12 trials (best-of-5) | 3.56 to 4.92 |
| the #1089 quadratic pattern this guards against | **16.7x**, bound catches it |
| 10 consecutive runs of the new test | 0 failures |
| new test with 4 CPU-hog processes running | passes |

That last row is the direct rebuttal of the original failure: the old assertion broke on a loaded runner, and the new one survives deliberate CPU contention.

The quadratic check matters most. A perf guard that only got harder to fail would be worthless, so I confirmed the bound still separates the regression it exists to catch (16.7x) from healthy behaviour (under 5x), with headroom on both sides.

## Summary by CodeRabbit

* **Tests**
  * Updated scan performance coverage to compare repeated minimum timings across 2,000- and 8,000-line inputs.
  * Retained ratio-based checks to help verify linear performance as input size increases.
  * Removed the fixed runtime ceiling to avoid flagging consistent slowdowns unrelated to scan scalability.

### DIFF
--- a/codebase_rag/tests/test_rust_attr_run_scan_perf.py
+++ b/codebase_rag/tests/test_rust_attr_run_scan_perf.py
@@ -43,14 +43,14 @@ def test_unbroken_attribute_run_scans_linearly() -> None:
     # at 3.6-4.9x and the #1089 pattern this guards against measures 16.7x, so
     # the bound separates them with room on both sides.
     assert large < small * 10, (small, large)
-    # A deliberate second bound, and NOT the kind just removed. The ratio cannot
-    # see a uniform slowdown: if some change made every scan 100x slower, 4x
-    # input still costs 4x time and the ratio stays healthy while the scan is
-    # unusable. This catches that. It survives where the old absolute bound did
-    # not because of the margin: `large` measures ~17ms, so the ceiling sits
-    # ~120x above it, against the ~6x inflation a loaded runner produced in
-    # issue #1382.
-    assert large < 2.0, large
+    # No absolute ceiling here on purpose. One used to sit below this line to
+    # catch a UNIFORM slowdown, which the ratio cannot see: if every scan got
+    # 100x slower, 4x input would still cost 4x time. It was removed anyway,
+    # because a sustained-contention runner can delay EVERY sample, so even a
+    # best-of-N minimum is not an uncontended measurement, and no wall-clock
+    # number distinguishes "the scanner regressed" from "the runner is busy".
+    # That coverage belongs in a controlled benchmark environment rather than a
+    # unit test on shared CI (issue #1382).
 
 
 def test_attribute_block_above_declaration_still_redirects() -> None:
--- a/codebase_rag/tests/test_rust_attr_run_scan_perf.py
+++ b/codebase_rag/tests/test_rust_attr_run_scan_perf.py
@@ -43,6 +43,13 @@ def test_unbroken_attribute_run_scans_linearly() -> None:
     # at 3.6-4.9x and the #1089 pattern this guards against measures 16.7x, so
     # the bound separates them with room on both sides.
     assert large < small * 10, (small, large)
+    # A deliberate second bound, and NOT the kind just removed. The ratio cannot
+    # see a uniform slowdown: if some change made every scan 100x slower, 4x
+    # input still costs 4x time and the ratio stays healthy while the scan is
+    # unusable. This catches that. It survives where the old absolute bound did
+    # not because of the margin: `large` measures ~17ms, so the ceiling sits
+    # ~120x above it, against the ~6x inflation a loaded runner produced in
+    # issue #1382.
     assert large < 2.0, large
 
 
--- a/codebase_rag/tests/test_rust_attr_run_scan_perf.py
+++ b/codebase_rag/tests/test_rust_attr_run_scan_perf.py
@@ -24,12 +24,25 @@ def _scan_time(source: str) -> float:
     return time.perf_counter() - start
 
 
+def _best_scan_time(source: str, repeats: int = 5) -> float:
+    # Take the FASTEST run rather than one sample. Scheduler noise only ever
+    # ADDS time, so the minimum is the closest estimate of the real cost, and
+    # one clean run among several is far likelier than a single clean sample
+    # (issue #1382: a loaded macOS runner inflated one sample ~6x and failed).
+    return min(_scan_time(source) for _ in range(repeats))
+
+
 def test_unbroken_attribute_run_scans_linearly() -> None:
-    small = _scan_time(_attr_run(1000))
-    large = _scan_time(_attr_run(4000))
-    # Quadratic behaviour makes 4x the input ~16x the time; linear stays
-    # around 4x. The bound leaves generous headroom for loaded runners.
-    assert large < max(small * 10, 0.05), (small, large)
+    # Compare a RATIO, never an absolute duration: the machine's speed cancels
+    # out, which an absolute ceiling cannot do. The previous absolute floor was
+    # the branch that actually fired, and 50ms of wall clock on a shared runner
+    # says nothing about complexity.
+    small = _best_scan_time(_attr_run(2000))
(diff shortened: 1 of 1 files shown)


vitali87:
@coderabbitai Keeping this one, with the reasoning now in the code (43f9b731). Explaining rather than silently declining, since the principle behind your finding is the same one this PR is built on.

You are right that absolute wall-clock gates are fragile, and that is exactly why the tight bound went. But the two assertions catch different failures, and the ratio cannot substitute for this one:

**The ratio is blind to a uniform slowdown.** If a change made every scan 100x slower, 4x input would still cost 4x time. The ratio stays a healthy ~4 while the scan is unusable. `assert large < 2.0` is the only thing that notices.

**The margin is the difference.** Measured on this branch:

| | value |
|---|---|
| `large` (n=8000, best-of-5) | ~16.5ms |
| headroom to the 2.0s ceiling | **121x** |
| worst inflation actually observed on a loaded runner (#1382) | ~6x |

The bound that flaked sat at 0.05s against a ~0.009s measurement, roughly 5x headroom, and 6x noise walked straight through it. This one has twenty times that margin. For it to flake, a runner would have to be ~20x more degraded than the worst case that caused this issue, and at that point the job is failing on timeouts regardless.

So: dropping it would remove the only check on catastrophic blowup to guard against a failure mode with 121x of margin. I have written that tradeoff into the comment so the next person does not have to re-derive it, and so nobody removes it thinking it is a leftover of the same kind as the one deleted directly above it.

If you would still prefer it gone, I would rather replace it with something deterministic (an operation or step count) than simply delete the coverage. Say the word.

vitali87:
@greptileai Could you review head 43f9b731? You passed 665a2866; the one commit since records why the `assert large < 2.0` backstop stays after CodeRabbit asked for its removal.

Short version: the ratio check cannot see a UNIFORM slowdown (100x slower everywhere still gives a healthy ~4x ratio), and that ceiling sits ~121x above the measured ~16.5ms, against the ~6x inflation a loaded runner actually produced. The bound that flaked had only ~5x headroom. The reasoning is now in the code so nobody removes it as a leftover of the kind deleted directly above it.

Gate: all checks green, Sonar 0 issues, no unresolved threads.

vitali87:
@greptileai @coderabbitai You were both right and I was wrong. Removed in 772cab0e.

I argued for keeping `assert large < 2.0` on a margin argument: ~121x headroom against ~6x observed inflation. Greptile then produced a measurement that breaks it, a 2.087s best-of-five minimum under sustained contention against 0.007s idle, roughly 290x.

## What I did with that

I tried to reproduce it before accepting or rejecting, and could NOT: 12 busy processes against 4 CPUs gave a contended best-of-five of 0.042s, well under the ceiling.

That does not refute the finding, and the mechanism explains why. My busy loops merely compete for CPU, so the scheduler still runs the scan often. A throttled CI container is different: under a hard CPU quota the process is stalled for whole scheduling periods, so EVERY sample is delayed and taking the minimum escapes nothing. My margin argument assumed a bounded worst case; the quota case has no such bound.

I also tried to keep the coverage rather than drop it, by normalising against a fixed reference workload measured in the same process:

| condition | scan | reference | scan/ref |
|---|---|---|---|
| idle | 15.39ms | 15.33ms | 1.00 |
| contended | 29.19ms | 16.93ms | **1.72** |

The ratio moves 72% under contention, because a regex scan and a pure-Python loop do not degrade alike. So that alternative needs generous margin too and buys complexity without stability. It is not a better answer.

## Outcome

The ceiling is gone. The ratio check remains and I re-confirmed it still fails on the #1089 quadratic pattern, so the regression this test exists to catch is still caught.

What is lost is uniform-slowdown detection: if every scan became 100x slower, this test would stay green. That is real, and I have written it into the comment rather than letting it vanish silently, with the note that it belongs in a controlled benchmark environment rather than a unit test on shared CI, which is what you suggested.

Worth saying plainly: I defended a wall-clock assertion in a PR whose entire purpose was removing a wall-clock assertion, on the strength of a margin I had not stress-tested to failure. The lesson is that "large headroom" is an assumption about the worst case, not a measurement of it.

vitali87:
@greptileai Could you review head 772cab0e? Your P1 is addressed there: the wall-clock ceiling is removed, exactly as you recommended.

I could not reproduce your 2.087s locally (12 busy processes on 4 CPUs gave 0.042s), but the mechanism you describe holds under CPU-quota throttling in a way plain contention does not, and my margin argument assumed a bounded worst case that quota throttling does not have. I also measured a normalised-reference alternative before giving up on keeping the coverage: its ratio moved 72% under contention, so it was no better. Details in the comment above.

The ratio check remains and still fails on the #1089 quadratic pattern (16.7x against a bound of 10), so the regression this test guards is still caught.

Gate: all checks green, Sonar 0 issues, no unresolved threads.

PR #1408, opened 2026-08-24
fix: warn about public compose ports when the stack is already running (#1380)

### LINKED ISSUE #1380: security: a running pre-#1012 stack is never warned about its exposed ports
Found while fixing #1372, in review with Greptile. Separate from the documentation fix because it is a behaviour defect, not a wording one.

## The gap

`_warn_if_ports_are_public` is the only thing that tells a user their stack is publishing unauthenticated Memgraph, Memgraph Lab and Qdrant on every interface. It is reachable from exactly one path:

```
ensure_compose_file()  <- warns here
  ^
  up()
  ^
  ensure_running()
```

and `ensure_running` returns before any of it when the stack is already healthy:

```python
def ensure_running(self) -> StackStatus:
    current = self.status()
    if current.state == cs.StackState.RUNNING:
        logger.info(cs.MSG_STACK_ALREADY_RUNNING)
        return current          # <- up() / ensure_compose_file() never run
    self.up()
```

`cgr daemon up` goes through `ensure_running` (`stack/cli.py::up_cmd`), as does `_maybe_start_stack` in the main CLI.

## Consequence

The users who are actually exposed are the ones least likely to hear about it. Someone who installed before #1012, whose stack has been up for weeks, gets the warning only if they happen to stop the stack and start it again. A long-running install is the exact profile of a legacy compose file, so the warning is silent precisely where it matters.

The exposure is real: Memgraph Bolt, Memgraph Lab and Qdrant carry no credential, so any host on the network can read the whole code graph.

## Suggested fix

Check the compose file independently of the start path. `ensure_running` could call `_warn_if_ports_are_public(self.compose_file)` before its early return when the file exists, which keeps `ensure_compose_file` untouched and costs one YAML parse. `cgr daemon status` reporting it too would help, since that is what someone runs when the stack is already up.

Worth considering separately: whether `daemon up` should offer to migrate a legacy file rather than only warn. It is the user's file and may carry edits, which is why the current code refuses to clobber it, but a prompt is not a clobber.

## Related

- #1012 shipped the loopback default for newly rendered files.
- #1372 / PR #1379 fixed the documentation, which had claimed the exposure no longer existed at all, and corrected the remediation: deleting the compose file while the stack is UP does nothing, since the containers keep their old bindings and `daemon up` returns early.

vitali87:
## Summary

- Warn about a compose file that publishes unauthenticated services on every interface even when the stack is already running: `ensure_running`'s early-return branch, `_maybe_start_stack`'s own early return, and `cgr daemon status` now all check the file, so long-lived pre-#1012 installs finally hear about their exposure.
- Harden the warning scan itself: a non-list `ports` value (e.g. `ports: 8080`) no longer crashes the start and status paths, and a long-form mapping with `host_ip: null` is now correctly classified as an all-interfaces binding.

## Type of Change

- [x] Bug fix

## Related Issues

Fixes #1380

## Test Plan

- [x] New tests added

All new behavior is covered red-first: warning fires for a public file when the stack is already up (manager, `daemon status`, and `_maybe_start_stack` paths), stays quiet for loopback and missing files, ignores scalar and string `ports` values, and reports a null long-form `host_ip` as public.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of services whose ports are exposed on all network interfaces, including configurations with an explicitly empty host address.
  * Added warnings when a running stack has publicly accessible ports, including during status checks and automatic startup.
  * Improved handling of invalid, unexpected, or unreadable port configuration formats.

* **Tests**
  * Added regression coverage for publicly exposed ports across stack startup and status workflows.
  * Added coverage for malformed and long-form Compose port configurations.

### DIFF
--- a/codebase_rag/cli.py
+++ b/codebase_rag/cli.py
@@ -201,6 +201,9 @@ def _resolve_active_projects(projects: str | None, default_project: str) -> list
 def _maybe_start_stack() -> None:
     mgr = StackManager()
     if mgr.status().state == StackState.RUNNING:
+        # This early return bypasses ensure_running, so it needs its own
+        # public-port check for a stack that is already up (issue #1380).
+        mgr.warn_if_ports_are_public()
         return
     try:
         mgr.ensure_running()
--- a/codebase_rag/stack/cli.py
+++ b/codebase_rag/stack/cli.py
@@ -57,7 +57,11 @@ def down_cmd() -> None:
 
 @cli.command("status", help=ch.CMD_DAEMON_STATUS, short_help=ch.CMD_DAEMON_STATUS)
 def status_cmd() -> None:
-    _print_status(StackManager())
+    mgr = StackManager()
+    # Status is what a user runs against an already-up stack, the one case the
+    # start path's warning never covers (issue #1380).
+    mgr.warn_if_ports_are_public()
+    _print_status(mgr)
 
 
 @cli.command("restart", help=ch.CMD_DAEMON_RESTART, short_help=ch.CMD_DAEMON_RESTART)
--- a/codebase_rag/stack/manager.py
+++ b/codebase_rag/stack/manager.py
@@ -22,8 +22,12 @@ def _publishes_on_all_interfaces(mapping: object) -> bool:
     even though it names a host.
     """
     if isinstance(mapping, dict):
+        # YAML parses `host_ip: null` (or a bare `host_ip:`) as None, which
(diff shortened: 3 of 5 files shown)


vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

vitali87:
@greptile review

PR #1430, opened 2026-08-25
fix: re-parse ast-grep and document files on watched edits

### LINKED ISSUE #1427: Realtime watcher re-parses only tree-sitter files; ast-grep and document tiers go stale on edit
## What

`realtime_updater.py` re-parses a changed file only when it resolves to a **tree-sitter** `LanguageSpec` with a loaded parser. Files handled by the other two tiers are never re-extracted on edit, so their definition nodes go stale until the next full index.

At `realtime_updater.py:290-309`:

```python
if event.event_type in (EventType.MODIFIED, EventType.CREATED):
    lang_config = get_language_spec(path.suffix)
    if (
        lang_config
        and isinstance(lang_config.language, SupportedLanguage)
        and lang_config.language in self.updater.parsers
    ):
        ... definition_processor.process_file(...)

    # Create File node for ALL files (code and non-code like .md, .json, etc.)
    self.updater.factory.structure_processor.process_generic_file(path, path.name)
```

Compare the batch path, `graph_updater.py:_process_single_file` (~:2478), which dispatches across three tiers: tree-sitter → dependency manifest → ast-grep → document. The realtime path implements only the first.

## Impact

- **ast-grep tier** (`.rb`, `.kt`, `.swift`, `.ex`, `.hs`, `.sol`, `.sh`, `.nix`): editing a Ruby file under `cgr start --watch` leaves its previous `Function`/`Class` nodes in place. Renaming a method leaves the old node and never adds the new one.
- **Document tier** (`.md`, `.markdown`, added in #1426): editing a Markdown file refreshes its `File` node but not its `Section` nodes, so headings added, renamed, or removed while watching are invisible until re-index.

Both are silent: the file **is** processed (it gets its `File` node), so nothing logs a skip and the graph looks updated.

## Not a regression

This predates the document tier — the ast-grep tier has had the same gap since it landed. #1426 adds a second tier with the behaviour rather than introducing it, which is why this is filed separately instead of being folded into that PR.

## Suggested fix

Factor the tier dispatch out of `GraphUpdater._process_single_file` into one helper both callers use, so a new tier is wired into batch and realtime indexing in a single place. That is better than adding two more `elif` branches to the watcher, which would leave the same trap for the next tier.

Worth checking as part of it: the delete/rename path. `Section` nodes hang off `Module` via `CONTAINS_SECTION` and nest recursively, so stale-node cleanup on edit needs to walk that chain the way it walks `DEFINES`.

## Tests

`codebase_rag/tests/test_realtime_event_filtering.py` already covers the "non-code file still gets a File node" behaviour (`test_markdown_file_creates_file_node:103`). The missing coverage is the positive case: edit a `.rb` and a `.md` under the watcher and assert the definition/section nodes actually change.

vitali87:
## Summary

Under `cgr start --watch`, editing a Ruby or Markdown file **emptied it in the graph** instead of refreshing it — and the handler logged `Graph updated successfully` while doing so.

The watcher's Step 1 always deletes the file's `Module` and everything hanging off it (`DEFINES` for the ast-grep tier, `CONTAINS_SECTION` for the document tier). Step 3 then re-parses **only** tree-sitter languages, so a file belonging to either other tier lost its definitions with nothing to re-create them. A full re-index restored everything, which is probably why this went unnoticed.

## Verified, not assumed

Dispatch comparison using the real registries — the exact condition from `realtime_updater.py` Step 3 against what `_process_single_file` would do:

```
file           realtime re-parses?    batch would use
m.py           True                   tree-sitter
app.rb         False                  ast-grep tier
d.md           False                  document tier
Main.kt        False                  ast-grep tier
doc.markdown   False                  document tier
```

End to end against a **real** `GraphUpdater` with a recording ingestor, indexing a file and then dispatching one `FileModifiedEvent` that adds a symbol:

```
                          before fix          after fix
markdown  Section    1 -> 0   [LOST]      1 -> 2   [OK]
ruby      Function   1 -> 0   [LOST]      1 -> 2   [OK]
```

Reverting the fix also fails 3 of the 5 new tests, so they are real regression coverage rather than tests that pass either way.

## The fix

The tier dispatch moves out of `GraphUpdater._process_single_file` into `GraphUpdater.process_with_secondary_tier`, which the batch path and the watcher now share. That is the point: a tier added later gets wired into **both** callers at once, instead of only into batch indexing — which is exactly how this bug arose when the ast-grep tier landed, and how the document tier inherited it in #1428.

The watcher calls it on the `else` branch of the existing tree-sitter check, so a Python file is not parsed twice (covered by a test).

## Scope

Affects every ast-grep tier language — Ruby, Kotlin, Swift, Elixir, Haskell, Solidity, Bash, Nix — plus Markdown via the document tier. Tree-sitter languages were always fine.

## Test Plan

- New `TestSecondaryTierReparse` in `codebase_rag/tests/test_realtime_event_filtering.py` (5 tests): Markdown edit, Ruby edit, created file, tree-sitter file taking the other branch, and deletion not re-parsing.
- 303 passed across every suite touching realtime updating, graph updating, the tiers, incremental indexing, and the document tier.
- Full suite reached 70% with zero failures before a local interruption; CI runs the complete matrix.
- `ruff check`, `ruff format --check` clean. `ty check` shows one pre-existing warning at `graph_updater.py:2750` (`fetch_all` on `Unknown | IngestorProtocol`), unrelated to this change and present on a clean tree.

## Related Issues

Fixes #1427.

Found while building #1428 and filed rather than folded in, which is why it lands separately.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Test updates

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated file-change handling so edits and newly created files in supported secondary formats are properly reprocessed.
  * Prevented Ruby definitions and Markdown sections from disappearing from search results after edits.
  * Ensured unsupported or unrecognized file types remain represented in the project index.
  * Preserved existing behavior for primary parsed languages and file deletions.
  * Improved refresh reliability for files handled by secondary language and document processing.
  * Improved handling of cached syntax information during file refreshes.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2470,6 +2470,12 @@ class GraphUpdater:
         file_bytes: bytes | None = None,
         pre_parsed: tuple[Node, dict[str, list] | None] | None = None,
     ) -> None:
+        """Ingest one file through the first tier that claims it.
+
+        Tries tree-sitter, then dependency manifests, then the secondary
+        tiers; every file gets a File node regardless, so an unclaimed
+        extension is still represented in the graph.
+        """
         if self._cpp_frontend_covered:
             rel = cached_relative_path(filepath, self.repo_path).as_posix()
             if rel in self._cpp_frontend_covered:
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2533,6 +2533,7 @@ class GraphUpdater:
         return False
 
     def _ast_for(self, file_path: Path) -> Node | None:
+        """The cached AST root for a file, or None when it is not cached."""
         entry = self.ast_cache.load(file_path)
         return entry[0] if entry else None
 
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2500,16 +2500,31 @@ class GraphUpdater:
                 self._parsed_files.append((filepath, language))
         elif self._is_dependency_file(filepath.name, filepath):
             self.factory.definition_processor.process_dependencies(filepath)
-        elif self.ast_grep_tier.handles(filepath.suffix):
-            self.ast_grep_tier.process_file(
-                filepath, self.factory.structure_processor.structural_elements
-            )
-        elif self.document_tier.handles(filepath.suffix):
-            self.document_tier.process_file(
-                filepath, self.factory.structure_processor.structural_elements
-            )
+        else:
+            self.process_with_secondary_tier(filepath)
 
         self.factory.structure_processor.process_generic_file(filepath, filepath.name)
+
+    def process_with_secondary_tier(self, filepath: Path) -> bool:
+        """Parse a file with whichever non-tree-sitter tier claims it.
+
+        Shared with the file watcher, which re-parses one changed file and
+        would otherwise handle only tree-sitter languages: it deletes a
+        file's Module (and everything hanging off it) before re-parsing, so
+        a tier it does not know about loses its symbols entirely rather than
+        merely going stale (issue #1427).
+
+        Returns True when a tier took the file, so the batch path can keep
+        its if/elif chain.
+        """
+        structural_elements = self.factory.structure_processor.structural_elements
+        if self.ast_grep_tier.handles(filepath.suffix):
+            self.ast_grep_tier.process_file(filepath, structural_elements)
+            return True
+        if self.document_tier.handles(filepath.suffix):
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

Issue #1426, opened 2026-08-25
Markdown structural tier: parse headings into nested graph sections (split from #164)


vitali87:
## Why

Issue #164 asks for Markdown support so the agent stops multiplying and desynchronising `.md` plan files. Its acceptance-criteria list, though, fuses two projects with very different risk profiles:

**Structural parsing** (parse to AST, identify header nodes, identify links to other files, chunk by header, reconstruct sections) — mechanical, testable, and identical in shape to every other language already in `LANGUAGE_SPECS`.

**Agent behaviour** (file-purpose awareness, deduplication logic, update synchronisation, plan consolidation, contextual file selection, metadata tagging) — heuristics and prompt/agent policy, with no deterministic oracle, sitting on top of whatever the first half produces.

The owner's own comment on #164 asks whether the whole thing is scope creep. This issue proposes splitting off the half that is not: the structural tier, which is a prerequisite for the behavioural half anyway and is worth having on its own (`.md` files become greppable/queryable structure in the graph instead of being invisible to it).

## Feasibility — verified, not assumed

`tree-sitter-markdown` 0.5.1 is on PyPI (`requires_python >=3.9`) with wheels covering the entire CI matrix — `macosx_10_9_x86_64`, `macosx_11_0_arm64`, `manylinux_2_17/2_28` x86_64 + aarch64, `musllinux_1_2` x86_64 + aarch64, `win_amd64`, `win_arm64`. It therefore fits the existing PyPI-grammar pattern in the `treesitter-full` extra; no git submodule and no `cgr language add-grammar` vendoring is needed.

Parsing the grammar against a representative plan file shows it already does the structural work the issue describes:

```
document
  section                     # <- sections NEST natively
    atx_heading
      atx_h1_marker  '#'
      inline         'Project Plan'
    paragraph
    section                   # <- '## Phase One' is a CHILD of the h1 section
      atx_heading
        atx_h2_marker '##'
        inline        'Phase One'
      section                 # <- '### Subtask A' nests one deeper
        atx_heading
```

- **Header dependency is free.** #164 asks that "header level2 is dependent on its parent header level1". The grammar's `section` nodes already nest exactly that way, so parent/child containment maps onto the existing `DEFINES`-style containment model without inventing a level-tracking pass.
- **Heading level and text** come from the `atx_h1_marker`…`atx_h6_marker` child and the sibling `inline` child.
[...]

PR #1437, opened 2026-08-25
docs: note that document module names carry their extension

vitali87:
## Summary

#1426 made a document's qualified name keep its suffix — `docs/guide.md` becomes `<project>.docs.guide_md` — so that `.md` and `.markdown` files sharing a stem cannot merge onto one `Module` node along with their identically-named sections.

That was recorded in the #1426 PR body, which is invisible after merge, and nowhere a user would look. This adds it to the "Document Support (document tier)" section of `docs/architecture/language-support.md`.

## Why it matters to users

A graph indexed before document support existed holds **unsuffixed** document module names and keeps them until re-indexed. A saved query written against the old shape silently stops matching a re-indexed module, with nothing explaining why.

## Consistency

Matches the phrasing `duplicates.md` already uses for the same class of change ("Graphs indexed before duplicate detection existed carry no fingerprints; re-index once with `--clean` to backfill them"), so the tiers read alike.

Verified `make readme` leaves it alone: `language-support.md` is a generator target, but the generator only replaces marked `<!-- SECTION -->` blocks and this prose sits outside them. Ran it and re-diffed — no change.

## Related

Retrospective documentation for #1426. Raised by the `fix-doc-qualified-name` session, which hit the identical gap for the ast-grep tier in #1435 and flagged that the document tier had it too.

## Type of Change

- [x] Documentation update

## Summary by CodeRabbit

* **Documentation**
  * Documented updated qualified-name handling for Markdown modules, including extension preservation and collision avoidance.
  * Clarified compatibility considerations for existing graphs and incremental synchronization.
  * Added guidance for rebuilding graphs and explained the effects of clean rebuild options.

### DIFF
--- a/docs/architecture/language-support.md
+++ b/docs/architecture/language-support.md
@@ -95,9 +95,14 @@ so a saved query written against the old names stops matching once the graph
 is rebuilt. An incremental sync will not move it: `.md` files were already
 hashed before this tier existed, so `--update-graph` sees them unchanged and
 skips them, leaving the graph exactly as it was. Rebuilding needs
-`cgr start --clean`, which is the remedy `cgr` itself recommends when parser
-code has changed — note that it clears **every** project in the shared graph,
-so run it only when that graph holds just this repository.
+`cgr start --clean --update-graph`: `--clean` on its own wipes the database,
+clears the embeddings, drops the hash cache and returns without indexing
+anything, so it would leave you with an empty graph rather than renamed
+document modules. Both flags together wipe and then re-index in one pass —
+and because the wipe drops the hash cache, the re-index treats every file as
+new, which is what re-emits the documents under their suffixed names. Note
+that the wipe clears **every** project in the shared graph, so run it only
+when that graph holds just this repository.
 Requires the `treesitter-full` extra.
 
 ## Language-Agnostic Design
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
## Why this says `--clean` and not `--update-graph`

Recording this because the gentler flag is the intuitively appealing edit, and making it would silently break the instruction.

An earlier draft of this paragraph said only "re-index to move them", naming no flag. That reads as `cgr start --update-graph`, which **does nothing here**: `.md` files were already hashed before the document tier existed (they always received `File` nodes), so the incremental path sees the content hash unchanged, skips them (`graph_updater.py:2288-2295`), and never runs the tier that would create `Section` nodes. The user gets a success message and an unchanged graph.

`cgr` already states this case by name (`logs.py:846`):

> Parser code changed since this graph was built. Incremental sync keeps results from the old parser for unchanged files, so the graph may be stale. Run 'cgr start --clean' to rebuild it from scratch.

Adding the document tier changes the parser fingerprint, so `_warn_if_parser_changed` emits exactly that — and it only warns, it does not force a re-parse.

So `--clean` is required, and the paragraph states its blast radius (it clears every project in the shared graph) rather than leaving the user to discover it. Softening the flag would produce an instruction that runs, reports success, and changes nothing — a worse failure than a destructive one, because it is silent.

Related: #1438 tracks the same missing blast-radius warning on `duplicates.md:378` and `:506`, where `--clean` is likewise correct and likewise unqualified.

vitali87:
@greptileai review

PR #1439, opened 2026-08-25
docs: state what --clean costs when backfilling fingerprints

### LINKED ISSUE #1438: duplicates.md recommends --clean for fingerprint backfill without warning it deletes every other project
## What

`docs/guide/duplicates.md:378` says:

> Graphs indexed before duplicate detection existed carry no fingerprints; re-index once with `--clean` to backfill them.

`--clean` is more destructive than that sentence suggests (`cli_help.py:264`):

> **DESTRUCTIVE: Delete every project from the shared graph** and clear the selected repository's sync cache. With `--update-graph`, rebuild after deletion. Asks for confirmation when other projects would be destroyed; use `--yes` to skip the prompt.

So a user with several projects in one shared graph, following this instruction to fix a **fingerprint** problem in **one** repository, destroys the indexed data for all the others. The confirmation prompt is the only thing standing between the doc's advice and that outcome, and `--yes` removes it.

## The advice is technically correct, which is why it survived

I checked whether the gentler flag would do, and it would not. Fingerprints are stamped at parse time (`parsers/ast_fingerprint.py` via `function_ingest.py`), so any re-parse backfills them — but the incremental path **skips unchanged files** (`graph_updater.py:2291-2295`, hash match → `skipped_count += 1; continue`). An old graph's files have not changed, so `--update-graph` alone re-parses nothing and backfills nothing. There is no force-reparse flag.

`--clean` really is the mechanism. The problem is purely that the doc states the remedy without its blast radius.

## Suggested fix

Keep the recommendation, add the consequence and the safe path — something like: `--clean` clears every project in the shared graph, so confirm the prompt only if this graph holds just this repository; otherwise re-index the affected project into its own graph. Worth checking whether a scoped variant is feasible, but the doc change stands on its own.

The same sentence appears again at `duplicates.md:506`.

## How this surfaced

The `fix-doc-qualified-name` session cited this line to me as a phrasing precedent while I was writing a migration note for #1437. It went to verify its own use of the precedent, found `--clean` was wrong for that purpose, and switched to `--update-graph`. My paragraph names no flag so it was unaffected, but the original line remains.

Worth recording the general shape: a convention copied from the codebase is evidence about **style**, not about **correctness**, and citing it as precedent propagates whatever was never checked the first time.

vitali87:
## Summary

`duplicates.md` told users to run `--clean` to backfill fingerprints into an older graph, without mentioning that `--clean` **deletes every project in the shared graph**:

> DESTRUCTIVE: Delete every project from the shared graph and clear the selected repository's sync cache... Asks for confirmation when other projects would be destroyed; use `--yes` to skip the prompt.

A user with several projects indexed, following that instruction to fix a fingerprint problem in **one** repository, loses the others. The confirmation prompt is the only guard, and `--yes` removes it.

## `--clean` stays — it is required

I checked whether the gentler flag would do, and it would not:

- Fingerprints are stamped **at parse time** (`parsers/ast_fingerprint.py` via `function_ingest.py`), so any re-parse backfills them.
- The incremental path **skips files whose content hash is unchanged** (`graph_updater.py:2288-2295` → `skipped_count += 1; continue`). An old graph's files have not changed.
- There is no force-reparse flag.

So `--update-graph` alone re-parses nothing and backfills nothing. Softening the flag would produce an instruction that runs, reports success, and changes nothing — a worse failure than a destructive one, because it is silent. The doc now says this explicitly so a future editor does not fix it that way.

`cgr` takes the same position elsewhere (`logs.py:846`): when parser code changes under an existing graph, it prints *"Incremental sync keeps results from the old parser for unchanged files... Run 'cgr start --clean' to rebuild it from scratch."*

## What I did NOT change, and why

The other "until re-indexed" sentence in this file (line ~506, editor links) is **correct as written** and needs no flag. That feature depends on the Project node's `root_path`, which `run()` writes unconditionally before any file processing (`graph_updater.py:853`), outside the per-file hash skip. A plain `--update-graph` restores it. Verified rather than assumed — the two sentences look alike and only one has the problem.

Confirmed `_confirm_destructive_clean` (`cli.py:272`) exists and is wired into both call sites, so the prompt the doc now mentions is real.

## Test Plan

- `make readme` is a no-op against this prose (it sits outside the marked `<!-- SECTION -->` blocks). Ran it and re-diffed.

## Related

Fixes #1438. Surfaced when the `fix-doc-qualified-name` session cited this line to me as a phrasing precedent while I wrote #1437, then found the precedent was wrong for its own case.

## Type of Change

- [x] Documentation update

## Summary by CodeRabbit

* **Documentation**
  * Updated prerequisite instructions to use `--update-graph` by default.
  * Clarified behavior for missing caches, wiped-graph caches, and populated project caches.
  * Explained when re-indexing happens automatically and when `--clean --update-graph` is required.
  * Documented that `--clean` affects all shared-graph projects, `--yes` skips confirmation, and `--clean` alone leaves an empty graph.
  * Clarified that unchanged files are not reparsed during incremental updates.

### DIFF
--- a/docs/guide/duplicates.md
+++ b/docs/guide/duplicates.md
@@ -375,8 +375,13 @@ cgr daemon up
 cgr start --repo-path /path/to/your/repo --update-graph --clean
 ```
 
-Graphs indexed before duplicate detection existed carry no fingerprints;
-re-index once with `--clean` to backfill them.
+Graphs indexed before duplicate detection existed carry no fingerprints.
+Backfilling them needs `--clean`: fingerprints are stamped while a file is
+parsed, and an incremental sync skips files whose content has not changed, so
+`--update-graph` alone re-parses nothing and backfills nothing. Note that
+`--clean` deletes **every** project in the shared graph, not just this one —
+it prompts before destroying others, so confirm only when that graph holds
+this repository alone.
 
 ## Basic Usage
 
--- a/docs/guide/duplicates.md
+++ b/docs/guide/duplicates.md
@@ -375,13 +375,22 @@ cgr daemon up
 cgr start --repo-path /path/to/your/repo --update-graph --clean
 ```
 
-Graphs indexed before duplicate detection existed carry no fingerprints.
-Backfilling them needs `--clean`: fingerprints are stamped while a file is
-parsed, and an incremental sync skips files whose content has not changed, so
-`--update-graph` alone re-parses nothing and backfills nothing. Note that
+Graphs indexed before duplicate detection existed carry no fingerprints, and
+fingerprints are stamped while a file is parsed, so backfilling them means
+re-parsing. Which command does that depends on the hash cache:
+
+- **No hash cache for the repo** (never synced from this machine, or the
+  cache was cleared): `--update-graph` re-parses everything, because a file
+  is skipped only when the cache already holds a matching hash. This
+  backfills fingerprints without touching any other project.
+- **A populated hash cache**: `--update-graph` skips every unchanged file and
+  so backfills nothing. Use `--clean --update-graph` together, which drops
+  the cache and re-indexes in one pass.
+
 `--clean` deletes **every** project in the shared graph, not just this one —
 it prompts before destroying others, so confirm only when that graph holds
-this repository alone.
+this repository alone. `--clean` on its own wipes without re-indexing and
+leaves an empty graph; pair it with `--update-graph`.
 
 ## Basic Usage
 
--- a/docs/guide/duplicates.md
+++ b/docs/guide/duplicates.md
@@ -372,25 +372,35 @@ Index the repository first, so the graph exists in Memgraph:
 
 ```bash
 cgr daemon up
-cgr start --repo-path /path/to/your/repo --update-graph --clean
+cgr start --repo-path /path/to/your/repo --update-graph
 ```
 
 Graphs indexed before duplicate detection existed carry no fingerprints, and
 fingerprints are stamped while a file is parsed, so backfilling them means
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
Heads-up: I duplicated part of this by mistake. I claimed #1438, wrote the same fix, and only found this PR when my push was rejected because the branch name already existed. My error — I claimed the issue without first checking whether a PR was already open against it. Claim released, branch dropped, and I am not touching this further.

One thing my version caught that this diff does not, offered rather than pushed:

The **Prerequisites** block just above the sentence still reads

```bash
cgr start --repo-path /path/to/your/repo --update-graph --clean
```

That is the copy-paste line for a *first-time* index, where `--clean` is unnecessary, and it is the more likely thing a reader actually runs. Every other guide's prerequisite uses `--update-graph` alone — `docs/index.md:40`, and `docs/guide/multi-project.md:19-20`, which is notable because that page is specifically about several projects sharing one graph, exactly the case where `--clean` is destructive.

So `duplicates.md` is the outlier, and the fenced command arguably matters more than the prose beneath it: someone skimming for the setup command never reaches the caveat.

Dropping `--clean` from that block would be a one-line change on top of what you have. Entirely your call whether it belongs here or in a follow-up.

Separately, `docs/guide/graph-export.md:14` also uses `--update-graph --clean`. I have not verified whether a full rebuild is intended there for a complete export, so I am flagging it rather than asserting it is wrong.

vitali87:
@greptileai review

vitali87:
@greptileai review

Issue #1447, opened 2026-08-26
Nine relationship types export as RELATIONSHIP_TYPE_UNSPECIFIED in the protobuf index


vitali87:
The protobuf relationship enum in `codec/schema.proto` has drifted behind `RelationshipType` in `codebase_rag/constants/graph.py`. Nine types exist in Python and not in the proto schema:

```
EXPOSES, FLOWS_TO, HAS_SMELL, HAS_VULNERABILITY, IMPLEMENTS_PATTERN,
INSTANTIATES, LINKS_TO, REFERENCES, RESOLVES_TO
```

`ProtobufFileIngestor.ensure_relationship_batch` (`codebase_rag/services/protobuf_service.py:153`) resolves the type by `getattr` on the generated enum and falls back to `RELATIONSHIP_TYPE_UNSPECIFIED` when the name is absent, logging a warning:

```python
resolved = getattr(pb.Relationship.RelationshipType, rel_type, None)
if resolved is None:
    logger.warning(ls.PROTOBUF_UNKNOWN_REL_TYPE.format(rel_type=rel_type))
    resolved = pb.Relationship.RelationshipType.RELATIONSHIP_TYPE_UNSPECIFIED
```

So the edge is written, but untyped. Anything reading the index back cannot tell a `FLOWS_TO` from an `EXPOSES` from an `INSTANTIATES` — they are all `UNSPECIFIED`. This affects the `cgr index`, `cgr verify-index` and `cgr diff-index` paths.

`FLOWS_TO` is the notable one: it is the newest and subtlest edge type, and it is invisible in any protobuf export today.

## Why it drifted

Nothing gates the two enums against each other. Adding a member to the Python enum is a one-line change with no signal that the proto schema also needs it, and the fallback is a warning rather than an error, so the loss is silent in normal use.

## Fix

Add the missing members to `codec/schema.proto`, regenerate the bindings, and add a parity test asserting that every `RelationshipType` member has a proto counterpart. The test is the part that matters: without it the next added relationship type drifts the same way. Discovered while reviewing #1445, which adds `LINKS_TO` and would otherwise have made the count ten.

vitali87:
Fixed and merged in #1445. Verified against current `main` rather than assumed:

```
missing from proto: NONE
orphaned in proto:  NONE
```

[...]

Issue #386, opened 2026-02-27
Support embedded databases as alternatives to Memgraph


vitali87:
## Task Description
Explore supporting embedded/lightweight databases as alternatives to Memgraph, removing the requirement for Docker and a running database server. This would significantly lower the barrier to entry for new users.

Suggested candidates:

| Database | Type | Notes |
|----------|------|-------|
| [Kuzu](https://github.com/kuzudb/kuzu) | Graph + Vector | Embedded, Cypher compatible, drop-in Neo4j alternative |
| [LanceDB](https://github.com/lancedb/lancedb) | Vector | Embedded, serverless |
| [Chroma](https://github.com/chroma-core/chroma) | Vector | Embedded, simple API |
| [Turso](https://github.com/tursodatabase/turso) | SQL + Vector | SQLite based, embedded |

Kuzu is particularly interesting because it supports Cypher queries natively, which could minimize changes to the existing query pipeline.

## Acceptance Criteria
- [ ] Evaluate at least one embedded graph DB (Kuzu) for compatibility with existing Cypher queries
- [ ] Evaluate at least one embedded vector DB for semantic search
- [ ] Prototype a backend that works without Docker
- [ ] Abstract the storage layer so users can choose between Memgraph and an embedded alternative

## References
- Original discussion: #61 (comment by @qdrddr)

## Priority
- [ ] High
- [x] Medium
- [ ] Low

## Estimated Effort
- [ ] Small (< 2 hours)
- [ ] Medium (2-8 hours)
- [x] Large (> 8 hours)

qdrddr:
Kuzu is dead. Long live https://github.com/LadybugDB/ladybug (a fork).

Turso is working on a graph extension (not released yet).

lvca:
Please consider also https://arcadedb.com open source dbms, Apache 2 license, graphs and vector as multi-model.

qdrddr:
A Fast LPG Graph DB with Cypher language.
Can run without a Docker container. Though it still requires a network connection and is technically not in-process / embeddable at the moment. The TurungDB team is working on releasing a fully **embeddable version soon**.
https://github.com/turing-db/turingdb


```shell
uv pip install turingdb
turingdb -demon
```
```python
# Set working graph
client.set_graph("mygraph")

# Create a new change on the graph
change = client.new_change()

# Checkout into the change
client.checkout(change=change)

# Create a node Person (Jane) - Edge (knows) - node (John)
[...]

PR #1453, opened 2026-08-26
fix: finding nodes were dropped from the protobuf index

### LINKED ISSUE #1452: Declare the graph node schema explicitly, so a schema-requiring backend can be built against it
Prerequisite for an embedded graph backend (#386), and a standalone correctness improvement.

## The gap

`NodeLabel` declares 21 node labels and `_NODE_LABEL_UNIQUE_KEYS` declares each one's unique key, so **identity** is explicit. What is not declared anywhere is each label's **property set**. A `Module` carries `qualified_name`, `name`, `path`, `decorators`, `rust_cfg_test_mods`, `flow_covered`, `generated`, `generator` and more, but nothing states that. The schema exists only as whatever the parsers happen to pass to `ensure_node_batch`.

Memgraph does not care, because it is schemaless and `SET n += row.props` accepts any map. So the omission is invisible in normal use.

## Why it matters now

Measured on #386: Kuzu requires an explicit table per label and **rejects any property not in it** (`Binder exception: Cannot find property not_in_schema for n`). It also has no `SET n += map`, so properties must be assigned individually — which requires knowing the property list per label. An embedded graph backend is therefore blocked on this, not on Cypher dialect differences.

## Why it matters independently of Kuzu

This is the same drift that let nine relationship types diverge from `codec/schema.proto` unnoticed (#1447): `RelationshipType` grew, the proto did not, and every one of those edges exported as `RELATIONSHIP_TYPE_UNSPECIFIED` with only a `logger.warning`. Nothing could catch it because nothing declared what the two were supposed to agree on. A parity test fixed that instance; the node side has the same shape and no such test.

Concretely, today:

- A typo in a property name creates a new property rather than an error.
- A property emitted by one parser and not another is indistinguishable from a property that is genuinely optional.
- `codec/schema.proto` already declares node payload messages (`Module`, `Class`, `Function`, ...), so a second, partial declaration of the node schema **already exists** and nothing checks it against what the parsers emit.

That last point is the strongest argument that this is worth doing regardless of #386: there are already two sources of truth and no gate between them.

## Work

1. Declare each `NodeLabel`'s property set and types in one place, with optionality explicit.
2. Add a parity test in both directions against `codec/schema.proto`'s node messages, mirroring the relationship parity test added in #1447. Assert parity rather than containment, and include a non-vacuity control so neither direction can pass on an empty read.
3. Consider validating emitted properties against the declaration in debug/test builds, so an undeclared property fails loudly where it is emitted rather than silently downstream.

## Acceptance

- [ ] Every `NodeLabel` has a declared property set with types and optionality.
- [ ] A test asserts parity between that declaration and `codec/schema.proto` node messages, in both directions, with a non-vacuity control.
- [ ] The declaration is sufficient to generate a Kuzu `CREATE NODE TABLE` statement per label (demonstrated, not necessarily wired up).

## Not in scope

Building the Kuzu backend itself (#386), or changing the Memgraph write path. This issue only makes the schema explicit and checked.

vitali87:
Fixes #1452.

## The bug

`Pattern`, `CodeSmell` and `SecurityIssue` — the ast-grep finding labels from #413 — have no payload message in `codec/schema.proto` and no `LABEL_TO_ONEOF_FIELD` entry. Measured before the fix:

```
NodeLabel count:        21
proto payload count:    18
in NodeLabel not proto: [CodeSmell, Pattern, SecurityIssue]
```

This is the node-side twin of #1447, but the failure mode is worse. An unknown *relationship* type still writes an edge, untyped. An unmapped *node* label hits this in `ensure_node_batch` (`protobuf_service.py:105`):

```python
payload_field_name = LABEL_TO_ONEOF_FIELD.get(node_label)
if not payload_field_name:
    logger.warning(ls.PROTOBUF_NO_ONEOF_MAPPING.format(label=label))
    return
```

It returns. The node never reaches the index. Every security finding, code smell and matched pattern was **silently absent** from `cgr index`, `verify-index` and `diff-index` output — with only a warning nobody sees during a run, and nothing downstream able to distinguish that from a repository that genuinely has no findings.

## Verified end to end

Not just through the tests. Ingesting a `SecurityIssue` and reading the written index back:

```
nodes stored: 1
 payload: security_issue -> sql-injection
```

Before the change the same call stored nothing at all.

## The parity test is the durable part

It asserts three separate things, because they fail identically and silently:

1. every `NodeLabel` has a proto payload message;
2. every proto payload message has a `NodeLabel` (the paired direction — otherwise this asserts containment, not parity, and an orphaned message nothing can populate would pass);
3. every `NodeLabel` has a **oneof mapping** — a message that exists but is unmapped is dropped exactly as an absent one is, so schema parity alone is not enough.

Plus a control asserting the descriptor read is non-empty, so none of the set comparisons can pass vacuously if the oneof lookup ever stops returning fields.

`_NOT_IN_PROTO` is empty and exists so that a future deliberate omission has to be recorded with its reason rather than silently rejoining the drift.

## Notes

- Field numbering is append-only from 19: a field number is wire format, so reusing one reinterprets existing indexes.
- Regenerated with `grpcio-tools==1.76.0`, per the pin documented in the proto header — generating with a newer protoc produces gencode the pinned runtime refuses to import.
- Found while evaluating Kuzu for #386, which needs an explicit schema and so surfaced the labels the proto did not know about.

## Verification

Full unit suite: **8035 passed**, 120 skipped. The single failure is the pre-existing `dotnet-trace` environment issue, which #1451 fixes and which is not on this branch.

## Summary by CodeRabbit

- **New Features**
  - Added support for representing pattern findings, code smells, and security issues in code analysis results.
  - Findings include names, messages, qualified identifiers, source locations, file paths, and code snippets.
  - These finding types can be ingested and processed alongside existing node types.

- **Quality Improvements**
  - Added validation to ensure supported finding types are consistently represented, mapped, and processed.


vitali87:

vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1461, opened 2026-08-26
fix: dispatch EMITTING_FRONTENDS generically by phase

### LINKED ISSUE #1460: EMITTING_FRONTENDS is not dispatched generically: a registered emitter is silently never called
Found by Greptile reviewing #1457, which registered a phase-tagged emitting frontend, exercised the real graph-build path, and confirmed the emitter was never invoked and its marker node never written.

## The defect

`register_emitting_frontend()` puts a frontend in `EMITTING_FRONTENDS`, and the registry is keyed by language, so it reads as a general mechanism. It is not. `graph_updater.py` looks up **one hardcoded key**:

```python
frontend = EMITTING_FRONTENDS.get(cs.SupportedLanguage.CPP)
```

A frontend registered for any other language is never called. Registration succeeds, `available()` and `applies()` are never consulted, `emit()` never runs, and the nodes and edges it would have contributed are simply absent. **Nothing errors and nothing warns.**

`FrontendPhase` has the same problem one level up. Each emitting frontend declares `BEFORE_DEFINITIONS` or `AFTER_DEFINITIONS`, but nothing reads it to decide ordering — the two call sites in `graph_updater.py` gate on `settings.CPP_FRONTEND` directly:

```python
if settings.CPP_FRONTEND != cs.CppFrontend.HYBRID:
    self._run_cpp_frontend()          # before Pass 2
...
if settings.CPP_FRONTEND == cs.CppFrontend.HYBRID:
    self._run_cpp_frontend()          # after Pass 2
```

So `phase` is currently descriptive rather than load-bearing.

## Why it matters

This is the same silent-loss shape as #1447 and #1453: the graph ends up smaller than the truth and nothing downstream can distinguish that from a repository which genuinely lacked those elements.

Here the cost is paid by the next contributor rather than by a user. #1178 asks for a "new-frontend doc"; a doc describing registration as sufficient would send someone to write a frontend that does nothing, with no error to diagnose. #1457 now carries an explicit warning about this instead, which is honest but is a workaround.

## Fix

Dispatch the registry generically:

1. Iterate `EMITTING_FRONTENDS`, filtered by `phase`, at each of the two existing call sites, rather than looking up `CPP`.
2. Consult `available()` and `applies()` per frontend, as the fact-provider loop already does for `FRONTENDS`.
3. Merge each `FrontendEmitResult` — `covered_files` unions, and the pending-call lists concatenate.

The per-mode `settings.CPP_FRONTEND` gating stays where it is: it decides whether the C++ frontend runs at all, which is a different question from which phase it belongs to.

## Acceptance

- [ ] A frontend registered for a language other than C++ is invoked in its declared phase.
- [ ] A test registers a dummy emitting frontend for each phase and asserts its marker node reaches the graph — a test asserting only "no crash" would pass today.
- [ ] `phase` decides ordering; removing a hardcoded language check does not change C++ behaviour.
- [ ] The warning in `docs/architecture/language-frontends.md` is removed once registration alone is sufficient.

Not blocking #1459, which wires C++ through the registry and keeps the existing hardcoded lookup, so behaviour is unchanged there.

vitali87:
Fixes #1460. Stacked on #1459 — that PR registers the C++ frontend, this one makes the registry actually dispatch. Base is `feat/cpp-emitting-frontend`; retarget to `main` once #1459 merges.

## How this was found

Greptile reviewed #1457 (the frontend docs) by **executing the checklist rather than reading it**: it registered a phase-tagged emitting frontend, ran the real graph-build path, and confirmed the emitter was never invoked and its marker node never written.

That is a defect in the code, not the doc. #1457 now carries a warning; this fixes the underlying gap.

## The defect

`EMITTING_FRONTENDS` is keyed by language and reads as a general mechanism. It was not:

```python
frontend = EMITTING_FRONTENDS.get(cs.SupportedLanguage.CPP)
```

One hardcoded key. A frontend registered for any other language was **silently never called** — registration succeeded, `available()` and `applies()` were never consulted, `emit()` never ran, its nodes were simply absent. Nothing errored.

`FrontendPhase` had the same problem one level up: declared on every emitting frontend and read by nothing, with ordering coming from two hardcoded `settings.CPP_FRONTEND` checks.

Same silent-loss shape as #1447 and #1453 — the graph ends up smaller than the truth and nothing downstream can distinguish that from a repository which genuinely lacked those elements.

## Three choices worth review

**C++ is explicitly skipped in the generic loop.** It owns the mode gating and compile-database discovery its two modes need, so dispatching it here as well would run it twice. #1459's behaviour is byte-identical.

**`available()` and `applies()` are wrapped.** A frontend whose *probe* raises would otherwise fail the whole index, and the protocol's invariant is that a missing toolchain degrades to the tree-sitter backbone — never worse.

**`covered_files` unions rather than overwrites**, since each frontend reports the files it owns and the definition pass must skip all of them.

## The tests assert the node reaches the graph

Not that indexing did not crash — **nothing crashed before either**, which was the entire failure mode. A test asserting the absence of an exception passes against the broken behaviour.

The two negative tests are paired controls: without them, a dispatch loop that ignored `available()` would pass both positives while breaking every machine lacking the tool.

Red first, and the red had the right shape: both positives failed (`emit() was never called`), both negatives passed.

## Verification

Full unit suite: **8044 passed**, 121 skipped, zero failures — 8040 plus the four new tests.

## Summary by CodeRabbit

* **New Features**
  * Added support for language frontends running at different processing phases.
  * Registered frontends can contribute results before and after definition processing.
  * Automatically falls back to tree-sitter when a frontend is unavailable or not applicable.

* **Bug Fixes**
  * Improved handling of frontend probe failures with clear warnings and continued processing.

* **Tests**
  * Added coverage for frontend dispatch across languages, phases, and fallback scenarios.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -61,6 +61,7 @@ from .parsers.frontends import (
     EMITTING_FRONTENDS,
     FRONTENDS,
     FrontendEmitContext,
+    FrontendPhase,
     SemanticFacts,
 )
 from .parsers.frontends.protocol import QueryCall
@@ -385,6 +386,49 @@ class GraphUpdater:
         self.finding_analyzer = FindingAnalyzer(
             self._sink, self.repo_path, self.capture
         )
+
+    def _run_emitting_frontends(self, phase: FrontendPhase) -> None:
+        """Run every registered emitting frontend declaring this phase.
+
+        C++ is dispatched by `_run_cpp_frontend`, which owns the mode gating
+        and the compile-database discovery its two modes need, so it is skipped
+        here to avoid running it twice (issue #1460).
+
+        A frontend that is unavailable or does not apply is skipped rather than
+        failing the run: a missing toolchain must degrade to the tree-sitter
+        backbone, which covers every file anyway.
+        """
+        for language, frontend in EMITTING_FRONTENDS.items():
+            if language == cs.SupportedLanguage.CPP:
+                continue
+            if frontend.phase != phase:
+                continue
+            try:
+                if not frontend.available() or not frontend.applies(self.repo_path):
+                    continue
+            except Exception:
+                logger.warning(ls.EMITTING_FRONTEND_PROBE_FAILED.format(lang=language))
+                continue
+            result = frontend.emit(
+                FrontendEmitContext(
+                    ingestor=self._sink,
+                    repo_path=self.repo_path,
+                    project_name=self.project_name,
+                    compdb_dir=self.repo_path,
+                    function_registry=self.function_registry,
+                    simple_name_lookup=self.simple_name_lookup,
+                    structural_elements=(
+                        self.factory.structure_processor.structural_elements
+                    ),
+                    exclude_paths=self.exclude_paths,
+                    unignore_paths=self.unignore_paths,
+                    owned_qns=self._frontend_owned_qns,
+                )
+            )
+            # Covered files union across frontends: each one reports the files
+            # it fully owns, and the definition pass must skip all of them.
+            if result.covered_files:
+                self._cpp_frontend_covered |= result.covered_files
 
     def _run_cpp_frontend(self) -> None:
         # Optional libclang C++ pre-pass when a compile_commands.json is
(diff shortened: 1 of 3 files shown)


vitali87:

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1468, opened 2026-08-26
feat: parse NODE_SCHEMAS into a machine-readable form

vitali87:
Addresses the remaining blocker on #386, and satisfies the last open acceptance criterion of #1452.

## The blocker

`NODE_SCHEMAS` declares each `NodeLabel`'s properties as a prose string:

```python
NodeSchema(NodeLabel.SECTION, "{qualified_name: string, name: string, heading_level: int, ...}")
```

Its consumers are README generation and LLM prompt text. **Nothing parses it**, so nothing can check it or generate from it.

That is the concrete blocker behind two issues:

- **#386** needs an explicit schema because Kuzu requires a table per label and **rejects any property not declared in it** (`Binder exception: Cannot find property ...`).
- **#1452** found `Pattern`, `CodeSmell` and `SecurityIssue` had drifted out of `codec/schema.proto` unnoticed. The declaration existed; it just was not in a form anything could compare against.

## Parses rather than replaces

The prose strings stay the source of truth. A second declaration to keep in sync is exactly the drift this exists to make detectable — adding one would recreate the problem in a new place.

## Measured before assuming a grammar

All 21 declarations parse, and the type vocabulary is closed:

```
string x73   int? x29   string? x19   list[string]? x13
boolean? x12   int x9   list[string] x6
```

Four base types plus an optional-marking `?`. Nothing else, so the parser can be strict.

## An unknown type raises rather than passing through

A property the parser cannot read is one a generated table would silently omit — and a backend that rejects undeclared properties would then fail at write time with no indication why. Passing it through would make the declaration *look* complete while the generated schema was short.

## Demonstrated against real Kuzu

#1452's final criterion is *"sufficient to generate a `CREATE NODE TABLE` per label (demonstrated, not necessarily wired up)"*:

```
accepted by Kuzu: 20 / 21
```

**The one failure is a genuine finding for any future backend**: `Union` is a **reserved word** in Kuzu's Cypher, so that table name needs backtick quoting. Verified that quoting fixes it.

Recording it rather than solving it, because this PR adds no backend. But a Kuzu implementation that did not know this would fail on exactly one label with a parser error naming nothing useful — the kind of thing that costs an afternoon.

## What this does not do

No backend, no wiring, no change to how anything is written today. It is the prerequisite #386's criterion 3 was blocked on, made available and demonstrated.

## Verification

Full unit suite: **8085 passed**, 121 skipped, zero failures. Lint clean, zero `ty` errors.

Seven tests, including a control that parses the **live** declarations rather than only synthetic strings — without it the parser could pass every unit case and still mis-read the schemas it exists to read.

## Summary by CodeRabbit

* **New Features**
  * Added structured parsing for declared node properties.
  * Supports scalar types, optional properties, and list element types.
  * Provides clear validation errors for malformed declarations and unsupported types.
  * Preserves property declaration order and exposes parsed schemas for all node labels.

* **Tests**
  * Added comprehensive coverage for valid schemas, optionality, lists, ordering, duplicate properties, and invalid declarations.

### DIFF
--- a/codebase_rag/schema_parse.py
+++ b/codebase_rag/schema_parse.py
@@ -89,14 +89,26 @@ def parse_properties(declaration: str) -> list[PropertySpec]:
     if not body:
         return []
     specs: list[PropertySpec] = []
+    seen: set[str] = set()
     for field in body.split(","):
         name, separator, raw_type = field.partition(":")
         if not separator:
             raise SchemaParseError(f"property has no type: {field.strip()!r}")
+        property_name = name.strip()
+        if not property_name:
+            # An unnamed property would generate an unnamed column, which the
+            # backend rejects with an error naming nothing useful.
+            raise SchemaParseError(f"property has no name: {field.strip()!r}")
+        if property_name in seen:
+            # Consumers key by name, so a duplicate silently loses one of the
+            # two declarations -- the declaration would look complete while
+            # the generated table was short by a column.
+            raise SchemaParseError(f"duplicate property: {property_name!r}")
+        seen.add(property_name)
         type_name, optional, element = _parse_type(raw_type)
         specs.append(
             PropertySpec(
-                name=name.strip(),
+                name=property_name,
                 type_name=type_name,
                 optional=optional,
                 element=element,
--- a/codebase_rag/schema_parse.py
+++ b/codebase_rag/schema_parse.py
@@ -1,0 +1,117 @@
+"""Machine-readable view of the declared node schemas (issue #386).
+
+`NODE_SCHEMAS` in `types_defs` declares each `NodeLabel`'s properties as a
+prose string -- `"{qualified_name: string, name: string, path: string?}"` --
+consumed by README generation and LLM prompt text. Nothing parses it, so
+nothing can check anything against it or generate anything from it.
+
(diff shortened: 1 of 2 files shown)


vitali87:

vitali87:
@greptileai review

vitali87:
Both review threads (empty and duplicate property names) are fixed in `29f5c995`, now pushed — the commit was written earlier but an unrelated working-tree conflict blocked the push until now. I resolved both threads; posting the reasoning here since the inline replies did not post.

Both cases now raise `SchemaParseError`:

- `{: string}` → *property has no name*. An unnamed property generates an unnamed column, and the backend then fails with an error naming nothing useful — a worse failure than refusing to parse.
- `{a: string, a: int}` → *duplicate property*. Consumers key by name, so one declaration is silently lost: the declaration reads as complete while the generated table is short by a column.

That second case is the exact shape this module exists to detect, so it should not have been able to produce it itself.

Three tests, including a **control against the real schemas** rather than only synthetic strings: `parsed_node_schemas()` now raises on a duplicate, so asserting the 21 shipped declarations parse clean is evidence they are genuinely duplicate-free — not merely that the new check is untested.

@greptileai review

Issue #1178, opened 2026-08-10
Extract a LanguageFrontend fact-bundle protocol from the Roslyn integration


vitali87:
## Problem

There is no language-frontend interface. The two existing compiler integrations are bespoke branches called directly from `GraphUpdater.run()` (`graph_updater.py:614-812` via `_run_cpp_frontend` at `:324` and `_run_csharp_frontend` at `:394`); `ProcessorFactory` has zero frontend references, and the `LanguageHandler` protocol (`parsers/handlers/protocol.py`) is purely syntactic with no hooks for semantic facts.

Worse, the two integrations follow opposite shapes:

- **csharp_frontend (Roslyn)** — the good shape: an external tool emits a *fact bundle* (`CSharpSemanticFacts`), keyed by `(rel_file, line, byte_col, name)`, consumed by the existing processors, with silent per-miss fallback to heuristics.
- **cpp_frontend (libclang)** — the fragile shape: writes graph nodes/edges directly via `IngestorProtocol`, and owns a parallel qn generator (`cpp_frontend/qn.py`) that must byte-for-byte reproduce `GraphUpdater._collect_eligible_files` walk order (#1025).

The roadmap adds frontends for Go, TypeScript/JS, Java, Dart, and more. Without a contract, each becomes a third/fourth/fifth special case.

## Proposal

Extract a `LanguageFrontend` protocol from the Roslyn integration:

```
run(repo_path, files) -> SemanticFacts
```

with generalized fact families (each optional per language):

- **resolved_call_sites** — the compiler's answer for which declaration a call binds to (overloads by argument type, extension/receiver binding, interface dispatch)
- **external_sites** — proof a call leaves the repo, suppressing fabricated first-party CALLS edges (Roslyn's false-positive killer; valuable in every language)
- **base_kinds** — INHERITS vs IMPLEMENTS classification by symbol identity
- **expression_types** — type-of-expression facts feeding each language's existing inference engine
- **invisible_constructs** — things tree-sitter cannot see: expanded macros, LINQ query operators, generated members

Contract details to standardize:

- **Join key**: `(rel_file, line, utf8_byte_col, name_token)` — lift the Roslyn tool's UTF-16→UTF-8 column re-measurement into documented shared convention so every new frontend matches tree-sitter coordinates.
- **Fallback**: every join miss silently degrades to the heuristic result — a missing toolchain must never make the graph worse than today.
[...]

PR #1469, opened 2026-08-26
docs: record the LIBCLANG disposition and correct a false dispatch warning

vitali87:
Part of #1178.

## What this fixes

Two problems in `docs/architecture/language-frontends.md`, both found while checking #1178's acceptance criteria one at a time.

### The dispatch warning was false

The guide carried a prominent warning saying `EMITTING_FRONTENDS` "is not yet dispatched generically", that registering an emitting frontend is **silently inert**, and that adding one means hand-writing dispatch in `graph_updater.py`.

That is no longer true. `_run_emitting_frontends` (`graph_updater.py:401`) iterates the registry and selects by `phase`. Registration is sufficient.

This was worse than ordinary staleness. It fails *safe-looking*: a new-language author reads it, hand-writes dispatch they do not need, their registration works anyway, and they conclude the doc was right. Nothing in the system would ever contradict it. A doc saying "this does not work yet" about a mechanism that works is more expensive than no doc at all.

What **is** true is narrower, and is now documented as the deliberate exception it is: the generic loop explicitly skips C++, which keeps its own call site at `:471`, because it needs the discovered `compdb_dir` and because HYBRID returns pending macro calls that must be narrowed. Marked "do not copy this" rather than left looking like a gap.

### The LIBCLANG disposition was recorded nowhere

#1178 requires that pure-LIBCLANG "either migrates too or is explicitly documented as a legacy full-replacement path". It **migrated** — it is a registered `EmittingFrontend`, and `phase` is read at access time so switching modes moves it between passes rather than down a different code path.

That was true in the code and stated in no document. Now written, together with the mode-to-phase table and the silent failure mode: getting the phase wrong does not raise, which is why `test_cpp_emitting_frontend.py` asserts the mapping per mode.

## What this does NOT do

#1178 stays open. Criterion 4 is untouched: `cpp_frontend/qn.py` still exists and `_frontend_owned_qns` still has 11 occurrences in `graph_updater.py`. The new text says so explicitly, so the guide does not read as though the migration finished.

## Verification

Every claim was checked against the code rather than written from recollection:

- `phase` read at access time — `parsers/frontends/cpp.py:46-52`
- per-mode assertions exist — `test_cpp_emitting_frontend.py:37,40`
- generic loop iterates and selects by phase — `graph_updater.py:401`
- the `continue` past CPP — `graph_updater.py:402`
- `qn.py` present, `_frontend_owned_qns` present — both confirmed on `origin/main`

Independently re-verified against `origin/main` by a second session before this was written.

Docs-only; no code paths change.

## Summary by CodeRabbit

- **Documentation**
  - Updated architecture guidance for emitting frontend dispatch and phase-based execution.
  - Clarified that registering a frontend is sufficient for it to run when its phase is reached.
  - Documented separate C++ dispatch and the available `treesitter`, `hybrid`, and `libclang` modes.
  - Added notes on the ongoing migration of `qn.py`.

### DIFF
--- a/docs/architecture/language-frontends.md
+++ b/docs/architecture/language-frontends.md
@@ -57,21 +57,31 @@ that emits its own definition nodes (the definition pass then skips the files it
 covered), or `AFTER_DEFINITIONS` for one that attaches to spans Tree-sitter
 already produced.
 
-!!! warning "Registration alone does not run an emitting frontend yet"
+Registration is enough to run it. `_run_emitting_frontends` iterates
+`EMITTING_FRONTENDS`, selects the entries whose `phase` matches the pass being
+run, and skips any that are unavailable or do not apply. Declaring `phase`
+correctly is therefore load-bearing rather than advisory: it is what decides
+when — or whether — your frontend is called.
 
-    `EMITTING_FRONTENDS` is not yet dispatched generically. The graph builder
-    looks up **C++ specifically**, and the ordering between the two phases comes
-    from two hardcoded call sites in `graph_updater.py` rather than from the
-    `phase` attribute.
+A frontend that raises from `available()` or `applies()` is skipped with a
+warning rather than failing the run, on the same principle as a join miss: a
+missing toolchain must never make the graph worse than the Tree-sitter
+backbone.
 
-    So registering an emitting frontend for a new language today is silently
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

Issue #385, opened 2026-02-27
Knowledge graph re-ranking for search results


vitali87:
## Task Description
Explore using the code knowledge graph to re-rank search results, improving relevance by incorporating structural relationships between code entities.

The idea is that when a user queries the codebase, the graph topology (call chains, imports, inheritance) can boost or demote results based on how closely related they are to the query context.

## Acceptance Criteria
- [ ] Research feasibility of graph-based re-ranking
- [ ] Prototype a re-ranking step that uses graph proximity/relationships
- [ ] Compare retrieval quality with and without graph re-ranking
- [ ] Integrate into the existing query pipeline if results are positive

## References
- Original discussion: #61 (comment by @qdrddr)
- Knowledge Graph Re-ranking notebook: https://github.com/Bavalpreet/MediumBlogs/blob/main/Knowledgegraph%20Reranking/Knowledge_graph_re_ranking.ipynb

## Priority
- [ ] High
- [ ] Medium
- [x] Low

## Estimated Effort
- [ ] Small (< 2 hours)
- [ ] Medium (2-8 hours)
- [x] Large (> 8 hours)

vitali87:
## Consolidating #387 into this issue

This issue and #387 track the same capability from two angles:

- This issue (#385): use graph topology to re-rank search results.
- #387: combine code embeddings with graph structure for retrieval (CodeFuse-CGM inspired).

Both ask for the same missing piece: a retrieval step that fuses vector similarity with graph-structure signals so structurally related code ranks higher even when lexically dissimilar. They reference the same discussion (#61) and were filed seconds apart.

Verified against current code: semantic search (`tools/semantic_search.py` -> `vector_store.py:136-154`) returns pure cosine similarity in Qdrant order, graph queries are a separate Cypher tool (`tools/codebase_query.py`), and nothing fuses them. So none of the criteria from either issue are met today.

Closing #387 as a duplicate and using this issue as the single tracker for the work.

### Unified acceptance criteria
- [ ] Research graph-based re-ranking and the CodeFuse-CGM approach; identify applicable techniques
- [ ] Prototype a retrieval step that combines embedding similarity with graph-structure scores (call chains, imports, inheritance, proximity/centrality)
- [ ] Benchmark retrieval quality with and without the graph signal
- [ ] Integrate into the query pipeline if results are positive

vitali87:
[...]

PR #1478, opened 2026-08-26
fix: make semantic recall@k order-sensitive so reranking can be measured

### LINKED ISSUE #1474: eval: semantic recall@k is order-insensitive, so reranking cannot be measured
Split out of a review finding on #1467 (CodeRabbit), verified and confirmed.

## The problem

`score_semantic` scores recall@k by **membership**:

```python
hits = {
    (case.query, case.expected_qn)
    for case in cases
    if case.expected_qn in ranking.get(case.query, [])
}
```

Any reranker returns a **permutation** of the candidate list. Reordering never changes what is in the list, so a baseline condition and a reranked condition produce identical scores for every input — by construction, not by coincidence.

Demonstrated over all six permutations of a three-hit list with the expected answer in the middle:

```
(a,b,c) -> recall 1.0     (b,c,a) -> recall 1.0
(a,c,b) -> recall 1.0     (c,a,b) -> recall 1.0
(b,a,c) -> recall 1.0     (c,b,a) -> recall 1.0
```

## Why it matters

This is not a latent inaccuracy — it already produced a wrong published conclusion. I reported on #385 that graph-proximity reranking measured **negative**, and offered that as grounds for a decision. The eval could not have detected a difference in **either** direction. That conclusion is retracted; #385 is now un-measured.

The failure mode is the one this repo keeps hitting: an assertion satisfied by both the working and the broken implementation. Here it was a *metric* satisfied by both the reranked and unreranked condition, which is worse, because a metric is trusted precisely when nobody is inspecting it.

## Two ways to fix it

1. **Truncate after reranking.** Retain candidates deeper than k, rerank the full list, then cut to k before scoring. A hit demoted out of the top k then genuinely costs recall. Smallest change; keeps recall@k as the headline.
2. **Score with an order-sensitive metric** — MRR or nDCG — where position is the thing being measured rather than a property the metric discards.

(2) is the more honest instrument for ranking work; (1) is cheaper and preserves comparability with the existing published retrieval numbers. They are not exclusive.

## Acceptance

- [ ] A reranker that demotes the expected answer out of the top k scores strictly worse than one that does not
- [ ] Verified by a test that FAILS against the current membership scorer — the guard must distinguish the two implementations, which is exactly what the present one cannot do
- [ ] #385 re-measured under the fixed metric, and its earlier retracted result replaced with a real one

Not claiming this. Filing so the retraction on #385 points at concrete follow-up rather than leaving the reranker permanently unmeasurable.

### LINKED ISSUE #1477: graph_rerank and the eval's proximity model disagree on multi-edge pairs
## The divergence

`codebase_rag/tools/graph_rerank.build_proximity_query` and `evals.semantic_search.proximity_edges` compute in-set proximity degree differently when two nodes are joined by **more than one** counted relationship.

The shipped Cypher counts **edges**: `UNWIND [id(a), id(b)]` yields both endpoints once per matching relationship, so a pair joined by `CALLS` and `OVERRIDES` contributes 2 to each endpoint. `proximity_edges` builds a **set** of neighbours, so the same pair contributes 1.

Measured against the real `_proximity_scores`, not a mirror of it. Normalisation hides the difference while a single pair dominates; it surfaces as soon as a multi-edge pair competes with a single-edge one. Over a-b (two edges) and b-c (one edge):

```
shipped: a=0.67  b=1.0  c=0.33
eval:    a=0.5   b=1.0  c=0.5
```

Same graph, different relative pressure on the ranking.

## Why it matters

This is the fourth member of the family fixed in review on #1467 (uncounted directed endpoint, double-counted self-loops, unpinned `CONTAINS_*` exclusion). All are divergences between the shipped query and the model the eval scores it against, rather than freestanding bugs. The consequence is the same: **the measurement stops describing the code that ships** — `graph_rerank.build_proximity_query`'s own docstring names this outcome.

Found by the guard added in #1474, which executes both paths over one fixture and asserts equal degrees. The existing tests could not catch it because they pin the Cypher *text* and the adjacency *shape* separately, each against its own expectations, with nothing comparing the two.

## The open question

Which count is correct is a question about the proximity model, not about scoring: **is a pair that both calls and overrides "closer" than a pair that only calls?** There is a real argument either way — multiple relationship types are evidence of a stronger association, or they are one association observed twice.

Not answered in #1474 deliberately: that issue changes eval scoring, and resolving this inside it would ship a silent behaviour change to the reranker under a metric fix.

## Current state

Pinned as a known divergence by `test_a_multi_edge_pair_is_a_known_divergence` in `codebase_rag/tests/test_graph_rerank.py`, which asserts the exact counts on both sides. It fails if either side changes, so the gap cannot widen unnoticed, and its message says to fold the pair back into the agreement test if the divergence is closed deliberately.

## Acceptance

- [ ] Decide whether proximity counts edges or distinct neighbours, and record the reason
- [ ] Make both implementations compute the chosen quantity
- [ ] Replace the known-divergence pin with a case in the agreement test
- [ ] Re-run the #385 comparison if the chosen semantics differ from what shipped

vitali87:
Closes #1474.

## The defect

`score_semantic` scored recall@k by **membership**. A reranker returns a *permutation*, and membership is invariant under permutation, so the baseline and reranked conditions scored identically for every input — by construction, not by coincidence.

Reproduced before any fix:

```
=== all permutations of a 3-hit list containing the expected answer ===
  ('iso.a', 'want', 'iso.c') -> recall 1.0
  ('iso.a', 'iso.c', 'want') -> recall 1.0
  ('want', 'iso.a', 'iso.c') -> recall 1.0
  ('want', 'iso.c', 'iso.a') -> recall 1.0
  ('iso.c', 'iso.a', 'want') -> recall 1.0
  ('iso.c', 'want', 'iso.a') -> recall 1.0

  expected kept at rank 1 : recall 1.0
  expected demoted to rank 3 (out of top 2): recall 1.0
```

## The fix

Both remedies the issue lists, since they are not exclusive:

* `score_semantic(cases, ranking, k=None)` truncates **before** the membership check, so a hit demoted out of the top k genuinely costs recall. `k=None` scores the whole ranking, preserving every existing call site and the published retrieval numbers.
* `score_semantic_mrr(cases, ranking, k=None)` is the order-sensitive instrument: it measures *where* the answer landed, so a promotion that crosses no cutoff still registers.

## The old guard asserted the bug

`test_the_comparison_scores_both_conditions_the_same_way` asserted `before.recall == 1.0` and `after.recall == 1.0` and offered that as evidence the comparison worked. A membership scorer returns 1.0 for every permutation, so the assertion held whether or not the metric could see order at all — the defect written into its own guard. It now asserts the two are *equal* (correct, since nothing crosses the cutoff there) and points at the test that actually distinguishes the implementations.

## Mutation proof

Green is not evidence for tests written after a fix. Reverting `score_semantic` to the membership check:

```
FAILED test_demoting_the_expected_answer_out_of_top_k_costs_recall
FAILED test_recall_at_k_is_not_invariant_across_permutations
FAILED test_reranking_that_demotes_the_answer_scores_worse_end_to_end
3 failed, 11 passed
```

with the message `reranking demoted the expected answer out of top-2 but recall did not move (1.0 -> 1.0)`. The other 11 still pass, so the guards are specific to this defect.

## Divergence guard, and a fourth divergence found

The three bugs fixed in review on #1467 were all divergences between the shipped Cypher and the adjacency model the eval scores it against. Nothing caught them because the existing tests pin the query *text* and the adjacency *shape* separately, each against its own expectations, with nothing comparing the two.

This adds tests that execute both paths over one fixture and assert equal in-set degrees. Writing them immediately surfaced a **fourth** divergence: the shipped Cypher counts *edges* (`UNWIND` yields both endpoints once per relationship) while `proximity_edges` builds a *set* of neighbours, so a pair joined by both `CALLS` and `OVERRIDES` contributes 2 in shipped code and 1 in the model measuring it. Verified against the real `_proximity_scores`:

```
shipped: a=0.67  b=1.0  c=0.33
eval:    a=0.5   b=1.0  c=0.5
```

Pinned as a known divergence rather than fixed here — which count is correct is a question about the proximity model, and answering it inside a metric fix would ship a silent behaviour change to the reranker. Filed as #1477.

## #385 re-measurement

Run on the fixed metric, retaining candidates deeper than k so reranking has somewhere to demote from:

```
condition     recall@k       MRR
baseline        1.0000    1.0000
reranked        1.0000    0.8333

  weight  recall@3       MRR
    0.05    1.0000    1.0000
    0.10    1.0000    1.0000
    0.25    1.0000    0.8333
    0.50    1.0000    0.8333
    1.00    1.0000    0.8333
```

Order changed on 3/3 queries, so the deltas measure something rather than reporting an inert comparison.

This is a **first measurement of post-#1467 code, not a correction of the retracted one**. Three correctness bugs were fixed after that measurement and the metric has changed, so two variables moved; the old number is void rather than superseded.

**Two caveats, both limiting:** the fixture's baseline is already perfect (MRR 1.0), so it can detect harm but never improvement — this is evidence against reranking helping *here*, not evidence it helps nowhere. And three cases on one synthetic repo is thin. What it does establish is that the instrument now moves, which is what #1474 was for. Note also that recall@k alone still reports `+0.0000`: all three answers stayed inside the top 3, and only MRR sees the demotion from rank 1 to rank 2 — the concrete argument for adding the order-sensitive metric rather than only truncating.

## Checks

`ruff check` clean, `ruff format --check` clean, 35 passed across the three affected test files. `ty` reports 2 diagnostics, both pre-existing on `main` (identical with these changes stashed, neither in a touched file).

## Summary by CodeRabbit

* **New Features**
  * Added a command-line workflow for evaluating semantic search.
  * Added baseline and graph-reranked scoring with recall@k and MRR.
  * Added configurable result cutoffs and validation for evaluation settings.
  * Added shared semantic evaluation cases for consistent scoring.

* **Bug Fixes**
  * Improved proximity ranking to count distinct relationship types accurately while ignoring repeated edges of the same type.
  * Added support for negative and zero cutoff validation.

* **Tests**
  * Expanded coverage for relationship types, cutoff behavior, CLI scoring, and invalid settings.

---

## Review round 2 — `8838d16c`

All three findings from the first round are addressed, and the scope grew to close #1477 as well.

### Greptile P1: the entry path did not use the new measurements

Correct, and the most important of the three. `k` and MRR existed but nothing invoked them, so the eval still ran the order-blind default — a capability nothing calls is indistinguishable from one never added.

There was no eval *workflow* to wire into: `semantic_search.py` had no `main`, unlike its siblings (`module_calls.py`, `flow_ground_truth.py`) which each expose `typer.run(main)`. Added one following that convention. It scores **both** conditions through the same scorer at the same cutoff, and retains candidates deeper than `top_k` so reranking has somewhere to demote from — truncating first would reintroduce #1474 at the call site.

Two guards on the entry point itself: `retain < 2` and `top_k < 1` are rejected with reasons, and the command says so explicitly when reranking changed no ordering, so an unchanged number cannot read as "reranking made no difference."

`SEMANTIC_CASES` moved into `evals/constants.py` (they previously lived only in a test module — a benchmark whose cases live only in its test cannot be run).

### CodeRabbit: negative `k` silently sliced

Correct. `hits[:-1]` means "all but the last", so a negative cutoff scored a nonsensical window and returned a plausible-looking number. Now raises `ValueError`, with tests covering `-1`/`-3` and confirming `k=0` remains a *valid* empty window.

### CodeRabbit: wrong issue reference

Correct, my slip — that block is #1477, not #1474. Fixed.

## Also closes #1477

The known-divergence pin is gone; both paths now agree. Measured on real code before choosing the semantics rather than guessing, across three subtrees (1115 adjacent pairs):

```
                                 pairs   repeated-same-type   distinct-types
codebase_rag/tools                 187                   48               21
codebase_rag/parsers/io_access     179                   94                0
evals                              749                  143               13
TOTAL                             1115                  285               34
```

Two distinct families, and only one was a judgement call:

* **285 pairs (25.6%)** are the *same* type repeated — a caller invoking a callee on three lines. All 51 multi-`CALLS` pairs in `codebase_rag/tools` were same-direction repeats, **zero mutual**. Counting these separately makes proximity depend on how often a caller happens to invoke a callee, which is indefensible under either model.
* **34 pairs (3.0%)** carry genuinely distinct types — the override that calls up (`super().handle()` is both `CALLS` and `OVERRIDES`). Credit to @vscode-link-terminal-scheme for isolating that case; verified through the real updater.

So both paths now count **distinct relationship types per pair**. This differs from *both* previous implementations: the shipped query over-counted the 25.6%, the eval under-counted the 3.0%.

## Mutation proof

The first mutation run exposed a real gap in my own tests: reverting the **eval-side production code** to presence-only counting left all 20 tests in `test_graph_rerank.py` green, because those compare two local *mirrors* of the models rather than the shipped functions. Added `test_the_eval_reranker_counts_distinct_types_not_neighbours`, which runs `reranked_semantic_ranking` itself. Re-verified:

```
mutation (eval counts neighbours, not distinct types) -> FAILED  (was: 20 passed)
mutation (shipped mirror counts edges again)          -> 7 failed
unmutated                                             -> 42 passed
```

## Checks

`ruff check` clean, `ruff format` clean, 42 passed. `ty` reports the same 2 diagnostics as `main` — verified pre-existing by stashing these changes and re-running; neither is in a touched file.

### DIFF
--- a/codebase_rag/tests/test_graph_rerank.py
+++ b/codebase_rag/tests/test_graph_rerank.py
@@ -209,7 +209,11 @@ def test_the_query_counts_both_endpoints_explicitly() -> None:
 
     query = build_proximity_query([1, 2])
 
-    assert "UNWIND [id(a), id(b)]" in query, query
+    # Both endpoints reach the UNWIND, via the normalised (low, high) pair the
+    # distinct-type counting in #1477 introduced. What matters is that neither
+    # endpoint is dropped, not which expression names them.
+    assert "UNWIND [low, high]" in query, query
+    assert "AS low" in query and "AS high" in query, query
     # A bare undirected match with a single projected endpoint is the shape
     # this replaced; it must not come back.
     assert "RETURN id(a) AS node_id" not in query, query
@@ -235,7 +239,7 @@ def test_self_loops_are_excluded_from_proximity() -> None:
     assert "id(a) <> id(b)" in query, query
 
 
-# --- the shipped query and the eval's adjacency must agree (issue #1474) ----
+# --- the shipped query and the eval's adjacency must agree (issue #1477) ----
 #
 # The three bugs fixed in review on #1467 -- one endpoint of a directed edge
 # going uncounted, self-loops being double-counted, and the CONTAINS_*
@@ -260,11 +264,14 @@ def _degrees_via_shipped_semantics(
     Mirrors `build_proximity_query` clause for clause over tuples:
     `MATCH (a)-[r:REL]->(b)` filtered to `_PROXIMITY_RELS`, both endpoints in
     the node-id list, `id(a) <> id(b)`, then `UNWIND [id(a), id(b)]` counting
-    each endpoint once per surviving edge.
+    each endpoint once per DISTINCT relationship TYPE joining the pair
+    (issue #1477).
     """
     from codebase_rag.tools.graph_rerank import _PROXIMITY_RELS
 
-    degrees: dict[str, int] = {}
+    # (unordered pair) -> the distinct counted types joining it, matching the
+    # query's `count(DISTINCT type(r))` per pair.
+    kinds: dict[tuple[str, str], set[str]] = {}
     for source, rel_type, target in rels:
         if rel_type not in _PROXIMITY_RELS:
             continue
@@ -272,8 +279,12 @@ def _degrees_via_shipped_semantics(
             continue
         if source == target:  # id(a) <> id(b)
             continue
-        for endpoint in (source, target):  # UNWIND [id(a), id(b)]
-            degrees[endpoint] = degrees.get(endpoint, 0) + 1
+        kinds.setdefault(tuple(sorted((source, target))), set()).add(rel_type)
+
+    degrees: dict[str, int] = {}
+    for (left, right), types in kinds.items():
+        for endpoint in (left, right):  # UNWIND [id(a), id(b)]
+            degrees[endpoint] = degrees.get(endpoint, 0) + len(types)
     return degrees
 
 
@@ -282,24 +293,32 @@ def _degrees_via_eval_adjacency(
 ) -> dict[str, int]:
     """In-set degree the way `evals.semantic_search.proximity_edges` does.
 
-    That helper builds an undirected adjacency SET keyed by qualified name;
-    `reranked_semantic_ranking` then counts `adjacency[qn] & in_set`.
+    That helper builds an undirected adjacency keyed by qualified name whose
+    values are the distinct counted relationship TYPES joining the pair;
+    `reranked_semantic_ranking` then sums those over in-set neighbours
+    (issue #1477).
     """
     from codebase_rag.tools.graph_rerank import _PROXIMITY_RELS
 
-    adjacency: dict[str, set[str]] = {}
+    adjacency: dict[str, dict[str, set[str]]] = {}
     for source, rel_type, target in rels:
         if rel_type not in _PROXIMITY_RELS:
             continue
         if source == target:
             continue
-        adjacency.setdefault(source, set()).add(target)
-        adjacency.setdefault(target, set()).add(source)
-    return {
-        qn: len(adjacency.get(qn, set()) & in_set)
-        for qn in in_set
-        if adjacency.get(qn, set()) & in_set
-    }
+        adjacency.setdefault(source, {}).setdefault(target, set()).add(rel_type)
+        adjacency.setdefault(target, {}).setdefault(source, set()).add(rel_type)
+
+    degrees: dict[str, int] = {}
+    for qn in in_set:
+        total = sum(
+            len(types)
+            for neighbour, types in adjacency.get(qn, {}).items()
+            if neighbour in in_set
+        )
+        if total:
+            degrees[qn] = total
+    return degrees
 
 
 def test_directed_edges_boost_both_endpoints_in_both_implementations() -> None:
@@ -382,30 +401,13 @@ def test_the_two_implementations_agree_over_a_mixed_graph() -> None:
     assert set(shipped) == in_set, shipped
 
 
-def test_a_multi_edge_pair_is_a_known_divergence() -> None:
-    """Two nodes joined by SEVERAL counted edges are scored differently.
+def test_distinct_relationship_types_between_one_pair_each_count() -> None:
+    """Two DIFFERENT relationship types joining a pair count as two (#1477).
 
-    Found by the comparison above rather than by inspection, which is the
-    point of comparing implementations at all. The shipped Cypher counts
-    EDGES -- `UNWIND` yields both endpoints once per matching relationship --
-    while `proximity_edges` builds a SET of neighbours, so every extra edge
-    between the same pair collapses to one.
-
-    A pair joined by both CALLS and OVERRIDES therefore contributes twice the
-    proximity in shipped code and once in the model that measures it.
-    Normalisation hides this while one pair dominates; it surfaces as soon as
-    a multi-edge pair competes with a single-edge one. Measured over
-    a-b (two edges) and b-c (one edge):
-
-        shipped: a=0.67  b=1.0  c=0.33
-        eval:    a=0.5   b=1.0  c=0.5
-
-    Pinned rather than fixed: which count is CORRECT is a question about the
-    proximity model (is a pair that both calls and overrides "closer" than one
-    that only calls?), not about this metric change, and answering it inside
-    an eval-scoring fix would ship a silent behaviour change to the reranker.
-    Tracked in issue #1477; this test fails the moment either side changes, so
-    the divergence cannot widen unnoticed.
+    An override that calls up -- `Child.handle` calling `super().handle()` --
+    carries both CALLS and OVERRIDES between the same pair. Those are two
+    genuinely different relationships, not one observed twice, so the pair is
+    more strongly associated than one joined by either alone.
     """
     rels = [("a", "CALLS", "b"), ("a", "OVERRIDES", "b")]
     in_set = {"a", "b"}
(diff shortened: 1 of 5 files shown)


vitali87:
@greptileai review

Addressed all three findings at `8838d16c`:
- **P1 (entry path unused)**: added `main` to `evals/semantic_search.py` following the sibling convention (`module_calls.py`, `flow_ground_truth.py`). It scores both conditions through the same scorer at the same cutoff, retains candidates deeper than `top_k` so reranking has somewhere to demote from, rejects `retain < 2` / `top_k < 1`, and reports explicitly when reranking changed no ordering.
- **Negative `k`**: now raises `ValueError`; `k=0` remains a valid empty window.
- **Issue reference**: corrected to #1477.

Scope also grew to close #1477 (the multi-edge divergence), with the semantics chosen from measurement rather than preference: 1115 adjacent pairs across three subtrees, 285 same-type repeats vs 34 genuinely-distinct-type pairs. Both paths now count distinct relationship types per pair.

vitali87:
@greptileai review

Sonar flagged one MAJOR code smell at `test_graph_rerank.py:216` (python:S9073, composite assertion). Fixed at `3062e761` by splitting it, so a failure names which endpoint alias is missing rather than reporting that one of them is.

No production code changed since your 5/5 at `8838d16c` — this is a test-readability fix only.

vitali87:
@greptileai review

Added one test at `96085d45`. No production code changed since your 5/5 at `3062e761`.

**Why:** applying @vscode-link-terminal-scheme's check — does a test's *stated* property entail the code's *defining* behaviour? — exposed a gap in my own guard. `test_the_eval_reranker_counts_distinct_types_not_neighbours` states only that a multi-type pair is weighed "more", which is satisfied by any implementation that distinguishes 1 from 2 and then stops. Measured: capping the degree at `min(2, ...)` passed **all 19** existing tests, because no fixture contained a node above degree 2. The difference is reachable — `hub` at degree 3 versus `rival` at degree 2 separate uncapped and tie when capped.

`test_eval_proximity_keeps_separating_above_two_types` closes it. Red proven first: under the capped mutation only the new test fails (1 failed, 19 passed); on real code it passes.

vitali87:
@greptileai review

Test-only change at `ea4aa0b5`; no production code since your 5/5.

Applying @vscode-link-terminal-scheme's third case — the *detector* that evaluates every assertion can itself be wrong — found circularity in my own comparison tests. `test_a_self_loop_contributes_nothing_in_both_implementations` and `test_containment_edges_change_no_degree_in_either_implementation` asserted only that the two mirrors agree, plus a `not in` membership check. Both are satisfied by a **shared** error: scaling both mirrors by 7 keeps them equal, keeps the excluded node absent, and changed no key — measured, those tests passed under it.

`test_the_two_implementations_agree_over_a_mixed_graph` had the same weakness via `set(shipped) == in_set`, which membership survives because scaling changes no keys.

All three now pin absolute degree values. Red proven: under the scaled mutation 7 tests fail (previously 5, with these 2 passing); on real code 20 pass.

Issue #1368, opened 2026-08-21
add support for ALL languages having Tree-sitter grammar.


vitali87:
See: https://github.com/vitali87/code-graph-rag/blob/main/docs/advanced/adding-languages.md
It is better to add all langauges ourselves in _development_ mode and get rid of this adding-languages option for users. Any thoughts?

vitali87:
Thoughts, as asked — with the caveat that the decision is yours and I have not implemented anything.

**Short version: I think you are right, and the evidence for it is already in the repo.**

## The current option promises more than it delivers

`docs/advanced/adding-languages.md` advertises exactly this:

```bash
cgr language add-grammar ruby
cgr language add-grammar kotlin
```

Both of those have **open issues asking for support** (#118, #1233). If `add-grammar` genuinely delivered support, those issues would not exist. That gap is the argument for your position, stated by the repo itself.

## What `add-grammar` actually produces

It wires node-type tuples into `language_spec.py` from `node-types.json`. What it cannot infer is everything language-specific:

- `codebase_rag/tools/language.py` contains **no reference to `LanguageHandler`** — the flow never produces one
- **9 handler registrations against 15 `SupportedLanguage` members**

So the output is a language that parses and yields definitions, and then diverges from a supported language in ways nobody sees until they look.

## I measured exactly this on Scala last week

Scala had node-type tuples wired and no handler. It produced classes, objects, traits, methods, modules and CALLS — and:

- **no INHERITS edges at all** — `extends A with B` produced nothing
- **no import resolution** — every form returned an empty mapping
- and once inheritance existed, `foo.Service[Int]` bound to a same-named **local** class, inventing an edge the source never expressed

None of that is inferable from `node-types.json`. Each needed reading the grammar and deciding what the language means. That is the work `add-grammar` cannot do and quietly implies it has done.

## Why "development mode only" is the sounder default

The failure mode of a self-service path is **silent partial support**: a user adds Kotlin, indexes, gets a graph, and has no way to know it is missing inheritance and imports. There is no error — just a quieter graph. That is the same silent-smaller-than-truth shape that has produced several bugs here already.

Doing it in development means each language arrives with a handler, fixtures and an oracle, and the support matrix means something. It also makes the demand question honest: the cost of a language is not "run a command", it is the handler work, so deciding which languages to do is a real prioritisation rather than a config toggle.

## Two caveats

[...]

Issue #1233, opened 2026-08-13
Language support request: Kotlin


rafaelcalves:
## Summary

Request official support for **Kotlin** as a parsed language.

## Current state

Kotlin isn't a supported language today. `.kt` files aren't parsed — the only place `kotlin` appears in the codebase is `codebase_rag/constants/ast_java.py` (`JAVA_PATH_KOTLIN`), where it's a JVM **source-root directory name** (`src/main/kotlin`) used by the Java parser to locate package roots. That does not parse Kotlin source.

The language matrix (`docs/architecture/language-support.md`) lists Scala as *In Development* and Ruby with structural support via the ast-grep tier, but Kotlin is absent.

## Motivation

Kotlin is the primary language for Android and a large share of modern JVM backends (Spring Boot, Ktor, Gradle build logic). A JVM-heavy monorepo today gets its Java graphed but its Kotlin invisible, leaving call/import/inheritance edges broken across the Java↔Kotlin boundary. Adding Kotlin closes that gap for mixed-JVM repos.

## Proposed scope

Start with structural coverage (parity with the Ruby ast-grep tier), then grow toward full support:

- Nodes: `Module`/`File`, `Class` (incl. `object`, `data class`, `interface`), `Function`, `Method`
- Edges: `IMPORTS`, `DEFINES`/`DEFINES_METHOD`, `INHERITS`/`IMPLEMENTS`, `CALLS`
- Later: type inference for `val`/`var`, extension functions, companion objects

## Implementation options

1. **Pluggable ast-grep tier** (like Ruby) — a single YAML pattern file, lowest friction to get structural nodes/edges.
2. **Full Tree-sitter parser** via `cgr language add-grammar kotlin` (the docs list this exact invocation) using a `tree-sitter-kotlin` grammar, for richer coverage.

## Notes

Filing per GOVERNANCE.md, which asks for new-language support to be proposed and discussed in an issue before implementation. Happy to contribute the ast-grep tier and tests if this is accepted and a grammar/tier direction is agreed.

PR #1479, opened 2026-08-26
docs: say what add-grammar does NOT do (#1368)

vitali87:
Relates to #1368, and lands the part of it that is true **whichever way that issue is decided**.

## The problem

`docs/advanced/adding-languages.md` lists five things that happen automatically, ending with:

> **Enables Parsing**: Makes the language available for codebase analysis

and mentions no limitation anywhere — **zero** occurrences of "handler", "limitation" or "does not" across 99 lines. It reads as *"you now support this language"*.

The page names `ruby` and `kotlin` as examples. Both have **open issues requesting support** (#118, #1233). That gap is stated by the repo itself.

## What `add-grammar` actually produces

Definitions and calls — everything derivable from the grammars own `tree-sitter.json` and `node-types.json`. It does not produce a `LanguageHandler`; `codebase_rag/tools/language.py` contains no reference to one. Inheritance, import resolution and qualified-name rules live there.

**Measured example, from this repo:** Scala shipped with node types wired and no handler. The result produced classes, objects, traits, methods, modules and CALLS edges — and **no inheritance edges at all**, plus no import resolution for any form. Neither is inferable from `node-types.json`; both took reading the grammar and deciding what the constructs mean.

## Why it is worth documenting rather than leaving implicit

**The failure is silent.** A partially-supported language produces a graph, not an error. Queries return fewer results, with nothing reporting that a whole relationship kind is absent — so the graph is thinner than the code and nothing says so.

## What this deliberately does not claim

Handler-less does **not** mean unsupported. C# and Dart have no handler entry and are `FULL` in the support matrix, because they compensate through frontends and dedicated extraction paths.

So the page points at the [support matrix](docs/architecture/language-support.md) for what each language actually provides, rather than implying a rule that would be wrong for two of them. Both added cross-references were checked to resolve.

Docs-only; no code paths change.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that language-specific extraction may be implemented through language handlers or dedicated language-specific paths.
  * Updated C# and Dart examples to reflect their extraction approaches.
  * Revised the Scala example to note that previous extraction gaps have been addressed, including mixins and import aliasing.
  * Directed readers to the support matrix as the authoritative source for language capabilities.

### DIFF
--- a/docs/advanced/adding-languages.md
+++ b/docs/advanced/adding-languages.md
@@ -50,21 +50,36 @@ Everything above is derived from the grammar's own `tree-sitter.json` and
 `node-types.json`. Anything that depends on what the language *means* is not,
 and cannot be, inferred from those files.
 
-A grammar added this way gets **definitions and calls**. It does not get a
-`LanguageHandler`, and several capabilities live there:
+A grammar added this way gets **definitions and calls**. Everything below the
+line needs language-specific code that someone has written:
 
-| Derived from the grammar | Needs a language handler |
+| Derived from the grammar | Needs language-specific code |
 |---|---|
 | Functions, methods, classes | Inheritance (`extends`, `with`, mixins) |
 | Modules and files | Import resolution and aliasing |
 | Call sites | Qualified-name construction rules |
 
-Measured example: Scala shipped with its node types wired and no handler. It
-produced classes, objects, traits, methods, modules and CALLS edges — and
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

Requesting a review of head `669ca56f8ecd`. The 5/5 gate timed out waiting for one, and the previous review anchors to an earlier commit.

PR #1482, opened 2026-08-27
test: separate degree-weighted proximity from a flat any-edge boost

### LINKED ISSUE #1481: Graph-rerank tests do not distinguish degree-weighted proximity from a flat any-edge boost
`test_graph_rerank.py` passes in full against an implementation that replaces degree-proportional proximity with a flat boost for any in-set edge.

## Reproduction

In `_proximity_scores`, replace the per-node normalised degree with membership:

```python
prox = 1.0 if proximity.get(node_id, 0.0) > 0 else 0.0
```

Result: `13 passed`. No test fails.

## Why this matters

`graph_rerank.py` exists to rank a hit by HOW connected it is to the rest of the result set. A flat any-edge boost ranks by WHETHER it is connected at all — a different quantity, and the same membership-vs-degree confusion recorded in #1474.

Every current test uses a fixture where the two agree:

- `{1: 4, 2: 0}` and `{1: 10, 2: 0}` — one connected node, one isolated. Normalised degree of the only non-zero node is 1.0, which is exactly what the flat boost gives.
- `{1: 3}`, `{1: 5}`, `{1: 2, 2: 1}` — the assertions read `proximity == 0.0` for the isolated node, or the max-degree node's score, both of which the flat version also satisfies.

The suite never contains **two nodes with different non-zero degrees whose relative order depends on the difference**. That single missing shape is what separates the intended implementation from the alternative.

## Suggested fix

A fixture where degree ordering decides the outcome, e.g. degrees `{1: 1, 2: 4}` with similarities chosen so the blended scores invert only under degree-proportional proximity:

```
similarity: node 1 = 0.80, node 2 = 0.70
flat boost:      1 -> 0.80+0.25 = 1.05,  2 -> 0.70+0.25 = 0.95   order [1, 2]
degree-weighted: 1 -> 0.80+0.0625 = 0.8625, 2 -> 0.70+0.25 = 0.95  order [2, 1]
```

The two implementations then produce different orderings, so the assertion separates them.

## How this was found

Not by reading the tests. By running the shipped code against a set of *plausible alternative implementations* someone might reasonably have written instead, and checking which the suite rejects. Three of four were caught — unnormalised raw degree, proximity replacing similarity, and self-loops included. The flat-boost alternative was not.

The general rule: a test is evidence only if it separates the chosen implementation from a named alternative. For a change, the alternative is the previous version and the check is mechanical. For new code — as `graph_rerank.py` was — the alternative has to be named by the author, and a test whose docstring cannot say what it would catch may catch nothing.

Related: #1474 (membership-vs-rank in `score_semantic`), #385 (the reranker's parent issue).

vitali87:
Fixes #1481

## The defect

`test_graph_rerank.py` passed in full against an implementation that gives every node with any in-set edge the same flat boost:

```python
prox = 1.0 if proximity.get(node_id, 0.0) > 0 else 0.0
```

`13 passed`. That implementation ranks by **whether** a hit is connected; `graph_rerank.py` exists to rank by **how** connected it is. Same membership-vs-degree confusion as #1474.

## Why nothing caught it

Every existing fixture used one connected node and one isolated node — `{1: 4, 2: 0}`, `{1: 10, 2: 0}`, `{1: 3}`. The connected node's normalised degree is 1.0 and the flat boost is also 1.0, so the two implementations agree on every case in the suite. No assertion could separate them because no fixture contained **two nodes with different non-zero degrees whose order depends on the difference**.

## The fix

Two tests, both in that missing shape.

`test_proximity_is_proportional_to_degree_not_mere_connectedness` picks similarities so the implementations disagree about the resulting **order**, not merely the scores:

```
flat boost:      1 -> 0.80 + 0.25   = 1.05     2 -> 0.70 + 0.25 = 0.95    order [1, 2]
degree-weighted: 1 -> 0.80 + 0.0625 = 0.8625   2 -> 0.70 + 0.25 = 0.95    order [2, 1]
```

Asserting order rather than exact scores keeps it meaningful if `DEFAULT_PROXIMITY_WEIGHT` is retuned. The sibling test pins the mechanism directly: the weaker node's boost is strictly between 0 and 1.

## Verification

Panel of four plausible alternative implementations, each anchored with `assert old in s` before applying:

|---|---|---|
| unnormalised raw degree | 2 failed | 3 failed |
| proximity replaces similarity | 3 failed | 3 failed |
| **flat boost for any edge** | **13 passed** | **2 failed** |
| self-loops included | 1 failed | 1 failed |

RED proven before the fix: both new tests fail against the flat-boost alternative while the 13 existing tests still pass under it — which is the original defect, measured rather than asserted.

## How this was found

Not by reading the tests. By running the shipped code against alternatives someone might reasonably have written instead, and checking which the suite rejects.

The rule this came from: a test is evidence only if it separates the chosen implementation from a **named alternative**. For a change, the alternative is the previous version and the check is mechanical. For new code — as `graph_rerank.py` was — the alternative has to be named by the author, and a test whose docstring cannot say what it would catch may catch nothing.

Related: #1474, #385.

## Summary by CodeRabbit

* **Tests**
  * Expanded graph-ranking coverage to verify that proximity reflects the relative degree of connected nodes.
  * Added checks confirming less-connected results receive a smaller boost than hub nodes.
  * Added validation for zero-degree scenarios to ensure similarity scores remain unchanged.
  * Added validation of the expected relationship set to prevent incomplete or unintended ranking behavior.
  * Added coverage confirming node identifiers are safely passed as query parameters.

### DIFF
--- a/codebase_rag/tests/test_graph_rerank.py
+++ b/codebase_rag/tests/test_graph_rerank.py
@@ -259,6 +259,37 @@ def test_the_proximity_query_filters_on_the_declared_relationships() -> None:
     assert set(match.group(1).split("|")) == set(_PROXIMITY_RELS), match.group(1)
 
 
+def test_node_ids_are_parameterised_not_interpolated() -> None:
+    """Ids reach the engine as parameters, never as query text.
+
+    The fifth axis of this query's contract, and the one nothing else covered:
+    replacing the `$0, $1` placeholders with the ids interpolated directly
+    left all 17 other tests passing.
+
+    Node ids are internal integers rather than user input, so this is not
+    today an injection vector. It is still the property worth pinning, because
+    the alternative is invisible in every other assertion -- the generated
+    Cypher reads correctly, returns the same rows, and the module keeps
+    working. Nothing downstream would report the change.
+
+    Interpolation would also defeat query-plan caching: every distinct result
+    set produces a textually different query, so the engine re-plans each
+    call rather than reusing one plan with new parameters.
+
+    Asserts BOTH halves. That the placeholders are present, and that no id
+    appears literally in the text -- because a query could carry `$0` and
+    still interpolate the rest, and the placeholder check alone would pass.
+    """
+    node_ids = [4321, 8765]
+
+    query = build_proximity_query(node_ids)
+
+    assert "$0" in query, query
+    assert "$1" in query, query
+    for node_id in node_ids:
+        assert str(node_id) not in query, (node_id, query)
+
+
 def test_all_zero_degrees_leave_every_similarity_unchanged() -> None:
     """The `highest <= 0` guard, which no other test enters.
 
--- a/codebase_rag/tests/test_graph_rerank.py
+++ b/codebase_rag/tests/test_graph_rerank.py
@@ -259,6 +259,36 @@ def test_the_proximity_query_filters_on_the_declared_relationships() -> None:
     assert set(match.group(1).split("|")) == set(_PROXIMITY_RELS), match.group(1)
 
 
+def test_all_zero_degrees_leave_every_similarity_unchanged() -> None:
+    """The `highest <= 0` guard, which no other test enters.
+
+    Found by replacing the guard's body with a raising assertion: the suite
+    still passed, so nothing exercised it.
+
+    Unlike the other branches this one is NOT reachable through the shipped
+    query -- `count(*)` does not yield 0 for a row that exists, and a graph
+    with no in-set edges returns no rows at all, which `if not degrees`
+    handles first. So it is a defensive guard against a malformed or foreign
+    row source rather than a production path.
+
+    Pinned anyway, because "unreachable" is a property of the CURRENT query
+    rather than a promise. If `build_proximity_query` ever emits an OPTIONAL
+    MATCH or a left join, zero-degree rows become real and this branch starts
+    running -- and without the guard the next line divides by `highest`.
+
+    Asserts behaviour rather than the early return: proximity is 0.0 and every
+    similarity survives untouched. An implementation that raised, or that
+    dropped the hits, would satisfy "does not divide by zero" too.
+    """
+    ingestor = _ingestor({1: 0, 2: 0})
+
+    ranked = rerank_by_graph_proximity(ingestor, [(1, 0.7), (2, 0.5)])
+
+    assert {hit.node_id: hit.proximity for hit in ranked} == {1: 0.0, 2: 0.0}
+    assert {hit.node_id: hit.score for hit in ranked} == {1: 0.7, 2: 0.5}
+    assert [hit.node_id for hit in ranked] == [1, 2]
+
+
 def test_the_declared_relationships_are_the_intended_six() -> None:
     """An ABSOLUTE list, because the sibling test compares two things that move together.
 
--- a/codebase_rag/tests/test_graph_rerank.py
+++ b/codebase_rag/tests/test_graph_rerank.py
@@ -136,54 +136,76 @@ def test_the_weight_bounds_how_far_proximity_can_move_a_result() -> None:
     assert boosted.score == 0.5 + DEFAULT_PROXIMITY_WEIGHT
 
 
+# A REACHABLE hub fixture: three hits, where node 1 is joined to both 2 and 3
+# and 2-3 are not joined to each other. `build_proximity_query` unwinds both
+# endpoints of every in-set edge, so the two edges 1-2 and 1-3 give:
+#
+#     degree: {1: 2, 2: 1, 3: 1}   normalised: {1: 1.0, 2: 0.5, 3: 0.5}
+#
+# Three nodes is the MINIMUM that can produce unequal degrees. With a two-node
+# result set every matched edge runs between those two nodes and UNWIND emits
+# both endpoints, so their degrees are necessarily equal -- an asymmetric
+# two-node fixture like `{1: 1, 2: 4}` describes a graph the production query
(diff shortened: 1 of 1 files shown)


vitali87:

vitali87:
@greptileai review d676108358e5a82a01ae32d5afe1e9bb67bcefc2

vitali87:
@greptileai review 8d9025e66692b23c948898cec052d277ca6cca9c

Both findings confirmed by reproduction and fixed.

**P1 — unreachable fixture.** Correct, and it is the defect this PR exists to fix, one level down. With a two-node result set every matched edge runs between those two nodes and `UNWIND [id(a), id(b)]` emits both endpoints, so their degrees are necessarily equal; `{1: 1, 2: 4}` describes a graph the production query cannot return. Three nodes is the minimum that can produce asymmetry. Replaced with a reachable hub fixture — node 1 joined to 2 and 3, 2-3 unjoined — giving `{1: 2, 2: 1, 3: 1}`.

**P2 — weight dependence.** Also correct, and my docstring claimed the opposite: it said asserting order 'keeps the test meaningful if DEFAULT_PROXIMITY_WEIGHT is ever retuned'. Measured, the ordering stops separating the implementations below w ≈ 0.133:

```
w=0.25  node1=0.8625 node2=0.9500 -> [2, 1]   discriminates
w=0.10  node1=0.8250 node2=0.8000 -> [1, 2]   does not
```

Now asserts proximity, which is computed before the weight is applied and is therefore weight-independent, plus an order assertion using equal similarities so the entire score gap comes from proximity. Verified RED against flat boost at weights 0.25, 0.10, 0.05 and 0.01 — both tests fail at every one.

**Third change, found while checking this suite for the same class of defect.** `test_the_proximity_query_filters_on_the_declared_relationships` compares the query's `r:` filter against `_PROXIMITY_RELS`, but the query is generated from that tuple. Removing CALLS from the tuple left all 15 tests passing while silently narrowing what 'graph proximity' means. Agreement between two representations of one decision cannot detect a change to the decision, so the six relationships are now pinned absolutely as a third reference.

vitali87:
@greptileai review 3aeb9f452abfd6d1e0db77759df09c7687d63e4d

Added one more test after auditing this module for the shape a peer session hit independently: a fixture that cannot produce the condition its property names.

Probed every guard in `graph_rerank.py` by replacing each body with a raising assertion and re-running. `highest <= 0` fired nothing — **16 passed** — so no test entered that branch.

Worth being precise about what that is, because it is the inverse of the peer's case rather than the same thing. Their untested branch was reachable in production, so it was a genuine coverage gap. Mine is not reachable through the shipped query: `count(*)` does not yield 0 for a row that exists, and a graph with no in-set edges returns no rows at all, which `if not degrees` handles first. It defends against a malformed or foreign row source.

Pinned regardless, because *unreachable* describes the current query rather than a promise. If `build_proximity_query` ever emits an `OPTIONAL MATCH` or a left join, zero-degree rows become real and this branch starts running — and without the guard the next line divides by `highest`.

The assertion is on behaviour, not on the early return: proximity 0.0 and every similarity untouched. An implementation that raised, or that dropped the hits, would also satisfy 'does not divide by zero'.

Verified RED: deleting the guard fails this test with `ZeroDivisionError`. 17 passed on real code.

vitali87:
@greptileai review b7b1b487ba9df127d2cf74dc36b1cd9fccf56a3a

One more, found by applying the lesson from your #1475 review to this module rather than waiting for it to be caught here.

That review found `_holds_ingestor_lock` had more contract axes than I had tested — it covered WHAT is locked and was blind to WHERE. So I enumerated `build_proximity_query`'s axes and probed each:

| axis | covered by |
|---|---|
| which relationships | pinned absolutely |
| both endpoints in set |  |
| self-loops excluded |  |
| both endpoints unwound |  |
| **parameterisation** | **nothing** |

Replacing the `$0, $1` placeholders with ids interpolated directly into the Cypher left **all 17 other tests passing**.

Node ids are internal integers, not user input, so this is not today an injection vector — I would rather state that than imply a severity the code does not have. It is still the property worth pinning, because the alternative is invisible in every other assertion: the generated query reads correctly, returns the same rows, and the module keeps working. Interpolation would also defeat query-plan caching, since every distinct result set yields a textually different query.

Asserts both halves — placeholders present, and no id appearing literally — because a query could carry `$0` and interpolate the rest, which a placeholder-only check would pass.

RED verified: interpolating fails only this test. 18 passed on real code.

vitali87:
@greptileai review b7b1b487ba9df127d2cf74dc36b1cd9fccf56a3a

Correcting the table in my previous comment — three cells came out empty because my shell interpreted the backticks. The content, properly:

| axis | covered by |
|---|---|
| which relationships | `test_the_declared_relationships_are_the_intended_six` |
| both endpoints in set | `test_the_query_counts_only_edges_inside_the_result_set` |
| self-loops excluded | `test_self_loops_are_excluded_from_proximity` |
| both endpoints unwound | `test_the_query_counts_both_endpoints_explicitly` |
| **parameterisation** | **nothing** |

The finding stands as described: replacing the `$0, $1` placeholders with ids interpolated directly into the Cypher left all 17 other tests passing.

Node ids are internal integers, not user input, so this is not today an injection vector — I would rather state that than imply a severity the code does not have. It is still worth pinning, because the alternative is invisible in every other assertion: the generated query reads correctly, returns the same rows, and the module keeps working. Interpolation would also defeat query-plan caching, since every distinct result set yields a textually different query.

Asserts both halves — placeholders present, and no id appearing literally — because a query could carry `$0` and interpolate the rest, which a placeholder-only check would pass.

RED verified: interpolating fails only this test. 18 passed on real code.

Issue #227, opened 2026-01-06
Traceback to graph correlation: rank root cause candidates via CALLS/FLOWS_TO, served over MCP


vitali87:
## Summary

Correlate a failure artefact (traceback, panic, paused frame stack) with the code graph, and walk the graph backwards to rank root cause candidates. This is a query layer over edges the project already builds.

## Proposal

Given a traceback (or panic output, or a DAP frame stack), the tool would:

1. **Map frames to graph nodes.** Each frame carries file path, line and symbol name; resolve these to `Function`/`Method` nodes via qualified name and line span, in the same way dead code analysis resolves definitions.
2. **Walk backwards from the failure site.** Use `CALLS` edges to enumerate paths that reach the failing frame, and `FLOWS_TO` edges to find where the failing value could have come from. The writers that can reach the failing read are the ranked root cause candidates.
3. **Return a compact, ranked report** suitable for an agent context window: candidate sites with file:line, the edge path connecting them to the failure, and source snippets.

## Surface

Expose as MCP tools on the existing server, no new product surface:

- `explain_traceback(traceback_text, project)`: frame resolution plus graph context per frame (callers, callees, flow sources).
- `rank_root_causes(traceback_text, project)`: the backwards walk with ranked candidates.

## Building blocks

- Typed `CALLS` edges across all supported languages (`codebase_rag/parsers/call_processor.py`)
- Opt in `FLOWS_TO` taint edges (`codebase_rag/parsers/flow_access/processor.py`)
- Reachability machinery in `codebase_rag/dead_code.py`
- MCP server (`codebase_rag/mcp/`) as the delivery surface

## Scope

In scope: parsing common traceback formats (Python first, then one compiled language to prove the shape generalises), frame to node resolution, backwards `CALLS`/`FLOWS_TO` walk, MCP tool surface, tests against fixture projects with known planted bugs.

Out of scope: breakpoints, stepping, watchpoints, live process attachment, any DAP server implementation.

## Acceptance criteria

- [ ] Python traceback text resolves to graph nodes with a measured resolution rate on fixture projects
- [ ] `rank_root_causes` places the planted defect in the top 3 candidates on fixture scenarios
- [ ] Both tools exposed via the MCP server with tests
- [ ] Graceful degradation when `FLOWS_TO` edges are absent (falls back to `CALLS` only walk)

## Extension: automatic diagnosis

[...]

PR #1483, opened 2026-08-27
feat: report a measured frame resolution rate on traceback correlation

vitali87:
Closes the last open acceptance criterion on #227.

## What was missing

Three of the four criteria were already met on `main` — I counted them against the checklist rather than estimating:

| Criterion | State |
|---|---|
| `rank_root_causes` places the planted defect in the top 3 | ✅ `test_rank_places_the_flow_writer_in_the_top_candidates` |
| Both tools exposed via MCP with tests | ✅ `test_mcp_crash_correlation.py` |
| Graceful degradation when `FLOWS_TO` is absent | ✅ `test_rank_degrades_to_calls_only_when_flow_is_absent` |
| **Python traceback resolves with a measured resolution rate** | ❌ **this PR** |

`FrameContext.unresolved_reason` already said *which* frames failed to resolve. Nothing said *how many*. A stack where one frame in three resolved and one where all three did produced reports that read alike, which is the number an agent needs to judge whether a report is worth acting on.

## The change

`FrameResolutionRate(total, resolved)` on `TracebackReport`, with `rate` as a property, mapped explicitly across the MCP boundary.

**Named `FrameResolutionRate`, not `ResolutionStats`** — `trace.resolution.ResolutionStats` is already imported in that module and counts something else (one frame's resolver attempts). My first version shadowed it and silently broke `_resolve_stack`; the type error surfaced it immediately.

**Mapped explicitly in `mcp/tools.py`** because that handler builds its response field by field and `rate` is a property, so `_asdict()` omits it. A measurement the MCP surface does not return is indistinguishable from one never computed — the defect Greptile caught on #1478.

## Mutation proof, including one my own test missed

Red proven first: all three tests failed with `AttributeError` before the implementation existed. Then each named alternative:

```
rate over RESOLVED (always 1.0)       -> test_..._counts_only_the_frames_that_resolved  FAILED
MCP drops the field                   -> test_..._exposes_the_resolution_rate_over_mcp  FAILED
vacuous 1.0 for an empty stack        -> 3 passed   NOT CAUGHT
```

The third **survived**, and its docstring claimed to guard exactly that case. The fixture had `total == 1`, so the `if self.total` guard was always taken and the empty branch never ran — a test asserting a property its fixture cannot reach.

Confirmed the branch is reachable (a frameless traceback yields `total == 0`) and added `test_resolution_rate_of_a_frameless_traceback_is_zero_not_one`. Re-verified: the vacuous-1.0 mutation now fails, and only that test.

## Checks

`ruff check` and `ruff format` clean. 20 passed across both crash-correlation test files. `ty` reports the same 2 diagnostics as `main`, neither in a touched file.

## Summary by CodeRabbit

* **New Features**
  * Added traceback resolution statistics showing total frames, resolved frames, and resolution rate.
  * Included resolution details in crash explanation results.
  * Resolution rates now support fully, partially, and unsuccessfully resolved tracebacks.

* **Bug Fixes**
  * Correctly reports a 0.0 resolution rate for empty or entirely unresolved tracebacks.
  * Ensures unresolved frames are not mistakenly counted as resolved.

* **Tests**
  * Added coverage for resolution statistics across common traceback scenarios and crash explanation responses.

### DIFF
--- a/codebase_rag/crash_correlation.py
+++ b/codebase_rag/crash_correlation.py
@@ -285,6 +285,15 @@ def explain_traceback(
         flow_gaps=graph.flow_gaps,
         resolution=FrameResolutionRate(
             total=len(contexts),
+            # Keyed on `qualified_name`, NOT on `unresolved_reason is None`.
+            # Today the two agree, because every failure path in
+            # `FrameResolver.resolve` records a reason before returning None.
+            # But `_resolve_stack` derives the reason with
+            # `next(iter(stats.unresolved), None)`, so a future failure path
+            # that forgets to record one would yield reason=None with no
+            # qualified name -- and the other predicate would then count an
+            # unresolvable frame as resolved. The qualified name is what the
+            # rate is ABOUT; the reason is diagnostic text alongside it.
             resolved=sum(1 for frame in contexts if frame.qualified_name is not None),
         ),
     )
--- a/codebase_rag/crash_correlation.py
+++ b/codebase_rag/crash_correlation.py
@@ -75,11 +75,43 @@ class FrameContext(NamedTuple):
     flow_sources: tuple[str, ...]
 
 
+class FrameResolutionRate(NamedTuple):
+    """How much of a traceback the graph could actually anchor (issue #227).
+
+    Named for the frames it counts rather than "resolution stats", because
+    `trace.resolution.ResolutionStats` is already imported here and counts
+    something else entirely -- the resolver's internal attempts for ONE frame.
+    Shadowing it silently broke `_resolve_stack`.
+
+    Per-frame ``unresolved_reason`` says WHICH frames failed to resolve; this
+    says HOW MANY, which is what decides whether a report is worth acting on.
+    A stack where one frame in three resolved and one where all three did are
+    different artefacts, and without the aggregate they read alike.
+    """
+
+    total: int
+    resolved: int
+
+    @property
+    def rate(self) -> float:
+        """Resolved frames over ALL frames, 0.0 when there are none.
+
+        The denominator is every frame in the traceback, including those
+        outside the repository. Dividing by the resolved count instead would
+        report 1.0 for every report ever produced, and skipping unresolved
+        frames would hide exactly the gap this measures. An empty stack
+        resolves nothing, so it scores 0.0 rather than a vacuous 1.0 -- the
+        same trap as an average over no cases.
+        """
+        return self.resolved / self.total if self.total else 0.0
+
+
 class TracebackReport(NamedTuple):
     exception_type: str
     exception_message: str
     frames: tuple[FrameContext, ...]
     flow_gaps: tuple[str, ...]
+    resolution: FrameResolutionRate
 
 
 class RootCause(NamedTuple):
@@ -251,6 +283,10 @@ def explain_traceback(
         exception_message=parsed.exception_message,
         frames=tuple(contexts),
         flow_gaps=graph.flow_gaps,
+        resolution=FrameResolutionRate(
+            total=len(contexts),
+            resolved=sum(1 for frame in contexts if frame.qualified_name is not None),
+        ),
     )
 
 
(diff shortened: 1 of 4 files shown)


vitali87:
@greptileai review

Test-only addition at `c196a117`, plus a clarifying comment on the predicate.

Applying @vscode-link-terminal-scheme's axis probe — enumerate a contract's axes and mutate each — found one uncovered on my own new code. Keying `resolved` on `unresolved_reason is None` instead of `qualified_name is not None` passed all 16 tests.

The two predicates agree today: every failure path in `FrameResolver.resolve` records a reason before returning `None`, so a frame with neither a qualified name nor a reason is not reachable through the shipped resolver. That makes this defensive rather than a coverage gap — so the test pins **behaviour**, not an early return, since "unreachable" describes the current resolver and not a promise. `_resolve_stack` derives the reason via `next(iter(stats.unresolved), None)`, so a single future failure path that forgets to record makes the divergence real, and the wrong predicate would then count an unresolvable frame as resolved.

The test patches the resolver to *be* that future path and asserts through `explain_traceback`, so it covers the production line that chooses the predicate. My first version asserted over a hand-built `FrameResolutionRate` and did **not** catch the mutation — it restated the rule instead of exercising the call site. Verified: the mutation now fails, and only this test.

vitali87:
@greptileai review

Test-only change at `7f8b06b2`.

Continuing the axis probe on my own change found a weak fixture in the MCP test. Swapping `total` and `resolved` in the response mapping passed all 4 tests, because the fixture was a **fully resolved** stack where `total == resolved == 2` — the swap was invisible.

The fixture now includes an out-of-repo library frame, so `total=2, resolved=1, rate=0.5` and the swap is observable. Verified: the swap now fails, and only that test.

Worth noting the mechanism, since it recurred three times on this PR: the fixture I reached for first was the *illustrative* one (a clean fully-resolved stack), and illustration selects for clarity rather than for separating candidates. Dropping any individual key was already caught — the gap was specific to values being right rather than present.

vitali87:
@greptileai review

Test-only change at `5c2c2cdd`.

**The number 0.5 was doing less work than it appeared to.** The MCP fixture asserted `total=2, resolved=1, rate=0.5`. Measured, five different formulas produce 0.5 at that input:

```
resolved/total  (correct)    = 0.5
1 - resolved/total           = 0.5
unresolved/total             = 0.5
resolved/(resolved+1)        = 0.5
1/total                      = 0.5
```

So the assertion held for four wrong implementations. The suite caught an inverted rate collectively — via the fully-resolved and nothing-resolved tests — but this test could not catch it alone.

Fixture now has two unresolvable frames, giving `total=3, resolved=1, rate=1/3`, which separates all five. Verified: the inverted-rate mutation now fails this test on its own, where before it passed.

**Also swept two things I had not:** `mcp/tools.py:517` is the only non-test consumer of `TracebackReport` (no CLI surface), so the hand-mapped `rate` property cannot be silently dropped elsewhere. And the `FrameResolver.resolve` monkeypatch does not leak — verified under random ordering, which is the CI condition.

vitali87:
@greptileai review

Test-only change at `43f9abf2`, and a **correction to my previous comment**: I claimed `1/3` "separates all five" formulas. It does not — `1/total` also yields 1/3 at `total=3`.

Searched for a ratio that collides with nothing, rather than iterating on near-misses:

```
total=5 resolved=2  rate=0.4     UNAMBIGUOUS
total=5 resolved=3  rate=0.6     UNAMBIGUOUS
total=6 resolved=2  rate=0.3333  UNAMBIGUOUS
```

Both partial fixtures now use **2/5**. Verified each wrong formula against the two partial tests *in isolation* rather than against the whole suite:

```
inverted           partial-only: 2 failed
unresolved/total   partial-only: 2 failed
1/total            partial-only: 2 failed
resolved/(r+1)     partial-only: 2 failed
hardcoded /2       partial-only: 2 failed
```

Previously those two assertions contributed nothing — the property was pinned only by the fully-resolved and nothing-resolved cases. A guard that discriminates only with help from its siblings loses coverage the moment one is weakened, and nothing records which test was load-bearing.

Credit to @vscode-link-terminal-scheme, whose independent probe enumerated six colliding formulas at `0.5` where I had found five, including the inverted rate — which reports the *failure* rate while looking like a success rate.

Issue #1185, opened 2026-08-10
PHP namespace-aware qualification, then php-parser NameResolver facts


vitali87:
## Why

PHP carries the repo's only `LIMITATION:` comment (`parsers/call_resolver.py:1173`): CGR qualifies PHP functions **by file path** and does not track namespaces, so cross-file call resolution in any namespaced codebase (i.e. all modern PHP) is systematically wrong. Total current PHP depth is a 62-LOC handler (`parsers/handlers/php.py`) plus three import helpers in `import_processor.py:3648-3703` — no `parsers/php/` package, no type inference.

### Stage 1 (pure tree-sitter): namespace-aware qualification

- A real `parsers/php/` package with a proper handler: parse `namespace_definition`, `use` / group-`use` / aliases, and build namespace-aware qns (`Vendor\Pkg\Sub\func` style mapped onto CGR's dotted qn scheme).
- Resolve calls through the current namespace, imported aliases, and the global fallback rule (unqualified function calls fall back to the global namespace — PHP's actual lookup order).
- This alone retires the `LIMITATION:` comment and is the prerequisite for everything below.

### Stage 2: php-parser `NameResolver` facts

nikic/php-parser's `NameResolver` visitor emits the fully-qualified name of **every** symbol occurrence — compiler-grade naming for the cost of a `php` binary. A small bundled PHP tool (seed: `evals/oracles/php_oracle/`) walks the resolved AST and emits `LanguageFrontend` facts: resolved call sites (including `ClassName::method`, `new`, instanceof), external proofs for symbols outside the repo, INHERITS/IMPLEMENTS by resolved name.

### Stage 3 (opt-in): PHPStan collector

For framework-heavy code (Laravel facades, Symfony containers) where names alone can't resolve dynamic dispatch, a PHPStan collector rule can export inferred types and call targets. Requires `composer install`, so strictly opt-in and last.

## Tests

Fixture PHP repo per stage: namespaced functions with `use` aliases and global fallback (stage 1); trait usage, static calls, interface implementation (stage 2). The existing `php_l1`/`php_retrieval` evals should show measurable precision gains after stage 1 alone.

Stages 2+ depend on the `LanguageFrontend` protocol issue; stage 1 does not.

vitali87:
claimed by feat-duplicates-clickable-locations

[...]

Issue #105, opened 2025-08-02
Support Scala (fully)


vitali87:

vitali87:
claimed by vscode-link-terminal-scheme (branch: feat/scala-baseline-parity)

This issue has an empty body, so treating #1186 as the specification — it lays out Scala support in two stages. Starting stage 1 (tree-sitter baseline parity: handler registration, import parsing, definitions, INHERITS/IMPLEMENTS).

Full support per this issue's title also requires stage 2 (SemanticDB semantic facts), which needs an opt-in compile via sbt/bloop/mill. This issue stays open until both land.

vitali87:
Keeping the `claimed` label. Stage 1 merged via #1472 and #1176; stage 2 (SemanticDB) is blocked on an absent toolchain — no sbt, bloop, mill, coursier or even a JVM on this machine, verified by execution. Detail on #1186, which is the specification this issue tracks.

vitali87:
Releasing this claim — I am not working it and have not been.

I claimed it while surveying the backlog, then went to bug work (#1492, #1494, #1495, #1498) and never came back. No open PR touches this issue, so the lock has been blocking availability rather than protecting work in progress.

Dropping the label. Anyone can take it.

PR #1484, opened 2026-08-27
feat: resolve PHP `use function` imports through declared namespaces

vitali87:
Addresses #1185 — **stage 1 only**. Stages 2 and 3 (php-parser `NameResolver` facts) remain open on the issue, matching how #105 tracks its two Scala stages.

## The defect

CGR qualifies PHP by file path and ignored the `namespace` declaration, so a `use function App\Text\format` import could never match the registered `project.text.format`. The miss fell through to the simple-name trie, which binds to whichever same-named function it reaches first:

```
before: ('php_ns.caller.render', 'php_ns.money.format')   ← never imported
after:  ('php_ns.caller.render', 'php_ns.text.format')
```

A **wrong** edge rather than a missing one. The graph gained a relationship the source never expressed, and nothing downstream could tell — a missing edge is visibly absent, a wrong edge is silently confident.

This is the repo's only `LIMITATION:` comment (`call_resolver.py`).

## Why not fold namespaces into qualified names

That would rewrite the identity of every existing PHP node in the graph, turning a resolution fix into a graph migration — and identity changes break incremental update, dead-code reachability, and every stored reference at once. PSR-4 also makes namespace and directory independent, so the qn could not be derived from the path even in principle.

Instead the declaration goes into a side map (`php_module_namespaces`, module qn → dotted namespace) consulted at import resolution: split the target into namespace and symbol, find the module *declaring* that namespace, look the symbol up there. Blast radius stays at the resolution step.

## Refusing to guess

`_php_target_for_namespace_import` returns `None` whenever the answer is not unique — no module declares the namespace, or several do and more than one defines the symbol. The trie fallback then applies exactly as before, so this can only replace an arbitrary binding with a determined one, never introduce a new arbitrary one.

Only top-level declarations are recorded, and the unnamed `namespace { }` global block is skipped: binding it to `""` would match the namespace part of every unqualified import target.

## Verification

Five alternative implementations. **Two initially passed** and are now covered:

|---|---|---|
| disable the fix entirely | caught | caught |
| take FIRST match, not unique | **5 passed** | caught |
| ignore the namespace match | caught | caught |
| record the empty namespace | 5 passed | *not a gap — see below* |
| drop the per-module reset | **5 passed** | caught |

The re-index gap is the one worth naming: no test parsed the same module twice, so **incremental re-index — the normal path in a running cgr — was untested**. Editing a file's `namespace` would have left both the old and new bound forever.

The empty-namespace alternative I investigated rather than "fixed". Removing either guard alone changes nothing because two guards enforce it redundantly; removing **both** makes the test fail. The behaviour is covered and the guards are defence-in-depth, so adding a test there would have produced a green assertion that cannot distinguish the implementations.

RED proven before the fix, then re-proven after: disabling the resolution fails exactly one test.

PHP and import suites: **554 passed, 2 skipped** — the skips are `torch` and Xdebug, both missing optional deps unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PHP namespaced function import resolution, including case-insensitive matching.
  * Imports now bind to the correct namespaced function when similarly named functions exist elsewhere.
  * Ambiguous or unavailable namespace matches remain unresolved instead of linking to an arbitrary function.
  * Re-indexing correctly updates or removes outdated namespace information.
  * Prevented non-PHP code and class-style imports from incorrectly resolving to PHP functions.
  * Improved handling of files containing multiple or global namespace blocks.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -1298,12 +1298,53 @@ class CallResolver:
 
         return self._try_resolve_wildcard_imports(call_name, import_map)
 
+    def _php_target_for_namespace_import(self, imported_path: str) -> str | None:
+        """Map a PHP `use function A\\B\\c` target onto the qn that registers it.
+
+        CGR qualifies PHP by file path, so the import target `App.Text.format`
+        never equals the registered `project.text.format`. Previously that miss
+        dropped through to the simple-name trie, which binds to whichever
+        same-named function it reaches first -- so `use function App\\Text\\format`
+        could resolve to `App\\Money\\format` (issue #1185). A wrong edge, not a
+        missing one.
+
+        Splits the target into the namespace it names and the symbol within it,
+        then finds the module that DECLARED that namespace and looks the symbol
+        up there. PSR-4 maps a namespace prefix to a directory root, so the two
+        cannot be derived from one another -- the declaration is the only link.
+
+        Returns None rather than guessing whenever the answer is not unique:
+        no module declares the namespace, or several do (PHP permits one
+        namespace across many files, and two files may both define `format`).
+        The trie fallback then applies exactly as before, so this can only
+        replace an arbitrary binding with a determined one, never introduce a
+        new arbitrary one.
+        """
+        namespace, separator, symbol = imported_path.rpartition(cs.SEPARATOR_DOT)
+        if not separator or not namespace or not symbol:
+            return None
+        namespaces = self.import_processor.php_module_namespaces
+        matches = [
+            f"{module_qn}{cs.SEPARATOR_DOT}{symbol}"
+            for module_qn, declared in namespaces.items()
+            if declared == namespace
+        ]
+        found = [qn for qn in matches if qn in self.function_registry]
+        if len(found) != 1:
+            return None
+        return found[0]
+
     def _try_resolve_direct_import(
         self, call_name: str, import_map: dict[str, str]
     ) -> tuple[str, str] | None:
         if call_name not in import_map:
             return None
         imported_qn = import_map[call_name]
+        if imported_qn not in self.function_registry and (
+            php_qn := self._php_target_for_namespace_import(imported_qn)
+        ):
+            logger.debug(ls.CALL_DIRECT_IMPORT, call_name=call_name, qn=php_qn)
+            return self.function_registry[php_qn], php_qn
         if imported_qn in self.function_registry:
             logger.debug(ls.CALL_DIRECT_IMPORT, call_name=call_name, qn=imported_qn)
             return self.function_registry[imported_qn], imported_qn
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -1288,7 +1288,7 @@ class CallResolver:
                 return None
             import_map = {}
 
-        if result := self._try_resolve_direct_import(call_name, import_map):
+        if result := self._try_resolve_direct_import(call_name, import_map, language):
             return result
 
         if result := self._try_resolve_qualified_call(
@@ -1335,13 +1335,25 @@ class CallResolver:
         return found[0]
 
     def _try_resolve_direct_import(
-        self, call_name: str, import_map: dict[str, str]
+        self,
+        call_name: str,
+        import_map: dict[str, str],
+        language: cs.SupportedLanguage | None = None,
     ) -> tuple[str, str] | None:
         if call_name not in import_map:
             return None
         imported_qn = import_map[call_name]
-        if imported_qn not in self.function_registry and (
-            php_qn := self._php_target_for_namespace_import(imported_qn)
+        # Gated on the CALLER's language. The namespace map is keyed by module
+        # qn and holds only PHP modules, but the import targets it is matched
+        # against are dotted strings that any language can produce -- a JS
+        # `import { format } from ...` recorded as `App.Text.format` would
+        # resolve straight into a PHP function. Reproduced in review on #1484:
+        # a JavaScript caller bound to a PHP target, an edge across languages
+        # that no source expressed.
+        if (
+            language == cs.SupportedLanguage.PHP
+            and imported_qn not in self.function_registry
+            and (php_qn := self._php_target_for_namespace_import(imported_qn))
         ):
             logger.debug(ls.CALL_DIRECT_IMPORT, call_name=call_name, qn=php_qn)
             return self.function_registry[php_qn], php_qn
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -1304,7 +1304,9 @@ class CallResolver:
                 return None
             import_map = {}
 
-        if result := self._try_resolve_direct_import(call_name, import_map, language):
+        if result := self._try_resolve_direct_import(
+            call_name, import_map, language, module_qn
+        ):
             return result
 
         if result := self._try_resolve_qualified_call(
@@ -1396,24 +1398,53 @@ class CallResolver:
             return None
         return found[0]
 
+    def _is_php_function_import(self, call_name: str, module_qn: str | None) -> bool:
+        """Whether `call_name` arrived via `use function`, not a class `use`.
+
+        Returns False when `module_qn` is unknown rather than defaulting to
+        True: an unidentified module cannot demonstrate the import was a
+        function binding, and the fallback's whole purpose is to avoid
+        asserting a target it cannot justify. The trie fallback still applies,
+        so this can only decline to add an edge, never remove a correct one.
+        """
+        if module_qn is None:
+            return False
+        return call_name in self.import_processor.php_function_imports.get(
+            module_qn, frozenset()
+        )
+
     def _try_resolve_direct_import(
         self,
         call_name: str,
         import_map: dict[str, str],
         language: cs.SupportedLanguage | None = None,
+        module_qn: str | None = None,
     ) -> tuple[str, str] | None:
         if call_name not in import_map:
             return None
         imported_qn = import_map[call_name]
-        # Gated on the CALLER's language. The namespace map is keyed by module
-        # qn and holds only PHP modules, but the import targets it is matched
-        # against are dotted strings that any language can produce -- a JS
+        # Gated on the CALLER's language AND on the import being a
+        # `use function` binding.
+        #
+        # The language gate: the namespace map is keyed by module qn and holds
+        # only PHP modules, but the import targets it is matched against are
+        # dotted strings that any language can produce -- a JS
         # `import { format } from ...` recorded as `App.Text.format` would
         # resolve straight into a PHP function. Reproduced in review on #1484:
         # a JavaScript caller bound to a PHP target, an edge across languages
         # that no source expressed.
+        #
+        # The `use function` gate: PHP keeps classes and functions in SEPARATE
+        # symbol tables, so `use App\Text\format` imports a CLASS while
+        # `use function App\Text\format` imports a function. Without this, a
+        # class or constant import whose name happens to match a registered
+        # PHP function resolved to that function -- also reproduced in review,
+        # and it SURVIVED the language gate because the caller is PHP. The two
+        # are independent axes. `php_function_imports` already records which
+        # local names arrived via `use function`, so this needs no new parsing.
         if (
             language == cs.SupportedLanguage.PHP
+            and self._is_php_function_import(call_name, module_qn)
             and imported_qn not in self.function_registry
             and (php_qn := self._php_target_for_namespace_import(imported_qn))
         ):
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -67,6 +67,22 @@ def _split_receiver_chain(expr: str) -> list[str]:
 
 PY_EXTERNAL_TARGET: tuple[str, str] = ("", "")
 
+# PHP folds A-Z only when comparing namespace and function names, so
+# `use function app\text\FORMAT` binds to `App\Text\format`.
+#
+# NOT `str.casefold()`, which folds the full Unicode range: PHP treats
+# identifiers differing outside ASCII as DISTINCT, so Unicode folding would
+# match names the language does not, trading a missed binding for a wrong one.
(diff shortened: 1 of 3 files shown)


vitali87:

vitali87:
@greptileai review 9a01a5121c87b62074aef40d2a9cbd9b0be4aa65

vitali87:
@greptileai review 3ff15fbe1feb1a614f232ac57dbb2f5602ae8fe7

Both P1s confirmed by my own reproduction before fixing, and both are real.

## P1-1 — cross-language leak

The worst finding against this change, because **the fix introduced a new class of wrong edge — exactly the defect it exists to remove.**

```
JS caller, target 'App.Text.format'  ->  proj.text.format
```

The namespace map holds only PHP modules, but the import targets matched against it are dotted strings any language can produce. I reasoned carefully about which *targets* could match and never about which *callers* could reach the helper — the reachability side of the guard went unasked.

Now gated on the caller's language. The test asserts JS, TS, Python and `None` all resolve to nothing **and** that PHP still resolves, so the guard scopes the feature rather than disabling it.

## P1-2 — multiple namespace blocks

Confirmed:

```
namespace App\First { function helper() }
namespace App\Second { function helper() }
->  recorded: {'proj.multi': 'App.First'}
```

Both blocks' functions share one module qn, so an import of `App\First\helper` binds to the `helper` defined in `App\Second`.

Worth stating plainly what happened here, since it is not a case of missing the risk. My own docstring said recording the first block was *"deliberate rather than a simplification: a file with two namespace blocks has no single answer, and guessing one would produce exactly the confident-wrong binding this whole change exists to remove."* I named the trap, in the code I was writing, and implemented the guess anyway.

Multi-block files now record **nothing** and fall through to the trie — where they were before this feature existed. Never worse than the status quo, and never asserting an answer it does not have. Same discipline as the Python frontend's `len(names) != 1` → no fact, never a guess.

## Verification

RED proven for both, each failing only the test that names it:

| mutation | result |
|---|---|
| remove the caller-language guard | fails only the cross-language test |
| revert to first-block-wins | fails only the multi-block test |

Scoped regression run over the shared call path — `_try_resolve_direct_import` is language-agnostic, so JS/TS and general resolution tests are what would catch a regression from threading `language` through: **1075 passed, 3 skipped**.

One correction to my own process worth recording: the first run of the new cross-language test failed with an `AttributeError`, not an assertion — my `SimpleNamespace` double lacked `commonjs_direct_exports`, which the non-PHP path falls through to. A fixture defect presenting as a code failure. Read at face value it would have had me modify working code to satisfy a broken test.

vitali87:
@greptileai review 982c960feda3cab86b9fb8721022868a473651db

Two further fixes since the P1 round.

## CodeRabbit's case-sensitivity finding — valid, fixed

PHP namespaces and function names are case-insensitive. Verified by executing it under PHP 8.5 rather than from the documentation:

```php
namespace App\Text; function format(string $s) { return "formatted:$s"; }
use function app\text\FORMAT;   // prints "formatted:x"
```

Reproduced the gap before fixing:

```
'App.Text.format' -> proj.text.format
'app.text.FORMAT' -> None
```

A valid import falling through to the simple-name trie — the same wrong-edge path this feature exists to prevent.

Fixed with an **ASCII-only** fold, which is the load-bearing detail. PHP folds `A-Z` only, so `str.casefold()` would match identifiers the language treats as **distinct**, trading a missed binding for a wrong one. `str.lower()` has the same defect (Turkish dotless i, Kelvin sign). Folding is for comparison only; the registered qualified name is returned unchanged, since it is the graph's real key.

Both halves are covered independently — one test varies only the namespace case, another only the symbol case — so an implementation folding just one half fails.

## A defect my own tests were hiding

`ty` caught the case-fold commit iterating `self.function_registry`. That object is a `FunctionRegistryTrieProtocol`: `__contains__`, `get`, `keys`, `items`, `find_with_prefix` — **no `__iter__`**. Not iterable at all.

The tests passed because my doubles were plain dicts, which do iterate. A double more permissive than the real object means the suite was green against an interface production never provides, and every mutation in my alternative panel would have passed too, since they all ran against that same permissive double.

Now uses `find_with_prefix`, the registry's own scoped query — iteration would have been a full registry scan per import even where it worked. The test doubles carry `find_with_prefix` so they exercise the method the resolver actually calls.

## Verification

| mutation | result |
|---|---|
| exact-case comparison | fails both case-insensitivity tests |
| `str.casefold()` instead of ASCII fold | fails only the ASCII-only test |

The ASCII-only test pins the boundary with the Kelvin sign plus a control asserting `str.casefold()` **does** match it, so it cannot pass vacuously.

Scoped regression over the shared call path: **1077 passed, 3 skipped** — skips are `torch`, `node`/`tsc` and Xdebug, all missing optional tooling.

One environment note: `ty check` initially failed in this worktree on unresolvable optional imports (`torch`, `numpy`, `qdrant_client`). It passes on pristine `main` elsewhere, and after `uv sync --all-extras` every hook passes here — a worktree venv difference, confirmed by installing rather than assumed.

vitali87:
@greptileai review 982c960feda3cab86b9fb8721022868a473651db

This P1 describes code that is no longer at the head — the review predates the fix rather than assessing it.

**Timing:**

| event | time |
|---|---|
| review posted | 2026-08-27T02:57:30Z |
| `982c960f` (the case-fold fix) authored | 2026-08-27T03:43:13Z |

46 minutes apart, so the review could not have seen it.

**Verified at `origin/feat/php-namespace-qualification` rather than from recollection:**

```
_php_fold / _PHP_ASCII_FOLD references          7
namespace compared via fold                     1   (if _php_fold(declared) == wanted)
test_a_mixed_case_import_resolves_like_php_does 1
```

The finding is nonetheless **correct about the defect**, and CodeRabbit reported the same thing independently — I fixed it in `982c960f` before this review landed. Both halves are handled:

- **namespace casing** — `_php_fold(declared) == wanted`
- **symbol casing** — `find_with_prefix` on the matched module, comparing `_php_fold(qn[len(prefix):]) == wanted_symbol`, so a registry entry differing only in symbol case is still found

The fold is **ASCII-only**, which the suggestion also called for and which matters: PHP folds `A-Z` only, so `str.casefold()` would match identifiers the language treats as distinct — trading a missed binding for a wrong one. The registered qualified name is returned unchanged.

The requested regression test exists, covering both halves independently — one case varies only the namespace casing, another only the symbol casing, so an implementation folding just one half fails. A second test pins the ASCII boundary using the Kelvin sign, with a control asserting `str.casefold()` **does** match it so the test cannot pass vacuously.

Re-review requested against `982c960f` explicitly.

vitali87:
@greptileai review 75ca89f2802bb671d7b20300c82cc04df98dd942

**Correcting my previous comment: the P1 was not stale, and I was wrong to call it so.**

I judged staleness by `createdAt` (02:57Z) against my push (03:43Z) and concluded the review predated the fix. It did not — Greptile **edits its summary comment in place**, so `createdAt` records the first post, not the review's currency. Reading the anchor instead shows it was already pointing at `982c960f`, my current head. My own project notes warn about exactly this, and I used the timestamp anyway.

The finding is live and correct.

## The defect

`find_with_prefix` splits the prefix on dots and walks a trie node per part, so a trailing separator produces a final empty part that is never a key:

```
find_with_prefix("proj.text.")  ->  []
find_with_prefix("proj.text")   ->  [('proj.text.format', Function)]
```

The case-insensitive **symbol** fallback was therefore dead code. A valid `use function app\text\FORMAT` still fell through to the simple-name trie whenever the symbol casing differed — the original wrong-edge defect in another shape.

## Why my tests missed it

`_FakeRegistry` subclassed `dict` and implemented `find_with_prefix` with `startswith`, which accepts a trailing separator. A double **more permissive than the real object** — the same failure as the previous commit's dict-that-iterates, one layer along. Every mutation in my alternative panel had been running against that stand-in rather than the shipped structure.

Replaced it with the real `FunctionRegistryTrie`. It is cheap to construct, so approximating it was never justified.

## Verification

RED against the real trie: restoring the trailing separator fails `test_a_mixed_case_import_resolves_like_php_does`. With the fake registry that same mutation **passed** — that contrast is the measurement showing the double was the problem, not the assertion.

The prefix boundary is now explicit: the remainder must start with the separator, which also rejects a sibling module whose name merely starts with this one (`proj.textutil` vs `proj.text`).

vitali87:
@greptileai review

Requesting a scored review of the current head `75ca89f2802bb671d7b20300c82cc04df98dd942`.

Your last full review is anchored to `982c960f`, one commit behind. The intervening commit is the fix for the P1 you raised — the trailing-separator prefix that made the case-insensitive symbol fallback unreachable — plus the replacement of the permissive test double with the real `FunctionRegistryTrie`.

Every substantive check passes at this head: CodeQL, CodeRabbit, Sonar Zero Issues Gate, Type Check, Lint & Format, the full platform matrix (Windows/macOS/Ubuntu on py3.12 and py3.13) and Integration Tests. The only red checks are `Greptile 5/5 Gate` and `All Checks Pass`, both waiting on a verdict for this head.

vitali87:
@greptileai review a8a75c38e03e70b0626b54b2519cfa5120b95a22

Addressed the remaining CodeRabbit thread — a **second, distinct** defect, not the case-sensitivity one.

## The finding

PHP keeps classes and functions in **separate symbol tables**: `use App\Text\format` imports a class, `use function App\Text\format` imports a function. The namespace fallback ignored the distinction. Reproduced before fixing:

```
class-style `use` (php_function_imports empty)
    ->  (NodeType.FUNCTION, 'proj.text.format')
```

A class or constant import whose name happens to match a registered PHP function bound to that function — an edge the source never expressed.

## Why the earlier guard did not cover it

It **survived the caller-language guard**, because the caller *is* PHP. The language gate and the binding-kind gate are independent axes, and covering one said nothing about the other — the same "predicate has more axes than the author enumerated" shape as the lock detector's WHAT-versus-WHERE on #1475.

Worth noting I nearly resolved this thread on assumption. It was marked outdated and I expected it to be the already-fixed case-sensitivity finding; reading it instead of assuming is what surfaced it.

## The fix

`php_function_imports` already records which local names arrived via `use function`, so the distinction needs no new parsing — `module_qn` is threaded through to reach it.

An unknown `module_qn` returns **False** rather than defaulting True: a module that cannot be identified cannot demonstrate the binding kind, and defaulting True would reinstate the defect wherever the caller module is not threaded through. Declining costs only the trie fallback, which is where such calls went before this feature existed.

## Verification

| mutation | result |
|---|---|
| remove the `use function` gate | 2 failed — both new tests |
| unknown module defaults to True | 1 failed — the unknown-module test |

Each new test is paired with a positive assertion, so it pins the gate rather than the fallback simply being switched off.

Hooks all pass including `ty check`. Scoped regression over the shared call path: **991 passed, 1 skipped** (the skip is `torch`, an absent optional dep).

vitali87:
@greptileai review 1b4828a40f68919a83c2936d1ee55f8689ddfe3b

Your P1 confirmed and fixed, plus a fourth axis I found by probing rather than waiting for the next round.

## The P1 — call-site casing

Verified against real PHP 8.5 before fixing:

```php
namespace App\Text; function format(...)
use function App\Text\format;
echo FORMAT("x");        // prints F:x
```

And reproduced my gap:

```
call as 'format' -> (Function, 'proj.text.format')
call as 'FORMAT' -> None
```

`_is_php_function_import` did an exact-case membership test, so it recognised the **import** but not the **call**, and the call fell through to the trie — the same wrong-edge path, reached through the alias instead of the target. Now folded, with a control asserting a never-imported name still declines.

## A fourth axis, found by probing

Three findings on this PR arrived as review P1s in sequence — import-path casing, binding kind, call-site casing — each surfacing only after I fixed the previous one. That pattern is itself the signal, so I enumerated the remaining axes of this function's contract and exercised them directly.

One was a real defect:

```
use function \App\Text\format;   ->  recorded as '.App.Text.format'
```

A fully-qualified import is idiomatic PHP and runs identically (verified: prints `B:x`). The import processor dots the path, so the leading backslash became a leading separator that never matched a declared `App.Text`. In PHP a `use` path is always resolved from the global namespace, so stripping it is a spelling normalisation, not a semantic change.

Two other axes probed and found **already correct**, so no change was made:

| axis | behaviour | verdict |
|---|---|---|
| `use function X as Y` | resolves through the rename | correct |
| target is a METHOD | declines | correct — `use function` names a free function |

## Verification

RED proven for both changes: reverting the fold fails only the call-site test; removing the strip fails only the fully-qualified test.

One scoping correction worth recording. My first version of the call-site test also asserted `Format`, which failed — but at the earlier `call_name not in import_map` check, a different layer from the gate under test. I narrowed it to the two spellings the finding is about rather than keep an assertion that passes for the wrong reason.

Hooks all pass including `ty check`. Scoped regression over the shared call path: **992 passed, 1 skipped** (`torch`, an absent optional dep).

vitali87:
@greptileai review e80c0bf705c96bb998352beb3d5bfca1290db8b1

Your P1 confirmed — and you were right where I was wrong. I had narrowed the `Format` case out of my test calling it "a different layer". It was a different layer, and it was the one that mattered.

Verified against the import processor rather than assumed:

```
use function App\Text\format;  ->  import_mapping keys: ['format']
                                    php_function_imports: {'format'}
```

Production keys the map by the **declaration** spelling, so a call written `FORMAT()` misses at `call_name not in import_map` and never reaches the folded binding-kind gate. My previous fix was unreachable on the real path.

**My fixture was hiding it.** It seeded both `format` and `FORMAT` into the import map — a state production cannot generate — so the assertion passed against an unreachable condition. That is the unreachable-fixture shape, and the second time on this PR that one of my doubles was more permissive than the real object.

Fixed at the layer where the miss happens: `_php_import_key` folds the map lookup itself. The fixture is now production-shaped (declaration spelling only) and covers `format`, `FORMAT` and `Format`.

`_php_import_key` returns None unless **exactly one** key folds to the call name. Two keys differing only in case cannot both be intended, and PHP rejects the duplicate `use` at compile time, so declining is correct rather than cautious.

## Verification

| mutation | result |
|---|---|
| `_php_import_key` always returns None | fails only the call-site test |
| drop the `len(matches) == 1` guard | fails only the ambiguity test |

The ambiguity guard was **uncovered** until this round — dropping it changed nothing, because no fixture had two keys folding to the same name. Found by isolating the mutation properly: my first attempt disabled the whole PHP path and broke 5 tests, which is too coarse to be a signal.

Hooks all pass including `ty check`. 15 tests in the PHP namespace suite.

Issue #1448, opened 2026-08-26
Markdown authoring awareness: file purpose, deduplication, and cross-file synchronisation


vitali87:
Splitting the second half of #164 into its own issue.

#164 carries eleven acceptance criteria. The first five are parser-level and are essentially complete: Markdown is parsed to a tree-sitter AST, headings are identified (ATX and setext), sections are chunked and nested by heading level, and #1445 adds `LINKS_TO` edges for links to other repository files.

The remaining six are a different feature. They are not about parsing Markdown; they are about an agent reasoning over a corpus of Markdown files it is authoring and maintaining:

- **File purpose awareness** — distinguish a plan from a summary from a TODO from a spec, so the agent knows what it is looking at.
- **Deduplication logic** — detect and prevent redundant content across files when generating or updating.
- **Update synchronisation** — when one file is updated, reflect consistent changes in related files where applicable (shared goals, task lists).
- **Plan consolidation** — offer to consolidate scattered plans into a single coherent file when several related ones exist.
- **Contextual file selection** — when asked to generate a plan, pick the most relevant existing file rather than creating a new one by default.
- **Metadata tagging** (marked optional in #164) — e.g. `purpose: planning`, `scope: service-X`, to assist selection and updates.

These share little with the parsing work beyond the file type, and each is substantial on its own. Keeping them on #164 makes that issue unclosable indefinitely and hides how far the parsing half has actually got.

## Worth settling before implementation

Several of these need a design decision that has not been made anywhere:

1. **Where purpose comes from.** Inferred from content and headings, declared in front-matter, or both with declared winning? Inference is the useful default but is the part that can be silently wrong.
2. **What "related" means for synchronisation.** The graph now has `LINKS_TO` edges from #1445, which is one plausible relation, but two plans sharing a goal need not link to each other.
3. **Whether the agent edits files unprompted.** Synchronisation and consolidation both imply writing to files the user did not name. That wants an explicit opt-in and a diff shown before it happens.

Filed rather than left implicit on #164, where the remaining criteria read as nearly-done alongside the parsing ones.

vitali87:
claimed by feat-duplicates-clickable-locations

Taking **the metadata-tagging bullet only** — the one item here that needs none of the three unsettled design decisions this issue names. Declared front-matter is not inferred, so "inference can be silently wrong" does not apply; and reading it writes nothing, so "whether the agent edits files unprompted" does not arise either.

[...]

PR #1488, opened 2026-08-27
feat: read declared Markdown front-matter onto the Module node

vitali87:
Addresses #1448 — **the metadata-tagging bullet only**. The other five remain open and unclaimed on that issue.

## Why this bullet is separable

#1448 names three design decisions nobody has made: where an inferred purpose comes from, what "related" means for synchronisation, and whether the agent may edit files unprompted. Five of its six bullets are gated on those.

Metadata tagging is not. Front-matter is **declared**, not inferred, so "inference can be silently wrong" does not apply; and reading it writes nothing, so the unprompted-edit question does not arise.

The precondition was verified rather than assumed — `tree-sitter-markdown` already exposes the block as a `minus_metadata` node, and `document_tier.py` ignored it.

## The schema audit corrected the design mid-build

The first version exploded front-matter keys onto the Module node. The repo's own audit rejected it:

```
Module 'md_fm.plan_md' has undocumented property 'purpose'
```

That is exactly right. The node schema is a fixed, audited property list, so free-form keys would let any document define any node property — including ones a future schema wants for something else. It is now **one declared `front_matter` property** holding sorted `key=value` entries, added to `NODE_SCHEMAS`.

I would not have chosen that shape unprompted; the audit is what made the constraint visible.

## Refusing rather than guessing

`parse_front_matter` returns nothing when the input is not unambiguously front-matter:

- no opening fence on the **first** line — a `---` elsewhere is a horizontal rule or a setext underline
- no closing fence — an unterminated block would swallow the document into node properties
- an empty key, or a key the ingestion layer owns (`path`, `qualified_name`, …)

A single malformed line is *skipped* rather than discarding the block, since front-matter is hand-written and a stray line is likelier than a wholly invalid block.

Absence stays absent: a document with no front-matter gains no property at all, so "declares no purpose" remains distinguishable from "declares an empty purpose".

## Verification

Alternative panel, four plausible implementations — **two initially passed** and are now covered:

|---|---|---|
| unterminated block swallows the file | caught | caught |
| split on every colon (`url: https://…` → `https`) | caught | caught |
| **accept a fence anywhere, not just line 1** | **10 passed** | caught |
| **allow reserved keys** | **10 passed** | caught |

The fence-position gap took two attempts. My first replacement fixture *also* failed to discriminate: its pairs sat after the first fence found, so a position-blind scan read a blank line and returned nothing either way — the fixture described the defect rather than separating the implementations. The discriminating shape needs a `key: value` on line 2 and a fence on line 3, which is a setext heading (`Title` underlined with `---`) and the realistic case where body text gets swallowed.

`ty` also caught a real annotation mismatch — `dict[str, str]` against a `dict[str, list[str]]` value. My first correction invented `dict[str, object]`; the repo already declares `PropertyDict`, which is what `ensure_node_batch` accepts.

12 tests in the new file. Regression over the shared emit path (`emit_flat_module` is used by the ast-grep tier too): **729 passed, 2 skipped**.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading scalar metadata from Markdown front matter.
  * Preserved valid front-matter metadata on imported document modules as sorted properties.
  * Added support for Rust module metadata fields.

* **Bug Fixes**
  * Improved handling of malformed, incomplete, quoted, and colon-containing front matter.
  * Prevented document metadata from overriding reserved identity fields.
  * Cleared previously stored metadata when front matter is removed.
  * Documents without valid front matter now receive an empty metadata value.


vitali87:

vitali87:
@greptileai review 63650a85

vitali87:
Both findings confirmed. One fixed, one I am scoping out with reasons rather than half-doing under review pressure.

## Fixed: CI base-install failure

Not in your review, but the more urgent problem — my own tests failed on `Unit Tests (base install)`:

```
AssertionError: no Module node emitted for plan.md
```

The two ingestion tests index a fixture through the real updater, so without `tree_sitter_markdown` no Module node exists at all — **a missing optional dependency reported as a product defect**. My worktree had the full extras, so local green said nothing about that platform.

Guarded with `skipif` on the ingestion class, matching the `pytest.importorskip` guard `test_document_tier.py` already uses for the same grammar. The **ten parser tests stay unguarded on purpose**: they are pure string handling, and the base install is exactly where the contract they pin is easiest to break unnoticed.

Pushed as `d69d1892`.

## Confirmed and scoping out: protobuf export

You are right, and I verified it rather than taking it on trust. `codec/schema.proto` lists Module properties **field by field** (`decorators = 4`, `rust_cfg_test_mods = 5`, … `generator = 9`), so a new schema property is silently dropped on export.

I also checked whether an existing guard would have caught it. It would not — `test_protobuf_relationship_parity.py` and `test_generated_sources.py` both pass unchanged against this branch, and neither compares `NODE_SCHEMAS` against the proto for node **properties**. So the drift is real *and* currently unguarded, which makes it a bigger finding than this PR.

Fixing it properly means a coordinated change across three artifacts — the proto (field 10 is free), the regenerated `schema_pb2.py`, and the codec mapping — plus the parity guard that should have caught it in the first place. That is a separate concern from reading front-matter, and bolting it on mid-review is how the last two PRs acquired their P1s.

**Proposed:** land this PR without the export path, and I will open the protobuf work as its own PR including the missing `NODE_SCHEMAS`-vs-proto parity test. If you would rather this PR carry both, say so and I will do it here instead — but I would want the parity guard in the same change either way, since without it the next property drifts the same way.

## Fixed: the nested/comment finding

Confirmed by reproduction, and worse than described — ```
parent:
  child: v   ->  {'parent': '', 'child': 'v'}
tags:
  - a          ->  {'tags': ''}
```

The nested case is the real damage: `child: v` was **hoisted to top level**, so a key declared under a parent became indistinguishable from one the author declared at the document level.

Three guards added, each RED-verified independently — dropping the indent check, the comment check, or the empty-value check each fails only its own test. The empty-value one matters beyond tidiness: recording `{"tags": ""}` asserts the author declared it empty, which is a different claim from declaring a list this parser does not represent, and a consumer cannot tell them apart afterwards.

15 tests now.

**Comment by vitali87** [comment]
Formatting correction to my previous comment — a code fence ran into the prose. The reproduction, properly:

```
parent:
  child: v      ->  {'parent': '', 'child': 'v'}
tags:
  - a           ->  {'tags': ''}
```

Content unchanged; all three guards are pushed at `96a90757` with 15 tests passing.

Also confirming `ty check` fails in this worktree only on unresolvable optional imports (`torch`, `numpy`, `qdrant_client`, `pymilvus`, `transformers`) — zero errors in the changed files.

**Comment by vitali87** [comment]
@greptileai review 96a90757

Two of your three findings are fixed since your review at `63650a85`; the third is scoped out with reasons in my earlier comment.

- **nested/comment lines** — three guards, each RED-verified independently. `child: v` under `parent:` was being hoisted to top level, which is the damaging case: a nested key became indistinguishable from a top-level declaration.
- **base-install failure** — my own ingestion tests failed CI without `tree_sitter_markdown`. Guarded with `skipif` on the ingestion class only; the ten parser tests stay unguarded deliberately, since the base install is where their contract is easiest to break unnoticed.
- **protobuf export** — confirmed real and left for a separate PR. It needs proto + regenerated bindings + codec mapping, *plus* a `NODE_SCHEMAS`-vs-proto parity guard that does not exist today.

15 tests, zero failures on this head.

**Comment by vitali87** [comment]
Followed the protobuf finding to its root and filed **#1490**.

Your finding is correct and it is **not specific to this PR**. Measured across the whole schema:

```
  Project.root_path      Module.front_matter    Class.modifiers
  Class.start_col        Function.modifiers     Function.start_col
  Function.name_start_line   Function.name_start_col   Function.is_macro
  Method.modifiers       Method.start_col       Method.name_start_line   …
```

Validated against a control first — `modifiers` appears nowhere in `schema.proto`, so these are real absences rather than an artefact of my matching.

And the guard genuinely does not exist. Verified empirically rather than by reading: `front_matter` is in `NODE_SCHEMAS`, absent from the proto, and **21 parity tests pass anyway**. The relationship side is guarded; the node side is not.

So `Module.front_matter` is the 25th instance of a systemic gap, not a defect this PR introduces. Fixing only the instance would leave the mechanism that produced the other 24 intact — which is why #1490 proposes a default-deny parity test with an explicit `_NOT_EXPORTED` allow-list carrying reasons, the same inversion the MCP lock guard needed in #1475.

That reinforces the split I proposed: land the front-matter reading here, do the export work under #1490 where it can be reviewed against the schema change it actually requires.

**Comment by vitali87** [comment]
@greptileai review 43e0dc08ddc3d3ca3e3267536b9318e4313e45e5

P1 confirmed and fixed. Reproduced all four shapes first:

```
tags: [a, b]   ->  {'tags': '[a, b]'}
meta: {k: v}   ->  {'meta': '{k: v}'}
note: |        ->  {'note': '|'}
note: >        ->  {'note': '>'}
```

The block-scalar cases are the worse half: `note: |` stored the literal string `"|"` — punctuation mistaken for content — while the actual text below was silently dropped, since this parser skips the indented lines carrying it.

All **six** markers are handled (`|` `>` with the `-` and `+` chomping variants). A guard matching only the bare forms would leave four spellings still storing punctuation.

The collection check tests the **first character only**, with a control asserting `title: Draft [v2]` survives. Rejecting any value containing a bracket would silently drop ordinary metadata — a fix worse than the defect, and the mutation confirms it: making the check match anywhere fails that control.

| mutation | result |
|---|---|
| allow block-scalar markers | fails only the marker test |
| allow flow collections | fails only the collection test |
| reject any bracket anywhere | fails only the control |

18 tests.

**Comment by vitali87** [comment]
Head correction: the review request above names `43e0dc08`, but a `ruff format` commit landed after it. Current head is **`0d032cab`** — same content plus formatting.

@greptileai review 0d032cab

**Comment by vitali87** [comment]
@greptileai review 0258a6cb377598e65068ca200144c8030f6484c7

CodeRabbit's stale-`front_matter` finding confirmed and fixed — a genuinely good catch, and the same defect class as the PHP re-index bug on #1484.

Verified the mechanism rather than taking it on trust: `cypher_queries.py:317` upserts with `SET n += row.props`, which **merges**. A key omitted on re-ingest keeps its previous value, so a document that dropped its front-matter kept the old metadata bound to its node — the graph asserting a declaration the file no longer makes.

Omission cannot express *"this document has no front-matter"*; it can only fail to contradict whatever was there before. `front_matter` is now emitted **always**, empty list included, because an empty list overwrites.

**My own test encoded the wrong contract.** `test_a_document_without_front_matter_gains_no_properties` asserted the property should be absent entirely — which reads correctly against a single ingest and leaves the re-ingest path broken, exactly where the defect lives. Rewritten to assert the empty list.

Added the regression the finding asked for: index with `purpose: planning`, drop the block, re-index, assert the second emission carries `[]`. Asserting the empty list rather than merely *"changed"*, since "changed" is also satisfied by writing some other wrong value.

RED-verified: reverting to omit-when-empty fails **two** tests — the contract and the regression. 19 tests.

**Comment by vitali87** [comment]
@greptileai review c4eb4e03d64c49def5a8267ab54b978aac25ce97

Block-scalar finding confirmed and fixed. Reproduced first — every header carrying an indentation digit was stored as its own value:

```
note: |2   ->  {'note': '|2'}
note: |2-  ->  {'note': '|2-'}
note: |-2  ->  {'note': '|-2'}
```

That is the same defect the guard was added to fix, in the six spellings the guard did not name. **The enumerated set was the wrong shape**: YAML allows an optional chomping indicator and an optional explicit indentation digit in either order, so a fixed list can only ever cover the spellings someone thought of.

Replaced with `^[|>](?:[-+]?\d*|\d*[-+]?)$` — a complete header and nothing else. Twelve forms now covered.

The control matters as much as the fix. A looser rule like "starts with `|` or `>`" would drop `>>= operator` and `a|b`, silently losing ordinary metadata — worse than the defect. Both directions are RED-verified:

| mutation | result |
|---|---|
| revert to the enumerated set | fails the marker test |
| over-broad prefix match | fails the control |

20 tests.

**On the protobuf half:** confirmed real, and deliberately not fixed here. It is not specific to this PR — 25 declared properties across 14 node labels have no proto field, and no guard existed. Filed as **#1490** with the guard in **#1491** (5/5), which is default-deny so a 26th cannot drift in unnoticed. Adding proto fields needs a per-label decision about what belongs in the export; I would be guessing.

**Comment by vitali87** [comment]
Head correction: my review request above names `c4eb4e03`; a ruff import-formatting commit landed after it. Current head is **`6c2123e8`** — same content.

@greptileai review 6c2123e8afd10982f5eb5f867d1e56d6d33b1677

**Comment by vitali87** [comment]
@greptileai review 6acda2a0

CodeRabbit's empty-after-unquote finding confirmed and fixed. Reproduced first:

```
k: ""   ->  {'k': ''}
k: ''   ->  {'k': ''}
```

The emptiness check ran **before** quote-stripping, so the quoted form slipped past it and stored an empty string — asserting the author declared the key empty, which is precisely the claim that guard exists to avoid making. My own docstring argued the point while the code allowed the exception.

The check now runs on the unquoted value. The control asserts `k: "x"` still yields `{'k': 'x'}`, since dropping every quoted string would be worse than the defect.

RED verified: moving the check back before unquoting fails exactly this test. **21 tests.**

Running total on this PR — six findings, all fixed:

| finding | source |
|---|---|
| base-install failure (my own tests) | CI |
| nested-key hoisting | Greptile |
| flow collections and block-scalar markers | Greptile |
| stale `front_matter` on re-ingest | CodeRabbit |
| block-scalar headers with indentation digits | Greptile |
| empty-after-unquote | CodeRabbit |

The protobuf export finding remains deliberately out of scope — it is not specific to this PR (25 properties across 14 labels), and the guard for it is **#1497**.

**Comment by vitali87** [comment]
@greptileai review 346a021fd98eb4c2945abb941ba4e2670da864ef

Inline-comment finding confirmed and fixed. Reproduced first:

```
note: | # explanation  ->  {'note': '| # explanation'}
note: >- # note        ->  {'note': '>- # note'}
note: |2 # why         ->  {'note': '|2 # why'}
```

The `$`-anchored pattern required the marker to END the line, so a header carrying a trailing comment stored the marker plus the comment while the indented text below was skipped — the same defect the pattern was written to fix, in the spellings it did not reach.

The comment is stripped **only** to test for a header, and only when preceded by whitespace. The stored value is always the full text.

**The control needed two attempts, and the second is the interesting one.** My first version asserted `C# notes`, `a#b` and `tag #1` survive — and an any-hash strip passed it, because none of those is a header, so mangling them inside the header test is invisible.

`|#x` is the discriminating case: YAML needs whitespace before `#` to start a comment, so it is a plain scalar, but an unconditional strip reduces it to `|`, which then matches the header pattern and the value is **dropped** rather than truncated.

| mutation | result |
|---|---|
| no comment stripping | fails the inline-comment test |
| strip on any hash | fails the `\|#x` control |
| store the stripped value | fails the preservation test |

**23 tests.** Seven findings on this PR now, all fixed.

**Comment by vitali87** [comment]
The one open thread — *"Commented block headers misclassified"* — is **already fixed**, at `346a021f`. Your review anchors to `6c2123e8`, two commits behind, so the thread predates the fix rather than describing current code.

Verified at origin's head rather than from recollection:

```
_without_trailing_comment references at origin/feat/markdown-front-matter:  2
parse_front_matter('note: | # explanation')  ->  {}
```

Not resolving the thread myself — a re-review at the current head is the honest way to close it, since I would otherwise be marking my own work as verified.

Current head is `08ac143f`, which also carries two SonarCloud fixes found by querying the issues API at the head rather than waiting for the gate:

- **S9073** — a composite assertion that could not say which half failed, where *"no Module node emitted"* is a different diagnosis from *"the value is wrong"*
- **S3776** — `parse_front_matter` reached complexity 19 as guards accumulated; extracted `_front_matter_pair` so each rule is named rather than nested

After that refactor I mutated **all seven** guards to confirm none had been silently disabled. Six discriminate; the empty-value guard no longer does, because the unquote check downstream subsumes it. I neither deleted it nor added a test that cannot discriminate — the redundancy is now documented with the measurement, since the two guards answer different questions and only coincide today.

23 tests, nine findings resolved on this PR.

vitali87:
Fixed at head — confirmed by running the reproduction rather than reading the diff.

\`\`\`
plain block scalar -> {'title': 'Hi'}
trailing comment   -> {'title': 'Hi'}
folded + comment   -> {'title': 'Hi'}
hash inside value  -> {'title': 'A # B'}
\`\`\`

`note: | # explanation` now drops the block scalar instead of storing `| # explanation` as a literal value.

The fourth case is the control that matters: `title: A # B` keeps its hash. YAML only begins a comment at a **whitespace-preceded** `#`, so stripping every `#` would have traded this bug for a value-corruption bug affecting far more documents. `_without_trailing_comment` strips only the whitespace-preceded form.

Pinned by a test over three header shapes — `| # explanation`, `>- # note`, and the indent-indicator `|2 # why` — so the indicator digit is not accidentally consumed as part of the comment.

vitali87:
@greptileai review

vitali87:
Fixed in e75428a0. Added \`repeated string front_matter = 10\` to the \`Module\` message and regenerated the bindings.

Two readback tests go through the real export path (\`GraphUpdater\` -> \`ProtobufFileIngestor\` -> read \`index.bin\` back), covering the populated and the empty case. Both fail against the previous bindings and pass against the new ones, so they discriminate this defect rather than just asserting the new behaviour. A mock would have agreed the property was passed and still missed the drop, since \`ensure_node_batch\` skips unmapped properties silently.

The empty case is part of the contract rather than an absence worth skipping: \`SET n += row.props\` merges, so omitting the property on re-ingest cannot express that a document no longer declares metadata.

Worth noting how it surfaced. Rebasing onto current main pulled in the node-property parity guard from #1490, and it failed immediately with:

\`\`\`
1 node propert(ies) are declared in NODE_SCHEMAS but have no field in
codec/schema.proto and are not listed in _NOT_EXPORTED:
  Module.front_matter
\`\`\`

That guard is default-deny by construction, so a newly declared property is refused until someone records its disposition. This is the first time it has caught a real gap.

vitali87:
@greptileai review

vitali87:
Fixed in 0e7b5ddf. Reproducing it turned up a second defect the finding did not mention, so both are covered.

Before:

\`\`\`
status: # planned      -> {"status": "# planned"}    invented metadata
status: planned # later -> {"status": "planned # later"}  comment kept
\`\`\`

After:

\`\`\`
status: # planned      -> {}            null value, declares nothing
status: planned # later -> {"status": "planned"}
lang: C#                -> {"lang": "C#"}       untouched
tag: a#b                -> {"tag": "a#b"}       untouched
note: "a # b"           -> {"note": "a # b"}    quoted, so content
\`\`\`

One cause behind both: \`_without_trailing_comment\` already existed with the right semantics but was only applied when testing for a block-scalar header, never to ordinary scalars.

Comments are stripped from **unquoted** values only, since quoting is how an author says the \`#\` is content. The leading-\`#\` case needed handling separately because that helper requires whitespace before the marker and cannot see a \`#\` at index 0.

**One pre-existing assertion changed, and I want to flag it rather than bury it.** \`test_a_hash_inside_an_ordinary_value_is_preserved\` asserted \`k: tag #1\` stored \`tag #1\`. I wrote that assertion, and it encoded this exact defect: by YAML rules the whitespace-preceded \`#\` IS a comment, so the declared value is \`tag\`. The other two assertions in that test were correct and are unchanged. I added the quoted form as the escape hatch so \`tag #1\` remains expressible.

Calling it out because "changed a test to make it pass" and "the test encoded the defect" look identical in a diff, and only one is legitimate.

Four mutations, each caught: strip from quoted values (2 failed), drop the leading-hash case (1), bare \`#\` split ignoring the whitespace rule (3), never strip from scalars (2).

vitali87:
@greptileai review

PR #1491, opened 2026-08-27
test: guard node properties against the protobuf schema

### LINKED ISSUE #1490: Node properties in NODE_SCHEMAS can be absent from schema.proto with nothing checking
`NODE_SCHEMAS` and `codec/schema.proto` declare node properties independently, and nothing checks that they agree. **25 declared properties currently have no proto field**, so they are silently dropped on export.

## Measured

```
  Project.root_path
  Module.front_matter
  Class.modifiers
  Class.start_col
  Function.modifiers
  Function.start_col
  Function.name_start_line
  Function.name_start_col
  Function.is_macro
  Method.modifiers
  Method.start_col
  Method.name_start_line
  …
```

Validated against a control before reporting: `modifiers` appears nowhere in `schema.proto`, so these are real absences rather than an artefact of the matching.

## Why this is not simply "the proto is a subset"

It may well be intended as one — the proto lists node fields individually and makes no lossless claim. But three things make the current state a defect rather than a design:

1. **Nothing records the intent.** There is no comment saying which properties are deliberately outside the export, so a reader cannot distinguish "excluded on purpose" from "forgotten".
2. **Nothing enforces it either way.** `test_protobuf_relationship_parity.py` and `test_generated_sources.py` both pass with a property missing — I verified this by adding `Module.front_matter` and watching them stay green. The relationship side *is* guarded; the node side is not.
3. **The failure is silent and one-directional.** A new property is added to `NODE_SCHEMAS`, ingestion writes it, queries read it, and export drops it. Nothing reports the loss.

The relationship messages carry `google.protobuf.Struct properties`, which would absorb arbitrary keys — nodes have no equivalent, so this cannot be solved the same way.

## Suggested fix

A parity test in the shape of the existing relationship one: for every `NodeSchema`, assert each declared property either has a proto field or appears in an explicit `_NOT_EXPORTED` allow-list carrying a reason. That makes the 25 a deliberate, documented set rather than an accident, and makes the 26th impossible to add unnoticed.

Default-deny rather than an inventory of what to check, for the reason the MCP lock guard needed the same inversion in #1475: a check that iterates a list of known properties cannot fail for the property nobody remembered to list.

## How this surfaced

Review on #1488 flagged that a newly declared `Module.front_matter` would not survive protobuf export. That is true, and checking whether an existing guard would have caught it turned up the other 24 — so the specific finding is an instance of a general gap, and fixing only the instance would leave the mechanism intact.

Filed separately rather than folded into #1488: the export work needs the proto, regenerated bindings, and the codec mapping changed together, and the parity guard is what stops the next property drifting the same way.

vitali87:
Fixes #1490

## The gap

`NODE_SCHEMAS` and `codec/schema.proto` declare node properties independently, and nothing checked they agree. A property added to the schema is written at ingestion, read by queries, and **silently dropped on export** — no error, no warning.

The sibling `test_protobuf_relationship_parity.py` guards the relationship *enum* for exactly this failure mode. The node side had no equivalent.

Measured across 14 node labels — the absences are real, verified against the generated descriptors rather than by grepping the proto text:

```
Project.root_path            Module.front_matter       Class.modifiers
Class.path                   Function.is_macro         Function.name_start_line
Method.overrides_external    Interface.docstring       Type.start_line   …
```

## How it surfaced

Review on #1488 flagged that `Module.front_matter` would not survive export. Checking whether an existing guard would have caught it turned up dozens of properties in the same position — so the specific finding was an instance of a general gap, and fixing only the instance would have left the mechanism intact.

## Default-deny

An inventory of "properties to check" cannot fail for the property nobody remembered to list — the same inversion the MCP lock guard needed in #1475. Every declared property must have a proto field **or** be named in `_NOT_EXPORTED`.

`_NOT_EXPORTED` records the **current state**, measured. No entry has been reviewed for whether it *should* be exported; #1490 carries that question. The value is that the set is now closed: a 26th absence fails this test and has to be argued for.

## Two controls keep the list honest

- An entry that gains a proto field must be **removed** — otherwise the list only grows and a reader cannot tell which entries are current.
- An entry naming a property that does not exist is **rejected** — otherwise a renamed property leaves an exemption that silently covers whatever later reuses the name.

## Verification

| mutation | result |
|---|---|
| drop `front_matter` from the allow-list | fails the default-deny test |
| add an already-exported property to the list | fails the staleness control |
| add a nonexistent property to the list | fails the existence control |

Each fails exactly one test. Scope is the guard only — adding proto fields for the existing absences needs someone to decide, per label, whether each belongs in the export, and I would be guessing.

## Summary by CodeRabbit

* **New Features**
  * Added support for reading YAML front matter from Markdown documents.
  * Preserves valid metadata as sorted `key=value` entries on the document record.
  * Ignores unsupported, reserved, malformed, or empty metadata entries.
* **Bug Fixes**
  * Prevents declared metadata from overwriting core document properties.
* **Tests**
  * Added coverage for front-matter parsing, ingestion, and property schema consistency.

### DIFF
--- a/codebase_rag/constants/graph.py
+++ b/codebase_rag/constants/graph.py
@@ -348,6 +348,8 @@ SCHEMA_OPTIONAL_SUFFIX = "?"
 NODE_PROJECT = NodeLabel.PROJECT
 
 KEY_PARAMETERS = "parameters"
+# Declared Markdown front-matter, as sorted "key=value" entries (issue #1448).
+KEY_FRONT_MATTER = "front_matter"
 KEY_DECORATORS = "decorators"
 # Target-module qn candidates of `#[cfg(test)] mod NAME;` declarations in a
 # Rust file, stored on the DECLARING module's node (issue #1010). The
--- a/codebase_rag/parsers/document_tier.py
+++ b/codebase_rag/parsers/document_tier.py
@@ -55,6 +55,78 @@ _MAX_HEADING_LEVEL = 6
 
 DOCUMENT_EXTENSIONS: frozenset[str] = frozenset({".md", ".markdown"})
 
+# YAML front-matter delimiter. The grammar exposes the whole block as a
+# `minus_metadata` node, but its contents are opaque -- tree-sitter-markdown
+# does not parse YAML -- so the pairs are read here.
+_FRONT_MATTER_FENCE = "---"
+
+# Property names a document may NOT declare, because the ingestion layer owns
+# them. A front-matter `path:` would otherwise overwrite the node's real path
+# and break every consumer that resolves a Module back to a file.
+_RESERVED_FRONT_MATTER_KEYS: frozenset[str] = frozenset(
+    {
+        cs.KEY_QUALIFIED_NAME,
+        cs.KEY_NAME,
+        cs.KEY_PATH,
+        cs.KEY_ABSOLUTE_PATH,
+        cs.KEY_START_LINE,
+        cs.KEY_END_LINE,
+        cs.KEY_HEADING_LEVEL,
+    }
+)
+
+
+def parse_front_matter(text: str) -> dict[str, str]:
+    """Read declared YAML front-matter into flat string properties.
+
+    Deliberately NOT a YAML parser. Only top-level `key: value` scalars are
+    read; nested structures, lists and multi-line values are skipped rather
+    than flattened, because a graph node property is a scalar and inventing a
+    representation for a list would be a schema decision this issue does not
+    have (#1448).
+
+    Declared metadata, not inferred: #1448 lists five other bullets that need
+    design decisions about inference and unprompted edits, and this one is
+    separable precisely because it reads what the author wrote.
+
+    Refuses rather than guesses in three cases, each of which would otherwise
+    put non-metadata into node properties:
+
+    - no opening fence on the FIRST line (a `---` elsewhere is a horizontal
+      rule or a setext underline, not front-matter)
+    - no closing fence (an unterminated block would swallow the document)
+    - a reserved key that the ingestion layer owns
+
+    A single malformed line is skipped rather than discarding the block:
+    front-matter is hand-written, so a stray line is likelier than a wholly
+    invalid block, and the valid pairs around it still carry meaning.
+    """
+    lines = text.splitlines()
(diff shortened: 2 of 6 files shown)


vitali87:

vitali87:
@greptileai review 5549c13a

vitali87:
@greptileai review d498d3e3

Your 5/5 is anchored to `5549c13a`; one commit has landed since, addressing CodeRabbit's finding on the same PR.

**The finding:** the guard did `if fields is None: continue`, so a node label with *no proto message at all* passed silently — the inventory-fails-open shape reproduced inside the test written to prevent it.

**It took three attempts, and the failures are the point:**

1. Asserted only on `_proto_fields`. Reverting to silent-skip **still passed** — no real label lacks a message, so the assertion could not reach the branch it was about.
2. Inlined a copy of the loop into the test. Worse: a **mirror** that passes against a broken production path. The mutation confirmed it — production reverted, test green.
3. Extracted `_unexported_properties` so the test calls the **real** comparison over a stand-in schema. Reverting to `continue` now fails exactly this test.

Only the mutation distinguished the three; each intermediate version looked correct and would have shipped a guard that could not fail.

4 tests. The scope is unchanged — the guard only, with `_NOT_EXPORTED` recording the current state so the set is closed. Adding proto fields for the existing absences still needs a per-label decision I would be guessing at.

vitali87:
Closing in favour of **#1497**, which contains the same guard built on `main`.

**This PR was branched from the front-matter work by mistake**, so it carried seven unrelated commits into a review meant to cover one test file. That is my error in worktree hygiene, not a review finding.

Worth recording how it surfaced, because the guard caught its own author. Rebuilding on clean `main` made `test_the_unexported_list_names_only_declared_properties` fail immediately:

```
_NOT_EXPORTED names Module.front_matter, which is not a declared property
```

`front_matter` exists only on the #1488 branch. The staleness control — written so an entry naming a nonexistent property is rejected — fired on its first real exposure, against the person who wrote it. That is exactly the behaviour it was for: without it, the guard would have shipped an exemption covering a property that does not exist on `main`, silently exempting whatever later reused the name.

Your 5/5 review at `d498d3e3` still applies to the guard's content; #1497 is the same four tests with the stale entry removed and no unrelated commits.

PR #1496, opened 2026-08-27
fix: report a rejected API key as an auth error, not a count failure

### LINKED ISSUE #1493: token count failed
Ask a question (Press Ctrl+J or Ctrl+E to submit, Enter for new line, Shift+Tab to toggle mode):
2026-08-27 13:21:37.607 | Context token count failed: 401:
{"type":"error","error":{"type":"authentication_error","message":"invalid ─────────────────────────────
x-api-key"},"request_id":"req_blabla"}r

vitali87:
Fixes #1493

## The report

```
Context token count failed: 401:
{"type":"error","error":{"type":"authentication_error",
 "message":"invalid x-api-key"}}
```

**The count was not broken — the message was.** An invalid API key is reported as an internal token-count failure, so the actual problem is not what the user is told.

## Two defects behind one symptom

`anthropic_token_counter.py` raised `TokenCountError(f"{status}: {text}")` for **every** status ≥ 400, and `main.py` caught it with a blanket `except Exception` at **debug** level. That collapses two cases needing opposite responses:

| status | meaning | user action |
|---|---|---|
| 401 / 403 | the credential is rejected | change the key — no retry ever helps |
| 429 / 5xx | transient | none; the next refresh probably works |

And debug is the wrong level for the permanent case: a rejected key recurs on **every** background refresh and silently disables the token counter for the whole session, so the user never learns why the count stopped.

## The fix

`TokenCountAuthError` subclasses `TokenCountError` deliberately — an existing `except TokenCountError` caller keeps working rather than having this escape as an unhandled exception in a background task.

The caller **warns once per session**, not per refresh. The counter runs on every turn, so repeating an unactionable-until-restart message would bury the rest of the log. Transient failures stay at debug.

The message leads with the actionable part and keeps the status for diagnosis, rather than echoing the provider's raw body.

## Verification

Five mutations, each caught by exactly one test:

| mutation | result |
|---|---|
| remove the auth branch | 3 failed |
| treat **all** 4xx as auth errors | 3 failed |
| drop the status from the message | 1 failed |
| warn every time (no once-guard) | 1 failed |
| let an auth failure fall through to debug | 1 failed |

The two **controls** carry the most weight. Treating all 4xx as auth errors fails the transient test, and warning every time fails the once-per-session test — without them this fix would invert the defect and nag about credentials during a rate limit.

Scoped regression: **83 passed, 4 skipped**.

## Two fixture defects found on the way

Both read as code failures, and I record them because each nearly sent me to fix working code:

- **`messages=[]` returns 0 before any HTTP call**, so the fixture never reached the code under test — an unreachable fixture. The failure said `DID NOT RAISE`, which looks like a missing raise rather than a test that never got there.
- **Patching `httpx.AsyncClient` with a lambda that calls `httpx.AsyncClient`** recursed infinitely. The resulting `RecursionError` looks like a product bug; the helper now captures the real class first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of API authentication failures during context token counting.
  * Added a clear warning when credentials are rejected, shown at most once per session.
  * Preserved quieter handling for transient service or rate-limit failures.
  * Error messages now identify authentication failures and include the relevant response status.

### DIFF
--- a/codebase_rag/constants/providers.py
+++ b/codebase_rag/constants/providers.py
@@ -87,6 +87,10 @@ FIELD_ENDPOINT = "endpoint"
 ANTHROPIC_COUNT_TOKENS_URL = "https://api.anthropic.com/v1/messages/count_tokens"
 ANTHROPIC_API_VERSION = "2023-06-01"
 ANTHROPIC_HEADER_API_KEY = "x-api-key"
+
+# HTTP statuses meaning the credential was REJECTED. Retrying cannot help,
+# so these are reported differently from transient failures (issue #1493).
+HTTP_AUTH_FAILURE_STATUSES: frozenset[int] = frozenset({401, 403})
 ANTHROPIC_HEADER_VERSION = "anthropic-version"
 HEADER_CONTENT_TYPE = "content-type"
 CONTENT_TYPE_JSON = "application/json"
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -147,6 +147,11 @@ EMBEDDING_SNIPPET_FAILED = (
 )
 EMBEDDING_BATCH_COMPUTE_FAILED = "Failed to embed batch of {count}: {error}"
 CONTEXT_TOKEN_COUNT_FAILED = "Context token count failed: {error}"
+# A rejected credential, distinct from a transient failure: the user must
+# change something, so it is warned rather than logged at debug (issue #1493).
+CONTEXT_TOKEN_COUNT_AUTH_FAILED = (
+    "Context token counting disabled for this session: {error}"
+)
 NO_SOURCE_FOR = "No source code found for {name}"
 EMBEDDINGS_COMPLETE = "Successfully generated {count} semantic embeddings"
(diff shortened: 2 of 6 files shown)


vitali87:

vitali87:
@greptileai review 629b098b

vitali87:
@greptileai review 407fd8e49ea780bc59bc7ac19dd263c0a465854f

Fixed four SonarCloud S5778 findings before the gate ran, caught by querying the issues API at the head rather than waiting for the check to go red.

All four were the same shape in my test file: `_messages()` was invoked **inside** the `pytest.raises` block, so a failure in the helper would have satisfied the expectation and the test would pass for the wrong reason — an assertion satisfied by something other than the behaviour it names. Hoisted to a local before each block.

Behaviour unchanged, and I re-verified the mutations still discriminate after the refactor:

| mutation | result |
|---|---|
| remove the auth branch | 3 failed |
| treat all 4xx as auth errors | 3 failed |

9 tests. Your 5/5 was anchored to `629b098b`; this is the same content with the test-structure fix.

PR #1497, opened 2026-08-27
test: guard node properties against the protobuf schema

### LINKED ISSUE #1490: Node properties in NODE_SCHEMAS can be absent from schema.proto with nothing checking

vitali87:
Fixes #1490

Replaces #1491, which was branched from the front-matter work by mistake and carried seven unrelated commits. Same guard, built on `main`, one file.

## The gap

`NODE_SCHEMAS` and `codec/schema.proto` declare node properties independently, and nothing checked they agree. A property added to the schema is written at ingestion, read by queries, and **silently dropped on export** — no error, no warning.

The sibling `test_protobuf_relationship_parity.py` guards the relationship *enum* for exactly this failure mode. The node side had no equivalent, and dozens of properties across 14 labels are currently absent.

## Default-deny

An inventory of "properties to check" cannot fail for the property nobody listed — the inversion the MCP lock guard needed in #1475. Every declared property must have a proto field **or** be named in `_NOT_EXPORTED`, which records the current state so the set is closed: a new absence fails and must be argued for.

Measured against the generated **descriptors**, not the `.proto` text — a text search answers "does this name appear in the file", which a comment satisfies.

## The controls caught their own author

Rebuilding on clean `main` made the existence control fail at once: `_NOT_EXPORTED` named `Module.front_matter`, which is declared only on the #1488 branch. Removed.

That is the behaviour the control was written for. Without it the guard would have shipped an exemption for a property that does not exist here, silently exempting whatever later reused the name.

## Verification

| mutation | result |
|---|---|
| drop an entry from the allow-list | fails the default-deny test |
| add an already-exported property | fails the staleness control |
| add a nonexistent property | fails the existence control |
| skip a label with no proto message | fails the missing-message test |

Each fails exactly one test. Scope is the guard only — adding proto fields for existing absences needs a per-label decision I would be guessing at.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that declared node properties are either represented in the protocol schema or explicitly exempted.
  * Added checks for missing protocol messages, hyphenated property names, and invalid exemption entries.
  * Improved detection and reporting of properties that are not exported.

### DIFF
--- a/codebase_rag/tests/test_protobuf_node_property_parity.py
+++ b/codebase_rag/tests/test_protobuf_node_property_parity.py
@@ -18,10 +18,10 @@
 # field or be named in `_NOT_EXPORTED` below, so forgetting fails.
 from __future__ import annotations
 
-import re
 from collections.abc import Iterable
 from typing import Any
 
+from codebase_rag.schema_parse import parse_properties
 from codebase_rag.types_defs import NODE_SCHEMAS
 from codec import schema_pb2 as pb
 
@@ -119,12 +119,20 @@ _NOT_EXPORTED: dict[str, frozenset[str]] = {
     "ModuleImplementation": frozenset({"absolute_path", "module_type"}),
 }
 
-# `{name: type, other: type?}` -> {"name", "other"}
-_PROPERTY_NAME = re.compile(r"(\w+):")
-
-
 def _declared_properties(schema_text: str) -> set[str]:
-    return set(_PROPERTY_NAME.findall(schema_text))
+    """Property names, read with the SCHEMA'S OWN parser.
+
+    An earlier version used `re.findall(r"(\\w+):")`, which silently mangles
+    any name containing a non-word character: `my-prop` matched as `prop`, so
+    the guard checked a name that does not exist while the real one went
+    unexamined -- the guard failing OPEN, which is the failure this file
+    exists to prevent (reported on #1497).
+
+    `parse_properties` is the parser `schema_parse` already provides for these
+    declarations. Writing a second, looser grammar for the same strings was
+    the mistake: two parsers for one format disagree exactly where it matters.
+    """
+    return {spec.name for spec in parse_properties(schema_text)}
 
 
 def _proto_fields(label: str) -> set[str] | None:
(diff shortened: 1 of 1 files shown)


vitali87:

vitali87:
@greptileai review 5d367f0e

vitali87:
@greptileai review 6e4d3dcdb7e4130461fcfddde1ec17d210459ae3

Confirmed and fixed — and this one is the guard failing in exactly the way it exists to prevent.

Reproduced first:

```
re.findall(r'(\w+):', '{qualified_name: string, my-prop: string}')
  ->  ['prop', 'qualified_name']
```

`my-prop` was read as `prop`, so the guard checked a name that does not exist while the real property went **unexamined**. A default-deny check that silently skips what it cannot parse fails open, which is the failure this file was written to stop.

**The fix is to stop having two parsers.** `schema_parse.parse_properties` already reads these declarations — it is the codebase's own grammar for the format, and it handles `my-prop` correctly. Writing a second, looser regex for the same strings was the mistake: two parsers for one format disagree precisely where it matters, and mine was the one that fell back to something plausible rather than failing.

Added a test that calls the real helper **and** the real comparison, so a change back to a hand-rolled grammar fails here rather than passing against a parser that disagrees with the schema's own.

RED verified: reverting to the regex fails two tests — the new one and the pre-existing existence control.

5 tests.

PR #1501, opened 2026-08-27
fix: reject unknown model ids at startup instead of at first query

### LINKED ISSUE #1492: cgr start must check for validity of supplied models for orch and cypher before launch
pydantic_ai.exceptions.ModelHTTPError: status_code: 404, model_name: opus-5, body: {'type': 'error',
'error': {'type': 'not_found_error', 'message': 'model: opus-5'}, 'request_id':
'req_011CeTEwfVnFZDMiJ3PLNi3y'}
An unexpected error occurred: status_code: 404, model_name: opus-5, body: {'type': 'error', 'error':
{'type': 'not_found_error', 'message': 'model: opus-5'}, 'request_id': 'req_011CeTEwfVnFZDMiJ3PLNi3y'}

vitali87:
Fixes #1492

## The report

```
pydantic_ai.exceptions.ModelHTTPError: status_code: 404, model_name: opus-5,
{'type': 'not_found_error', 'message': 'model: opus-5'}
```

`opus-5` is not a model id. The real one is `claude-opus-5`.

## Why startup did not catch it

`_validate_provider_config` already ran for both roles at startup, so the seam existed. But it calls `provider.validate_config()`, which **takes no model id** — it checks credentials. It could not have caught this. The user paid for a full startup and got a raw provider traceback at the first query.

Now:

```
Orchestrator configuration error: Unknown anthropic model 'opus-5'.
Did you mean 'claude-opus-5'?
```

## The constraint that shaped the design

The obvious implementation — reject anything not in pydantic-ai's catalogue — breaks the project's default configuration.

| provider | catalogued ids |
|---|---|
| anthropic | 17 |
| openai | 73 |
| google | 20 |
| **ollama** | **0** |
| **azure** | **0** |
| **litellm_proxy** | **0** |
| **minimax** | **0** |

Ollama is the default provider, and its model space is whatever the user has pulled locally. Azure deployment names, LiteLLM routes and MiniMax ids are equally user-defined. A strict allow-list turns a rare typo into a **universal outage**.

So a provider with no catalogue is not gated, and the enforced set is derived from the catalogue itself rather than hand-maintained — a provider added later is not gated until pydantic-ai knows its models.

I verified every model id in this repo's own documentation still passes. The MiniMax examples survive **only** through that exemption.

## A false rejection found before review

`provider=openai` pointed at vLLM or an OpenAI-compatible proxy is a documented setup, and the catalogue only describes what OpenAI's *own* endpoint serves:

```
provider=openai, model_id=Qwen2.5-Coder-32B-Instruct,
endpoint=http://localhost:8000/v1   ->  REJECTED
```

Validation is now suppressed when the endpoint differs from the provider's default, normalised for a trailing slash so a stray `/` cannot silently disable the check.

## Two more from an adversarial review

**Case failed open**, which is the worse direction. An unmatched prefix yields an empty set, which the caller reads as "no catalogue" and skips entirely — so `provider="Anthropic"` did not mis-key the lookup, it switched the whole check off. `"Anthropic"` + `"opus-5"` passed.

**One vendor can span several prefixes.** `openai-chat` carries five ids absent from `openai` (the search-preview variants and `gpt-3.5-turbo-16k`), so consulting a single prefix rejected real models. Measured, not assumed.

## Verification

Seven mutations, each caught:

| mutation | result |
|---|---|
| remove the startup call | 1 failed |
| gate every provider | 3 failed |
| accept everything | 3 failed |
| drop the `.lower()` | 2 failed |
| consult only the first prefix | 3 failed |
| suppress for all endpoints | 8 failed |
| drop the suggestion | 1 failed |

The last one **initially survived**. `test_the_error_names_the_real_model_id` asserted `"claude-opus-5" in message`, and the fallback branch lists the entire catalogue — which contains that id — so the assertion was true of the working and the degraded message alike. It now asserts the suggestion form and counts candidates.

Scoped regression: **280 passed**. Cost is 0.03 ms per call, twice per startup.

## Also

The no-close-match message listed all 78 openai ids at 6,368 characters, burying the error it was attached to. Truncated to 20, now 429.

## Summary by CodeRabbit

* **New Features**
  * Added provider-specific model ID validation during application startup.
  * Invalid model names now provide close-match suggestions or available model listings.
  * Supports provider-prefixed configurations and case-insensitive provider names.
  * Custom endpoints, Ollama, and Vertex Model Garden IDs remain supported without catalogue checks.

* **Bug Fixes**
  * Prevents mistyped model IDs from failing during the first query.

* **Tests**
  * Added coverage for valid, invalid, custom-endpoint, and provider-specific configurations.

### DIFF
--- a/codebase_rag/constants/core.py
+++ b/codebase_rag/constants/core.py
@@ -18,6 +18,9 @@ ARG_DRY_RUN = "dry_run"
 
 SEPARATOR_DOT = "."
 SEPARATOR_SLASH = "/"
+# Splits "provider:model" both in user-supplied settings and in the model
+# names pydantic-ai enumerates.
+MODEL_STRING_SEPARATOR = ":"
 # Disambiguates definitions that share one qualified name (if/else import
 # fallbacks, typing.overload, try/except fallbacks): "<qn>@<start_line>".
 DUP_QN_MARKER = "@"
--- a/codebase_rag/constants/providers.py
+++ b/codebase_rag/constants/providers.py
@@ -46,6 +46,16 @@ OLLAMA_HEALTH_PATH = "/api/tags"
 GOOGLE_CLOUD_SCOPE = "https://www.googleapis.com/auth/cloud-platform"
 V1_PATH = "/v1"
 
+# The endpoint each provider serves its own published catalogue from. A
+# config pointing somewhere else (vLLM, an OpenAI-compatible proxy) serves
+# whatever that host chooses, so model-id validation does not apply to it.
+# Providers absent here have no single canonical endpoint and are never
+# gated on model id.
+PROVIDER_DEFAULT_ENDPOINTS: dict[str, str] = {
+    Provider.OPENAI: OPENAI_DEFAULT_ENDPOINT,
+    Provider.MINIMAX: MINIMAX_DEFAULT_ENDPOINT,
+}
+
 HTTP_OK = 200
 
 UNIXCODER_MODEL = "microsoft/unixcoder-base"
--- a/codebase_rag/constants/providers.py
+++ b/codebase_rag/constants/providers.py
@@ -56,6 +56,18 @@ PROVIDER_DEFAULT_ENDPOINTS: dict[str, str] = {
     Provider.MINIMAX: MINIMAX_DEFAULT_ENDPOINT,
 }
 
+# Catalogue prefixes to consult per provider, where pydantic-ai splits one
+# vendor across several. `openai-chat` carries five ids absent from
+# `openai` (the search-preview variants and gpt-3.5-turbo-16k), so
+# consulting only `openai:` would reject real models.
+PROVIDER_CATALOGUE_PREFIXES: dict[str, tuple[str, ...]] = {
+    Provider.OPENAI: ("openai", "openai-chat"),
+    Provider.GOOGLE: ("google", "google-cloud"),
+}
+
+# How many model ids to list when nothing resembles what the user typed.
+MODEL_ID_SUGGESTION_LIMIT = 20
+
 HTTP_OK = 200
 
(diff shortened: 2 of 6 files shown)


vitali87:
claimed by feat/duplicates-clickable-locations

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Good catch, fixed in 92e71a3a. Confirmed the rejection before changing anything:

\`\`\`
provider=google, provider_type=vertex,
model_id=meta/llama-3.3-70b-instruct-maas   ->  REJECTED
\`\`\`

This is the same open-model-space argument as a custom endpoint, reached through a different door: Model Garden models are chosen by the deployment rather than published by the vendor, so they will never appear under \`google:\`.

Scoped to \`provider_type=vertex\` rather than exempting Google wholesale, since GLA serves Google's own published models and its catalogue is authoritative there. Exempting the whole provider would switch validation off for the common case in order to serve the rare one. A control test pins that distinction, and it fails against the wholesale exemption.

Also covered \`publishers/anthropic/models/claude-sonnet-4-5\`, the other Model Garden id shape.

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Sonar S5778 cleared in 99eebb73 — five occurrences, and the rule was right on the substance.

Each flagged block built the `ModelConfig` inside the `with pytest.raises(ValueError)`, so a failure in the CONSTRUCTOR would have satisfied the assertion without `validate_model_id` ever running. That is an assertion satisfied by both the working code and a broken one, which is the defect class I have been auditing elsewhere in this branch.

Every block now constructs the config first and calls only the function under test inside, so the raise can come from exactly one place. 26 tests still pass.

PR #1502, opened 2026-08-27
fix: up arrow recalls the previous message in interactive chat

### LINKED ISSUE #1495: up arrow does not paste the previous message in interactive mode chat

vitali87:
Fixes #1495

> up arrow does not paste the previous message in interactive mode chat

The issue body was empty, but the title is a complete report and reproduces.

## Two independent causes

Either one alone breaks recall, so fixing only one would have looked like a fix and changed nothing.

**1. The session was thrown away every turn.** `get_multiline_input` called the bare `prompt()` function. From prompt_toolkit's own source:

```python
def prompt(message=None, *, history: History | None = None, ...):
    """The global `prompt` function. This will create a new `PromptSession`"""
    session: PromptSession[str] = PromptSession(history=history)
```

A new session per call, with `history=None`. Every turn started empty, so there was nothing to recall whichever key was pressed.

**2. The arrow keys were unbound.** This prompt is genuinely multiline — Enter inserts a newline, Ctrl+J and Ctrl+E submit — so up moved the cursor within the buffer.

## The fix

The session is built once and reused, submitted text is appended to it, and up/down are bound to history **only at the buffer edges**:

| cursor | up arrow does |
|---|---|
| row 0 (single or multiline) | recall history |
| any row below | move the cursor |

Binding them unconditionally would make a half-written multiline question uneditable — trading the reported bug for a worse one.

Blank entries and immediate repeats are not remembered. Blanks would push the real previous message out of reach for anyone holding Ctrl+J on an empty buffer.

## Two test defects caught by mutation

Both would have shipped a green suite that proved nothing:

**The multiline tests never called the code.** They drove a bare `Buffer` and asserted on `cursor_position_row`, so they passed against an implementation that *always* navigates history — precisely the regression they existed to prevent. Rewritten to invoke the real handler and record which action it took.

**`assert session.history is not None` was true either way.** prompt_toolkit supplies an `InMemoryHistory` by default, so that assertion held for a correct session and for one built with no history argument. Replaced with a write on one lookup and a read on a second, which is what actually distinguishes a persistent history from a per-turn one.

## Verification

| mutation | result |
|---|---|
| drop the `lru_cache` (fresh session per turn) | 10 errors |
| return a new session each call | 2 failed |
| never append to history | 4 failed |
| remember blanks too | 1 failed |
| remove the up binding | 1 failed |
| up always navigates history | 3 failed |
| up always moves the cursor (the reported bug) | 2 failed |

The last two are the pair that matters: both directions of the arrow logic are pinned, so neither the bug nor its over-correction can return silently.

Scoped regression: **405 passed**. One unrelated `test_cli_smoke.py::test_version_flag` subprocess timeout under `-n auto` contention; it passes in isolation in 11s and this branch does not touch that path.

## Summary by CodeRabbit

* **New Features**
  * Added persistent history for multiline chat input across turns.
  * Up and down arrows now navigate previous entries at buffer boundaries.
  * Blank entries are ignored, and consecutive duplicate questions are not added to history.

* **Bug Fixes**
  * Prompts now correctly connect to the active console during each interaction.
  * Preserved normal cursor movement within multiline input.

### DIFF
--- a/codebase_rag/constants/cli.py
+++ b/codebase_rag/constants/cli.py
@@ -18,6 +18,8 @@ class KeyBinding(StrEnum):
     ENTER = "enter"
     CTRL_C = "c-c"
     SHIFT_TAB = "s-tab"
+    UP = "up"
+    DOWN = "down"
 
 
 class PermissionMode(StrEnum):
--- a/codebase_rag/main.py
+++ b/codebase_rag/main.py
@@ -1248,14 +1248,23 @@ def _input_history() -> History:
     return InMemoryHistory()
 
 
-@lru_cache(maxsize=1)
 def _input_session() -> PromptSession[str]:
-    """The chat prompt, built once so its history survives the turn.
+    """The chat prompt, built fresh each turn over the persistent history.
 
     `get_multiline_input` previously called the bare `prompt()` function,
     whose own docstring says it "will create a new PromptSession" and which
     passes `history=None`. Every turn therefore started with an empty
-    history and the up arrow had nothing to recall (issue #1495).
+    history and the up arrow had nothing to recall (issue #1495). Passing
+    `_input_history()` is what fixes that; the session itself is deliberately
+    NOT cached.
+
+    `PromptSession` binds its `Application` to the ambient app session's
+    input and output at construction, so a cached one keeps reading the
+    console it was born under. Across two app sessions that means the
+    second prompt reads the first one's input; when that input is a closed
+    pipe, POSIX returns EOF but a Win32 pipe blocks, which timed out the
+    Windows CI job at thirty minutes. Rebuilding per turn is what the bare
+    `prompt()` did all along and costs nothing at human typing speed.
     """
     return PromptSession(history=_input_history())
 
--- a/codebase_rag/main.py
+++ b/codebase_rag/main.py
@@ -17,13 +17,15 @@ from collections.abc import Callable, Coroutine
 from contextlib import contextmanager
 from dataclasses import replace
 from decimal import Decimal
+from functools import lru_cache
 from html import escape as html_escape
 from pathlib import Path
 from typing import TYPE_CHECKING
 
 from loguru import logger
-from prompt_toolkit import PromptSession, prompt
+from prompt_toolkit import PromptSession
 from prompt_toolkit.formatted_text import HTML
+from prompt_toolkit.history import History, InMemoryHistory
 from prompt_toolkit.key_binding import KeyBindings
 from prompt_toolkit.shortcuts import print_formatted_text
 from pydantic_ai import (
@@ -1186,7 +1188,7 @@ def _thinking_with_status_bar(message: str):
             refresh_task.cancel()
 
 
-def get_multiline_input(prompt_text: str = cs.PROMPT_ASK_QUESTION) -> str:
+def _input_keybindings() -> KeyBindings:
     bindings = KeyBindings()
 
     @bindings.add(cs.KeyBinding.CTRL_J)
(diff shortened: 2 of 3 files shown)


vitali87:
claimed by feat/duplicates-clickable-locations

vitali87:
@greptileai review

vitali87:
The Windows failures were not test failures — the jobs were **cancelled**, and `gh pr checks` renders cancelled identically to fail. Both runs stopped at 30:19, which is a job timeout.

Diagnosis by comparison rather than guesswork: #1501's Windows jobs **passed** on the same runners in the same window, so it was specific to this branch. The only Windows-relevant thing this change adds is that three tests called `_input_session()`, which constructs a real `PromptSession` and attaches to the console. On a headless runner that blocks until the job times out.

Fixed in 96e55a28 by giving the history its own seam. `_input_history()` owns the `InMemoryHistory`; `_input_session()` reads it. Persistence across turns — the property #1495 is actually about — is now exercised with no prompt involved.

Splitting them creates a new way to be wrong: the tests could pass while the session reads some *other* history. So one test asserts the wiring (`_input_session().history is _input_history()`) and is the only one here that builds a real prompt. Confirmed by mutation — pointing the session at a fresh `InMemoryHistory` fails exactly that test and nothing else.

Stating the limit plainly: I cannot run Windows locally, so this is inference from the evidence available (same branch cancels twice at the same duration, sibling branch passes, console attachment is the only Windows-specific addition). If it cancels again the hypothesis is wrong, and the next step is bisecting which test hangs rather than re-running.

Rebased onto main, which now carries #1501. 13 tests pass locally.

vitali87:
@greptileai review

vitali87:
Tested this against the actual code path and it does not reproduce — one entry, not two:

```
returned: 'question'
history : ['question']
entries : 1
```

driving the real `get_multiline_input` with `prompt` returning `"  question  "`.

The reasoning is right for an ordinary prompt and wrong for this one. `Buffer.append_to_history` runs when a buffer is **accepted**; this prompt never accepts one. Enter is bound to insert a newline, and Ctrl+J calls `app.exit(result=...)`, which bypasses buffer acceptance entirely. So the session does not append, and `_remember_input` is the only writer.

I have pinned it rather than just rebutting, in eaa67157, because the property belongs to **these key bindings** rather than to prompt_toolkit — it would quietly stop holding if submission ever moved to a normal accept, and the symptom would be exactly the double entry described.

Two tests, not one: counting alone cannot distinguish "recorded once, stripped" from "recorded once, raw", so the second asserts the stored form. Verified by mutation — adding a second append reproduces the reported duplicate and fails both.

Also worth recording that `_remember_input` is not redundant with prompt_toolkit's own logic even though it looks it. `append_to_history` skips blanks and immediate repeats too, but it is never reached here, so removing the helper would lose both behaviours rather than defer them.

vitali87:
@greptileai review

vitali87:
My first hypothesis was **wrong**, and I said on this PR what I would do if it was: bisect rather than re-run.

The jobs cancelled a third time, at 30:22 and 30:23 — the same duration as the previous two. Moving the history into its own seam did not change it.

Reading the file rather than the diff showed why: **three tests still built a real `PromptSession`.** Two patched `type(_input_session())` to stub `prompt`, which constructs one just to reach its type, and the wiring test constructed one deliberately. I had reduced the count from three to three and reported it as a fix.

Now zero, in 36fcf812:

- the two submission tests use a stub carrying only what `get_multiline_input` touches
- the wiring test reads the **source** of `_input_session` rather than calling it, proving the same property — that it passes `_input_history()` and not a fresh history — with nothing to attach to

Both mutations still fail exactly one test each, so the source-inspection form discriminates as well as the runtime one did.

**Stating the uncertainty rather than implying confidence.** I cannot run Windows. The console hypothesis is now *unproven*, not confirmed — it survived one falsification only because my previous change did not actually test it. What justifies this change independently is that a unit test of history behaviour has no business attaching to a terminal on any platform.

If the jobs cancel a fourth time, the cause is elsewhere and the next step is deleting the test file outright to confirm the branch is even responsible — the peer's #1443 passes Windows on the same runners, so it is branch-specific, but I have not proven it is this file.

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Root-caused the Windows timeout (both jobs cancelled at the 30-minute mark, at 99%, with no failing assertion) and fixed it in 2a98692b.

**Cause.** `PromptSession` binds its `Application` to the ambient app session's input and output at *construction* time (`Application.__init__`: `self.input = input or get_app_session().input`). Caching the session with `lru_cache` therefore pinned the prompt to whichever console existed on the first call. The pre-existing `test_multiline_input_keybindings.py` drives `get_multiline_input` inside `create_app_session(input=create_pipe_input(), ...)`, once per test, so from its second test onward the prompt was reading the *first* test's pipe, which had already closed. POSIX returns EOF there, which is why Linux and macOS stayed green; a Win32 pipe blocks instead, and the job sat until it timed out.

This also explains why the earlier attempts did not help: eaa67157 and 36fcf812 only stopped the *new* tests from constructing a session, but the hang came from the product change reaching the old test file.

**Fix.** `_input_session` is no longer cached; it is rebuilt per turn over the still-cached `_input_history()`. That is exactly what the bare `prompt()` did before this PR (so it restores the Windows behaviour main already has) while keeping the persistent history that fixes #1495. Rebuilding costs nothing at human typing speed.

**Tests.** Added `TestThePromptFollowsTheCurrentConsole`: the binding test asserts the prompt attaches to the current pipe across two app sessions, and fails RED against the cached version with `[True, False] == [True, True]`; the control asserts history still survives the rebuilt session. Updated the autouse fixture and the wiring test, which referenced `cache_clear`/`__wrapped__` on the now-uncached function.

Verified locally: both input test files pass (25), and the 356-test input/prompt/cli subset is green apart from a known `test_cli_smoke` subprocess timing flake that passes in isolation.

vitali87:
@greptile-apps please review — your only review on this PR is anchored to 96e55a28 from yesterday, so the 5/5 currently showing predates the fix entirely and the gate has been waiting on a review that never arrived.

Worth recording why: the two earlier re-invocations on this PR were addressed to `@greptileai`, which is a different, unrelated GitHub account, not this reviewer. The correct mention is the bot's login minus the `[bot]` suffix — `@greptile-apps`. Anything sent to the other handle notifies a stranger and never reaches the reviewer, which is why this PR sat with a stale review and a pending gate while looking, from the checks column, like it was simply waiting.

Since that review: 2a98692b removes the `lru_cache` from `_input_session` so the prompt is rebuilt per turn over the still-cached `_input_history()`. Both Windows jobs now pass (py3.13 16m30s, py3.12 14m8s) where both previously hung to a 30-minute cancellation.

vitali87:
Correcting my previous comment: I was wrong about the handle.

`gh api users/greptileai` returns `type=Organization name=Greptile` — it is Greptile's own organization account, not an unrelated third party, and `greptile-apps` is simply the bot user that posts. A peer session produced a direct counterexample on PR #1506: an `@greptileai` invocation at 00:26:28Z drew a review at 00:33:33Z, seven minutes later. So mentions to that handle do reach the reviewer, and my claim that the earlier re-invocations here "never reached" anything is not established.

What I actually had was a correlation — 14 invocations on #1503, 4 on #1485, 2 here, no review appearing — and I wrote it up as a mechanism without checking the account. The competing explanation, which fits the same data and is still live, is that a re-invocation is a no-op when a review already exists for that state: on #1487 a 5/5 landed on push and five subsequent mentions, including a correctly-addressed one, triggered nothing.

The part of that comment that stands, and the reason this PR is not merged yet: the only Greptile review on this PR is anchored to 96e55a28 from yesterday, so the 5/5 currently displayed predates 2a98692b and has never seen the fix under test. That is a property of the review record, independent of any handle theory, and it is the thing blocking the gate.

PR #1480, opened 2026-08-26
refactor: remove duplicated code found by cgr duplicates

vitali87:
`cgr duplicates` on this repo reported **123 clone groups** (93 exact-fingerprint, 30 near-duplicate ≥0.8). This collapses the groups where the duplication was real, and leaves alone the ones that only look alike structurally (per-language type-inference bodies, the lazy-property block, 2-member 0.80 pairs).

Net **−200 lines** across 52 files, with 4 new shared modules.

## Production code

| Group | Change |
|---|---|
| `vector_store` | Both backends' `delete_project_embeddings` and `store_embedding_batch` shared a guard/log/swallow wrapper. Extracted; only the client call stays per backend. |
| `language_spec` | Four `*_file_to_module` differed only in which package sentinel or source root to strip → one `_file_to_module` + `partial` bindings. |
| `_dart_get_name` | Existed **twice** with the subtle constructor-naming rule restated in both. `language_spec` now delegates to the single impl in `parsers/dart/utils.py`, via the deferred import it already uses for the JS/Rust helpers. |
| `java/variable_analyzer` | Three identical recursive walkers differing only in (node type, processor) → one `_traverse_for`. |
| `parsers/utils` | Four `*_invoked_parameter_names` differed only in scan config + scope rule; moved to the dispatch site. |
| frontends | go/csharp/java `_build_tool`'s freshness→lock→compile dance and the javac `-version` probe → `build_cached_artifact` / `toolchain_runs` in `build_lock`. The four `resolve_*_frontend` → `resolve_frontend_mode` in a new leaf module (`parsers.frontends` can't be imported from a frontend without closing a cycle). |
| `providers/base` | OpenAI/Anthropic/Azure/MiniMax restated the same `api_key` check → new `ApiKeyProvider` base that names its error. `validate_config` **stays abstract** on `ModelProvider`. |
| `protobuf_service` | `_flush_joint`/`_flush_split` shared the write-then-unlink-the-other-layout logic → `_write_layout`. A new layout would previously have needed a third copy that remembered to delete both others. |
| `mcp/tools` | `index_repository`/`update_repository` shared the lock/thread/error protocol. |

## evals and benchmarks

| Group | Change |
|---|---|
| `benchmarks/harness.py` | `run_benchmark` + `print_results` were copied into **all ten** bench scripts (the latter differing only in column width). |
[...]

PR #1510, opened 2026-08-29
ci: cap the Greptile gate poll at 20 minutes so it stops gating other jobs' logs

### LINKED ISSUE #1507: A polling Greptile gate withholds the failure logs of every other job in the run
## What happens

`gh run view --job <id> --log-failed` returns

```
run <id> is still in progress; logs will be available when it is complete
```

for **every** job in a run until the **whole run** completes. `Greptile 5/5 Gate` polls an external service for a review of the head commit, and when that review never starts, the gate stays `in_progress` until it ages out at roughly one hour.

The consequence is not the gate's own red. It is that **a review bot can withhold the failure logs of every unrelated job in the run**.

## Measured on PR #1480, head `df56b363`

```
19:54:36Z   completed = 26     pending = 2
            PENDING  Greptile 5/5 Gate   started 19:16:31Z   (38 min, output.summary = NULL)
            PENDING  Greptile Review     started 19:52:02Z
```

At that moment **26 of 28 checks were finished**, including nine failures across all seven unit-test platforms and both Sonar checks. Those nine had been complete for over an hour. None of their logs were retrievable.

There was **no `Greptile Review` check run on the head at all** until it was requested manually; it then spawned within seconds, which is what distinguishes "never started" from "ran and failed quietly".

## Why this is worth fixing rather than working around

The nine failing jobs are the thing a contributor needs to act on. Their diagnosis was blocked for over an hour by a job that has nothing to do with them, and the blocking is invisible: the symptom is a progress message that reads like ordinary slowness. A `grep` over that output returns empty, which reads like "the log is empty" rather than "the log is unavailable".

This also interacts badly with the gate's two terminal states, which need different remedies:

| gate state | remedy |
|---|---|
| `in_progress` (still polling) | request the review; the live poller picks it up |
| `completed/failure` (aged out) | request the review **and** `gh run rerun --failed`; a terminated job never re-evaluates |

Getting that wrong either restarts a job about to succeed, or leaves a mergeable PR permanently red.

## Suggested directions

Either would remove the coupling; the first is smaller:

1. **Cap the gate well under the log-availability horizon.** A gate that cannot succeed after N minutes of a missing review should fail fast rather than hold the run open. It fails closed either way, so nothing merges unreviewed.
2. **Fail the gate immediately when no `Greptile Review` check run exists on the head**, rather than polling for one that was never created. Absence at start is distinguishable from pending.

A third, if the gate must keep polling: have it emit the reason into `output.summary` rather than leaving it `NULL`, so the state is diagnosable without correlating job lists by hand.

## Other instances

Three across two PRs and two sessions today:

- **#1480** (this report): gate `in_progress` 38+ min, no review run existed, 26/28 complete, logs gated
- **#1499**: gate `completed/failure` at 1h0m23s with `output.summary` null, no review run on head; needed both a review request and `gh run rerun --failed`
- **#1503**: same shape caught early — gate polling, no review run on head, review requested at ~10 min rather than discovered at 60

## Not the ask

This is not "make Greptile faster" or "loosen the 5/5 requirement". The gate doing its job is fine. The issue is that log availability is coupled to run completion, so any long-polling required check becomes a diagnosis blocker for everything beside it.

Reported by session `fix-doc-qualified-name`; instances on #1499 and #1503 observed by session `cgr-other`.

vitali87:
Closes #1507.

## The problem

GitHub refuses the logs of **every** job in a run until the **whole run** completes. `Greptile 5/5 Gate` polled for up to 60 minutes waiting for a review of the head commit, so a job waiting on an external service was withholding the failure logs of the unit-test matrix beside it.

Measured on #1480: 26 of 28 checks finished, including nine failures across all seven unit-test platforms, and none of their logs were retrievable for over an hour.

Reproduced fresh while writing this, on an unrelated run:

```
run 33225205863 — 9 of 16 jobs completed
gh run view --job 99027370629 --log
  -> run 33225205863 is still in progress; logs will be available when it is complete
```

A finished job's log, refused because a sibling was still polling.

## The change

Poll deadline 3600s -> 1200s, and `timeout-minutes` 65 -> 25. The failure message now names the remedy, including the correct bot handle and the rerun command.

## Why 20 minutes costs no legitimate pass

Measured over **48 gate runs**:

| outcome | durations (minutes) |
|---|---|
| success | 0 0 0 0 0 0 2 3 3 4 4 9 9 11 · 28 46 54 |
| failure | 0 0 7 9 10 12 12 18 · 60 |

Every success that did not need human intervention finished within **11 minutes**; 8 of 9 failures within 18.

The three successes past 20 minutes each had a review **requested mid-poll**, so none would have passed unaided:

| head | gate window | review requested |
|---|---|---|
| `2a98692b` (#1502) | 00:26–00:54 | 00:46, mid-poll |
| `df56b363` (#1480) | 19:16–20:02 | mid-poll |
| `0f4d03ff` (#1480) | 00:05–00:59 | mid-poll |

The remedy after a fast fail is the same one those three used during a long poll: request the review, then re-run the job. The only difference is that everyone else's logs are readable while you do it.

**This fails closed.** Nothing merges without a 5/5 review of head. The change moves the waiting out of a job that blocks other jobs' diagnosis.

## Withdrawn: the issue's suggestion 2

#1507 also proposed failing immediately when no `Greptile Review` check run exists on the head, citing `df56b363` as having none. That evidence was wrong and I have [corrected it on the issue](https://github.com/vitali87/code-graph-rag/issues/1507#issuecomment-5459376472). Both reviewed heads now carry a completed `Greptile Review`, each started ~41 minutes post-commit and only after a manual request; `ba9f4e30` has none at all because none was ever requested. Absence is the normal state of every head until someone asks, so a gate failing on it would fail every PR at the moment CI starts.

## The Sonar gate is deliberately untouched

`Sonar Zero Issues Gate` has the same polling shape, but different measured behaviour: successes run 14–19 minutes, clustered at the top of that range. A 20-minute cap would sit on top of its normal completion time and fail legitimate passes. Its 40-minute deadline is proportionate to a service that does respond unaided, which Greptile does not.

## Verification

- YAML parses; all 12 jobs intact
- Gate script passes `bash -n`
- Failure branch executed directly: `$HEAD_SHA` expands, `$GITHUB_RUN_ID` stays literal for copy-paste

## Summary by CodeRabbit

* **Chores**
  * Reduced the automated review wait period from 60 minutes to 20 minutes.
  * Added safeguards to prevent stalled review requests and network calls from exceeding the time limit.
  * Updated timeout guidance with clearer instructions for requesting a review and rerunning failed checks.

* **Tests**
  * Expanded coverage for delayed, unresponsive, and numerous review candidates.
  * Confirmed that completed reviews continue to pass the automated gate as expected.
  * Improved compatibility by skipping shell-based checks when required tools are unavailable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,7 +542,18 @@ jobs:
                 # Resolve to the full SHA so the comparison below is exact
                 # equality, never a prefix guess; an unresolvable reference
                 # just skips this candidate.
-                reviewed=$(timeout 60 gh api "repos/$REPO/commits/$reviewed" \
+                #
+                # The bound is the REMAINING budget, not a fixed 60s. A
+                # per-call bound does not bound the loop: one stalled call per
+                # abbreviated candidate accumulates, and with enough candidates
+                # the total runs past the deadline and into the job timeout,
+                # which kills the job without the remedy below. Bounding each
+                # call by what is left makes the aggregate unable to exceed the
+                # deadline however many candidates there are.
+                remaining=$(( deadline - $(date +%s) ))
+                [ "$remaining" -lt 1 ] && break
+                [ "$remaining" -gt 60 ] && remaining=60
+                reviewed=$(timeout "$remaining" gh api "repos/$REPO/commits/$reviewed" \
                   --jq .sha 2>/dev/null || true)
               fi
               [ "$reviewed" = "$HEAD_SHA" ] || continue
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -494,6 +494,18 @@ jobs:
           # Failing closed is safe here. Nothing merges without a 5/5 review of
           # head; a fast fail only moves the wait out of a job that is blocking
           # everyone else's logs.
+          # `timeout` bounds the API calls below. If it were ever missing, each
+          # call would fail, `|| true` would swallow it, and the gate would
+          # report "no review yet" forever -- a false statement that blocks
+          # every merge. Fail loudly here instead: a missing tool is a broken
+          # gate, not an unreviewed pull request.
+          command -v timeout >/dev/null || {
+            echo "❌ 'timeout' (GNU coreutils) is not available on this runner."
+            echo "The gate bounds its API calls with it; without it a stalled"
+            echo "request cannot be distinguished from a missing review."
+            exit 1
+          }
+
           deadline=$(( $(date +%s) + 1200 ))
           while :; do
             # --jq would run per page; slurp the pages so every scored comment
@@ -502,11 +514,22 @@ jobs:
             # the globally latest comment instead would let a late update to
             # an old-commit review mask a valid 5/5 review of head. Bodies are
             # multiline, so ship them through base64.
-            candidates=$(gh api "repos/$REPO/issues/$PR/comments" --paginate \
+            # `timeout` bounds the request itself. The deadline below is only
+            # consulted AFTER this call returns, so an unbounded `gh api` that
+            # stalls runs past the deadline and into the job timeout, which
+            # kills the job WITHOUT printing the remedy underneath -- the whole
+            # point of failing fast. Capping the poll at 20 minutes made that
+            # more reachable, not less: the job timeout is now 25 minutes.
+            # A stalled or failed request is treated as a poll that found no
+            # review, so the loop proceeds to the deadline check and either
+            # retries or fails with guidance. `|| true` keeps a non-zero exit
+            # (124 on timeout) from aborting the loop.
+            candidates=$(timeout 120 gh api "repos/$REPO/issues/$PR/comments" --paginate 2>/dev/null \
               | jq -rs 'add
                         | [.[] | select(.user.login == "greptile-apps[bot]")
                                | select(.body | contains("Confidence Score:"))]
-                        | sort_by(.updated_at) | reverse | .[].body | @base64')
+                        | sort_by(.updated_at) | reverse | .[].body | @base64' \
+              2>/dev/null || true)
             score=""
             for encoded in $candidates; do
               body=$(printf '%s' "$encoded" | base64 -d)
@@ -519,7 +542,7 @@ jobs:
                 # Resolve to the full SHA so the comparison below is exact
                 # equality, never a prefix guess; an unresolvable reference
                 # just skips this candidate.
-                reviewed=$(gh api "repos/$REPO/commits/$reviewed" \
+                reviewed=$(timeout 60 gh api "repos/$REPO/commits/$reviewed" \
                   --jq .sha 2>/dev/null || true)
               fi
               [ "$reviewed" = "$HEAD_SHA" ] || continue
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -452,7 +452,8 @@ jobs:
     name: Greptile 5/5 Gate
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
-    timeout-minutes: 65
+    # Comfortably over the poll deadline below; the deadline is the real bound.
+    timeout-minutes: 25
     permissions:
       # contents: read backs the commits API call that resolves an
       # abbreviated reviewed-commit reference to a full SHA.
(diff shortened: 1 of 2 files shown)


vitali87:
@greptile-apps review

vitali87:
Fixed the P1, and found two more defects in the fix while proving it.

## 1. The P1 as reported

Both `gh api` calls are now bounded with `timeout`, and a stall is treated as a poll that found no review, so the loop always reaches the deadline branch and prints the remedy.

## 2. A silent-failure mode in that fix

With `timeout` absent from PATH, `2>/dev/null || true` swallows "command not found", `candidates` comes back empty, and the gate reports:

```
No Greptile review of <sha> yet, waiting ...
```

forever, while a valid 5/5 review sits on the pull request. Reproduced outside pytest: hand-tracing each stage of the loop gave correct values, and the assembled assignment returned empty.

It does not bite on `ubuntu-latest`, which ships coreutils. But the same swallowing that makes a stall non-fatal makes a missing tool invisible, and the symptom is a confident false statement rather than an error. Added a `command -v timeout` preflight that fails loudly: a missing tool is a broken gate, not an unreviewed pull request.

## 3. A bug in that preflight

I first wrote the tool name in backticks inside a double-quoted `echo`, so bash ran `timeout` as a command substitution:

```
❌  (GNU coreutils) is not available on this runner.
```

An error message that deletes the name of the missing thing. Single-quoted now, verified by executing the branch with `timeout` off PATH.

## The guard

`codebase_rag/tests/test_greptile_gate_deadline.py` extracts the gate's own shell from `ci.yml` and **executes** it against a stubbed `gh`. A structural check for the word `timeout` would pass on one bounding the wrong command, or one whose non-zero exit aborts the loop — neither emits the message this protects.

Two tests, and the second is why the first means anything:

| test | proves |
|---|---|
| stalled API reaches the deadline message | the bound exists and the deadline stays reachable |
| a 5/5 review of head still passes | the guard is not satisfied by a gate that fails everything |

**Mutation-verified.** Removing *only* the bound from the candidates call:

```
FAILED  test_a_stalled_api_call_still_reaches_the_deadline_message
        (hangs to the wall clock — the reported defect)
PASSED  test_a_five_of_five_review_of_head_still_passes
```

The asymmetry is the point: the failure is attributable to the missing bound rather than to a broken harness.

The test supplies its own `timeout` stub rather than depending on the host's, because macOS has no GNU `timeout` and a skipped test is a false green. The stub genuinely kills an overrunning child and exits 124, so the stall test cannot pass vacuously.

## Verification

- 2 passed, ruff clean, YAML parses with all 12 jobs, gate script passes `bash -n`
- Preflight failure branch executed with `timeout` off PATH
- Deadline failure branch executed: `$HEAD_SHA` expands, `$GITHUB_RUN_ID` stays literal for copy-paste

## Also updated on the issue

#1507's suggestion 2 was mine and my withdrawal of it was right in conclusion, wrong in reason. `cgr-other` measured seven consecutive heads: two spawned a review **automatically in 62 and 64 seconds**, so "absence is normal until someone asks" is false — I generalised from three heads that all happened to need asking. A [threshold](https://github.com/vitali87/code-graph-rag/issues/1507#issuecomment-5459513452) rather than an instant check is what survives, and that is a follow-up to this PR, not part of it.

vitali87:
@greptile-apps review

vitali87:
Fixed at `c6152be6`. The finding is correct and it is the same defect one level down.

## What was wrong

I bounded each `gh api` **call** and called the job done. The candidate loop makes one bounded resolution request per abbreviated candidate, and the deadline check sits **outside** that loop, so per-call bounds impose no aggregate bound: enough stalled candidates accumulate past the 20-minute deadline and into the 25-minute job timeout, which kills the job without the remedy.

Verified the structure before accepting the report — `timeout 60 gh api` inside `for encoded in $candidates`, deadline check after `done`.

## The fix

Each resolution call is bounded by the **remaining budget** rather than a fixed 60s:

```sh
remaining=$(( deadline - $(date +%s) ))
[ "$remaining" -lt 1 ] && break
[ "$remaining" -gt 60 ] && remaining=60
reviewed=$(timeout "$remaining" gh api ...)
```

The aggregate then cannot exceed the deadline however many candidates there are, which is a stronger property than adding a second deadline check inside the loop: it bounds the time spent, not just the number of checks.

## The guard

`test_many_stalled_candidates_cannot_outlast_the_deadline` drives **20 abbreviated candidates** whose resolution calls all hang, and requires the deadline message anyway. The stub answers the comments call and hangs only on `*commits*`, so the accumulation is per-candidate exactly as in the report.

This test earns its place because the two existing ones **pass with the defect present** — they use a full-length SHA, so no resolution call happens at all.

**Mutation-verified.** Reverting only this hunk to the fixed 60s bound:

```
PASSED  test_a_stalled_api_call_still_reaches_the_deadline_message
PASSED  test_a_five_of_five_review_of_head_still_passes
FAILED  test_many_stalled_candidates_cannot_outlast_the_deadline   (hangs to the wall clock)
```

Only the new test flips, so the failure is attributable to the missing aggregate bound rather than to the harness.

## Verification

3 passed, ruff clean, YAML parses with all 12 jobs, gate script passes `bash -n`.

## Note on the review shape

For anyone reading the check list: `Greptile Review` shows `completed/failure` while its own `output.summary` reads *"Greptile reviewed this pull request successfully — this check reflects your team's confidence threshold, not a review failure. The review scored 4/5."* A failed check run there means the score was below the bar, not that the bot errored. Same check-run/comment split that #1507 is about.

vitali87:
@greptile-apps review

vitali87:
@greptile-apps review

Issue #1025, opened 2026-08-01
Watch re-parse prefix sweep deregisters sibling files' inline-mod functions sharing the qn prefix


vitali87:
## Summary

`GraphUpdater.remove_file_from_state` clears every `function_registry` entry whose qn starts with the touched file's module-qn prefix. When another file's inline-mod chain shares that prefix (the issue #1017 shape: one qn, two modules), a watch event on the touched file deregisters the sibling file's functions even though that file was never re-parsed, so they are never re-registered.

## Repro

```text
src/lib.rs      pub mod beta; pub mod a;
src/beta.rs     pub fn helper() -> u32 { 2 }
src/a.rs        #[cfg(feature = "inline")]
                pub mod b {
                    pub mod c {
                        use crate::beta::helper;
                        pub fn go() -> u32 { helper() }
                    }
                }
                #[cfg(not(feature = "inline"))]
                pub mod b;
src/a/b.rs      pub fn wrap() {}
```

Index fully, then dispatch a watch `FileModifiedEvent` on `src/a/b.rs` with unchanged content. `remove_file_from_state` sweeps the prefix `project.src.a.b`, which removes `project.src.a.b.c.go` registered from `src/a.rs`. The repo-wide CALLS recompute then derives a degraded caller qn for `go`: the recorded edge becomes `(project.src.a.go, project.src.beta.helper)` instead of `(project.src.a.b.c.go, project.src.beta.helper)`, and the `go` Function node is gone from the registry until `src/a.rs` itself is touched.

## Expected

A watch event on one file must only drop registrations that file actually produced. Registrations owned by other files that happen to nest under the same qn prefix should survive, or be re-derived from their own (untouched, still cached) ASTs.

## Notes

Observed on `main` and on the #1007 branch while pinning watch-path behaviour there; the sweep predates both. Ownership tracking per file (which file registered which qn) is likely the root fix, since prefix matching cannot distinguish the two writers by construction.

PR #1511, opened 2026-08-29
refactor: share one repository walk so the C++ qn map cannot drift from the indexer

vitali87:
Closes the last open acceptance criterion of #1178 for the `_eligible_rel_files` half of `qn.py`, and removes the duplication behind the #1025 coupling.

## The problem

`cpp_frontend/qn.py` carried its own copy of `GraphUpdater._collect_eligible_files`' walk. Its comment states the requirement plainly: *"Reproduce `GraphUpdater._collect_eligible_files`' ordering exactly"* — because the module-qn disambiguation rule hands the base qn to whichever file is seen **first** and appends an extension to the loser (`foo.cpp` -> `proj.foo`, `foo.h` -> `proj.foo.h`). A divergence in walk **order**, not just in which files are eligible, silently reassigns qualified names across the graph. Nothing raises. That is what #1099 was, twice.

The two copies shared their filter predicates (the #1088 fix) but not the loop, so nothing forced the orderings to stay equal.

## Why the existing tests did not cover it

`test_cpp_qn_map_exclude_unignore.py::TestParityWithTheIndexerWalk` compares `set(qn_map)` against the indexer's keys. A set comparison cannot see order, so a walk yielding the same files in a different order satisfies every one of those assertions.

Measured, not assumed. Reversing the `dirnames` sort in `qn.py` on `main`:

| mutation | existing suite | order guard added here |
|---|---|---|
| `dirnames` reversed | 9 passed | 3 failed |
| `os.walk(topdown=False)` | 9 passed | 3 failed |

Under the first mutation the two walks genuinely disagree (`ORDERS AGREE: False`, `SETS AGREE: True`).

To be precise about severity: that divergence is **latent today, not an active defect**. A basename collision needs an identical relative path minus suffix, hence the same directory, and within a directory the filename order is pinned by a hardcoded expectation in the existing suite. It stops being latent if the disambiguation rule ever widens, or if a caller starts depending on map insertion order.

## The change

One `walk_eligible_files` generator in `utils/path_utils.py`, next to the predicates it already shares. Both `_collect_eligible_files` and `build_module_qn_map` consume it, so they cannot drift by construction. The indexer's `_collected_dir_mtimes` bookkeeping is preserved through an `on_dir` hook. Net 72 lines of duplicated walk removed.

## Tests

`test_qn_walk_order_parity.py`, five tests in two classes:

- **`TestTheQnMapWalksInIndexerOrder`** — the qn map's key order equals the indexer's file order, under no filters, `--exclude`, and `.cgrignore` rescue.
- **`TestTheSharedWalkOrderIsPinnedAbsolutely`** — pins the order against a literal expectation.

The second class exists because of something the mutation testing showed: once both callers share one walk, **parity alone stops being a guard**. Mutating the shared definition moves both consumers together, so they keep agreeing and every parity assertion stays green. Verified: the `dirnames`-reversed mutation applied to the shared walk leaves all three parity tests passing. The absolute-order tests catch all three mutations (`dirnames` reversed, `filenames` reversed, `topdown=False`), each failing on the assertion that owns that property.

## Verification

- `8499 passed, 49 skipped` (full unit suite)
- 514 passed across the walk/ignore/exclude/incremental subset
- `ruff check` clean; `ty check` shows the same 2 pre-existing warnings as `main` (both at `graph_updater.py:2808`, unrelated)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency between repository indexing and C++ module mapping.
  * Preserved deterministic, top-down file traversal order across supported operations.
  * Ensured exclusions, unignored paths, and skipped files are handled consistently.
  * Standardized module naming for regular files and package entry points.

* **Tests**
  * Added coverage verifying traversal order and parity under exclusion and unignore configurations.
  * Added checks for consistent module naming across supported file types.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -100,6 +100,7 @@ from .utils.path_utils import (
     should_keep_dir,
     should_skip_path,
     should_skip_rel_file,
+    walk_eligible_files,
 )
 from .utils.source_extraction import extract_source_with_fallback
 
@@ -2244,43 +2245,23 @@ class GraphUpdater:
             return []
 
         eligible: list[tuple[Path, str]] = []
-        state_filenames = cs.CGR_STATE_FILENAMES
-        repo_str = str(self.repo_path)
-        repo_prefix_len = len(repo_str) + 1
-        exclude_paths = self.exclude_paths
-        unignore_paths = self.unignore_paths
         self._collected_dir_mtimes = {}
-        for dirpath, dirnames, filenames in os.walk(repo_str):
-            if len(dirpath) < repo_prefix_len:
-                rel_dir = ""
-                dir_parts: tuple[str, ...] = ()
-                dir_key = cs.ROOT_DIR_KEY
-            else:
-                rel_dir = dirpath[repo_prefix_len:].replace(os.sep, "/")
-                dir_parts = tuple(rel_dir.split("/")) if rel_dir else ()
-                dir_key = rel_dir or cs.ROOT_DIR_KEY
-            dir_prefix = f"{rel_dir}/" if rel_dir else ""
+
+        def _record_dir_mtime(dir_key: str, dirpath: str) -> None:
             try:
                 self._collected_dir_mtimes[dir_key] = os.stat(dirpath).st_mtime
             except OSError:
                 pass
-            dirnames[:] = sorted(
-                d for d in dirnames if self._should_keep_dir(d, dir_prefix)
-            )
-            for fname in sorted(filenames):
-                if fname in state_filenames:
-                    continue
-                dot = fname.rfind(".")
-                suffix = fname[dot:] if dot != -1 else ""
-                rel_path_str = f"{dir_prefix}{fname}"
-                if not should_skip_rel_file(
-                    rel_path_str,
-                    dir_parts,
-                    suffix,
-                    exclude_paths=exclude_paths,
-                    unignore_paths=unignore_paths,
-                ):
-                    eligible.append((Path(f"{dirpath}/{fname}"), rel_path_str))
+
+        # One shared walk with the C++ module-qn map, so the two cannot drift
+        # in ORDER as well as in membership (issues #1025, #1099).
+        for dirpath, fname, rel_path_str in walk_eligible_files(
+            self.repo_path,
+            self.exclude_paths,
+            self.unignore_paths,
+            on_dir=_record_dir_mtime,
+        ):
+            eligible.append((Path(f"{dirpath}/{fname}"), rel_path_str))
         return eligible
 
     def _process_files(self, force: bool = False) -> None:
--- a/codebase_rag/parsers/cpp_frontend/qn.py
+++ b/codebase_rag/parsers/cpp_frontend/qn.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from ... import constants as cs
-from ...utils.path_utils import walk_eligible_files
+from ...utils.path_utils import base_module_qn, walk_eligible_files
 from ..cpp.utils import convert_operator_symbol_to_name
 from . import constants as fc
 
@@ -31,15 +31,6 @@ def _eligible_rel_files(
     ]
 
 
-def _base_module_qn(rel: str, project_name: str) -> str:
-    rel_path = Path(rel)
-    if rel_path.name in (cs.INIT_PY, cs.MOD_RS):
-        parts = rel_path.parent.parts
-    else:
-        parts = rel_path.with_suffix("").parts
-    return cs.SEPARATOR_DOT.join([project_name, *parts])
-
-
 def build_module_qn_map(
     repo_path: Path,
     project_name: str,
@@ -52,7 +43,7 @@ def build_module_qn_map(
     claimed: dict[str, str] = {}
     result: dict[str, str] = {}
     for rel in _eligible_rel_files(repo_path, exclude_paths, unignore_paths):
-        base = _base_module_qn(rel, project_name)
+        base = base_module_qn(Path(rel), project_name)
         existing = claimed.get(base)
         if existing is None or existing == rel:
             final = base
--- a/codebase_rag/parsers/cpp_frontend/qn.py
+++ b/codebase_rag/parsers/cpp_frontend/qn.py
@@ -1,11 +1,10 @@
 from __future__ import annotations
 
-import os
 from pathlib import Path
 from typing import TYPE_CHECKING
 
 from ... import constants as cs
-from ...utils.path_utils import should_keep_dir, should_skip_rel_file
+from ...utils.path_utils import walk_eligible_files
 from ..cpp.utils import convert_operator_symbol_to_name
 from . import constants as fc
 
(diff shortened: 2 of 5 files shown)


vitali87:
Pushed a second commit extending the same idea one level down.

`_base_module_qn` in `qn.py` and the module-qn construction in `DefinitionProcessor.process_file` computed the identical value with the identical `__init__.py`/`mod.rs` special case, spelled out separately in each. Both now call `base_module_qn` in `utils/path_utils.py`.

Checked equivalence before unifying rather than after: the two implementations agree on all ten probe cases including `weird.tar.gz` (only the final suffix strips), `dir.with.dots/f.py` (dotted directories keep their dots), `.hidden.py`, and both package forms. Zero divergences. Also verified the one substitution the extraction makes -- `relative_path.name` where the original read `file_path.name` -- is safe, since `cached_relative_path` preserves the filename (confirmed on the `__init__.py` and `mod.rs` cases the branch actually depends on).

Three tests added, and deliberately NOT as a parity comparison between the two call sites, per the finding in the PR description: after unification such a comparison cannot fail. They pin the rule itself -- stem retention, package naming, and single-suffix stripping.

Full unit suite: 8501 passed, 50 skipped. `ty check` shows the same 2 pre-existing warnings as `main`.

vitali87:
@greptile-apps review

vitali87:
@greptile-apps review

vitali87:
@greptile-apps review

Four commits have landed since your review of `0dd87bae`, and the summary above is anchored to that first commit only. Requesting coverage of the current head `bd8c9a0f`:

- `6c60c0a4` extracts `base_module_qn` so the indexer and the C++ qn map share the `__init__.py`/`mod.rs` rule instead of each spelling it out
- `113135e0` splits `_walk_dir_keys` out of the shared walk (Sonar S3776, cognitive complexity 23 vs 15)
- `17de2d66` adds two tests for the unignore rescue quantifier
- `bd8c9a0f` drops an unused local the previous commit left behind (Sonar S1481)

The commit most worth your attention is `17de2d66`. Mutating `any` to `all` in `should_keep_dir` passed the entire 8502-test suite: every existing fixture supplies a single unignore pattern, and with one element the two quantifiers return the same value, so nothing in the suite could separate them. The two new tests use two patterns of which one is relevant, plus the same-fixture-minus-one-property control.

vitali87:
@greptile-apps review bd8c9a0f9e6017a9c51f8218cafd4a367109b3fc

vitali87:
**State: 28 checks pass, 2 fail, and the 2 are the Greptile gate and the `All Checks Pass` aggregate downstream of it.** Everything substantive is green -- unit tests across all six platform/version combinations, integration, type check, lint, both Sonar gates, CodeQL, osv-scanner, smoke tests. Full suite locally: 8503 passed, 50 skipped.

The gate is **correct to fail**: there genuinely is no Greptile review of `bd8c9a0f`, and a gate timeout is a refusal rather than a flake. Not merging past it.

**Diagnosis, since the failure looks like a bot outage and is not one.** `Greptile Review` spawned automatically on the first commit (`0dd87bae`, 10:25:24Z) and has not spawned since, across four further pushes and four explicit requests over 90 minutes:

```
10:50:02  @greptile-apps review                    -> no spawn
11:35:18  @greptile-apps review                    -> no spawn
11:40:00  @greptile-apps review (naming commits)   -> no spawn
11:57:52  @greptile-apps review <full head SHA>    -> no spawn
```

The bot is not down: it reviewed #1503 at 11:50:49Z, inside that window. The last form is the one that produced a review in 21 seconds on #1503, so the difference is not the handle and not the SHA.

Recording the honest conclusion rather than a tidy one: on this repository **neither pushing nor commenting reliably triggers a re-review**. Measured across both PRs today, push-triggered spawns at 62s and 64s twice and produced nothing twice; comment-triggered at 21s and 9s and produced nothing four times. Anyone debugging this should read `gh api repos/OWNER/REPO/commits/$SHA/check-runs` per head rather than inferring from the summary comment, which Greptile edits in place -- `created_at == updated_at` is how you tell it never re-reviewed.

I have not used the `app.greptile.com/api/retrigger` link in the bot's own comment. It is an undocumented third-party endpoint, and firing it is not a call I will make on the assumption that it is simply a re-review button.

vitali87:
**Correcting my own diagnosis above.** I concluded that "neither pushing nor commenting reliably triggers a re-review." Every measurement in that comment is accurate, but the frame was wrong: I was comparing trigger mechanisms on a PR where no trigger applies.

Checking `Greptile Review` check-runs across **every** head this PR has had:

```
0dd87bae  runs = 1     <- automatic, 10:25:24Z
6c60c0a4  runs = 0
113135e0  runs = 0
17de2d66  runs = 0
bd8c9a0f  runs = 0

summary comment: created=10:31:56Z  updated=10:31:56Z   (byte-identical)
```

Greptile **did** review this PR, once, on the first commit, and has not touched its comment since. That is "reviewed once, will not revisit" rather than "never engaged" — a distinction the four failed requests cannot make on their own, and which retro-explains all of them without the handle or the SHA mattering at all.

So the honest statement is narrower than what I posted: on this PR the bot entered an already-reviewed state after the first commit, and no request form moves it. My cross-PR trigger data (push-spawns at 62s and 64s, comment-spawns at 21s and 9s, absences in both modes) still stands as data, but it does not diagnose *this* PR, and presenting it as the diagnosis attributed the cause to the wrong thing.

Leaving both comments rather than editing the first, since the reasoning error is the more useful record: I measured the right quantity on the wrong subject, and a tidy-sounding conclusion is exactly what makes that hard to notice afterwards.

The practical position is unchanged. The gate is correct to fail, this PR needs a review of `bd8c9a0f` that I cannot summon, and 28 of 30 checks pass with the 2 failures being the gate and its downstream aggregate.

vitali87:
**Withdrawing my second diagnosis too.** I posted "reviewed once, will not revisit" as a correction to an earlier wrong diagnosis. That explanation is also wrong, and two wrong causes in a row is worth recording rather than quietly posting a third.

A peer session counted `Greptile Review` check-runs across every head of a PR running alongside this one:

```
1e3e5054  1     b8a2e74b  1
2b297a52  1     06e0a377  1
3aabe7fe  1     88d14756  1
dc94c571  0     a482e17f  1
```

**Seven of eight heads on a single PR over ~17 hours.** So per-head re-reviewing is the norm on this repository, and there is no general already-reviewed rule to appeal to.

**What still stands, unchanged**, because it was measured rather than inferred:

- `Greptile Review` ran once on `0dd87bae` (10:25:24Z) and on none of the four later heads
- the summary comment is byte-frozen: `created_at == updated_at == 10:31:56Z`
- four requests in four forms over 90 minutes produced no spawn
- the bot was demonstrably alive throughout, reviewing another PR at 11:50:49Z

So this PR is **anomalous rather than governed**. I do not know why, and I would rather say that than offer a third cause. A timing story does not fit either: the first head here was reviewed about a minute after its push, and the second head landed over an hour later with no response.

The reasoning error is the same one twice, and it is worth naming since it is the useful part of this record. Both times I had a true observation and attached an invented explanation to it — first "the trigger mechanism is unreliable" (true of the evidence, wrong frame), then "the bot will not revisit a reviewed PR" (refuted by the PR next door). A tidy cause is the most dangerous thing to add to an accurate measurement.

Practical position is unchanged: 28 of 30 checks pass, the 2 failures are the gate and its downstream aggregate, and the gate is **correct** to fail because no review of `bd8c9a0f` exists. Not merging past it.

vitali87:
@greptile-apps review bd8c9a0f9e6017a9c51f8218cafd4a367109b3fc

vitali87:
**CodeRabbit's finding is correct and is fixed in `e96fb95e`.** Reproduced first:

```
repo='/'      dirpath='/tmp'      -> rel='mp'     <- first character eaten
repo='/repo'  dirpath='/repo/tmp' -> rel='tmp'    <- correct
```

`repo_prefix_len = len(repo_str) + 1` adds one for a separator the repo path may already carry. At the filesystem root it is counted twice, so every child directory silently loses its first character and every qualified name derived from it is corrupted — no exception, no log.

**Pre-existing, not introduced here.** The same arithmetic is on `main` in both `graph_updater.py:2249` and `cpp_frontend/qn.py:31`. This PR merged those two copies, which is what makes it a one-line fix instead of two.

Fix: `len(repo_str) + (0 if repo_str.endswith(os.sep) else 1)`. Preferred over `os.path.relpath`, which also works but touches the filesystem once per directory in the walk's hot path.

## The test is partial, and the docstring says so

Three attempts, each of which passed while surviving a mutation that restored the defect:

1. **Recomputed the prefix arithmetic** and asserted on the helper — testing its own copy of the logic rather than the shipped line.
2. **Used `Path(f"{tmp_path}{os.sep}")`** — `Path` normalises a trailing separator away, so `Path("/repo/")` is `"/repo"` and the fixture never reached the branch. The root is the only path in the filesystem that keeps one.
3. **Passed the root's prefix length to `_walk_dir_keys`** — that helper receives the prefix as a parameter and never reads the mutated line, which lives in the caller.

Reaching the line for real means calling `walk_eligible_files` with `/` as `repo_path`, which would `os.walk` the entire disk. So the shipped tests assert the **contract** the arithmetic must satisfy (given the root's correct prefix length, the helper returns the child's full name), and the class docstring states in as many words that a mutation of the caller's arithmetic does **not** fail them and that the line is covered by review and reasoning rather than by a test.

A fourth attempt that went green was clearly available. It would have proved nothing, and a false green is worse than a documented gap.

Full unit suite: 8506 passed, 50 skipped.

vitali87:
@greptile-apps review e96fb95e

vitali87:
Audited `main` for other copies of this walk, since a fix that leaves siblings behind is worth knowing about. **There are none of the same kind.**

Eight non-test `os.walk` sites exist. Classified by whether they share the *eligible-files contract* this PR unifies:

| site | shares the contract? |
|---|---|
| `graph_updater.py:2253` | **yes** — merged here |
| `cpp_frontend/qn.py:34` | **yes** — merged here |
| `import_processor.py:2427` | no — a Rust `#[path]` redirect prescan. Correctly shares `should_keep_dir`/`should_skip_rel_file` (the #1088 fix) but answers a different question, so it should not share the whole walk |
| `java_generated.py:31` | no — annotation-processor output discovery, prunes only |
| `contracts.py:54`, `js_ts/module_paths.py:32` | no — filtered file sweeps, no qn derivation |
| `python_source_roots.py:84` | no — builds setuptools `package-dir` **import names**, driven by `pyproject.toml` rather than walk order |
| `graph_updater.py:882` | no — separate purpose |

The distinguishing property is whether a walk's output feeds a **module qualified name**, since that is what makes ordering load-bearing: the disambiguation rule gives the base qn to whichever file is seen first. Only the two merged here do that.

One correction to my own audit worth recording: I first checked this with a grep for qn-building symbols and read an empty result as "no hazard". `python_source_roots.py` *does* build dotted names (`str(rel_dir).replace(os.sep, SEPARATOR_DOT)`) and my grep missed it. Reading it showed the dotted name is an **import name** from setuptools remaps, not a module qn — right conclusion, but the grep had not established it.

Issue #1188, opened 2026-08-10
Overload- and template-aware call facts in the C++ hybrid frontend


vitali87:
## Why

The hybrid C++ frontend currently layers exactly two things over tree-sitter: macro `Function` nodes (+ macro-expanded CALLS re-attribution) and `#include` IMPORTS. The heavyweight precision problems in C++ — overload selection and templates — are still handled by tree-sitter heuristics (`parsers/cpp/type_inference.py`, 354 LOC), even when a `compile_commands.json` and libclang are available and a full AST with resolved references is already being built per TU.

## Plan

After the hybrid mode migrates onto the `LanguageFrontend` fact-bundle protocol (prerequisite issue), extend the per-TU cursor walk to emit:

- **resolved_call_sites**: `CursorKind.CALL_EXPR` → `cursor.referenced` gives the exact overload libclang selected — argument-type-accurate CALLS for free where compile flags are known; includes operator overloads (`operator()`, `operator<<`) which currently resolve by name pattern.
- **Template instantiation mapping**: map calls into template specializations back to the primary template's graph node (`cursor.specialized_template`), with the specialization recorded as an edge property — so `vector<int>::push_back` and `vector<string>::push_back` both land on the `push_back` definition users actually navigate to, without fabricating per-instantiation nodes.
- **Implicit calls**: constructor/conversion calls libclang materializes (implicit conversions, copy/move ctors at call boundaries) — extending the existing implicit base ctor/dtor lifecycle work in `parsers/cpp/`.
- **external_sites**: referenced declarations living outside the repo (system headers) → suppress fabricated first-party CALLS, the same false-positive killer Roslyn provides for C#.

## Performance guardrails

TU parsing is the expensive step and the current hybrid already pays it — the marginal cost is the extra cursor visits. Still: per-TU time budget, cache facts keyed by file content hash + compile flags (compose with `parser_fingerprint.py`), and skip TUs whose facts are cache-hot.

## Tests

Fixture C++ repo exercising: an overload set differing by parameter type, a class template with two instantiations, an operator-overload call chain, an implicit conversion at a call boundary, calls into `<vector>` (must become external proofs). Assert graph parity when libclang is absent (pure fallback) — a missing toolchain never makes the graph worse.

Depends on: the `LanguageFrontend` protocol issue and the libclang packaging issue.

vitali87:
[...]

PR #1514, opened 2026-08-29
test: pin that a C++ template specialization is unreachable from calls

vitali87:
Characterisation only — **no production change**. Pins a C++ limitation found while investigating #1188, plus the reason a tempting fix is wrong.

## The finding

Given a primary template and an explicit specialization, the graph contains both classes and both methods, and the specialization gets its own `INHERITS` edge — but every call resolves to the primary:

```
CLASSES : Proc, Proc<int>
INHERITS: Proc -> Base, Proc<int> -> Base
CALLS   : use -> Proc.f            <- on a Proc<int> receiver
```

So `Proc<int>.f` is reachable structurally and never by a call, and a call on a specialized receiver is silently attributed to the primary template's method.

The cause is on the **call** side. `CppTypeInference` maps a receiver to its "bare C++ type name", stripping template arguments, because `_resolve_class_name` takes bare names — a contract documented in three places with 15 call sites through it. `Proc<int>` and `Proc` reduce to the same key.

## Why I am not fixing it here, and why the obvious fix is wrong

`extract_cpp_class_name` returns `Proc<int>` verbatim: tree-sitter names a specialization with a `template_type` node rather than a `type_identifier`, so the extractor's scan misses it and the `name`-field fallback returns the raw text. Meanwhile `_scope_segment_name` states the invariant that template arguments are text "no registry class QN holds".

That looks like the bug. I implemented the unwrap, and **three tests I had written first went red then green** — and it was still wrong:

```
before: CLASSES: Box, Box<char>     CALLS: use -> Box.put
after : CLASSES: Box, Box@3         CALLS: use -> Box.put     <- unchanged
```

The specialization then collides with the primary's bare name and takes the `@line` disambiguation suffix. Same orphan, less informative name. It also breaks `test_cpp_inheritance_edge_cases`, which asserts the bracketed form on an `INHERITS` source. Reverted.

My assertions were "no angle brackets in a class qn" and "none in a method qn" — both true of the broken graph *and* of the working one. Red/green proves a test is coupled to your change; it does not prove the test measures the defect, and when test and fix come from one mental model in one sitting they inherit its misconception together. Probing the actual graph after green is what caught it.

## What this PR adds

Two tests, pinning the limitation from both sides so it cannot drift silently:

- `test_a_call_on_a_specialized_receiver_is_attributed_to_the_primary` — asserts **today's** behaviour, so the limitation is visible rather than folklore.
- `test_a_call_on_a_specialized_receiver_should_reach_the_specialization` — `xfail(strict=True)`, asserts the **intended** behaviour. Strict so it fails loudly the day dispatch starts working instead of rotting into a silent pass.

The module docstring records the failed approach and its measurements, so the next person does not re-derive it or re-attempt the unwrap.

## Also on #1188

Three of that issue's four bullets do not describe current `main`: overload selection and external-callee suppression are already implemented via `cursor.referenced`, and the template bullet prescribes `cursor.specialized_template`, which does not exist on this libclang's `Cursor` at all (verified by `dir()`). Details are on the issue; it needs a scope decision rather than an implementation.

## Verification

Full unit suite green with this branch (the one `test_cli_smoke::test_version_flag` failure is a `subprocess.TimeoutExpired` under `-n auto` that passes in isolation and touches nothing here). `ruff` clean.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for C++ template specialization dispatch.
  - Verified current call attribution for primary and specialized template receivers.
  - Added an expected-failure test documenting the intended resolution of calls to specialized implementations.
  - Improved regression coverage for distinguishing primary-template and specialization targets.
  - Added checks confirming that calls from both receiver types are detected and attributed consistently.

### DIFF
--- a/codebase_rag/tests/test_cpp_template_specialization_qn.py
+++ b/codebase_rag/tests/test_cpp_template_specialization_qn.py
@@ -1,0 +1,89 @@
+"""An explicit C++ template specialization is unreachable from calls (#1188).
+
+Measured on `main` rather than argued. Given a primary template and an explicit
+specialization, the graph contains BOTH classes and both methods, and the
+specialization gets its own `INHERITS` edge -- but every call resolves to the
+primary:
+
+    CLASSES : Proc, Proc<int>
+    INHERITS: Proc -> Base, Proc<int> -> Base
+    CALLS   : use -> Proc.f            <- on a `Proc<int>` receiver
+
+So `Proc<int>.f` is reachable structurally and never by a call. The cause is on
+the CALL side, not in the node name: `CppTypeInference` maps a receiver to its
+"bare C++ type name", stripping template arguments, because `_resolve_class_name`
+takes bare names -- a contract documented in three places with 15 call sites
+through it. `Proc<int>` and `Proc` therefore reduce to the same key.
+
+A note for anyone tempted by the node name, since I tried it and was wrong.
+`extract_cpp_class_name` returns `Proc<int>` verbatim because tree-sitter names
+a specialization with a `template_type` node rather than a `type_identifier`,
+and `_scope_segment_name` states that template arguments are text "no registry
+class QN holds". Unwrapping it does NOT fix the orphan: the specialization then
+collides with the primary's bare name and is disambiguated to `Proc@3`, which is
+the same orphan with a less informative name, and it breaks
+`test_cpp_inheritance_edge_cases`, which asserts the bracketed form. The
+bracketed qn is at least self-describing.
+"""
+
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+from codebase_rag.tests.conftest import get_relationships, run_updater
+
+_SOURCE = (
+    "template<typename T>\n"
+    "struct Box { void put(T v) {} };\n"
+    "template<>\n"
+    "struct Box<char> { void put(char v) {} };\n"
+    "void use() {\n"
+    "    Box<double> bd; bd.put(1.0);\n"
+    "    Box<char>   bc; bc.put('x');\n"
+    "}\n"
+)
+
+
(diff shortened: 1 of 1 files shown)


vitali87:
**The finding is correct and is now fixed in `7204aab4`.** Reproduced it before acting on it, and it is the same defect this file's own docstring warns about, one level up — which is the part worth recording.

Ran the exact mutation:

```
removed `Box<char> bc; bc.put('x');` from the fixture
-> 1 passed, 1 xfailed     (identical to the unmutated result)
```

So the test did not require the specialized receiver's call to exist at all. The assertion was on the set of call *targets*, and both receivers resolve to the same target under today's behaviour, so a set collapses two calls into one element — satisfied equally by "both calls resolved to the primary" and by "only one call exists".

That is precisely the trap the module docstring describes: my first attempt at this file asserted "no angle brackets in a class qn", which was true of the broken graph and the working one. I wrote that up as a warning and then shipped the same class of assertion in the test beneath it.

**Fix:** `_put_calls` now returns a **list** rather than a set, with the reason in its docstring, and the characterisation test asserts both the count and the targets:

```python
assert len(calls) == 2, calls
assert {target for _caller, target in calls} == {f"{temp_repo.name}.box.Box.put"}, calls
```

Verified against your mutation — removing the `Box<char>` call now fails `test_a_call_on_a_specialized_receiver_is_attributed_to_the_primary`, where before it passed. Restore sha-verified; full suite 8495 passed, 49 skipped, 1 xfailed.

The `xfail(strict=True)` companion is unchanged: it asserts two distinct `put` targets, which is the intended behaviour, and still fails as expected. The pair now fails from one side or the other for both the fixture shrinking *and* dispatch changing.

vitali87:
@greptile-apps review 7204aab4

Issue #1177, opened 2026-08-10
Default CPP_FRONTEND=HYBRID is inert because libclang ships only in the test extra


vitali87:
## Problem

The default C++ frontend mode is HYBRID (`codebase_rag/config.py:198`, `CPP_FRONTEND: cs.CppFrontend = HYBRID`), but `libclang` is only listed in the `test` extra of `pyproject.toml` (~line 94). For a normal `pip install code-graph-rag`, `cpp_frontend_available()` fails its `clang.cindex.Index.create()` probe and the frontend silently degrades to pure tree-sitter.

Net effect: the macro-expansion `CALLS`, `#include` `IMPORTS`, and macro `Function` nodes the hybrid mode advertises never materialize for most users — the default configuration is inert, and nothing tells the user why.

## Fix

1. **Packaging**: add an optional extra — `pip install "code-graph-rag[cpp]"` — carrying `libclang`, and mention it anywhere `CPP_FRONTEND` is documented. (Consider whether the wheel size cost justifies making it a default dependency; the extra is the conservative first step.)
2. **Visibility**: when the repo contains C/C++ files and `CPP_FRONTEND` is HYBRID/LIBCLANG but the probe fails, log a one-time, actionable hint: install the `[cpp]` extra; and if bindings are present but `find_compile_commands()` finds nothing, hint `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON` or `bear -- make`.
3. **Docs**: a short "C/C++ semantic mode" page covering the three modes (`treesitter | libclang | hybrid`), what each adds, and the `compile_commands.json` requirement.

## Notes

- `parser_fingerprint.py:37-42` already includes the frontend mode, so availability flipping after the extra is installed correctly invalidates the incremental graph — verify with a test.

## Acceptance

- Fresh venv + `pip install ".[cpp]"` + a repo with `compile_commands.json` → hybrid facts appear (macro Function nodes, `#include` IMPORTS) with no further setup.
- Fresh venv without the extra → clear log line explaining what's missing and how to enable it.

alloutflo:
I’m working on this. I’ll keep the change scoped to the `[cpp]` extra, actionable one-time fallback diagnostics, focused documentation, and regressions for the availability/compile-database paths and parser fingerprint.

PR #1516, opened 2026-08-29
test: pin that the C++ frontend hints stay silent when they would be noise

vitali87:
Adds the missing half of an existing guard's coverage. Test-only, no production change.

## What was covered and what was not

#1177 fixed "the default configuration is inert, and nothing tells the user why": `CPP_FRONTEND` defaults to HYBRID while libclang shipped only in the `test` extra, so the macro `Function` nodes and `#include` `IMPORTS` that mode advertises silently never appeared. The fix added a `cpp` extra plus two actionable warnings, and `test_cpp_frontend_setup.py` pins all three — the extra declaration, the libclang install hint, and the compile-database hints.

What nothing pinned is the **silence** half. `_run_cpp_frontend` has two early exits whose whole job is *not* warning:

```python
if settings.CPP_FRONTEND not in (LIBCLANG, HYBRID):
    return
if not self._repo_has_c_or_cpp_files():
    # HYBRID is the default, so a repo with no C/C++ sources must
    # skip silently instead of warning ... on every index of a Python/Go project.
    return
```

That comment states the requirement and nothing enforced it. HYBRID being the **default** is what makes this consequential: a leaked warning fires on every index of every Python or Go repository, and the two requirements pull against each other — making the hints louder risks the noise, making them quieter risks the original defect. Only the loud half was pinned.

## Added

- `test_a_repo_with_no_c_or_cpp_sources_warns_about_nothing` — both downstream conditions forced to fail, so without the early exit either hint would fire.
- `test_the_treesitter_mode_warns_about_nothing` — opting out is a choice, not a degradation to report, even in a C++ repo with no libclang and no compile database.

Both assert `warning_messages == []` rather than the absence of one specific string, so a *new* warning added to this path also has to justify itself.

## Verification

Mutation-tested, each guard against the test that owns it, restores sha-verified:

| mutation | result |
|---|---|
| remove the no-C/C++ early exit | `test_a_repo_with_no_c_or_cpp_sources_warns_about_nothing` fails, 4 pass |
| remove the mode gate | `test_the_treesitter_mode_warns_about_nothing` fails, 4 pass |

Each mutation kills exactly one test, so neither is riding on the other's coverage.

Full unit suite: 8492 passed, 50 skipped. The one `test_cli_smoke::test_help_command_works` failure is a `subprocess.TimeoutExpired` under `-n auto`; that file passes 15/15 in isolation and this branch touches nothing it exercises.

## A note on how this was found

I first wrote a separate test file asserting on the log-constant names and reported to myself that the warnings had no coverage at all. Wrong: `grep` for the constant *names* missed `test_cpp_frontend_setup.py`, which asserts on message *text*. The empty result was a claim about my query rather than about the repository. Re-checked by behaviour, the real gap was narrower and is what this PR fills.

## Summary by CodeRabbit

* **Tests**
  * Added coverage confirming hybrid indexing produces no warnings for repositories without C/C++ source files.
  * Added coverage confirming Tree-sitter mode produces no warnings when indexing C++ repositories without libclang or a compilation database.

### DIFF
--- a/codebase_rag/tests/test_cpp_frontend_setup.py
+++ b/codebase_rag/tests/test_cpp_frontend_setup.py
@@ -69,3 +69,45 @@ def test_missing_compile_database_logs_generation_hints(
         if "CMAKE_EXPORT_COMPILE_COMMANDS=ON" in message and "bear -- make" in message
     ]
     assert len(matching) == 1, warning_messages
+
+
+def test_a_repo_with_no_c_or_cpp_sources_warns_about_nothing(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    warning_messages: list[str],
+) -> None:
+    """The silent half of the same guard, and the consequential one.
+
+    HYBRID is the DEFAULT mode, so a warning that leaked here would fire on
+    every index of every Python or Go repository -- the two hints above and
+    this silence pull against each other, and only the loud half was pinned.
+    """
+    (tmp_path / "main.py").write_text("x = 1\n", encoding="utf-8")
+    monkeypatch.setattr(gu.settings, "CPP_FRONTEND", cs.CppFrontend.HYBRID)
+    # Both downstream conditions fail: without the no-C/C++ early exit, either
+    # hint would fire.
+    monkeypatch.setattr(gu, "cpp_frontend_available", lambda: False)
+    monkeypatch.setattr(gu, "find_compile_commands", lambda _path: None)
+    updater = gu.GraphUpdater(MagicMock(), tmp_path, {}, {})
+
+    updater._run_cpp_frontend()
+
+    assert warning_messages == []
+
+
+def test_the_treesitter_mode_warns_about_nothing(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    warning_messages: list[str],
+) -> None:
+    # Opting out of the frontend is a choice, not a degradation to report --
+    # even in a repo full of C++ with no libclang and no compile database.
+    repo = _cpp_repo(tmp_path)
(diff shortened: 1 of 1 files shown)


vitali87:
**Correcting the last section of the PR description**, and reporting a near-miss that is more useful than the PR itself.

I wrote that my grep for the log-constant names missed `test_cpp_frontend_setup.py`. True, but incomplete: it also missed **`test_cpp_frontend_hints.py`**, a second existing file from the same #1177 fix (`3953b83e`) covering the same two hint cases through a full `run_updater` rather than a direct call. So the loud half had *two* files of coverage, not one, and my "the warnings had no test" reading was further from the truth than I said.

**The near-miss:** I had scratched my first draft into a file with that same name, decided it was redundant, and ran `rm -f` on it. That deleted the real tracked file. It surfaced only as `Warning: 1 uncommitted change` on `gh pr create`, which I nearly read as the usual parser-fingerprint noise. `git log --all -- <path>` showed it belonged to `3953b83e`; restored, and `git diff main --stat` now shows exactly the 42 lines this PR adds and nothing removed.

Two things I would take from it:

1. **A filename I chose is not evidence the file is mine.** The collision is not a coincidence — I picked that name because it is the obvious one for this subject, which is exactly why an existing file had it already.
2. `rm -f` on a path I did not verify as untracked is the deletion equivalent of the mutation-restore problem: the post-condition check (`git status`) is what caught it, not the intention.

**None of this changes the PR's substance.** Both existing files cover only the loud half; neither asserts the warnings stay silent, which is the requirement the code comment states and the one that matters at the default setting. The two tests here remain additive, and the mutation results stand:

```
remove the no-C/C++ early exit  -> test_a_repo_with_no_c_or_cpp_sources_warns_about_nothing fails, 4 pass
remove the mode gate           -> test_the_treesitter_mode_warns_about_nothing fails, 4 pass
```

Both hint files pass together: 7 passed.

Issue #1527, opened 2026-08-29
[Epic: Agent IDE] T6: Return types and parameter types as first-class graph facts


vitali87:
Part of epic #1521 — **T6**.

## Dependencies
- Depends on: nothing — foundation ticket
- Unblocks: #1533 (T11), #1536 (T14)

## Problem
`parameters` / `positional_params` are captured, but return types and parameter type annotations are mostly not graph facts, so "what does this return / what can I pass" is a source read. `change_signature` and the context builder both need types they can query.

## Scope
- Node properties on Function/Method: `return_type` (string as written + resolved qn when resolvable), `param_types` (parallel to `parameters`).
- `RETURNS` / `ACCEPTS` edges to Class/Interface/Type/Enum nodes when the annotation resolves in-project.
- Languages: Python, TypeScript, Go, Java, Rust, C#, Kotlin first (typed or annotation-bearing); others best-effort.

## Acceptance
- For each covered language, a fixture function with an annotated return and params yields the properties and, where in-project, the edges.
- Schema docs updated; `graph_audit` accepts the new properties/triples.

Issue #1521, opened 2026-08-29
Epic: CGR as an IDE for AI agents + a graph-native edit algebra


vitali87:
## Why
Every structural pain in a codebase — dead code, duplicates, broken callers, import cycles, "which tests do I run" — is the same pain: the code changes faster than anyone's model of it. AI agents write code from text and local snapshots, which is why duplicated code is rising sharply in AI-assisted repos. CGR already holds the whole structural graph across languages. This epic turns it into (1) the agent's IDE — deterministic resolve/navigate/context/feedback — and (2) an edit algebra — rename, change_signature, move, extract, inline — defined as graph transformations that cannot leave dangling callers.

## Build order
Tickets are listed in dependency order. Pick the lowest unblocked ticket; do not start a ticket whose "Depends on" issues are open.

### Foundation (parser + graph fidelity)
- T1 — call-site and import-site locations on edges
- T3 — single-file scoped re-ingest with latency budget
- T5 — edge resolution confidence + trace write-back (needs T1)
- T6 — return/param types as graph facts

### IDE layer
- T2 — deterministic query tools (needs T1)
- T4 — structural delta after every write (needs T2, T3)
- T14 — `context()` budgeted slice (needs T2, T6)

### Edit-algebra infrastructure
- T7 — transactional multi-file edits
- T8 — syntax-preserving patchers per language
- T10 — import statement rewriting (needs T1, T8)
- T13 — postcondition contract (needs T4)

### Edit-algebra operations
- T9 — `rename` (needs T1, T2, T5, T7, T8, T10, T13) — first op, proves the pipeline
- T11 — `change_signature` (needs T9, T6)
- T12 — `move` (needs T9, T10)
- T15 — `extract` / `inline` (needs T9)

## Dependency graph
```
T1 ──┬──> T2 ──┬──> T4 ──> T13 ──┐
     │         │                 │
     ├──> T5 ──┼─────────────────┼──> T9 ──┬──> T11 (also needs T6)
     │         │                 │         ├──> T12 (also needs T10)
     └──> T10 ─┼─────────────────┘         └──> T15
 T3 ──────────>┘ (T4)
 T6 ──> T14 (also needs T2), T11
 T7 ──> T9
 T8 ──> T10, T9
```

## Tickets

- [x] #1522 **T1** Store call-site and import-site locations on CALLS/REFERENCES/IMPORTS edges (depends on: —)
[...]

PR #1548, opened 2026-08-30
feat: change_signature edit operation with per-site argument mapping (#1533)

### LINKED ISSUE #1533: [Epic: Agent IDE] T11: Edit algebra op 2: `change_signature(qn, new_params, arg_mapping)`
Part of epic #1521 — **T11**.

## Dependencies
- Depends on #1532 (T9)
- Depends on #1527 (T6)
- Unblocks: —

## Problem
Signature changes are the most-cited cross-package breakage: the definition changes, callers in other packages do not. With call-site locations, arg counts/kwarg names and types in the graph, every call site can be rewritten or explicitly flagged.

## Scope
- `change_signature(qn, new_params, mapping)` where `mapping` says for each new parameter: from old positional index / old kwarg / default literal / "unmapped".
- Rewrite each call site per mapping (positional and keyword forms); rewrite overrides/implementors on hierarchies; leave `unmapped` sites untouched and list them.
- Type checks where types are known: refuse a default literal incompatible with the declared param type.
- Transaction + postcondition contract as in `rename`.

## Acceptance
- Add a required parameter with a default mapping: every call site gains the argument, tests pass.
- Reorder two parameters: positional callers reordered, keyword callers untouched.
- A site through a `heuristic` edge is listed as unmapped rather than rewritten.

vitali87:
Implements T11 of epic #1521 (closes #1533). Stacked on #1547 (postcondition contract) and #1539 (type facts); this branch merges both, so retarget once they land.

## What

- `codebase_rag/editing/signature.py`: `change_signature(repo_root, fetch_all, project, qn, new_params, allow_heuristic=, dry_run=, reingest=)`. `new_params` is the new parameter list in order, each spec `name[:annotation][=default][@source]`: `@2` takes the value from old positional index 2, `@old` from the old parameter `old`, `=literal` without a source is inserted where a site passes nothing (and becomes the definition's default), a bare name is unmapped.
- Definitions of the whole override hierarchy are rewritten (old parameter text carried over, new ones rendered per grammar: Python/TS `name: T = d`, Go `name T`, Java `T name`, Rust `name: T`). Python refuses a required parameter after a defaulted one.
- Call sites are rewritten from the tree-sitter argument list per the mapping: positional values follow the new order, keyword values keep their keyword, and once a site relied on an old default that is no longer last every later value is spelled by name (Python; other languages leave such a site unmapped). A keyword-only site that keeps every name is left as written. Sites the mapping cannot complete (no value for a required parameter, splat/spread, unknown keyword), sites without a location, and sites resolved by guesswork (unless `allow_heuristic`) are left untouched and listed as unmapped with the reason.
- A default literal is checked against the declared type (the new annotation or the graph's `param_types` from #1527) for the builtin names and refused on mismatch before anything is written.
- One `EditTransaction`; applied changes are held to the postcondition contract (`change_signature_expectation` with the unmapped sites; a rewritten site counts as mapped) through the scoped re-ingest, with undo on failure. The verdict rides on the report.
- MCP tool `change_signature` and `cgr change-signature QN -p SPEC ...`; the rename tool shares the new `_reingest_for_contract` helper.
- Fixes carried from the stack: the arity verdict counted keyword arguments twice (`arg_count` already includes them; also pushed to #1546), and the contract treats a rewritten site as mapped (also pushed to #1547).
- Docs: `docs/architecture/change-signature.md`.

## Tests

`codebase_rag/tests/test_change_signature.py` (12): add a required parameter with a default mapping (every site gains it, keyword site stays keyword, fixture tests pass), reorder two parameters (positional callers reordered, keyword caller untouched), a heuristic site listed as unmapped rather than rewritten, incompatible default literal refused, unmapped parameter leaves sites untouched and lists them (contract passes), method hierarchy rewritten together, TypeScript sites, contract failure on a caller the graph never saw (undone, files restored), spec parsing, refusals, and the MCP tool. Unit suite: 8959 passed.

## Summary by CodeRabbit

- **New Features**
  - Added Change Signature support through the CLI and MCP tools.
  - Update function or method parameters and known call sites using explicit mappings.
  - View rewritten and unmapped call sites, with dry-run support and safety checks.
  - Rename results now include verdict information and appropriate failure status.

- **Documentation**
  - Added architecture and MCP guide documentation for signature changes.

- **Tests**
  - Added coverage for parameter changes, call-site updates, validation, rollback, and unmapped cases.


vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Pushed `af92bbad`, fixing the three runtime defects from the 2/5 review. Each was reproduced against `a01ce699` before being fixed, not taken on the report's word.

**Duplicate source mappings dropped caller arguments.** `_map_arguments` consumes the site's value with `values.pop()`, so the first mapping took it and every later one read as omitted: `['left@0','right@0']` on `def helper(value)` rewrote the definition to two parameters and left `helper(7)` a one-argument call. Simulated the exact loop — `left` takes `'7'`, `right` skipped.

Refused rather than emitted twice. `_resolve_specs` already refuses duplicate parameter NAMES a few lines above (`SIGNATURE_DUPLICATE_PARAM`), so the duplicate-SOURCE guard is the same shape in the same function, and refusing gives the caller a message where the alternative gives them a `TypeError` at runtime.

**Override callers were matched against the base's parameter names.** `old_names` was derived once from the selected method and reused for every hierarchy member's call sites, so `Square.area(self, factor)` overriding `Base.area(self, scale)` had `Square().area(factor=2)` read as an unknown keyword and left alone — against a definition that had just been rewritten. Each member now carries its own pre-edit names and keyword-only set.

Worth noting why this survived: `test_method_hierarchy_is_rewritten_together` spells the parameter `scale` in BOTH classes, so the names have to differ for the bug to appear at all.

**Python's `/` was dropped on rebuild.** Confirmed the round trip: `def target(a, b, /, c)` parsed to `[('a',False),('b',False),('c',False)]`, so the rebuild emitted `def target(a, b, c)` and began accepting `target(a=1, b=2, c=3)` — a call the original rejects. That silently widens the callable contract, which is worse than the mapping simplification is worth.

The skip itself is correct and its comment is accurate: `/` neither takes a value nor shifts positional indices, so it rightly stays out of the MAPPING model. The error was using that same model to REBUILD. The boundary is now recorded separately and re-emitted, and a rewrite that would move a parameter across it is refused rather than answering a binding question nobody asked.

Verification: `test_change_signature.py` 25 passed, 0 failed, six new tests. Each fix mutation-tested individually — disabling the duplicate-source guard reddens exactly its own cases and nothing else.

One process note for whoever reviews: this PR's 30 green checks were genuine and current (run 34116746285, head_sha a01ce699) while all three of these bugs were live. Green here meant the suite never exercised these paths, not that they worked. Each new test was written to fail against the unfixed code first.

vitali87:
Superseded by #1912, a from-scratch rebuild of change_signature on today's main. This PR closes once #1912 merges.

vitali87:
Superseded by a `pr-split` DAG of draft PRs, all rebased onto the current `feat/postcondition-contract` base and kept under 350 diff lines per group:

- #2042 — feat: add change-signature types and constants
- #2043 — feat: parse and render signature parameters
- #2041 — feat: map change-signature call arguments
- #2047 — feat: rewrite signature definitions and call sites
- #2049 — feat: plan change-signature graph edits
- #2044 — feat: expose change-signature via API, CLI, and MCP
- #2045 — test: cover change-signature units
- #2046 — test: cover change-signature acceptance flows
- #2048 — test: cover change-signature regressions
- #2050 — docs: document change-signature architecture

Validation on the split source: `uv run pytest -q codebase_rag/tests/test_change_signature_acceptance.py codebase_rag/tests/test_change_signature_units.py codebase_rag/tests/test_change_signature_regressions.py` → 24 passed, 1 skipped.

PR #1549, opened 2026-08-30
feat: move edit operation with importer rewriting and cycle refusal (#1534)

### LINKED ISSUE #1534: [Epic: Agent IDE] T12: Edit algebra op 3: `move(qn, target_module)`
Part of epic #1521 — **T12**.

## Dependencies
- Depends on #1532 (T9)
- Depends on #1530 (T10)
- Unblocks: —

## Problem
Moving a symbol out of a "shared dumping ground" into the one package that uses it is the refactor that shrinks affected sets and ownership confusion, and it is tedious and error-prone by hand: definition, its own imports, every importer, re-exports.

## Scope
- `move(qn, target_module)`: cut the definition span (plus decorators/docstring/adjacent comments), paste into target with the imports it needs, rewrite all importers via the import rewriter, optionally leave a deprecation re-export at the old location (`keep_alias=True`).
- Refuse when the move would create an import cycle (check before applying); report the cycle.
- Transaction + postcondition contract.

## Acceptance
- Move a function used by three modules in a fixture: all importers updated, no cycle, tests pass; with `keep_alias` the old import path still works.
- Attempting a move that creates a cycle fails before touching files and names the cycle.

vitali87:
> [!WARNING]
> **Resolving this branch's conflicts can silently reintroduce a fixed defect.**
>
> This branch is `DIRTY` against its base: 44 conflict markers across 11 files.
> `codebase_rag/structural_delta.py` is one of them, with 12 differing hunks —
> and it contains a fix that only exists on this side.
>
> The base still has the regex-scanning `_is_variadic` / `_VARIADIC` (4 mentions).
> This branch replaces those internals with `_header_is_variadic`, which parses
> with `ast`. Resolving that file is therefore a choice between a regex scan and
> a parse. **Take-theirs, or any scripted resolution, restores the bug**: the
> too-many-arguments suppressor goes back to reading `b=2*3` as `*args` and
> missing a real `*rest` after a tuple default. Both directions are reproduced
> in the comments below.
>
> `codebase_rag/editing/move.py` does **not** conflict, so the import-detection
> fix is safe either way.
>
> **Check any resolution with:**
> ```
> pytest codebase_rag/tests/test_variadic_header_detection.py \
>        codebase_rag/tests/test_move_import_presence.py
> ```
> 21 tests, all confirmed to fail against the unfixed code before the fixes
> landed (reverting the variadic fix reddens 6 of 13; reverting the import fix
> reddens 4 of 8). A green run here is meaningful, not merely absent-of-red —
> which matters because this branch receives no automatic CI.
>
> **A symbol-inventory check will NOT catch a bad resolution here.** The two
> versions use different NAMES: this branch delegates `_is_variadic` to a new
> `_header_is_variadic` (ast-based); the base has no `_header_is_variadic` at
> all and keeps the regex body inside `_is_variadic`. Both sides therefore
> still define a variadic helper, so a resolution favouring the base imports
> cleanly and passes any "does a variadic helper exist" check while having
> silently restored the defect. Only behaviour distinguishes them.
>
> Two checks that do work:
>
> 1. **Write the expected reds down before resolving.** `def f(a, b=2*3)` must
>    be non-variadic (the regex says varargs); `def f(a=g(), *rest)` must be
>    variadic (the regex misses it). Record those, resolve, then check — so a
>    wrong resolution fails a prediction rather than needing to be spotted by
>    eye in a 12-hunk merge.
> 2. **Prove the tests can still go red.** Passing is not evidence on a branch
>    with no automatic CI: mutate the resolved helper back to the regex
>    behaviour for one run, confirm red, revert. "The tests exist", "the tests
>    ran" and "the tests could have failed" are three different claims.
>
> Then verify:
> ```
> grep -c '_header_is_variadic' codebase_rag/structural_delta.py   # expect 2
> ```
>
> Resolve by hand with the file open, and diff the symbol inventory against the
> base afterwards. A scripted resolution on a 12-hunk file is the shape that has
> silently deleted functions in this repo before.

Implements T12 of epic #1521 (closes #1534). Stacked on #1548 (change_signature); retarget once the stack lands.

## What

- `codebase_rag/editing/move.py`: `move(repo_root, fetch_all, project, qn, target_module, keep_alias=, dry_run=, reingest=)`.
  - Cuts the definition's whole lines with its decorators (or `export` wrapper) and the comment lines immediately above; methods are refused (move the class).
  - Pastes it into `target_module` (dotted name or repo-relative path; created when missing, appended when present) with the old module's import statements the moved text uses, narrowed to the needed name and respelled when the specifier is relative (JS/TS), plus an import from the old module for the names it still takes from there.
  - Rewrites every importer through the import rewriter (`from pkg.util import helper, other` splits into one statement per module); Python uses through a module import (`pkg.util.helper(...)`) follow and gain `import pkg.core`; statements the rewriter cannot retarget are listed as `unchanged_importers`.
  - The old module imports the name back when it still uses it; `keep_alias` always leaves that import marked as a re-export (`# noqa: F401`, or `export { name } from './core'`), so the old import path keeps working.
  - Refuses before touching any file when the simulated module import graph gains a strongly connected component containing the old or new module, naming the cycle (`MoveRefused.cycle`).
  - One `EditTransaction` (the new file staged alongside the patches); applied moves are held to the postcondition contract (`move_expectation`) through the scoped re-ingest, with undo on failure.
- Structural delta: a removed and an added definition with the same name and fingerprint in different files pair as a move, and `import_cycles` is public (both also pushed to #1546).
- The in-memory graph emulator keeps every site of an edge (`edge_sites`, `sites_of`) instead of the last one, so a statement importing two names yields two importer rows, as the real store does; sites are dropped with their edge on delete. It also answers `CYPHER_GRAPH_IMPORTS_OF` (what a module imports, with sites).
- MCP tool `move` and `cgr move QN TARGET [--keep-alias]`; docs in `docs/architecture/move.md`.

## Tests

`codebase_rag/tests/test_move_op.py` (8): a function used by three modules moves with its comment, docstring and the one import it needs while the unrelated import stays; both `from` importers and the module-attribute user are updated and the fixture tests pass; `keep_alias` keeps `from pkg.util import helper` working; a move that would create a cycle is refused before writing with the cycle named and nothing on disk changed; moving into an existing module appends; an old module that still uses the name imports it back; refusals and dry run; the MCP tool. Unit suite: 8965 passed.

## Summary by CodeRabbit

* **New Features**
  * Added a `move` operation for relocating module-level functions or classes while updating imports and references.
  * Added CLI and MCP support, including dry-run mode and optional alias preservation.
  * Added safeguards that detect import cycles and roll back unsuccessful changes.

* **Documentation**
  * Added architecture and MCP usage documentation for the move operation.

* **Tests**
  * Added coverage for moves, import rewriting, aliases, cycle detection, dry runs, and rollback behavior.


vitali87:
@coderabbitai review

vitali87:
Reviewed this by hand because CI has never run on it — `ci.yml` fires only on `pull_request: branches: [main, master, develop]`, and this PR targets `feat/change-signature`, so its 7 green checks are CodeQL and CodeRabbit only. No unit tests, lint, type check, Sonar or Greptile have ever run against these 12,361 lines. The layer below (#1548) turned out to carry three runtime bugs the moment anyone looked, so I went looking here too.

**One confirmed defect, `codebase_rag/editing/move.py:537`.**

`_retarget_attribute_uses` rewrites `pkg.old.helper(...)` to `pkg.new.helper(...)` at each call site, then decides whether the file needs `import pkg.new` with a TEXT search:

```python
if not _uses(text, f"import {new_spelled}"):
```

`_uses` is a regex over the whole file, so it answers "does this string appear" rather than "does this module get bound". Three inputs make it answer yes when the module is NOT bound, and in each the rewritten call site then raises `NameError`:

```
import pkg.new as n        guard=True  -> no import added   <-- pkg is never bound
# import pkg.new           guard=True  -> no import added   <-- a comment
s = "import pkg.new"       guard=True  -> no import added   <-- a string literal
```

The alias case is the realistic one and I ran it against a real interpreter rather than reasoning about it:

```
pkg/new.py:  def helper(): return 1
user.py:     import pkg.new as n
             print(pkg.new.helper())

NameError: name 'pkg' is not defined
```

So the move reports success and the edited program dies on the line the move itself rewrote. Same shape as the three defects in #1548: a successful-looking edit that leaves the project broken at runtime.

**Fix:** ask the AST, which the function already has — `self._parse(path, source)` is called two lines above. A check that walks top-level `import_statement` nodes and requires a `dotted_name` equal to the module (or a prefix of it), while rejecting `aliased_import`, gets all five cases right:

```
alias      -> False (import needed)      plain     -> True  (no import needed)
comment    -> False (import needed)      submodule -> True  (no import needed)
string     -> False (import needed)
```

**Checked and NOT a defect**, listed so nobody re-files them:

- `import pkg.new.sub` returning True is correct, not a substring bug. A submodule import binds the parent package, so `pkg.new.helper()` resolves — verified by running it.
- `_cut_span` is right. It captures leading comments and decorators, computes the moved text BEFORE swallowing trailing blank lines, and leaves a clean residue. I expected an off-by-one here and there is none.
- `_uses` is properly `re.escape`d with `(?<![\w.])%s(?!\w)`, so it does not match `foo.bar` when searching for `bar`. The bug above is not the regex, it is asking a regex a question only the AST can answer.
- `_refuse_cycles` diffs cycles before and after and refuses only on fresh ones containing the moved modules. Sound.

A broader sweep of `imports.py`, `structural_delta.py`, `contract.py`, `graph_query.py` and `structural_check.py` is still running; I will add anything it confirms.

vitali87:
Second confirmed defect, `codebase_rag/structural_delta.py:394-425`. Independently reproduced against real files on disk with the real `Definition` type, not reasoned about.

`_is_variadic` decides whether a callee accepts `*args` by regex-scanning the raw header text. It is wrong in **both** directions, and at line 450 it is the *only* thing that can suppress a `too_many` verdict:

```python
if passed > verdict.declared_count:
    if _is_variadic(definition, repo_root):
        return declared_count, cs.DELTA_ARITY_OK      # <-- sole suppressor
    return declared_count, cs.DELTA_ARITY_TOO_MANY
```

Measured against `ast` as ground truth, 4 of 7 realistic headers disagree with Python:

```
header                       scan   python
def f(a, b=2*3):             True   False   <-- false positive
def f(a, doc='a*b'):         True   False   <-- false positive
def f(a=g(), *rest):         False  True    <-- false negative
def f(a=(1,2), *rest):       False  True    <-- false negative
def f(a, *rest):             True   True    ok
def f(a, b):                 False  False   ok
def f(a, *, kw=1):           True   True    ok
```

Two independent mechanisms:

1. `_VARIADIC = re.compile(r"(?<!\*)\*(?!\*)")` searches raw text, so a `*` in a default expression (`b=2*3`) or inside a string default (`doc='a*b'`) reads as `*args`.
2. `close_at = text.find(")", open_at + 1)` stops at the FIRST `)`, which may be inside a default (`a=g()`), so a real `*rest` after it is never scanned.

Run against files on disk rather than in isolation:

```
def f(a, b=2*3):        _is_variadic -> True   (Python: False)
def g(a=dict(), *rest): _is_variadic -> False  (Python: True)
```

Both directions are user-visible. The false positive lets a genuinely broken call site pass as OK. The false negative is worse in practice: it propagates through `_arity_findings` to `contract._check_sites_mapped`, where `too_many` is deliberately not suppressible by `rewritten`, so a correct edit against a legitimately variadic callee **fails its contract and rolls back**.

**Fix** — parse rather than scan; the header is already text in hand:

```python
node = ast.parse(textwrap.dedent(text)).body[0]
return bool(node.args.vararg or node.args.kwonlyargs)
```

That also removes a latent second trigger in the same function: the header window starts at `start_line` and stops at the first line containing `)`, so any `)` before the parameter list truncates it. `function_ingest.py:812` sets `start_line` to the `def` line rather than the decorator, so a decorated definition does not currently reach it from this repo's own ingest — latent rather than live, and fixed for free by the same change.

**Also swept and clean**, so nobody re-reviews them: `editing/imports.py` (duplicate bindings round-trip; `_relative_specifier` correct across 7 path shapes; the Java/Rust `rpartition` widening correctly declines non-matching packages; every generated Python rewrite re-parsed under `ast.parse`), `editing/contract.py`, `graph_query.py`, `structural_check.py`.

Two near-misses chased and cleared rather than filed: `use crate::util::{self, helper}` splitting to `use crate::util::self;` is deprecated-but-legal and compiles under `rustc --edition 2021`; and `rename_in_all`'s whole-file regex is genuinely fragile but is pre-existing code this layer does not touch.

Caveat on coverage: `structural_delta.py`'s graph-facing paths (`snapshot`, `_tests_reaching`, `_new_duplicates`) were exercised with hand-built fakes only, as no populated graph was available.

vitali87:
Independent reproduction of both defects reported above, from a separate clone against `76fce9ad`. Executed rather than reasoned about, since the same subsystem's sibling PR yielded three defects that all looked fine by inspection.

## 1. `move.py:537` — the import-presence check confirmed

`_uses(text, f"import {new_spelled}")` is `re.search` over the raw file text (`move.py:101-102`), so any textual occurrence counts as "the import is already there".

```
does _uses think 'import pkg.new' is already present?
  aliased import  (import pkg.new as n)   -> True
  in a comment    (# TODO: import pkg.new) -> True
  in a string     (DOC = "import pkg.new") -> True
  genuinely absent                         -> False
  genuinely present                        -> True
```

Three false positives. The aliased-import case is the dangerous one, because it is idiomatic rather than contrived: `import pkg.new as n` binds only `n`, so after the call sites are rewritten to `pkg.new.helper(...)` the name `pkg` is unbound.

Ran the result against a real interpreter with a real package on disk:

```
$ python after.py
    pkg.new.helper()
    ^^^
NameError: name 'pkg' is not defined
```

Both the before and after files parse cleanly, so a parse-based postcondition would not catch it — the move reports success and the program dies at runtime on the line the move just rewrote.

## 2. `structural_delta.py:397` — the variadic detector confirmed

Two independent flaws: `close_at = text.find(")", open_at + 1)` stops at the first close-paren rather than the matching one, and `_VARIADIC` then regex-scans that slice.

Compared against `ast` on realistic headers:

| header | regex | ast | |
|---|---|---|---|
| `def f(a):` | False | False | ok |
| `def f(a, *rest):` | True | True | ok |
| `def f(a, *, k):` | True | True | ok |
| `def f(a, b=2*3):` | **True** | **False** | **disagree** |
| `def f(a, b=(1,2), *rest):` | **False** | **True** | **disagree** |
| `def f(a, cb=lambda x: (x)):` | False | False | ok |
| `def f(*args, **kw):` | True | True | ok |

Both directions reproduce. `b=2*3` — a `*` in an arithmetic default — reads as `*args`. `b=(1,2), *rest` — a real `*rest` after a tuple default — is missed because `find(")")` stops inside the tuple.

`_is_variadic` is the sole suppressor of `DELTA_ARITY_TOO_MANY` (`structural_delta.py:450-452`), so:

- the **false negative** makes a correct edit fail its own postcondition and roll back;
- the **false positive** silently suppresses a genuine too-many-arguments verdict.

(I measured 2 of 7 on my sample where the original report measured 4 of 7 — different headers, same defect. The direction that matters is that both error modes are reachable.)

## Common root cause

Both defects, and the three in #1548, are the same shape: **asking a regex a question only the parser can answer**. The repo already has tree-sitter parsing in hand at both sites — `move.py:536` parses the file immediately before the text check, and `structural_delta` has the definition node available.

Worth noting for whoever picks this up: this PR is not verifiably gated, so nothing above was going to be caught by CI. `uv run python scripts/check_pr_gated.py 1549` reports 5 reasons, including "no CI run exists at the head SHA 76fce9ad" and "required context absent at the head: ['All Checks Pass']". Tracked at #1582.

vitali87:
## The phase-order CI failure at this head is a stale test, not a code defect

The manually-dispatched run at `76fce9ad` (still this PR's head) fails `test_graph_updater_phase_order.py` on all six platform/version combos plus base install, with twelve phases reported as "pinned by this test but no longer called in `GraphUpdater.run`". Diagnosed rather than worked around, because the test's own message asks for that:

**The phases were not deleted.** All twelve still exist and still run. They were extracted into a new method:

- `_resolve_deferred_definitions()` at `graph_updater.py:904-1030` now contains the calls — e.g. `_rehydrate_registry_from_graph()` at line 927, `resolve_deferred_cpp_methods()` at line 935.
- `GraphUpdater.run()` calls `_resolve_deferred_definitions()`.

So execution order is preserved; the phases are simply one level of indirection below `run()`.

The test walks `run()` and collects direct attribute calls (33 distinct ones at this head) and asserts each pinned phase appears among them. After the extraction, none of the twelve do — hence twelve simultaneous failures on every platform, which is the signature of a structural change rather than a flake.

**This is the test needing to follow the refactor**, not the refactor being wrong. The right fix is to make the phase-detection walk transitively through methods `run()` calls, rather than deleting the pins — the test explicitly warns against dropping a constraint without confirming the recorded dependency is genuinely gone, and here it is not gone.

I have **not** made that change: it belongs to whoever owns this PR's content, and the check should be strengthened rather than relaxed, which deserves its own review.

## Note for anyone reading a red on this branch

Two commits were added on top of `76fce9ad` fixing the import-detection and variadic-detection defects reported above. They touch `editing/move.py` and `structural_delta.py` only, and neither goes near `GraphUpdater`. The phase-order failure predates them and reproduces at `76fce9ad` itself, so a red here should be attributed carefully.

Also reported at this branch's sibling: `#1550` at `0e7aadd5` fails one base-install test with `FileNotFoundError` on `.git/objects/maintenance.lock` under `popen-gw3` — that one does look like a git-maintenance race under xdist rather than a code defect, but it is equally invisible on the PR page.

Both failures come from **manually dispatched** runs, which is why neither appears in the PR's check list.

vitali87:
## Correction to my comment above: no fix is needed here at all

I said the test "needs to follow the refactor" and recommended teaching it to walk transitively. That work is **already done on `main`** — I was recommending something that exists. Correcting rather than editing, since the wrong recommendation is the more useful half to record.

Verified statically (no suite run needed):

- `main`'s `test_graph_updater_phase_order.py` defines `_DELEGATES = ("_resolve_deferred_definitions",)` and splices that delegate's body into `run()` at its call site, so phases inside it are visible to the walk.
- This branch's copy of that test file contains **no `_DELEGATES` at all** — 0 occurrences against 3 on `main`.
- The delegate it names is precisely the method I found the twelve phases had moved into (`graph_updater.py:904-1030`).

So the sequence was: the extraction landed, `main` later taught the test to follow the delegate, and this branch — **771 commits behind `main`** — carries the pre-teaching test against post-extraction code.

**Nothing in this PR needs changing, and `_ORDER` must not be touched.** The failure resolves when the branch is merged up or rebuilt. Given the chain is scheduled to be rebuilt anyway, it will resolve as a side effect.

Prediction worth recording for whoever gets there first: swapping in `main`'s version of that one test file, with `graph_updater.py` untouched, should turn all three green. I could not run it — the host test lock was held throughout — so this is **stated as unverified**, on static evidence only.

Credit to the peer session that caught my error; my original diagnosis (phases extracted into `_resolve_deferred_definitions`, behaviour intact) was right about the shape and wrong about who needed to change.

The two defect fixes on this branch are unaffected by any of this.

vitali87:
## Prediction confirmed by execution

My previous comment stated the stale-test conclusion on static evidence and flagged it as unverified because the host test lock was held. The lock cleared; here is the experiment.

Same `graph_updater.py`, untouched, only the test file swapped:

```
branch's test_graph_updater_phase_order.py  ->  3 failed in 2.47s
main's   test_graph_updater_phase_order.py  ->  3 passed in 1.91s
```

Test file restored byte-identical afterwards; working tree clean.

So it is settled: **no production code is at fault, and nothing on this PR needs changing.** The branch carries the pre-`_DELEGATES` test against post-extraction code, and merging `main` up (or the scheduled rebuild) resolves it.

## Full suite on this branch, with both defect fixes applied

```
3 failed, 8967 passed, 125 skipped, 1 xfailed  (320s, -n 4, not integration)
```

The 3 are exactly the stale phase-order tests above. **Nothing else fails**, so the two fixes introduce no regression across the whole suite.

For completeness, the fixes' own evidence:

- Reverting `_module_bound` to the text search reddens 4 of its 8 tests.
- Reverting `_header_is_variadic` to the regex scan reddens 6 of its 13 tests.
- Both source files restored byte-identical after each mutation.

Worth noting this suite run is the only verification these fixes will get: this PR's base is a feature branch, so the run at its head is manually dispatched rather than automatic, and nothing runs on a push here.

vitali87:
## Warning for whoever resolves this branch's conflicts

This branch is `DIRTY` against its base — 44 conflict markers across 11 files — and **one of the conflicting files contains a fix on this branch that a careless resolution would silently undo**.

`codebase_rag/structural_delta.py` conflicts, with 12 differing hunks. The base branch still has the old regex-scanning `_is_variadic` / `_VARIADIC` (4 mentions); this branch replaces that internals with `_header_is_variadic`, which parses with `ast`.

So resolving that file is a choice between a regex scan and a parse. Taking the base's side, or resolving mechanically, restores the defect: the too-many-arguments suppressor goes back to reading `b=2*3` as `*args` and missing a real `*rest` after a tuple default. Both directions of that defect are documented above with reproductions.

`codebase_rag/editing/move.py` does **not** conflict, so the import-detection fix is safe regardless.

**How to check the resolution afterwards**, since this branch gets no automatic CI:

```
pytest codebase_rag/tests/test_variadic_header_detection.py \
       codebase_rag/tests/test_move_import_presence.py
```

21 tests. They were confirmed to fail against the unfixed code before the fixes landed — reverting `_header_is_variadic` to the scan reddens 6 of 13, reverting `_module_bound` to the text search reddens 4 of 8 — so a green run is meaningful here rather than merely absent-of-red.

Per the repo's own guidance, this conflict wants resolving by hand with the file open, and a symbol-inventory diff against the base afterwards. A scripted resolution on a 12-hunk file is exactly the shape that has silently deleted functions in this repo before.

vitali87:
Superseded by a `pr-split` DAG of smaller reviewable PRs:

- #2072 — `feat: add move public API and constants`
- #2070 — `feat: parse move spans and imports`
- #2075 — `feat: expose import graph sites for move`
- #2071 — `feat: detect move contract deltas`
- #2080 — `feat: retarget move imports safely`
- #2074 — `feat: implement move refactoring operation`
- #2073 — `feat: expose move through CLI`
- #2078 — `feat: expose move through MCP`
- #2076 — `test: cover move operation behavior`
- #2079 — `test: cover move MCP wiring`
- #2077 — `docs: document move operation`

Native stacks registered where the DAG is linear:

- #2083: #2072 → #2070
- #2082: #2080 → #2074 → #2073

The remaining PRs are DAG fan-out or multi-parent nodes, so they keep branch-base dependencies rather than being forced into a native linear stack.

Validation run on the sanitized source branch:

- `uv run ruff format ...`
- `uv run ruff check --fix ...`
- `python3 -m compileall -q ...`

Targeted pytest for the move split was not started locally because another host-wide pytest suite was already running; replacement PR CI should be treated as the next test signal.

vitali87:
Closed as superseded by the split replacement PR DAG listed above.

PR #1550, opened 2026-08-30
feat: extract and inline edit operations (#1535)

### LINKED ISSUE #1535: [Epic: Agent IDE] T15: Edit algebra ops 4–5: `extract(qn, span, new_name)` and `inline(qn)`
Part of epic #1521 — **T15**.

## Dependencies
- Depends on #1532 (T9)
- Unblocks: —

## Problem
Extract and inline are the two operations that reduce duplication in practice (`cgr duplicates` finds clones; nothing can act on them). They complete the minimal algebra: rename, change_signature, move, extract, inline.

## Scope
- `extract(qn, (start_line, end_line), new_name)`: compute free variables in / values flowing out of the span (tree-sitter scope analysis; reuse flow_access where present), create the new function with those as params/returns, replace the span with a call. Refuse spans with early `return`/`break` that cannot be expressed as a single call.
- `inline(qn)`: only for single-return bodies with no side effects on parameters; substitute at each call site with argument-to-parameter binding, delete the definition when no callers remain (dead-code check).
- Both under transaction + postcondition contract; `inline` reports if the target has `heuristic`/`dynamic` callers and refuses.

## Acceptance
- Extract a 10-line block with two inputs and one output into a function in Python and TypeScript fixtures; tests pass.
- Inline a trivial wrapper with three callers; definition removed; tests pass.

vitali87:
Implements T15 of epic #1521 (closes #1535). Stacked on #1549 (move); retarget once the stack lands.

## What

`codebase_rag/editing/extract.py`:

- `extract(repo_root, fetch_all, project, qn, (start_line, end_line), new_name, dry_run=, reingest=)`: the span must be whole statements of the function's body; a span that cuts a statement, holds none, or leaves the function early (`return`/`yield` anywhere, `break`/`continue` whose loop is outside the span) is refused. Tree-sitter scope analysis in statement order yields the inputs (names read before being bound that the function binds) and outputs (names bound in the span that later statements read); attribute/property names and keyword-argument names are not reads, nested functions are descended. The new function goes right after the enclosing definition (after decorators/export) at the same indentation with the span dedented, returning `a, b` / `{ a, b }`; the span becomes `a, b = new_name(x, y)` / `const { a, b } = new_name(x, y);` (plain assignment when the outputs existed before). A Python method's span becomes a method with `self` carried; TypeScript parameter annotations travel with the inputs.
- `inline(repo_root, fetch_all, project, qn, dry_run=, reingest=)`: single-return bodies only (docstring allowed). Each graph-known call site is replaced by the returned expression with arguments bound positionally and by keyword, defaults filled, `self` bound to the receiver, parenthesised where precedence could change, substituted by token position so `a` never touches `obj.a`. Callers resolved by guesswork or trace only refuse the operation with the sites named. Once every site is rewritten the definition is removed and the import entries binding its name dropped (`from pkg.util import wrapper, other` keeps `other`).
- Both run as one `EditTransaction` and are held to the postcondition contract (extract: new symbol added, nothing else moves; inline: callee removed, callers of it are the plan, no dangling check) through the scoped re-ingest, with undo on failure.
- MCP tools `extract` and `inline`; `cgr extract QN START END NEW_NAME` and `cgr inline QN`; docs in `docs/architecture/extract-inline.md`.

## Tests

`codebase_rag/tests/test_extract_inline.py` (8): a nine-line accumulation with two inputs and two outputs extracted in Python (fixture tests pass) and in TypeScript, refusals (early exit, split statement, empty span, unknown symbol), a method's span extracted as a method, a trivial wrapper with three callers inlined and its definition and import entry removed (fixture tests pass), refusals for guessed callers and multi-statement bodies, dry run, and the MCP tools. Unit suite: 8973 passed.

## Summary by CodeRabbit

* **New Features**
  * Added `extract` and `inline` refactoring commands through the CLI and MCP tools.
  * Supports dry-run plans, safe transactional changes, validation, rollback, and clear refusal messages.
* **Bug Fixes**
  * Improved handling of edits without available language parsers and formatter drift reporting.
  * Fixed cross-platform cleanup and line-ending behavior.
  * Evaluation and indexing now honor configured ignore rules.
* **Documentation**
  * Added guidance for extract and inline refactoring tools.
* **Chores**
  * Removed automated issue-vetting workflow and related tooling.
  * Updated the project version to 0.0.823.


vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

vitali87:
**First full verification of this branch.** Like #1551, nothing had ever run against it: `.github/workflows/ci.yml` triggers only on `pull_request: branches: [main, master, develop]`, so a PR based on a feature branch never starts the workflow. This PR shows 7 checks against 27 for a comparable PR targeting `main` — and the 20 missing include `All Checks Pass`, which is the **only** check the `main` ruleset requires. A feature-branch PR therefore cannot satisfy the required check at all, independent of coverage.

Ran the missing checks locally against `0e7aadd5` in a detached worktree:

| Check | Result |
|---|---|
| Full unit suite (`-n 4`, `-m 'not integration'`) | **8944 passed, 126 skipped, 1 xfailed, 0 failed** |
| `ruff check` | clean |
| `ruff format --check` | 1108 files clean |
| `ty check codebase_rag --exclude tests/` | 15 diagnostics (same as #1551) |

The type-check number is **not** comparable to `main`'s 19: this branch predates #1814, which added the optional `neo4j` import diagnostics.

**Also reviewed `editing/extract.py` (957 lines) by hand**, since the sibling #1548 turned up three reproduced runtime defects in the same subsystem the moment someone looked:

- **The guessed-caller guard is real and non-vacuous.** The module docstring promises that a callee with `heuristic` or `dynamic` callers is refused. Those words appear only in the docstring, which initially looked like an unimplemented promise — but the guard exists under different vocabulary: `extract.py:641-648` filters call sites whose `resolution` is in `_AMBIGUOUS` and raises `InlineRefused(cs.INLINE_GUESSED_CALLERS)`. `_AMBIGUOUS` is `frozenset({'heuristic', 'dynamic', 'overload'})` — non-empty, and covering one case the docstring does not mention.
- **Refusal is by exception, not by sentinel return.** `ExtractRefused` / `InlineRefused` with zero `return cs.<CONSTANT>` early-exits, so a refusal cannot be mistaken for a successful edit by a caller that ignores a return value. A different shape from `signature.py`'s constant returns, and the safer one.

**No defects found.**

One methodological note worth recording, because it nearly produced a false finding here: my first grep searched for the *docstring's* vocabulary (`heuristic`, `dynamic`) in the code and found nothing outside the docstring. That reads exactly like "the promised guard is missing". The guard was there under the name `INLINE_GUESSED_CALLERS`. A grep that cannot see the shape it is checking for is not a check — in either direction.

Tracking the CI gap at #1582.

vitali87:
Superseded by the `pr-split` replacement DAG:

- #2056 — feat: add extract/inline public types and constants
- #2057 — feat: analyze extract/inline scopes and transactions
- #2062 — feat: implement extract refactoring operation
- #2058 — feat: implement inline refactoring operation
- #2063 — feat: expose extract/inline through CLI
- #2061 — feat: expose extract/inline through MCP
- #2059 — test: cover extract refactoring
- #2060 — test: cover inline, MCP, and CRLF regression
- #2064 — docs: document extract/inline architecture
- #2065 — docs: list extract/inline MCP tools

Native stacks registered where linear:

- Stack #2067: #2056 → #2057 → #2062 → #2058 → #2063, based on `feat/move-op`.
- Stack #2068: #2061 → #2065, based on `pr-split/split-pr1550-source/pr-4`.

#2060 remains a multi-parent DAG PR and targets `feat/move-op`, carrying its ancestors until they merge.

Validation on the sanitized split source:

```bash
uv run pytest -q codebase_rag/tests/test_extract_operation.py codebase_rag/tests/test_inline_operation.py codebase_rag/tests/test_extract_inline_mcp.py codebase_rag/tests/test_extract_inline_regressions.py
# 9 passed, 1 skipped, 1 warning
```

vitali87:
Closing as superseded by the pr-split replacement DAG listed above.

PR #1561, opened 2026-08-30
fix(evals): emulate the affected-caller query so the incremental emulation re-parses dependents like production

### LINKED ISSUE #1560: evals: _StatefulIngestor never answers CYPHER_AFFECTED_CALLER_PATHS, so the incremental emulation re-parses no dependents
Found while probing clean-index vs incremental-run equality across languages after #1552.

## What happens

`evals.cgr_graph._StatefulIngestor` is the in-memory stand-in for the graph store that `evals/incremental.py` and the incremental tests use to compare an incrementally mutated graph against a clean re-index. Its `fetch_all` emulates `CYPHER_INBOUND_EDGES`, `CYPHER_ALL_DEFINITION_QNS`, `CYPHER_ALL_INHERITS`, `CYPHER_ALL_MODULE_QNS` and the path queries, but has no case for `CYPHER_AFFECTED_CALLER_PATHS`, so it returns `[]` for it.

`GraphUpdater._affected_caller_keys` therefore never finds a dependent under the emulator, and every cross-file edge into a re-indexed file has to survive through the verbatim inbound restore. Production Memgraph answers that query and re-parses the dependents instead (issue #1229 phase 4), so the emulation under-reports edges a real run keeps.

## Measured

Two cases where the emulation loses an edge a production run recomputes:

- Go, same package, no import statement: `base.go` defines `func (b Base) Run()`, `derived.go` calls `d.Run()`. Touch `base.go`: `derived.Make CALLS base.Base.Run` is gone. Production re-parses `derived.go` (it holds a CALLS edge into `base.go`); the emulator restores nothing because `Base.Run` is a deferred Go receiver method not yet registered when Pass 2 restores.
- C++: `shape.h` declares `geo::Shape::area`, `shape.cpp` defines it out of class and calls it. Touch `shape.h`: `use CALLS geo.Shape.area` is gone. The header's delete cascades through `DEFINES_METHOD` into the `area` node whose `path` is `shape.cpp`, so the path-keyed capture never sees the edge; production re-parses `shape.cpp` (it IMPORTS the header).

## Expected

The emulator answers `CYPHER_AFFECTED_CALLER_PATHS` the way the production query does: distinct `caller_path` for every `CALLS|REFERENCES|INSTANTIATES|IMPORTS|INHERITS` edge whose target's `path` is in the re-indexed set, whose caller's `path` is set and not in that set, and whose both qns start with the project prefix. With that, both cases above equal a clean index (the C++ one additionally needs #1552).

vitali87:
Closes #1560. Stacked on #1559 (the C++ case's equality assertion needs that fix; this PR retargets to `main` once it merges).

## What happens

`evals.cgr_graph._StatefulIngestor`, the in-memory graph store the incremental eval and the incremental tests compare against a clean re-index, has no `fetch_all` case for `CYPHER_AFFECTED_CALLER_PATHS`, so it answers `[]`. `GraphUpdater._affected_caller_keys` therefore never finds a dependent under the emulator and every cross-file edge into a re-indexed file has to survive through the verbatim inbound restore. Production Memgraph answers the query and re-parses the dependents (issue #1229 phase 4), so the emulation under-reports edges a real run keeps.

Two measured losses, both absent from a production run:

- Go, same package: touching `base.go` dropped `derived.Make CALLS base.Base.Run`. The restore cannot carry it because `Base.Run` is a deferred receiver method not yet registered when Pass 2 restores; production re-parses `derived.go` instead.
- C++: touching `shape.h` dropped `use CALLS geo.Shape.area`. The header's delete cascades through `DEFINES_METHOD` into the `area` node whose `path` is `shape.cpp`, so the path-keyed capture never sees the edge; production re-parses `shape.cpp` (it IMPORTS the header).

## Fix

The emulator now answers `CYPHER_AFFECTED_CALLER_PATHS` the way the production query does: distinct `caller_path` for every `CALLS|REFERENCES|INSTANTIATES|IMPORTS|INHERITS` edge whose target's `path` is in the re-indexed set, whose caller's `path` is set and not in that set, and whose both qns start with the project prefix (`REFERENCES` in, `OVERRIDES` out, as in the query).

## Test

`test_incremental_inbound_deferred_targets.py` indexes each fixture, touches the defining file, runs an incremental update and asserts the cross-file CALLS edge is present and the snapshot equals the clean index. Both cases red before the emulator change, green after. The existing 86 emulator and incremental tests still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved incremental indexing reliability when files define deferred cross-file call targets.
  * Preserved inbound call relationships and complete code graphs after incremental updates.

* **Tests**
  * Added regression coverage for incremental reindexing scenarios in Go and C++ projects.
  * Expanded validation of affected dependency paths during reindexing.

### DIFF
--- a/codebase_rag/tests/test_incremental_inbound_deferred_targets.py
+++ b/codebase_rag/tests/test_incremental_inbound_deferred_targets.py
@@ -9,6 +9,7 @@
 # index and a production run keep both edges (issue #1560).
 from __future__ import annotations
 
+import os
 from pathlib import Path
 
 import pytest
@@ -70,6 +71,16 @@ def _index(
     ).run(force=force)
 
 
+def _touch_after_cache(path: Path, root: Path) -> None:
+    # The incremental pass skips a cached file whose mtime is not newer than
+    # the hash cache's, without hashing it; on a coarse-timestamp filesystem
+    # a write landing in the cache's own tick would be skipped and the test
+    # would pass without re-parsing anything. Place the edit past the cache.
+    cache_mtime = (root / cs.HASH_CACHE_FILENAME).stat().st_mtime
+    path.write_text(path.read_text(encoding="utf-8") + "// touched\n")
+    os.utime(path, (cache_mtime + 1, cache_mtime + 1))
+
+
 @pytest.mark.parametrize(
     ("language", "fixture", "edited", "call"),
     [
@@ -97,8 +108,7 @@ def test_incremental_reindex_keeps_inbound_calls_to_deferred_targets(
     # A trailing comment changes the hash but not the AST: the defining file
     # re-parses and the caller joins it as a dependent, so its edge into the
     # re-indexed file is recomputed rather than restored.
-    path = root / edited
-    path.write_text(path.read_text(encoding="utf-8") + "// touched\n")
+    _touch_after_cache(root / edited, root)
     _index(store, root, language, force=False)
     after = _snapshot(store)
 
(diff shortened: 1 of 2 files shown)


vitali87:
@coderabbitai review

PR #1600, opened 2026-09-01
fix: preserve rejected highlights in release news

vitali87:
## Summary

- emit one `Release Summary` bullet when every generated highlight is rejected by the feature-theme filter
- preserve the theme and substance of every rejected highlight in that single bullet
- keep release-workflow reruns idempotent while allowing later releases to add distinct summaries
- parse both `*` and `-` highlight markers and ignore empty highlight bodies

## Red/green TDD

1. **Red** (`c2c251e`): added regression tests first; 7 tests failed against the existing implementation.
2. **Green** (`ddfaf34`): implemented the fallback summary and idempotence fix; the same suite passes.

## Verification

- `uv run pytest codebase_rag/tests/test_update_news.py codebase_rag/tests/test_release_news_derivation.py` — 26 passed
- `uv run ruff check scripts/update_news.py codebase_rag/tests/test_update_news.py` — passed
- `git diff --check` — passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release highlight detection for both dash- and asterisk-style bullets.
  * Excluded empty highlight entries from generated news.
  * Added a consolidated **Release Summary** when no feature highlights are available.
  * Prevented duplicate summaries when news updates are repeated.
* **Tests**
  * Added coverage for highlight extraction, aggregation, filtering, and repeated updates.

### DIFF
--- a/codebase_rag/tests/test_update_news.py
+++ b/codebase_rag/tests/test_update_news.py
@@ -2,7 +2,9 @@ from __future__ import annotations
 
 from scripts.update_news import (
     LATEST_RELEASE_MARKER,
+    create_aggregated_bullet,
     existing_themes,
+    extract_all_highlights,
     extract_bullets,
     is_feature_theme,
     prepend_news,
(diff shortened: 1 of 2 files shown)


vitali87:
Heads-up on an overlap and a defect that survives this PR's rewrite of `HIGHLIGHT_BULLET`. Flagging rather than changing anything, since this is your branch.

**The defect.** Both the current `main` pattern and this PR's replacement require the colon **outside** the bold, but the highlights generator emits it **inside**:

```
this PR:  ^[-*]\s+\*\*(?P<theme>[^*]+?)\*\*:\s*(?P<text>\S.*)$

'* **Parsing Improvements:** PHP imports resolve.'   -> NO MATCH
'* **Web Search**: the agent searches.'              -> MATCH
```

Every v0.0.820 highlight used the first form, so none parsed. That is why `NEWS.md` has not changed since v0.0.770 while releases kept shipping — release commit `97dc6a31` contains only `pyproject.toml`, `server.json` and `uv.lock`, and the step still reported success. Filed as #1605.

The workflow prompt asks for "a short bold theme followed by a colon", which the model satisfies both ways, so neither spelling is wrong.

**The overlap.** I have a fix on `fix/news-bullet-colon-placement` (unpushed) that accepts either placement:

```
^- \*\*(?P<theme>[^*]+?)(?::\*\*|\*\*:) +(?P<body>\S.*)$
```

The colon is anchored to the **first** bold deliberately. A looser `:?\*\*:?` lets the theme group end at a later bold, so `- **Improvements** to **CI automation**: ...` captures `theme="Improvements"` and leaves `CI automation` in the body where `is_feature_theme` never inspects it — a CI entry then reaches Latest News, which is the thing that filter exists to prevent. I introduced exactly that leak on my first attempt and it is now pinned by regression tests in both directions.

**Ordering.** We touch the same line, so whichever merges second will conflict. If this PR lands first I will rebase onto it and apply the colon fix to whatever pattern survives (this PR collapses `for pattern in [HIGHLIGHT_BULLET, BULLET_PATTERN]` to `HIGHLIGHT_BULLET` alone, so there would be one place rather than two). Happy either way — just do not want the resolution to silently drop the colon fix, since the pipeline stays dead without it.

Issue #1609, opened 2026-09-01
extract_bullets and extract_all_highlights disagree on a missing space after the colon, silently demoting or dropping a feature highlight


vitali87:
The two extractors in `scripts/update_news.py` disagree about a bullet with no space after the colon. `BULLET_PATTERN` requires at least one space; `HIGHLIGHT_BULLET` uses `\s*` and accepts none. Same input, same file, two answers — and the disagreement is user-visible in the rendered README.

## The inputs

```python
extract_bullets("- **Theme**:No space here.")        # []
extract_all_highlights("* **Theme**:No space here.") # [('Theme', 'No space here.')]
```

Both spellings of the colon behave the same way, so this is about the space, not the colon placement.

## User-visible symptom

`prepend_news` feeds `extract_bullets` first and falls back to the aggregate built from `extract_all_highlights`. So a feature highlight written with no space after the colon is **silently reclassified**, and which way depends on what else is in the release. Measured:

**A — it is the only feature bullet in the release:**

```python
prepend_news(NEWS, "## Highlights\n* **Voice Input**:Speak to the agent.\n")
# inserted: ['- **Release Summary**: Voice Input: Speak to the agent.']
```

It is demoted out of its own entry into the catch-all Release Summary.

**B — another well-formed feature bullet exists:**

```python
prepend_news(NEWS, "## Highlights\n"
                   "* **Voice Input**:Speak to the agent.\n"
                   "* **Web Search**: Search the web.\n")
# inserted: ['- **Web Search**: Search the web.']
```

`extract_bullets` returns a non-empty list, so the aggregate fallback never fires and **the Voice Input highlight disappears entirely** — no entry, no aggregate, no warning.

A feature is therefore dropped from Latest News over a single whitespace character, and the release log says nothing (see #1605 for why that log cannot distinguish this from a legitimate no-op).

## Pre-existing

Verified against clean `origin/main` (35bc841f) in an isolated checkout: identical divergence there, so this is not introduced by #1606.

## Recommended direction

Unify on one rule — either `BULLET_PATTERN`'s (a space is required, and a no-space bullet is rejected consistently by both) or `HIGHLIGHT_BULLET`'s (no space is tolerated, and both accept it). One decision, applied to both patterns.

[...]

Issue #1608, opened 2026-09-01
Latest News filter inspects only the theme, so a non-feature word in the body reaches the README


vitali87:
`is_feature_theme` is passed the bullet's **theme** and nothing else, so a non-feature word living in the **body** is never inspected. A bullet whose theme reads as a feature can carry CI, tooling or refactor content straight into Latest News.

## Reproduction

```python
extract_bullets("- **Improvements**: to **CI automation**: pipelines are faster.")
# ['- **Improvements**: to **CI automation**: pipelines are faster.']

is_feature_theme("Improvements")     # True   <- the only thing checked
is_feature_theme("CI automation")    # False  <- never reached
```

## This is pre-existing, not a regression

Verified against clean `origin/main` (35bc841f) in an isolated checkout: the colon-outside form above already leaks there.

A related but **distinct** shape is fixed by #1606: where the theme group could run past a *later* bold, so `- **Improvements** to **CI automation**: ...` captured `theme="Improvements"` and left the CI text in the body. Anchoring the colon to the first bold closes that one. It does not and cannot close this one, because here the colon genuinely is on the first bold and the body simply is not inspected.

Both halves are pinned in `test_the_filter_inspects_the_theme_only_which_a_later_bold_exploits`, which asserts the fixed case returns `[]` and the leaking case still returns the bullet, so the boundary between them is testable rather than implied.

## The design question

Closing this means inspecting the body as well as the theme, which **redefines what counts as a feature entry**:

* A legitimate feature bullet may mention CI in passing (*"...and the CI matrix now covers it"*), and a body filter would drop it.
* Matching anywhere in the body is strictly more aggressive than today's theme-only rule, so some entries that render now would stop rendering.
* The alternative is leaving the theme as the sole contract and treating the body as prose the author is trusted with, which is the status quo.

That is a product decision about Latest News, not a mechanical fix, so it wants the repo owner's call rather than an agent's initiative. Filing it so the limit is recorded and testable either way.

Found while fixing #1605.

vitali87:
Still open and unchanged by #1614 or #1618, but here is production evidence of the impact, gathered while scoping the NEWS.md backfill.

Now that the parser fix (#1614) is on main, I replayed the real v0.0.820 highlights (the release the outage silently skipped) through the shipped code. All five themes parse, where the old parser produced zero. But **`is_feature_theme` returns True for all five**, including:

```
[...]

Issue #1605, opened 2026-09-01
Release news pipeline reports success while writing nothing: continue-on-error plus a log line that cannot distinguish dedupe-all from parse-nothing


vitali87:
`NEWS.md` last changed at v0.0.770. Releases have shipped continuously since. The release job reported **success** for every one of them, including for runs whose commit contains no `NEWS.md` at all.

## What the pipeline reported vs what it did

Release run for **v0.0.820** (`33393822143`), step list:

```
8.  Generate release notes                -> success
10. Generate highlights                   -> success
12. Update NEWS.md and regenerate README  -> success
18. Create release                        -> success
```

The commit that run produced, `97dc6a31`:

```
 pyproject.toml | 2 +-
 server.json    | 4 ++--
 uv.lock        | 2 +-
```

No `NEWS.md`, no `README.md`, though the commit step stages both. Step 12's own log line reads:

```
no new themes to add, NEWS.md unchanged
```

## Three separate things make this invisible

**1. `continue-on-error: true` on both news steps.** Steps 10 and 12 report `success` regardless of outcome, so the job summary cannot distinguish a working pipeline from a dead one.

**2. An `exit 0` guard that looks like a skip.** Step 10 opens with:

```bash
[ -n "$HIGHLIGHTS_API_URL" ] || exit 0
```

A clean exit 0 renders as a green tick. (Not the active cause here: all three `HIGHLIGHTS_*` secrets are configured.)

**3. The log message has no discriminating power.** `scripts/update_news.py` prints `no new themes to add, NEWS.md unchanged` in two entirely different situations:

* every parsed highlight was already present (correct no-op), and
* **zero highlights parsed at all** (pipeline broken).

These are the working and the broken case, and they emit the same bytes. No amount of reading that log tells you which one happened.

## Suggested change

Print the three counts separately so the states stop collapsing:

```
parsed: N   deduped: M   inserted: K
```

and fail the step when a highlights file that is **non-empty and contains bullet-shaped lines** parses to zero bullets. The guard must be conditional in exactly that way: a release with intentionally empty highlights must still bump the version, so "empty file" and "file full of bullets none of which parsed" have to be treated differently.

## Scope note

This is a change to `.github/workflows/version-bump.yml` plus `scripts/update_news.py`. Filing separately from the parser fix so the workflow edit can be reviewed and merged on its own by the repo owner.

vitali87:
Part 1 landed in #1618 (merged as `bc84bfab`).

[...]

PR #1614, opened 2026-09-01
fix: parse release highlights whose colon sits inside the bold theme

vitali87:
## Summary

- Release highlights using `**Theme:**` (colon inside the bold) never parsed, so ~50 releases inserted no news while the step reported success. Both patterns now accept either colon placement.
- The theme group excludes a trailing colon, so `**X:**` and `**X**:` produce the same dedup key and still match entries already in NEWS.md.
- The colon must sit on the first bold. `main` already rejects a later-bold theme and this change preserves that rejection rather than introducing it; only the colon-inside spelling is new.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Related to #1608, #1609, #1605 (all filed from this work, none fixed here; see Known issues below).

## Test Plan

- [x] New tests added
- [ ] Integration tests pass (`make test-integration`, requires Docker) - Docker not available on this host; integration tests unchanged by this PR
- [x] Manual testing (described below)

- 33 passed on `test_update_news.py` + `test_release_news_derivation.py` at this SHA
- 54 passed on the wider `-k "news or readme or release or highlight"` selection
- End-to-end: a v0.0.820-shaped fragment yields `[]` before this change and two
  correct entries after
- Mutation-checked, all four killed: `group('text')→group('body')` (18 failures),
  colon-outside-only revert (4), `:?\*\*:?` loosening (2), space-requirement
  relaxation (1)
- Everything #1600 introduced verified present and semantically unchanged

## Checklist

- [x] No new comments or docstrings (code should be self-documenting) - no new functions; the last commit only corrects the wording of existing comments

### The outage

Every v0.0.820 highlight used `**Theme:**` (colon inside the bold), and
`BULLET_PATTERN` only accepted `**Theme**:` (colon outside). Not one bullet
parsed, the release inserted no news, and the step still reported success.

The generator's prompt asks for "a short bold theme followed by a colon" and the
model satisfies that both ways, so the two spellings have always been in play.
This is why release highlights stopped reaching NEWS.md.

### The fix

Both patterns accept both placements via `(?::\*\*|\*\*:)`, and the theme group
excludes a trailing colon, so `**X:**` and `**X**:` produce the same dedup key
and still match entries already in NEWS.md.

The colon must sit on the **first** bold, never merely somewhere after it. A
looser `:?\*\*:?` lets the theme group end at a later bold, so
`- **Improvements** to **CI automation**: ...` would capture `theme="Improvements"`
and leave `CI automation` in the body where `is_feature_theme` never inspects it.
A CI entry would then reach Latest News, which is the one thing this module
exists to prevent. That anchoring is load-bearing: loosening it fails two
tests. To be precise about what is new here: `main` already rejects that
line, so this change **preserves** the rejection rather than introducing
it. Only the colon-inside spelling is new.

### Three findings from the rebase, each of which reading the diff would have passed

Recorded because in all three the diff looked right and running the function
did not.

1. **Near-revert of #1600.** Taking my side of the conflict wholesale would have
   silently dropped that PR's `[-*]` marker support and its `\S.*` body
   requirement. The conflict resolved cleanly either way; only diffing against
   what #1600 introduced showed one resolution was a revert. Fixed by layering
   the colon alternation on top rather than replacing the line.

2. **Missing named group.** The pattern used `(?P<text>...)` while the consumer
   still read `group('body')` - an `IndexError` on every well-formed bullet, i.e.
   total failure of the thing being fixed. Every regex-level test still passed,
   because none of them called `extract_bullets`. Caught only by an end-to-end
   test that drives the function.

3. **Dropped double space.** A whitespace detail lost in the resolution, invisible
   in review and caught by the same end-to-end test.

### Known, filed, not fixed here

- #1608 - `is_feature_theme` inspects only the theme, so a non-feature word in
  the body is never filtered. Confirmed pre-existing on `main`; this branch adds
  only its colon-inside twin. Closing it changes what counts as a feature entry.
- #1609 - `BULLET_PATTERN` requires a space after the colon while
  `HIGHLIGHT_BULLET` accepts none, so `- **Theme:**No space` is dropped by one
  and seen by the other. Identical on `main`, now pinned by a test.
- #1605 - the release step reports success when zero bullets parse, which is what
  let this run silently for ~50 releases. Needs its own PR; it touches CI.

## Summary by CodeRabbit

* **New Features**
  * News bullet formatting now supports colons placed either inside or outside bold theme labels.
  * Both formats are normalized consistently for display in the news feed.

* **Bug Fixes**
  * Improved theme handling prevents formatting issues and duplicate theme entries.
  * Highlights are correctly recognized when using either supported colon placement.
  * Missing spaces after the theme colon are handled more reliably.

### DIFF
--- a/codebase_rag/tests/test_update_news.py
+++ b/codebase_rag/tests/test_update_news.py
@@ -113,13 +113,15 @@ class TestExtractBullets:
         "- **Improvements**: to **CI automation**: ..."; accepting the
         colon-inside spelling necessarily admits its twin. Anchoring the
         colon to the FIRST bold is what stops the theme group itself from
-        running past a later bold, which is the part this change does fix.
+        running past a later bold; main already does that and this change
+        preserves it, so the anchoring is load-bearing but not new here.
 
         The residual hole is the filter's input, not the pattern, and
         closing it means inspecting the body too - a behaviour change to
         what counts as a feature entry, out of scope here.
         """
-        # Fixed by this change: the theme group cannot end at a LATER bold.
+        # Preserved by this change, also [] on main: the theme group cannot
+        # end at a LATER bold.
         assert extract_bullets("- **Improvements** to **CI automation**: x.") == []
         assert extract_bullets("* **Improvements** to **CI automation**: x.") == []
         # Not fixed, and equally true before it: a colon directly after the
--- a/codebase_rag/tests/test_update_news.py
+++ b/codebase_rag/tests/test_update_news.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 from scripts.update_news import (
+    BULLET_PATTERN,
+    HIGHLIGHT_BULLET,
     LATEST_RELEASE_MARKER,
     create_aggregated_bullet,
     existing_themes,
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

PR #1618, opened 2026-09-02
fix: report highlight parse failures instead of exiting 0 silently

vitali87:
## Summary

- `scripts/update_news.py` exited 0 whenever nothing was inserted, so a highlights format change that made every bullet unparsable looked exactly like "no new themes to add". That is what let the v0.0.820 outage run silently for ~50 releases.
- Bullet-shaped lines that none of the patterns parse now exit 1 with a diagnostic naming `HIGHLIGHT_BULLET` and the fragment path. Genuinely empty input, prose-only input and fully-deduped releases still exit 0.
- The parsed/deduped/inserted counts are reported separately, and the aggregation fallback added by #1600 is labelled rather than counted as a negative dedup.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Part 1 of #1605. The workflow half (dropping `continue-on-error` on the two news steps) is deliberately a separate PR because it touches release CI.

## Test Plan

- [x] New tests added
- [ ] Integration tests pass (`make test-integration`, requires Docker) - Docker unavailable on this host; integration tests untouched by this PR
- [x] Manual testing (described below)

- 8 new tests. 3 were confirmed red against the current script before
  implementing; 2 of the 4 state tests passed beforehand by design, pinning
  behaviour this change must not break; and the 2 covering the regression above
  were confirmed red before its fix, both failing with
  `assert True is False` on `has_unparsable_bullets`.
- All four states verified by driving `main()` end to end at the head commit:
  non-feature run 1 rc=0 / run 2 rc=0, broken shape rc=1, empty rc=0,
  prose-only rc=0.
- The stderr message named `BULLET_PATTERN` after the guard moved to
  `HIGHLIGHT_BULLET`, so an operator debugging a real failure would have
  inspected the wrong regex. Corrected, after confirming no test asserts on
  that string.

## Checklist

- [ ] No new comments or docstrings (code should be self-documenting) - this PR adds one helper with a docstring and explanatory comments, deliberately: they record why the guard keys on parse success rather than feature filtering, which is the exact regression this branch already went through once

### What was silent

`scripts/update_news.py` exited 0 whether it inserted news or understood
nothing at all. When every v0.0.820 highlight used a colon spelling the parser
rejected, zero bullets parsed, no news was inserted, and the release step
reported success. That is how the outage ran for roughly fifty releases without
anyone seeing it.

### What this changes

**Counts are reported separately** on every run: `parsed N, deduped N,
inserted N`. They fail for different reasons, and only the first distinguishes
a broken parser from a quiet release. Collapsing them into "nothing added" is
what hid the outage.

**Exit 1 is reserved for one case**: the fragment contains bullet-shaped lines
AND none of them parsed. That is the format-divergence signature. Everything
else stays exit 0.

Four states, all pinned by tests:

| fragment | exit | why |
|---|---|---|
| bullet-shaped, none parse | **1** | parser and generator have diverged |
| empty | 0 | nothing offered, so nothing missing; a release with intentionally empty highlights must still bump |
| prose only, no bullets | 0 | same reason |
| bullets parse, all themes already in NEWS.md | 0 | a normal rerun, not a failure |

That last state is why the condition is an AND. A naive `inserted == 0 means
failure` check turns every workflow rerun red.

### `BULLET_SHAPED` is deliberately looser than `BULLET_PATTERN`

`has_unparsable_bullets` asks whether a line was *meant* to be a bullet, so a
shape the real patterns reject still counts as one. A stricter matcher here
would return `False` on precisely the malformed input this exists to detect.
The docstring says so, because the natural review instinct is to tighten it.

### Aggregation-fallback counts

`prepend_news` falls back to a single `Release Summary` bullet (added by #1600)
that `extract_bullets` never produced, so `inserted` can exceed `parsed`. The
first draft computed `deduped = max(parsed - inserted, 0)`, which clamped -1 to
0 and printed the self-contradictory `parsed 0, deduped 0, inserted 1`. That
path is now labelled `(aggregated fallback)` and pinned by a test.

### The regression this went through, and the evidence it is closed

The first version of the guard keyed on `extract_bullets`, which both parses
**and** applies `is_feature_theme`. So it returned `[]` for two unrelated
reasons, and a release whose highlights are all non-feature themes (CI, docs)
was reported as a parse failure even though every bullet parsed perfectly.

It misfires only on a **rerun**. The first run is masked because the #1600
aggregation fallback makes `inserted` non-empty, so the guard is never reached.
On the rerun the summary is already in NEWS.md, nothing is inserted, and the
guard fires.

Measured on both revisions, same fragment, two consecutive runs:

| revision | run 1 | run 2 |
|---|---|---|
| base `900fe04a` | rc=0 | rc=0 |
| this branch, first attempt | rc=0 | **rc=1** (regression) |
| this branch, fixed | rc=0 | rc=0 |

The guard now keys on `extract_all_highlights`, which uses `HIGHLIGHT_BULLET`
with no feature filter and is the actual "did the patterns understand this"
question. The two return values separate the cases that `extract_bullets`
cannot:

| fragment | `extract_bullets` | `extract_all_highlights` |
|---|---|---|
| non-feature, parses fine | 0 | 2 |
| v0.0.820 broken shape | 0 | 0 |

Detection is preserved: the broken shape still exits 1. Two tests pin this - an
end-to-end rerun, and a unit test asserting the discrimination directly so a
future change that reconflates them fails at the unit level rather than only in
the rerun path. Both were confirmed red before the fix.

Worth noting the CI consequence, since it was not cosmetic: Actions `run:` blocks
execute under `bash -e`, so that exit 1 aborted the step before the README
regeneration, skipping the re-render on an otherwise healthy release.

### This is not yet fatal in CI

The non-zero exit is **recorded on the step and swallowed** by
`continue-on-error: true` at `.github/workflows/version-bump.yml:230`, which
covers the call at line 241. So part 1 makes the failure **visible in the log**;
part 2 (a separate PR, owned by the user because it touches the release
workflow) removes `continue-on-error` and makes it **fatal**. Stating this
explicitly because a PR that reads as "this stops the silent failure", while the
failure is still swallowed one layer up, would be the same class of bug it fixes.

### Stale docstring

The module docstring claimed "a malformed or empty AI response degrades to a
no-op" and "Exit code 0 always signals NEWS.md is in a valid state". This change
falsifies both. Found by grepping the *concept* (`exit code`, `no-op`, `always`)
rather than the symbol edited, because the claim sits at lines the diff never
displays. Rewritten to state both exit codes and what each means.

## Summary by CodeRabbit

* **New Features**
  * Added clear update summaries showing parsed, deduplicated, and inserted release counts.
  * Added an “aggregated fallback” indicator when that fallback is used.
  * Update processing now reports a failure when bullet-formatted content cannot be parsed.

* **Bug Fixes**
  * Distinguished legitimate no-op updates from malformed release content.
  * Correctly handles reruns, fully deduplicated releases, and releases containing only non-feature changes.

### DIFF
--- a/codebase_rag/tests/test_update_news.py
+++ b/codebase_rag/tests/test_update_news.py
@@ -440,9 +440,16 @@ class TestParseFailureIsReported:
             "- **Graph Export**  the graph exports to GraphML.\n"
         )
         exit_code, news = self._run(tmp_path, monkeypatch, fragment)
+        captured = capsys.readouterr()
         assert exit_code != 0
         assert "Web Search" not in news
-        assert "parsed 0" in capsys.readouterr().out
+        assert "parsed 0" in captured.out
+        # The diagnostic must name the regex an operator has to inspect, and the
+        # file to inspect it against. An earlier review found this message naming
+        # BULLET_PATTERN after the guard had moved to HIGHLIGHT_BULLET, which would
+        # send someone to the wrong regex mid-incident; nothing asserted it then.
+        assert "HIGHLIGHT_BULLET" in captured.err
+        assert str(tmp_path / "highlights.md") in captured.err
 
     def test_an_empty_highlights_file_stays_a_clean_noop(
         self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
--- a/codebase_rag/tests/test_update_news.py
+++ b/codebase_rag/tests/test_update_news.py
@@ -14,6 +14,7 @@ from scripts.update_news import (
     existing_themes,
     extract_all_highlights,
     extract_bullets,
+    has_unparsable_bullets,
     is_feature_theme,
     prepend_news,
 )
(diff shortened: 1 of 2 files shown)


vitali87:
@coderabbitai review

vitali87:
@greptileai review

Issue #1606, opened 2026-09-01
Changing --exclude between runs is invisible to the sync check, so the run is skipped and the excluded file stays indexed


vitali87:
## What

`GraphUpdater._is_already_in_sync` does not consider the effective exclusion set. When a file is excluded via the **CLI `--exclude` flag** on an otherwise unchanged repo, nothing on disk changes, the sync check returns True, `run()` returns at graph_updater.py:925 before any work, and the previously-indexed file keeps its `Module`, `Class` and `Method` nodes.

## Probe

Same fixture, same exclusion, two exclusion sources. Second run is incremental (`force=False`):

```
CGRIGNORE-FILE: exclude=['src/Base.java'] _process_files_ran=True  in_sync_skip=False Base_after=False
CLI-FLAG-ONLY : exclude=['src/Base.java'] _process_files_ran=False in_sync_skip=True  Base_after=True
```

The `.cgrignore` row is the control and it **works correctly** — editing `.cgrignore` changes a file that is itself indexed and hashed, so the hash loop in `_is_already_in_sync` returns False, `_process_files` runs, the excluded key is in `old_hashes` and not in `current_file_keys`, and the reconciliation at graph_updater.py:2486 deletes it.

The CLI-flag row leaks: no on-disk change, so nothing tells the sync check anything happened.

## Mechanism

`_is_already_in_sync` (graph_updater.py:2147-2190) checks:

- `_delombok_state_changed`
- `_single_file`
- hash-cache presence
- directory mtimes
- per-file hashes

It never looks at `exclude_paths` or `unignore_paths`. The exclusion set is not part of the sync fingerprint, so any exclusion source that does **not** pass through a hashed file is invisible to it. `.cgrignore` is covered only incidentally, by virtue of being an indexed file.

## Scope

Narrow: it needs the exclusion set to change between runs on an otherwise unchanged repo, via `--exclude` rather than `.cgrignore`. `--force` is an effective workaround today.

## Suggested fix

Persist the effective exclusion set alongside the existing cache fingerprint and compare it in `_is_already_in_sync`, so a changed exclusion set forces a run the same way a changed parser fingerprint does. Small and reviewable. Until then, `--force` after changing `--exclude` patterns is the documented escape hatch.

## Not the following, all checked

[...]

PR #1624, opened 2026-09-02
fix: MCP index and update honour .cgrignore and .gitignore like the CLI (#1616)

### LINKED ISSUE #1616: MCP index/update ignore .cgrignore and .gitignore, diverging from the CLI
## Problem

The MCP index and update tools (`codebase_rag/mcp/tools.py`, `_index_repository_sync` around line 721 and `_update_repository_sync` around line 766) construct `GraphUpdater` without any exclude set, so `.cgrignore` and `.gitignore` patterns are ignored entirely when a repository is indexed or updated over MCP. The CLI path honours both files and the `--exclude` flags.

Two consequences:

- An MCP-driven index parses everything the ignore files were meant to skip (vendored code, build output, generated sources), producing a different graph from the CLI for the same repository.
- Alternating CLI and MCP runs on the same repository flips the eligible file set on every switch, so each run after a switch is a full re-parse rather than a fast-path no-op.

## Expected

MCP index and update resolve the same exclude and unignore sets the CLI resolves (`.cgrignore`, `.gitignore`, interactive-setup choices) and pass them to `GraphUpdater`, so both entry points produce the same graph and share the fast path.

## Notes

Found while reviewing #1606. The exclusion-state stamp introduced there records the eligible set per run, so once MCP honours the ignore files the CLI/MCP flip disappears without further change.

vitali87:
Closes #1616.

## Problem

`_index_repository_sync` and `_update_repository_sync` constructed `GraphUpdater` without `exclude_paths` or `unignore_paths`. Both parameters default to `None`, so an MCP-driven index or update parsed everything `.cgrignore` and `.gitignore` were meant to skip, while the CLI honoured both files at each of its two call sites.

Two consequences, per the issue: an MCP index produces a different graph from the CLI for the same repository, and alternating CLI and MCP runs flip the eligible file set on every switch, so each run after a switch is a full re-parse rather than a fast-path no-op.

## Change

A `_ignore_sets()` helper on `MCPToolsRegistry` resolves the sets through `load_ignore_patterns`, the same loader both CLI sites use, and both sync methods now pass them.

`load_ignore_patterns` is the correct loader rather than `load_cgrignore_patterns`: it unions `.cgrignore` with the root `.gitignore` minus negations, whereas the latter would drop the `.gitignore` half that the issue explicitly names.

MCP has no `--exclude` flag and no interactive setup (both tool schemas are `properties={}, required=[]`), so that loader is the whole contract here. The CLI's additional `cli_excludes` union is a no-op when the flag is absent, and its `prompt_for_unignored_directories` branch has no MCP counterpart.

## Tests

`codebase_rag/tests/test_mcp_ignore_parity.py`, six cases parametrised over both sync methods.

They assert on the kwargs `GraphUpdater` is **called with**, not on the resulting graph. The defect is an omission at the call site: a missing argument leaves the graph well-formed but wrong, so an end-to-end assertion would pass against the broken code. The exclude test also guards its own fixture (`{"vendor/", "generated/", "dist/", "build/"} <= expected`) so an empty expectation cannot produce a vacuous pass.

Confirmed red first against the unfixed code (`KeyError: 'exclude_paths'`), then green. Three mutants, each caught by the test written for it, with apply-and-restore verified both ways:

- drop `exclude_paths` at the update site only -> only the `_update_repository_sync` cases fail, so the parametrisation discriminates per call site
- drop the `or None` idiom -> the empty-repo cases fail
- swap to `load_cgrignore_patterns` -> the exclude cases fail, catching the lost `.gitignore` half

## Notes

`or None` mirrors the CLI's own idiom and the parameters' defaults. It is call-shape parity rather than behaviour: every consumer gates on truthiness (`if exclude_paths and ...` in `should_skip_path`, `should_skip_rel_file`, `should_keep_dir`), so an empty frozenset and `None` are indistinguishable downstream. The comments say so explicitly, because an earlier draft claimed a behavioural distinction that does not exist.

No interaction with #1606's exclusion-state stamp: `.cgrignore` and `.gitignore` are themselves eligible hashed files and are not in `CGR_STATE_FILENAMES`, so editing either already defeats the in-sync skip. MCP has no flag channel, so the only exclusion source on this path is covered.

Local: 8903 passed excluding `tests/integration/` (the 257 integration errors are Memgraph-connection setup failures with no local DB, pre-existing and unrelated). ruff, ruff format and `ty` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Repository indexing and updates through MCP now respect `.cgrignore` and root `.gitignore` rules.
  * Excluded files are no longer processed during MCP-driven ingestion, matching CLI behavior.
  * Ignore rules with negation patterns are handled consistently.

### DIFF
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -11,6 +11,7 @@ from rich.console import Console
 from codebase_rag import constants as cs
 from codebase_rag import logs as lg
 from codebase_rag import tool_errors as te
+from codebase_rag.config import load_ignore_patterns
 from codebase_rag.graph_updater import GraphUpdater
 from codebase_rag.models import ToolMetadata
 from codebase_rag.parser_loader import load_parsers
@@ -707,6 +708,22 @@ class MCPToolsRegistry:
             logger.error(lg.MCP_ERROR_WIPE.format(error=e))
             return cs.MCP_WIPE_ERROR.format(error=e)
 
+    def _ignore_sets(self) -> tuple[frozenset[str] | None, frozenset[str] | None]:
+        # The CLI resolves `.cgrignore` and the root `.gitignore` through
+        # `load_ignore_patterns` before every ingest; MCP omitted it entirely,
+        # so an MCP-driven run parsed what the ignore files exclude and built a
+        # different graph from the CLI for the same repo (#1616). MCP has no
+        # `--exclude` flag and no interactive setup, so that loader is the whole
+        # contract here -- the CLI's extra `cli_excludes` union and
+        # `prompt_for_unignored_directories` branch have no MCP counterpart.
+        # `or None` mirrors the CLI's own `... or None` idiom and the
+        # parameters' default. It is call-shape parity, not behaviour: every
+        # consumer gates on truthiness (`if exclude_paths and ...` in
+        # should_skip_path/should_skip_rel_file/should_keep_dir), so an empty
+        # frozenset and None are indistinguishable downstream.
+        patterns = load_ignore_patterns(Path(self.project_root))
+        return patterns.exclude or None, patterns.unignore or None
+
     def _index_repository_sync(self) -> str:
         # Same collision-resistant derivation as the CLI: a bare directory
         # name would let two repos named alike delete each other's graphs.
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

Issue #1615, opened 2026-09-01
Incremental update forgets deletions when the process dies between hash-cache save and graph flush


vitali87:
## Problem

`GraphUpdater._save_hash_cache` runs inside `_process_files` (graph_updater.py, around line 2650), before the final `flush_all` in `run` (around line 1190). Any deletion issued in that window is committed to the hash cache before the corresponding `CYPHER_DELETE_MODULE` reaches the graph. If the process dies between the cache save and the flush (crash, SIGKILL, database outage), the next incremental run reads a cache that already says the file is gone, computes an empty `deleted_keys`, and never re-issues the delete. `preexisting_paths` is populated only on full builds, so incremental reconciliation does not consult the graph either. The subtree stays in the graph permanently until a full rebuild.

This applies to plain on-disk deletions as well as to files newly excluded via `--exclude` or ignore patterns.

## Reproduction

1. Run 1: index a repo containing `module_a.py`.
2. Run 2: delete `module_a.py` on disk (or add it to `--exclude`), run an incremental update with `flush_all` patched to raise after `_process_files` returns.
3. Run 3: run the same incremental update again.

Observed on run 3: `_process_files` runs, but the set of `CYPHER_DELETE_MODULE` paths does not contain `module_a.py`; the Module node and its children remain in the graph.

Expected: run 3 deletes `module_a.py` from the graph.

## Proposed fix

Either save the hash cache only after the flush commits (the same post-flush commit point the exclusion-state stamp uses), or, when the previous run did not reach its commit point, populate `preexisting_paths` from `_existing_module_paths()` on the next incremental run so the graph's own module list joins reconciliation. The second option is cheaper and also covers the exclusion-change case.

A regression test must run the successor update to completion and assert on the deleted module paths, not only on the sync check flipping to false.

Found while reviewing #1606, which addresses the exclusion-set fingerprint but not this window.

PR #1626, opened 2026-09-02
fix: reset skipped_because_in_sync per run so a reused updater does not report a stale skip (#1620)

### LINKED ISSUE #1620: skipped_because_in_sync is sticky across run() calls on a reused GraphUpdater
## Problem

`GraphUpdater.skipped_because_in_sync` is set to `True` on the in-sync early return in `run()` and is only ever cleared in `__init__`. It is never reset at the top of `run()`, so on an instance that is reused for a second `run()` the flag stays `True` even when the second run does real work.

This matters now that `run()` resets its per-run exclusion memo at the top (fix for #1606), which makes instance reuse a supported shape: the memo is per-run by construction, but this flag is not.

Observed: instance runs once in sync (flag `True`), the stamp is rewritten underneath it, second `run()` reconciles and re-parses, flag still reads `True`.

No production caller reuses an instance today (`cli.py`, `mcp/tools.py` construct one and call `run()` once), so this is latent.

## Expected

Reset `skipped_because_in_sync = False` at the top of `run()` next to the exclusion-memo reset, with a test that runs one instance twice and asserts the flag reflects the second run.

## Notes

Found during review of the #1606 fix.

vitali87:
Closes #1620.

## Problem

`GraphUpdater.skipped_because_in_sync` is set to `True` on the in-sync early return in `run()` and was cleared only in `__init__`. On an instance reused for a second `run()` the flag stayed `True` even when that run did real work.

`cli.py:379` reads the flag to choose between "already in sync, no changes detected" and the normal completion message, so a stale `True` reports a run that re-parsed the repository as having done nothing.

No production caller reuses an instance today (`cli.py` and `mcp/tools.py` each construct one and call `run()` once), so this is latent. It matters because #1606 made instance reuse a supported shape: `run()` already resets its per-run state at the top, and this flag was the one piece that was not.

## Change

One assignment beside the existing per-run reset at the top of `run()`, above every branch point that can set or read the flag.

## Tests

`test_skipped_flag_is_per_run_not_per_instance` in the existing `TestFastPathInSync` class: one instance, two runs to reach the in-sync fast path, then a third forced run that does real work.

The order of assertions is load-bearing. The flag is asserted `True` first, because without that a run which never set it would satisfy the second assertion on its own and the test would pass against the unfixed code.

Verified red against the unfixed code, both before and after rebasing onto current `main`: it fails at the second assertion with `assert True is False`, which also proves the first assertion held and the fast path was genuinely taken rather than the test passing vacuously.

## Scope

Deliberately limited to the reset and its test, with nothing else in `run()` touched, so the pending #1615 work in the same function rebases cleanly. #1615 moves the hash-cache save to the post-flush commit point (~line 1220s); this inserts at ~957, a different region.

## Notes

Three staleness axes were checked beyond the reported one, all covered by a reset at the top of `run()` rather than needing a `finally`:

- **exception mid-run** — the flag reads `False` after a run that raises following a prior skip, because the reset happens before any code that can raise past it
- **single-file path** — `_is_already_in_sync` returns `False` unconditionally when `_single_file is not None`, so that path can never set the flag `True`, and the reset still clears a prior `True` on it
- **external setters** — none exist; the only three assignment sites are all in `graph_updater.py` (init, this reset, and the in-sync set), so the reset cannot clobber a caller-set value

Local: full suite 8965 passed, 0 failed (integration errors are Memgraph-connection setup failures with no local database, pre-existing). ruff, ruff format and `ty` clean; the one `ty` warning in this file is pre-existing on `main` and exits 0.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected status reporting when the graph updater is reused across multiple runs. Results now accurately indicate whether the current run was skipped because everything was already synchronized.

- **Tests**
  - Added regression coverage for status reset behavior across unchanged and forced update runs.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -954,6 +954,12 @@ class GraphUpdater:
         # Reset per-run parse tracking so a reused updater does not reprocess
         # a previous run's files in Pass 3.
         self._parsed_files.clear()
+        # Per-run for the same reason: set on the in-sync early return below
+        # and previously cleared only in `__init__`, so a reused instance kept
+        # reporting a previous run's skip. `cli.py` reads it to decide what to
+        # report, so a stale True describes a run that did real work as
+        # already in sync (#1620).
+        self.skipped_because_in_sync = False
         self._sink.ensure_node_batch(
             cs.NODE_PROJECT,
             {
--- a/codebase_rag/tests/test_graph_updater_incremental.py
+++ b/codebase_rag/tests/test_graph_updater_incremental.py
@@ -968,6 +968,41 @@ class TestFastPathInSync:
             updater2.run(force=True)
             spy_calls.assert_called_once()
 
+    def test_skipped_flag_is_per_run_not_per_instance(
+        self, py_project: Path, mock_ingestor: MagicMock
+    ) -> None:
+        """A reused instance must not carry a previous run's skip into this one.
+
+        `skipped_because_in_sync` is set on the in-sync early return and was
+        only ever cleared in `__init__`, so once an instance skipped it kept
+        reporting True for every later `run()` however much work that run did.
+        `cli.py` reads the flag to decide what to tell the user, so a stale
+        True reports "already in sync" for a run that re-parsed the repo.
+
+        The order matters: the flag must be observed True first, or a run that
+        never set it would satisfy the second assertion on its own and the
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

Issue #1632, opened 2026-09-02
csharp_oracle_available() checks that dotnet exists, not that it can build net10.0, so an older SDK turns skips into failures


vitali87:
`csharp_oracle_available()` tests whether the `dotnet` binary exists, but the oracle project targets `net10.0`. On a machine with an older SDK installed the guard reports the toolchain as available, the build fails, and the C# oracle tests go red for a reason that has nothing to do with the code under test.

Found while running the suite during unrelated work.

## Reproduction

On a box with .NET SDK 8.0.130:

```bash
uv run --frozen pytest codebase_rag/tests/test_csharp_retrieval_eval.py::test_oracle_captures_first_party_csharp_calls -q
```

```
subprocess.CalledProcessError: Command '['/usr/bin/dotnet', 'build', '.../csharp_oracle',
'-c', 'Release', '-o', '.../bin/oracle', '--verbosity', 'quiet']' returned non-zero exit status 1
```

The same build run directly gives the actual reason:

```
error NETSDK1045: The current .NET SDK does not support targeting .NET 10.0.
Either target .NET 8.0 or lower, or use a version of the .NET SDK that supports .NET 10.0.
```

## Cause

`evals/oracles/csharp_oracle.py:31`

```python
def csharp_oracle_available() -> bool:
    return shutil.which(ec.DOTNET_BIN) is not None
```

`evals/oracles/csharp_oracle/Oracle.csproj:5`

```xml
<TargetFramework>net10.0</TargetFramework>
```

Presence is being used as a proxy for "can build this project", and the two differ for every SDK below 10.

## Why it is not caught

CI never exercises this path. There is no `setup-dotnet` step and no reference to `dotnet` anywhere under `.github/`, so on the runners `shutil.which` returns `None`, the tests skip, and the suite is green. The failure only appears on a developer machine that has .NET installed at a version below 10 -- where the guard is least likely to be questioned, because the toolchain visibly *is* installed.

The two states are not equivalent from the guard's point of view but produce opposite outcomes:

```
dotnet absent              -> skipif fires  -> green
dotnet present, SDK  < 10  -> skipif passes -> red, with a CalledProcessError
dotnet present, SDK >= 10  -> skipif passes -> green
```

## Scope

[...]

PR #1633, opened 2026-09-02
fix(fingerprint): treat capture selection as parser identity

### LINKED ISSUE #1630: Changing CGR_CAPTURE on an indexed project is silently a no-op: capture selection is missing from the parser fingerprint
Changing `CGR_CAPTURE` on an already-indexed project is silently a no-op. The incremental run reports "already in sync", skips every pass, and the newly enabled edges never appear. Nothing warns that the capture selection changed.

Found while dogfooding cgr on open-source repositories.

## Reproduction

```bash
git clone --depth 1 https://github.com/psf/requests.git
cgr start --repo-path ./requests --update-graph                 # default capture
CGR_CAPTURE="io" cgr start --repo-path ./requests --update-graph # enable io
```

The second run:

```
Knowledge graph already in sync (hash cache matches every file). Skipping passes.
Knowledge graph already in sync for 'requests__6601d2ee' (0.53s, no changes detected)
```

```
Resource nodes: 0      READS_FROM: 0      WRITES_TO: 0
```

The user enabled a feature, the tool said everything was fine, and produced nothing.

## The same repo at a fresh path, so no cache entry exists

```bash
cp -r ./requests ./requests-io
CGR_CAPTURE="io" cgr start --repo-path ./requests-io --update-graph
```

```
requests__6601d2ee     CALLS 1462, IMPORTS 568, ... (14 types, no io edges)
requests-io__33f0cb60  CALLS 1462, IMPORTS 568, ... + READS_FROM 93, WRITES_TO 40
```

Same bytes, same commit, same capture setting. The only difference is that one path had a cache entry and the other did not. 12 `Resource` nodes appear too, and they are good ones -- `resource::ENV::REQUESTS_CA_BUNDLE`, `resource::ENV::NETRC`, `resource::SOCKET::<dynamic>`.

So the io group works correctly on real code. It is unreachable on any project already in the graph.

## Why this is the fingerprint's job, by its own rule

`parser_fingerprint.py` exists precisely for this class of input. Its header:

> the incremental hash cache keys only the source files, so a parser or config [change is missed] ... fingerprint keys the other inputs

`compute_parser_fingerprint` hashes parse-relevant sources, grammar versions, repo frontend inputs, frontend settings and tool versions. The frontend-settings entry carries the criterion for inclusion:

> The active frontend selection changes which edges are produced for unchanged sources (e.g. the C# Roslyn hybrid rewrites INHERITS/IMPLEMENTS), so it is part of the parser identity and must trip the staleness warning.

The capture selection meets that criterion exactly -- it changes which edges are produced for unchanged sources -- and it is not hashed. `grep -n capture codebase_rag/parser_fingerprint.py` matches only `capture_output=True` from a subprocess call.

This looks like an omission from an existing rule rather than a new design question.

## Scope

The direction that silently under-produces is the dangerous one, but the reverse is also unsound: going from `CGR_CAPTURE=io` back to default leaves the io edges in the graph, so the graph then shows edges the current setting says should not exist. Either way the graph stops matching the configuration that produced it, with no warning.

## Suggested fix

Add the resolved capture selection to `compute_parser_fingerprint`, alongside `_frontend_settings()`. The existing staleness path then handles it with no new machinery. Hash the resolved selection rather than the raw `CGR_CAPTURE` string, so `io` and `io,io` do not read as different identities.

## Not part of this issue

`CGR_CAPTURE="default,io"` logs `Ignoring unknown capture token: default`. There is no `default` token, so building on the defaults requires knowing that `io` already means "defaults plus io" -- which it does; I verified the default edge counts are identical with `io` set. Cosmetic next to the above, and worth filing separately if wanted.

## Checked and not a defect

Zero `FLOWS_TO` on requests is correct, not a symptom. FLOWS_TO needs a source-to-sink path inside one function, and a library reads env vars for config without sinking them. Verified with an end-to-end positive control through the same CLI path:

```bash
printf "import os\n\n\ndef leak():\n    x = os.getenv('K')\n    print(x)\n" > flowprobe/m.py
CGR_CAPTURE="io" cgr start --repo-path ./flowprobe --update-graph
```

```
('resource::ENV::K', 'resource::STDOUT::<dynamic>', 'resource')
```

The edge persists to Memgraph through `cgr start`, so the flow pipeline is healthy end to end.

## Environment

```
cgr             0.0.838
repo            vitali87/code-graph-rag at e92a81ad
target          psf/requests at depth-1 HEAD
graph           Memgraph via `cgr daemon up`
command         CGR_CAPTURE="io" cgr start --repo-path <path> --update-graph
```

vitali87:
Closes #1630.

Enabling a capture group on an already-indexed project reported "Knowledge graph already in sync" and emitted nothing. The incremental hash cache keys file contents, so unchanged sources skip every pass, and the capture selection was not part of the parser fingerprint that exists to catch exactly this class of input.

`parser_fingerprint.py` already hashes the frontend mode under a criterion it states in its own comment:

> The active frontend selection changes which edges are produced for unchanged sources ... so it is part of the parser identity and must trip the staleness warning.

The capture selection meets that criterion and was not hashed. This adds it.

## What this does and does not do

It converts a **silent** no-op into a **warned** one. The run still reports "already in sync" and skips passes -- the fingerprint is a staleness *warning*, and #1630 asked for the existing path with no new machinery.

What actually makes a newly enabled group appear is deleting that one repository's `.cgr-hash-cache.json`, `.cgr-dir-mtimes.json` and `.cgr-parser-fingerprint`, then indexing again. Verified end to end:

```
before   requests__6601d2ee   14 relationship types, no io edges
after    requests__6601d2ee   same 14 + READS_FROM 93, WRITES_TO 40
         projects in shared graph: 9 before, 9 after
```

That remedy is now what both the warning and `docs/architecture/language-support.md` prescribe. Previously both pointed at `cgr start --clean --update-graph`, which wipes **every** project in a shared database -- disproportionate when the change only adds edges.

## Why the selection is threaded rather than read from settings

`cli.py:_capture_selection` merges `CGR_CAPTURE` with `--capture` tokens and hands `GraphUpdater` a resolved `CaptureSelection`:

```python
return resolve_capture([*split_spec(settings.CGR_CAPTURE), *(capture or [])])
```

A fingerprint consulting the environment alone would have stayed blind to `cgr start --capture io`, fixing the bug for one entry point and leaving it for the other. `compute_parser_fingerprint` takes the selection and both call sites in `graph_updater.py` pass `self.capture`, so what is stamped is what built the graph.

The RESOLVED selection is hashed rather than the raw spec, so `io,io` and `io` are one identity. Entries are sorted because the selection holds frozensets, whose iteration order is not stable across processes, and counted so a relationship named like a node label cannot move between the two lists unnoticed.

## Sibling sweep

Every `settings.*` read by `codebase_rag/parsers/` and `graph_updater.py`, classified:

| Input | Status |
|---|---|
| `CSHARP_FRONTEND`, `CPP_FRONTEND`, `GO_FRONTEND`, `PYTHON_FRONTEND`, `JAVA_FRONTEND` | hashed (`_frontend_settings`) |
| `LOMBOK_JAR` version | hashed (`LOMBOK=` in `_frontend_settings`) |
| go / javac / dotnet versions | hashed (`_tool_versions`) |
| `compile_commands.json` selection and contents | hashed (`_repo_frontend_inputs`) |
| parser sources, grammar wheel versions | hashed (`_fingerprint_sources`, `_grammar_versions`) |
| capture selection (`CGR_CAPTURE` + `--capture`) | **added here** |
| `CAPTURE_FUNCTION_LOCAL_DEFINITIONS` | **added here** |
| exclusions (`--exclude` / `--unignore`) | not a member: own `.cgr-exclusion-state.json` and `_exclusions_match_last_run` |
| delombok state | not a member: own `.cgr-delombok-state.json` |
| `SKIP_EMBEDDINGS`, `QDRANT_BATCH_SIZE`, `FILE_FLUSH_INTERVAL`, `EMBEDDING_PROGRESS_INTERVAL`, `CGR_HOME` | not a member: none changes which nodes or edges a source file produces |

`CAPTURE_FUNCTION_LOCAL_DEFINITIONS` is the one the sweep turned up. `parsers/class_ingest/mixin.py:126` uses it to pick between two predicates for whether a nested definition is ingested as a method, so flipping it changes which `Method` nodes unchanged sources produce. Same criterion, also unhashed, fixed here with its own test.

## Tests

Seven new tests. The behavioural red was real: before the fix, `CGR_CAPTURE=""` and `CGR_CAPTURE="io"` produced the identical digest `a3ea0db9dadd1bdba593c5835a6c0fab`.

Two of them exercise the updater rather than the hash function, and neither call site can be reverted silently. Verified with frame-scoped mutations that strip the kwarg at exactly one site, each asserting it was applied so a mutation that failed to take could not read as a pass:

- reverting only the stamp site (`graph_updater.py:2705`) fails `test_same_capture_group_twice_does_not_warn`, and that test alone
- reverting only the compare site (`graph_updater.py:2259`) fails **both** new tests

The stamp-site case is the one worth having: run 1 stamps the default digest while running io, run 2 compares the io digest, so the graph warns on **every** subsequent run. A "does not warn" control is the only thing that catches permanent false staleness -- an assertion that it *does* warn is satisfied by both the working and the broken code.

Mutating `_capture_entries` to `return []` fails four of the seven capture tests; `test_same_capture_group_twice_does_not_warn` survives, as a silence control must.

## Not covered

Disabling a group leaves its edges in the graph. The warning now says so and points at `--clean --update-graph` for the case where nodes must be removed rather than added.

## Verification

```
codebase_rag/tests/test_parser_fingerprint.py + test_clean_remedy_strings.py    60 passed
fingerprint + robustness + tool_versions + ast + flow_edges + local_definitions 168 passed
-k "capture or fingerprint or incremental or sync" -n 2                         716 passed, 1 env failure
ruff check .                                                                    All checks passed
```

The one failure is `test_csharp_retrieval_eval.py::test_oracle_captures_first_party_csharp_calls`, environmental and unrelated: `dotnet build` fails with `NETSDK1045` because the oracle targets .NET 10.0 and this box has SDK 8.0.130. Reproduced outside pytest and filed separately as #1632.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parser fingerprints now account for capture selections, helping detect stale indexes when capture settings change.
  * Improved mismatch warnings provide repository-scoped rebuild guidance and clarify when a clean graph update is required.

* **Documentation**
  * Clarified cache invalidation and graph rebuild behavior for capture changes, node renames, and newly generated relationships.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2257,7 +2257,7 @@ class GraphUpdater:
         # unknown (pre-fingerprint) parser: treat it as stale too, without
         # paying for a fingerprint computation that cannot match.
         if stored is None or stored != compute_parser_fingerprint(
-            repo_path=self.repo_path
+            repo_path=self.repo_path, capture=self.capture
         ):
             logger.warning(ls.PARSER_FINGERPRINT_MISMATCH)
 
@@ -2702,7 +2702,9 @@ class GraphUpdater:
         if is_full_build:
             _save_parser_fingerprint(
                 self.repo_path / cs.PARSER_FINGERPRINT_FILENAME,
-                compute_parser_fingerprint(repo_path=self.repo_path),
+                compute_parser_fingerprint(
+                    repo_path=self.repo_path, capture=self.capture
+                ),
             )
 
     def _pre_parse_changed_files(
--- a/codebase_rag/logs.py
+++ b/codebase_rag/logs.py
@@ -876,11 +876,18 @@ HASH_CACHE_ORPHANED = (
 )
 PARSER_FINGERPRINT_SAVE_FAILED = "Failed to save parser fingerprint to {path}: {error}"
 PARSER_FINGERPRINT_MISMATCH = (
-    "Parser code changed since this graph was built. Incremental sync keeps "
-    "results from the old parser for files not touched since the last sync, so "
-    "the graph may be stale. Run 'cgr start --clean --update-graph' to rebuild "
-    "it from scratch. '--clean' on its own deletes the graph without rebuilding "
-    "it, and deletes every project in a shared database, not just this one."
+    "A parser input changed since this graph was built: parser code, a grammar "
+    "or toolchain version, a frontend mode, or the capture selection. "
+    "Incremental sync keeps results from the old inputs for files not touched "
+    "since the last sync, so the graph may be stale. To rebuild THIS repository "
+    "only, delete its '.cgr-hash-cache.json', '.cgr-dir-mtimes.json' and "
+    "'.cgr-parser-fingerprint' and index again with 'cgr start --update-graph': "
+    "every file is then treated as new, and other projects in a shared database "
+    "are untouched. That adds what the new inputs emit but does not remove nodes "
+    "the old ones left behind, so when names or node kinds changed rather than "
+    "edges being added, run 'cgr start --clean --update-graph' instead. "
+    "'--clean' on its own deletes the graph without rebuilding it, and deletes "
+    "every project in a shared database, not just this one."
 )
 
 REHYDRATE_QUERY_FAILED = (
--- a/codebase_rag/parser_fingerprint.py
+++ b/codebase_rag/parser_fingerprint.py
@@ -10,11 +10,15 @@ import subprocess
 from collections.abc import Callable
 from importlib import metadata
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from loguru import logger
 
 from . import constants as cs
 from . import logs as ls
+
+if TYPE_CHECKING:
+    from .capture import CaptureSelection
 
 _FILE_HASH_CHUNK_SIZE = 1024 * 1024
 
@@ -190,7 +194,9 @@ def _tool_versions(repo_path: Path | None = None) -> list[str]:
 
 
 def compute_parser_fingerprint(
-    package_root: Path | None = None, repo_path: Path | None = None
+    package_root: Path | None = None,
+    repo_path: Path | None = None,
+    capture: "CaptureSelection | None" = None,
 ) -> str:
     root = package_root if package_root is not None else Path(__file__).resolve().parent
     hasher = hashlib.md5(usedforsecurity=False)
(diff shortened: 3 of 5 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

PR #1653, opened 2026-09-02
fix: do not paste a failed advisory API response into release notes

vitali87:
## Problem

The v0.0.845 release shipped with this as its entire Security section:

```
## 🔒 Security

{"message":"Resource not accessible by integration","documentation_url":"...","status":"403"}

**Upgrading is strongly recommended.**
```

`gh api` writes its HTTP error body to **stdout**, so `2>/dev/null || true` suppressed the stderr and the exit code but still captured the 403 JSON into `DETAIL`. The following `[ -n "$DETAIL" ]` test then passed, and the error blob was published as the advisory detail line. The intended fallback ("see advisory GHSA-…") was unreachable whenever the fetch failed.

This is not a rare edge case: `GITHUB_TOKEN` can never read a **draft** advisory, and an advisory is normally still a draft when the security release is cut, so the fetch fails on essentially every security release. v0.0.670 avoided it only because its Security section was replaced by hand.

## Fix

```diff
-              2>/dev/null || true)
+              2>/dev/null) || DETAIL=""
```

`gh api` exits non-zero on 403, so `|| DETAIL=""` clears the captured body and the existing GHSA-pointer fallback runs as designed.

## Verification

Reproduced with a stub `gh` that prints a 403 body to stdout and exits 1, run under Actions' real invocation (`bash --noprofile --norc -eo pipefail`):

- old form → raw 403 JSON in the Security section
- new form → `This release contains a security fix; see advisory [GHSA-…](…).`
- success path → detail line emitted verbatim, unchanged
- 404, jq filter error, and exit-0-with-empty-stdout all fall back correctly
- errexit unaffected: `V=$(...) || V=""` is a `||` list, suppressing `-e` exactly as `|| true` did
- `/tmp/security.md` stays non-empty in all branches, so the section is never dropped

v0.0.845's notes have already been corrected by hand; this prevents a recurrence.

## Known limitation

This fixes the symptom, not the ordering. The workflow reads advisory details at a moment when the advisory is usually still a draft and therefore unreadable, so the enriched section will rarely appear — you get the pointer line instead. Regenerating the notes on the advisory `published` event would be the real fix; out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved security notice generation when security-advisory details cannot be retrieved, ensuring an appropriate fallback notice is used.
  * Added validation to confirm the fallback notice displays correctly when advisory details are unavailable.

### DIFF
--- a/.github/workflows/version-bump.yml
+++ b/.github/workflows/version-bump.yml
@@ -176,7 +176,7 @@ jobs:
           if [ -n "$GHSA" ]; then
             DETAIL=$(gh api "repos/${GITHUB_REPOSITORY}/security-advisories/${GHSA}" \
               --jq '"\(.summary) (\(.severity) severity, CVSS \(.cvss.score // "n/a")), advisory [\(.ghsa_id)](https://github.com/'"${GITHUB_REPOSITORY}"'/security/advisories/\(.ghsa_id))\(if .cve_id then " / " + .cve_id else "" end)."' \
-              2>/dev/null || true)
+              2>/dev/null) || DETAIL=""
           fi
           {
             echo "## 🔒 Security"
--- a/codebase_rag/tests/test_release_news_derivation.py
+++ b/codebase_rag/tests/test_release_news_derivation.py
@@ -11,8 +11,10 @@
 
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 
+import pytest
 import yaml
 
 _WORKFLOW = (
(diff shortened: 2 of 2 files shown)


vitali87:
Adopted by the coordinating session: no session claimed this branch, so it was rebased onto current main (content unchanged, one-line diff) and pushed with --force-with-lease to clear the stale CI failure from the pre-#1662 base. Bots re-triggered on the new head.

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

PR #1656, opened 2026-09-02
fix: publish the hash cache atomically with its stamp (#1646)

### LINKED ISSUE #1646: Hash cache is published unstamped, then backdated, so an interrupted run can leave it hiding edits
The hash cache is written and then backdated by a separate `os.utime` call. A process that stops between the two leaves the cache carrying its physical write time, which is exactly the state #1615's restamp exists to prevent.

## The window

At the commit point in `run()`:

```python
_save_hash_cache(*self._pending_hash_cache)     # file now has its WRITE time
_save_dir_mtimes(*self._pending_dir_mtimes)
...
os.utime(path, (observed_at, observed_at))      # backdated to the observed instant
```

Between those calls the cache on disk is a fully valid, loadable cache whose mtime is later than any edit made while the run was hashing. If the process stops there, the successor loads it, compares `file_mtime <= cache_mtime` for each cached file, and skips the edited one. The graph keeps pre-edit content until something else advances that file's mtime.

The window is small but it is the same window the branch already treats as worth closing: #1615 moved the cache write after the flush for a crash that is no more likely than this one, and the restamp itself exists because a late stamp hides edits.

## Suggested fix

Publish atomically: write each cache to a temporary path in the same directory, `os.utime` the temporary file to `observed_at`, and `os.replace` it over the live path only once both have succeeded. `os.replace` is atomic within a filesystem, so a successor either sees the previous cache or a fully stamped new one, never an unstamped new one.

That also simplifies the failure handling this introduced. Today a failed `os.utime` means the live cache is already mis-stamped and must be removed (see #1644 for the crash that removal caused); with a temporary file, a failed stamp means discarding the temporary and leaving the previous cache untouched, which needs no unlink of live state at all.

A regression test should interrupt between the save and the stamp and assert the successor does not skip a file edited during the run.

## Provenance

Found by CodeRabbit as an outside-diff comment on #1635. Verified against the merged code: the two operations are separate statements at the commit point with no atomicity between them. Filed rather than fixed in that PR because it is outside its diff and the temporary-file rewrite is a larger change than the one under review.

vitali87:
Closes #1646.

## The defect

The hash cache was written and then stamped as two separate filesystem
operations:

```python
_save_hash_cache(*self._pending_hash_cache)   # visible immediately
...
os.utime(path, (observed_at, observed_at))    # a separate syscall later
```

Between those two calls the cache is on disk carrying its OWN write time
rather than the instant observed before hashing. `_is_already_in_sync` skips
any file whose mtime is at or below the cache's, so a run that stops in that
window hands its successor a cache that hides every edit made while the run
was hashing. The stamp exists precisely to prevent that, and the gap
republishes the bug it fixes.

## The fix

`_publish_hash_cache` builds the whole artifact on a `.<pid>.tmp` sibling —
write, then `os.utime` — and `os.replace`s it into place. `os.replace` is
atomic within a filesystem, so a successor sees either the previous cache or
a fully stamped new one, never an unstamped new one.

On any failure the temporary is removed and the PREVIOUS cache is left
untouched. That is the safe direction: a cache one run out of date costs a
rescan, a mis-stamped one loses an edit.

## Publish order at the commit point

The hash cache is now published FIRST and the directory mtimes LAST. A stop
between the two leaves a mismatched pair that is only detected in one
direction:

- **Stale map, fresh cache** (the order here): every entry still holds its old
  mtime, so an addition surfaces and the next run reconciles it.
- **Fresh map, stale cache** (the previous order): the map records every
  directory as current, so nothing is compared, and the file loop only walks
  keys the hash cache already names. A file added in the gap is never indexed.

### What actually detects an addition

The comment justifying that order has been wrong twice, each version correct
for the one shape probed and false as a general claim. The rule that survived
all shapes tried: `_is_already_in_sync` iterates only the entries the map
RECORDS, so it never reaches a directory with no entry. What catches an
addition is the recorded entry for the directory that DIRECTLY CONTAINS it,
whose stored mtime no longer matches disk.

Measured by refreshing one entry at a time — refreshing the containing entry
alone hides the addition, refreshing any other leaves it visible:

| Shape | Containing entry | Result |
|---|---|---|
| New subdirectory | the parent's entry | refresh parent → hidden; refresh others → visible |
| File added to a directory already recorded | that directory's own entry (parent contributes nothing) | refresh it → hidden; refresh others → visible |
| File added at the repo root | `'.'` (recorded keys were `['.']`) | stale `'.'` → `in_sync=False`; refresh `'.'` only → `in_sync=True` |

Each prediction was written before the probe ran, and all three matched.

## Tests

- `test_edit_during_the_deferred_window_is_not_skipped` — retargeted to
  `_publish_hash_cache`; the wrapper sleeps, then edits, and the edit must
  survive into the next run.
- `test_a_read_only_tree_still_completes_the_run` — temp write, `utime` and
  `replace` all refuse with `EROFS`; the run still finishes.
- `test_a_temp_that_cannot_be_cleaned_up_does_not_end_the_run` — `os.replace`
  refuses `EROFS`, then the `.tmp` unlink refuses `EACCES`. Asserts the run
  completes, the existing cache is byte-identical, both sites were reached
  (`reached == {"replace", "cleanup"}`), and the warning names the CLEANUP
  failure ("Permission denied") rather than the publish failure. That last
  assertion pins a regression this rewrite introduced and round 2 caught: an
  `except OSError:` with no `as` reported `e`, naming a reason unrelated to
  the removal.
- `test_a_stop_between_the_two_cache_publishes_is_still_detected` — patches
  both publishers, lets the first through and stops the second; asserts
  `order == ["hash_cache", "dir_mtimes"]` and `_is_already_in_sync() is False`.
  Keyed on whichever publish fires second rather than on a function name: an
  earlier version keyed by name passed with the order reversed.

Warnings are asserted through a loguru sink, not `caplog` — this project logs
via loguru, and `caplog` captures none of it.

## Logging

`CACHE_STAMP_FAILED` is deleted (no callers remain; verified with a grep whose
command shape returns hits for the sibling constant, so the empty result is a
real absence rather than a broken search). `CACHE_STAMP_CLEANUP_FAILED` is
retargeted to the temp-file removal and now says what is true of a `.tmp`
leftover: it is inert, because nothing reads that path.

## Suite

`9050 passed, 257 errors`, measured before the rebase onto current `main`.
The rebase changed no content (all four blobs are byte-identical and the two
diffs compare equal), and `main` had not touched any of these four files, so
the result carries. All 257 are Docker-unavailable SETUP errors in
`@memgraph-integration`:

```
_ ERROR at setup of TestCachelessRebuild.test_rebuild_without_cache_drops_renamed_symbols _
self = <docker.transport.unixconn.UnixHTTPConnectionPool ...>
method = 'GET', url = '/version'
```

`origin/main` on the same host and the same configured checkout shows an
identical `9050 passed, 257 errors, 257 memgraph-tagged, 0 untagged`, so this
branch introduces none of them.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved incremental updates by publishing cache changes atomically, reducing the risk of incomplete or corrupted cache state.
  - Runs now complete gracefully when cache files cannot be written or updated, including in read-only environments.
  - Interrupted cache updates are detected on the next run, allowing the system to recover and resynchronize safely.
  - Temporary cache files left after a failed update are treated as inactive and can be removed without affecting normal operation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ evals/results/**/agentic_qa*records*.jsonl
 evals/results/logs/
 .cgr-parser-fingerprint
 .cgr-exclusion-state.json
+.cgr-*.tmp
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -268,8 +268,13 @@ def _save_hash_cache(cache_path: Path, hashes: FileHashCache) -> None:
 
 def _publish_hash_cache(
     cache_path: Path, hashes: FileHashCache, observed_at: float | None
-) -> None:
+) -> bool:
     """Write the hash cache and its timestamp as one visible step.
+
+    Returns True when the cache is published, False when it is not. The
+    caller needs that answer: the directory-mtime map must not advance past
+    a hash cache that failed to publish, or the pair on disk becomes the
+    fresh-map/stale-cache combination that hides an addition entirely.
 
     `_is_already_in_sync` skips any file whose mtime is at or below this
     file's, so a cache carrying its own write time hides every edit made
@@ -293,6 +298,7 @@ def _publish_hash_cache(
             os.utime(tmp_path, (observed_at, observed_at))
         os.replace(tmp_path, cache_path)
         logger.info(ls.HASH_CACHE_SAVED, count=len(hashes), path=cache_path)
+        return True
     except OSError as e:
         logger.warning(ls.HASH_CACHE_SAVE_FAILED, path=cache_path, error=e)
         try:
@@ -307,6 +313,7 @@ def _publish_hash_cache(
             logger.warning(
                 ls.CACHE_STAMP_CLEANUP_FAILED, path=tmp_path, error=cleanup_error
             )
+        return False
 
 
 def _load_parser_fingerprint(stamp_path: Path) -> str | None:
@@ -1332,13 +1339,21 @@ class GraphUpdater:
         # compared, and the file loop only walks keys the hash cache already
         # names, so a file added in the gap is never indexed. Publishing the
         # hash cache first leaves the harmless window.
+        # A failed publish leaves the PREVIOUS hash cache on disk, so advancing
+        # the map anyway would build exactly the fresh-map/stale-cache pair the
+        # ordering above exists to avoid: nothing would be compared, and a file
+        # added during the run would never be indexed. Withholding the map
+        # keeps both artefacts one run behind, which only costs a rescan.
+        hash_cache_published = True
         if self._pending_hash_cache is not None:
             cache_path, new_hashes = self._pending_hash_cache
-            _publish_hash_cache(cache_path, new_hashes, observed_at)
+            hash_cache_published = _publish_hash_cache(
+                cache_path, new_hashes, observed_at
+            )
             self._pending_hash_cache = None
-        if self._pending_dir_mtimes is not None:
+        if self._pending_dir_mtimes is not None and hash_cache_published:
             _save_dir_mtimes(*self._pending_dir_mtimes)
-            self._pending_dir_mtimes = None
+        self._pending_dir_mtimes = None
         self._pending_cache_observed_at = None
 
         if self._single_file is None:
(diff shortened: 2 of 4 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

PR #1689, opened 2026-09-03
fix: sweep deleted-file state on recorded module qns, not path-derived (#1676)

### LINKED ISSUE #1676: remove_file_from_state keys its registry sweep on a path-derived qn that ignores the mod.rs collapse
## Defect

`GraphUpdater.remove_file_from_state` derives `module_qn_prefix` from the file path (`project.<parts>`) without the `mod.rs` collapse that `base_module_qn` applies (`path_utils.py`, the `MOD_RS` branch). For `src/a/mod.rs` the recorded module qn is `proj.src.a` but the prefix computed here is `proj.src.a.mod`, so the function-registry prefix sweep and the `commonjs_direct_exports.pop` in that function are keyed on a qn nothing recorded. The Rust import-state block below it already re-derives the recorded qn from `module_qn_to_file_path` for exactly this reason, and its comment says so; the two earlier sweeps in the same function do not.

## Effect

Deleting or re-parsing a `mod.rs` on a reused updater (watcher, MCP server) leaves its function-registry entries and simple-name lookup entries in place, so Pass 3 can still resolve calls to definitions that no longer exist.

## Fix

Key every sweep in `remove_file_from_state` on the qns the parse recorded for that path (the `module_qn_to_file_path` entries), falling back to the path-derived prefix only when the file was never parsed. Add a test that reuses one updater across a `mod.rs` deletion and asserts the registry no longer holds its definitions.

Found by the local review round on #1576.

vitali87:
Closes #1676.

## What

`remove_file_from_state` derived its sweep prefix from the file path, special-casing only `__init__.py` and so missing the `mod.rs` collapse that `base_module_qn` applies. For `src/a/mod.rs` the parse records `proj.src.a` but the sweep used `proj.src.a.mod`:

| path | recorded qn | path-derived prefix | |
|---|---|---|---|
| `src/a/mod.rs` | `proj.src.a` | `proj.src.a.mod` | **mismatch** |
| `src/a.rs` | `proj.src.a` | `proj.src.a` | ok |
| `pkg/__init__.py` | `proj.pkg` | `proj.pkg` | ok (already special-cased) |

So the function-registry prefix sweep and the `commonjs_direct_exports.pop` were keyed on a qn nothing had recorded, and `mod.rs` was the only shape affected. Deleting or re-parsing one on a reused updater (watcher, MCP server) left its registry and simple-name entries in place for Pass 3 to keep resolving calls into definitions that no longer exist.

## Fix

Both sweeps now key on the qns `module_qn_to_file_path` recorded for that path — which is exactly what the Rust and C# import-state blocks a few lines below already do, and why their comments say so. `touched_modules` was computing that identical set separately, so it now reuses it.

The path-derived form survives only as the fallback for a file that was never parsed: it recorded no qn, so it owns no state these sweeps could wipe. On the real deletion path this fallback is not reached — `reingest` calls `_seed_module_qns_from_graph` for the gone paths first, so the recorded qn is present.

## Verification

`test_a_deleted_mod_rs_is_swept_on_its_recorded_qn` seeds the recorded qn directly rather than parsing a real `mod.rs`. The collapse is a property of `remove_file_from_state`, and driving it through a parse makes the test **silently skip** wherever the Rust grammar is absent — which is any Python-only venv, where it would report as a pass while never running.

Pinned both ways against a fix-reverted copy, on an interpreter with all 15 grammars loaded:

- with the fix: **12 passed**
- reverted: the new test **fails** (`the deleted mod.rs left registry entries behind`)

Also checked, with the disambiguation shape (`src/a.rs` owning `proj.src.a` beside a `mod.rs`): the multi-prefix sweep cannot over-remove another file's qns — the `foreign_qns` ownership guard (#1025) still holds in both deletion orders, and the reverted code fails both.

`ruff check` / `ruff format --check`: clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale in-memory entries remaining after deleting files whose recorded module identifiers differ from their path-derived identifiers.
  * Improved cleanup for collapsed module files, disambiguated identifiers, CommonJS exports, and related function lookups.

* **Tests**
  * Added coverage verifying that deleted module files are fully removed from function registries and name lookups.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2217,10 +2217,29 @@ class GraphUpdater:
             if file_path.name == cs.INIT_PY
             else relative_path.with_suffix("").parts
         )
-        module_qn_prefix = cs.SEPARATOR_DOT.join([self.project_name, *path_parts])
-        self.factory.import_processor.commonjs_direct_exports.pop(
-            module_qn_prefix, None
-        )
+        path_derived_qn = cs.SEPARATOR_DOT.join([self.project_name, *path_parts])
+        # Sweep on the qns the parse RECORDED for this path, not on the
+        # path-derived form. The two differ whenever `base_module_qn` applies
+        # its mod.rs collapse (src/a/mod.rs records as proj.src.a, while the
+        # path yields proj.src.a.mod) or `_disambiguate_module_qn` suffixes a
+        # qn an earlier-parsed sibling already owns. Keying on the path-derived
+        # form swept a qn nothing ever recorded, so a mod.rs deletion left its
+        # registry and simple-name entries behind and Pass 3 kept resolving
+        # calls into definitions that no longer exist. The Rust and C# import
+        # blocks below already re-derive the recorded qn for this reason.
+        # A file that was never parsed recorded nothing; the path-derived form
+        # is the only prefix available for it, and it owns no recorded qn that
+        # this could wipe.
+        recorded_qns = {
+            qn
+            for qn, path in (
+                self.factory.definition_processor.module_qn_to_file_path.items()
+            )
+            if path == file_path
+        }
+        module_qn_prefixes = recorded_qns or {path_derived_qn}
+        for prefix in module_qn_prefixes:
+            self.factory.import_processor.commonjs_direct_exports.pop(prefix, None)
         # A deleted file is no writer: its mod-scope registry entries must
         # not weigh in the next arbitration (a modified file re-drops this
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@coderabbitai review

PR #1690, opened 2026-09-03
fix: give the scoped re-ingest its own cache-stamp log constants

vitali87:
## Problem

Main fails the type check: `ty` reports `Module codebase_rag.logs has no member CACHE_STAMP_FAILED`. #1656 removed `CACHE_STAMP_FAILED` and reworded `CACHE_STAMP_CLEANUP_FAILED` for the atomic hash-cache publish, while #1538's `GraphUpdater._reingest_update_hashes` still logs both when backdating the cache after a scoped re-ingest. The two merged independently, so no PR saw the conflict; every PR's pre-commit and Type Check job now fails on it.

## Fix

The re-ingest backdate step names its own two constants, `REINGEST_CACHE_STAMP_FAILED` and `REINGEST_CACHE_STAMP_CLEANUP_FAILED`, with the wording #1538 shipped; the publish path keeps its own. No behaviour change.

## Verification

- pre-commit on the branch, including `ty check`: passed.
- `codebase_rag/tests/test_reingest.py`: 68 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved warning messages for cache timestamping and cleanup failures during scoped re-ingest.
  - Clarified when cache entries are removed after timestamping failures or may affect subsequent update runs.
  - Added coverage for scenarios where cache timestamping or cleanup operations fail.

PR #1693, opened 2026-09-03
fix: let a run finish and rebuild when an orphaned cache cannot be discarded (#1647)

### LINKED ISSUE #1647: Discarding an orphaned hash cache crashes the run on a read-only path
`_drop_cache_if_graph_lost` unlinks both cache files without guarding `OSError`, so on a read-only repository path the discard raises out of `GraphUpdater.run()` instead of degrading.

## The code

```python
logger.warning(ls.HASH_CACHE_ORPHANED.format(project=self.project_name))
cache_path.unlink(missing_ok=True)
(self.repo_path / cs.DIR_MTIMES_FILENAME).unlink(missing_ok=True)
```

Neither call is inside a `try`. On `main` these are lines 2309-2310.

## Why it matters

Every other filesystem writer on this path swallows `OSError` — `_touch_empty_json`, `_save_hash_cache`, `_save_dir_mtimes`, `_save_exclusion_state` and `_save_parser_fingerprint` all log and return. So a read-only repository reaches this function normally, having degraded quietly everywhere else, and then crashes here.

The condition is reachable: with the graph reporting zero modules for the project on a read-only path, the run raises `OSError: [Errno 30] Read-only file system` out of `run()`. The user loses the whole indexing run because the tool could not delete a cache file it only wanted to discard.

Discarding is already best-effort in intent — `missing_ok=True` says a cache that is not there is fine. A cache that cannot be removed should be the same: log it and continue, since the next run re-reads the graph and reaches the same conclusion.

## Suggested fix

Wrap both unlinks in `try/except OSError`, log that the stale cache could not be discarded, and continue. The same shape as the guard added in #1644 for the restamp cleanup path.

A regression test can refuse `Path.unlink` with `EROFS` for the cgr cache files and assert the run completes, with a guard asserting the refusal was actually reached so the test cannot pass without exercising the path.

## Provenance

Introduced by `6279102c` ("discard the hash cache when the graph no longer holds the project"), unrelated to #1615 or #1635. Found by a local reviewer while checking whether #1644's fix had siblings elsewhere on the same path, and verified reachable on the merged `main`. Filed separately because it is outside #1644's diff and predates it.

vitali87:
Closes #1647.

## The crash

`_drop_cache_if_graph_lost` ended with two unguarded `unlink` calls:

```python
cache_path.unlink(missing_ok=True)
(self.repo_path / cs.DIR_MTIMES_FILENAME).unlink(missing_ok=True)
```

On a read-only tree that raises `OSError: [Errno 30]` straight out of
`GraphUpdater.run()`. Every other filesystem writer on that path
(`_touch_empty_json`, `_save_hash_cache`, `_save_dir_mtimes`,
`_save_exclusion_state`, `_save_parser_fingerprint`) logs and returns, so such
a tree reaches this function having degraded quietly everywhere else, then
loses the whole indexing run to a file the tool only wanted to delete.

## The part the issue did not predict

Guarding the unlinks alone is not enough, and the first version of this branch
was wrong in a way worth writing down.

Refusing to DELETE the cache must not resurrect TRUSTING it. The incremental
pass short-circuits on mtime before any hash comparison:

```python
if file_mtime <= cache_mtime:
    new_hashes[file_key] = old_hashes[file_key]
    continue
```

A real completed run stamps the cache at or after its sources
(`os.utime(observed_at)` at the commit point), so a cache that survives the
discard makes every file look unchanged against hashes already known dead. The
run indexes nothing. The graph stays empty, so the next run reaches the same
discard, fails the same way, and indexes nothing again — permanently.

Measured, three consecutive runs on a read-only tree with an orphaned cache:

| Cache mtime | Discard | run1 | run2 | run3 |
|---|---|---|---|---|
| Realistic (>= sources) | refused | `[]` | `[]` | `[]` |
| Realistic | allowed (control) | `['proj.m']` | `['proj.m']` | — |
| Backdated 60s | refused | `['proj.m']` | — | — |

The control isolates the cause: the read-only tree parses fine, so the mtime
fast path is the only thing stopping it.

So the fix has two halves: guard the unlinks, and set
`_cache_discarded_in_memory` so this run ignores the cache it could not delete.
That gives the run the full rebuild the discard intended.

The flag is reset immediately BEFORE the discard that sets it, not with the
other per-run resets further down `run()` — those execute after
`_drop_cache_if_graph_lost` and would clear the flag just raised, making the
whole fix dead code.

## How the fixture nearly hid it

The first version of the test backdated the cache mtime 60s, copying the
sibling test in the same file. That made it pass while production indexed
nothing: the table above is exactly the difference between the two fixture
choices. The test now stamps the cache AFTER the sources, which is what a real
run leaves behind, and the comment says why so nobody re-introduces the
backdate.

Credit where due: the local reviewer asked whether the backdating was masking
a product bug rather than arranging a fixture. It was.

## Discrimination

Every mutation of the production code turns this test red:

| Mutation | Result |
|---|---|
| Remove the guard entirely | red |
| Guard only the FIRST unlink, leave the second bare | red |
| Swap the loop order | red |
| Drop the `logger.warning`, swallow silently | red |
| Keep the guard, ignore `_cache_discarded_in_memory` | red |

The last two matter most. Without the log assertions, deleting the warning line
kept the test green, so nothing pinned the new constant. Without the realistic
mtime, the in-memory discard could be removed and the test still passed.

## Note on CI

`ty check` fails on this branch with `Module codebase_rag.logs has no member
CACHE_STAMP_FAILED` at `graph_updater.py:3574`. That is pre-existing on `main`
(the reference is there and the constant is defined zero times) and is being
fixed by #1690; it is unrelated to this diff and will clear once that merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when stale cache files cannot be removed.
  * The current run now ignores undeletable cache data and rebuilds instead of failing or using stale results.
  * Added warnings identifying cache cleanup failures while allowing indexing to continue.
  * Improved accuracy of type-annotation information for unchanged, re-parsed, and removed files.
  * Single-file indexing now preserves existing cache data and avoids unnecessary project-wide cleanup.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -479,6 +479,10 @@ class GraphUpdater:
         # them: the reconciliation it was meant to do may not have happened,
         # so that run must not stamp its exclusion set as reconciled.
         self._graph_state_unknown: bool = False
+        # Set when an orphaned cache could not be deleted: the file is still on
+        # disk, so this run must ignore it in memory rather than trust it
+        # (issue #1647).
+        self._cache_discarded_in_memory: bool = False
         # Written at the post-flush commit point in `run`, never inside
         # `_process_files` (issue #1615).
         self._pending_hash_cache: tuple[Path, FileHashCache] | None = None
@@ -1178,6 +1182,10 @@ class GraphUpdater:
         )
         logger.info(ls.ENSURING_PROJECT, name=self.project_name)
 
+        # Reset BEFORE the discard below, which is what SETS it: a reset after
+        # that call would clear the flag it just raised and make the in-memory
+        # discard dead code (issue #1647).
+        self._cache_discarded_in_memory = False
         if not force and self._single_file is None:
             self._drop_cache_if_graph_lost()
             self._warn_if_parser_changed()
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

Issue #1641, opened 2026-09-02
Method calls named like a type become INSTANTIATES edges: err.Error() reports as constructing gin's Error struct


vitali87:
A method call whose **name** matches a first-party type name is recorded as an `INSTANTIATES` edge to that type, even when the receiver is an unrelated type from the standard library. On gin, `err.Error()` on a stdlib `error` produces `INSTANTIATES -> gin.errors.Error`, and `buf.String()` on a `bytes.Buffer` produces `INSTANTIATES -> gin.render.text.String`.

The result is that "what constructs `X`" returns functions that never construct `X`. Project-wide there are **27** `INSTANTIATES` edges to `errors.Error` against **6** genuine `Error{` composite literals in the whole repository.

Go is unusually exposed because the two most idiomatic interface methods in the language, `error.Error()` and `fmt.Stringer.String()`, have exactly the names projects give their own types. I have not established whether the resolver behaves the same in other languages; the evidence below is all Go.

Found while dogfooding cgr on open-source repositories.

## Reproduction

```bash
git clone --depth 1 https://github.com/gin-gonic/gin.git
cgr start --repo-path ./gin --update-graph
```

```cypher
MATCH (a)-[:INSTANTIATES]->(b)
WHERE b.qualified_name ENDS WITH '.errors.Error'
RETURN split(a.qualified_name,'.')[1] AS module, count(*) AS c ORDER BY c DESC
```

## The cleanest case: a package with zero literals and nine edges

The `render/` package contains **no** `Error{` composite literal anywhere (`grep -rn "Error{" render/` returns nothing), yet it produces **9** `INSTANTIATES` edges to `gin.errors.Error`. It contains exactly **9** `err.Error()` calls, and the per-function multiplicity matches one-for-one:

| function | `.Error()` calls in source | `Error{` literals | INSTANTIATES edges |
|---|---|---|---|
| `render_test.TestRenderJsonpJSONError` | 4 (lines 195, 202, 209, 216) | 0 | **4** |
| `render_test.TestRenderBSONWriteError` | 1 (441) | 0 | **1** |
| `render_test.TestRenderXMLError` | 1 (466) | 0 | **1** |
| `render_test.TestRenderDataError` | 1 (582) | 0 | **1** |
| `render_test.TestRenderWriteError` | 1 (806) | 0 | **1** |
| `render_msgpack_test.TestRenderMsgPackError` | 1 | 0 | **1** |

[...]

Issue #1639, opened 2026-09-02
Oracle skip guards check for a binary on PATH, not a working toolchain: 24 failures + 4 errors on a clean main


vitali87:
The oracle tests already carry skip guards for a missing toolchain, and every one of them is `shutil.which(...) is not None`. That answers "is this binary on PATH", not "can this toolchain run the oracle", so a toolchain that is present but too old, or a launcher shim whose component is not installed, sails past the guard and the test hard-fails instead of skipping.

On a machine with Node 18, .NET SDK 8 and a rustup `rustfmt` shim, that is **24 failures and 4 errors** in the unit suite on a clean `origin/main`, none of them related to any code change. CI is green because its runners happen to have newer toolchains.

Found while checking whether local suite failures were attributable to #1636. They were not, but establishing that took a full revert-and-compare run, which is the cost this defect imposes on every such check.

## Reproduction

```bash
git checkout main
uv run pytest codebase_rag/tests/test_csharp_retrieval_eval.py \
              codebase_rag/tests/test_csharp_structure_oracle.py \
              codebase_rag/tests/test_patcher.py \
              codebase_rag/tests/test_ruby_structure_oracle.py -q -n 2
# 24 failed, 20 passed, 1 skipped, 4 errors
```

## The three guards and how each fails open

| Guard | Location | Sees | Misses |
|---|---|---|---|
| `node_oracle_available()` | `evals/oracles/_common.py:29` | `node` and `npm` on PATH | that Node 18 cannot `require()` an ESM-only package |
| `csharp_oracle_available()` | via `test_csharp_structure_oracle.py:64` | `dotnet` on PATH | that SDK 8.0 cannot target the oracle's `net10.0` |
| `shutil.which(cs.PATCH_RUSTFMT)` | `codebase_rag/tests/test_patcher.py:429` | `/root/.cargo/bin/rustfmt` exists | that it is a rustup shim for an uninstalled component |

**Ruby — 8 failures.** `_require_ruby` skips on `node_oracle_available()`, which passes. Then:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../node_modules/@ruby/prism/src/index.js
from .../evals/oracles/ruby_oracle/ruby_ast.js not supported.
Node.js v18.19.1
```

**C# — 15 failures and 4 setup errors.** `_require_csharp` skips on `csharp_oracle_available()`, which passes. Then:

```
[...]

PR #1694, opened 2026-09-03
fix: a nested class is not a self-dispatch target, so self.Inner() emits no CALLS to it (#1650)

### LINKED ISSUE #1650: self.Inner() emits an invalid (Method)-[:CALLS]->(Class) edge alongside the correct __init__ call
Constructing a nested class through `self` emits an extra `CALLS` edge pointing at the **Class** node, which the schema does not allow. `CALLS` may target a Function, Method, Enum or Type — never a Class; construction is what `INSTANTIATES` is for, and the correct `CALLS` edge to `__init__` is emitted alongside it.

`codebase_rag/types_defs.py:1052`:

```python
RelationshipSchema(
    (NodeLabel.MODULE, NodeLabel.FUNCTION, NodeLabel.METHOD),
    RelationshipType.CALLS,
    (NodeLabel.FUNCTION, NodeLabel.METHOD, NodeLabel.ENUM, NodeLabel.TYPE),
),
```

The repo's own test-suite audit already rejects it, so any test exercising this shape aborts:

```
AssertionError: (Method)-[:CALLS]->(Class) is not documented in RELATIONSHIP_SCHEMAS
```

(`codebase_rag/tests/conftest.py:497`, `AuditCheck.UNDOCUMENTED_RELATIONSHIP`.)

Pre-existing on `main`; found while adding tests for #1641.

## Minimal fixture

```python
class Outer:
    class Inner:
        def __init__(self, v):
            self.v = v

    def make(self):
        return self.Inner(1)


def via_instance():
    o = Outer()
    return o.Inner(3)
```

Edges emitted:

```
(Function) via_instance     -> (Method) Outer.Inner.__init__     CALLS          correct
(Function) via_instance     -> (Class)  Outer.Inner              INSTANTIATES   correct

(Method)   Outer.make       -> (Method) Outer.Inner.__init__     CALLS          correct
(Method)   Outer.make       -> (Class)  Outer.Inner              INSTANTIATES   correct
(Method)   Outer.make       -> (Class)  Outer.Inner              CALLS          INVALID
```

`via_instance`, whose receiver is a typed local, is clean and produces exactly the correct pair. Only the `self.`-receiver path emits the third edge, and it is a duplicate: the correct `CALLS -> __init__` is already there beside it.

## Why it has gone unnoticed

No test in the suite constructs a nested class through `self`, so the audit never fires. Adding such a test is what surfaced it — I had to drop that shape from the #1641 fixture and assert on the resolver predicate instead, because an end-to-end assertion fails on this unrelated edge.

## Effect

`CALLS` is what "who calls this" and reachability walk. An edge into a Class node is not answerable as a call, and `dead_code.py` traverses `CALLS` for reachability, so the duplicate is at best inert and at worst double-counts a construction as both a call and an instantiation.

## Suggested direction

The `self.`-receiver path should take the same constructor redirect as every other class callee: `call_processor.py:3947` sets `callee_type = METHOD` and retargets to `__init__` when it exists. `via_instance` proves that redirect works; the `self.` path appears to emit the raw class edge in addition to it rather than instead of it.

## Not assumed

I have not located the emit site for the extra edge and am not claiming the redirect is bypassed rather than duplicated — only that the observable result is one valid pair plus one invalid edge, and that a receiver-typed local in the same file produces the pair alone.

## Environment

```
cgr        0.0.842
repo       vitali87/code-graph-rag at main 81c35e5f
harness    codebase_rag/tests/conftest.py create_and_run_updater, CGR_AUDIT_SWEEP=1 to observe rather than abort
```

vitali87:
Closes #1650.

`self.Inner()` — constructing a nested class — emitted a `CALLS` edge pointing at the **Class** node, alongside the correct `CALLS -> __init__` and `INSTANTIATES`. `RELATIONSHIP_SCHEMAS` allows `CALLS` to target a Function, Method, Enum or Type, never a Class; construction is what `INSTANTIATES` records. So the edge was both a duplicate and undocumented, and the suite's own `_audit_recorded_graph` rejects it.

## Root cause

`self_dispatch_targets` fans `self.M()` out to the enclosing class's `M` and every concrete override, so a method reached only polymorphically is not reported dead. For `self.Inner(1)` the name resolves through `_try_resolve_method` to `Outer.Inner`, which is a **Class** node, and the fan-out emitted a `CALLS` edge to it like any other dispatch target.

The fix drops Class-labelled targets from that fan-out. Construction already reaches the graph through the paths beside it.

## Effect, measured on the same file

| caller | before | after |
|---|---|---|
| `Outer.make` (`self.Inner()`) | `CALLS -> Class`, `CALLS -> __init__`, `INSTANTIATES` | `CALLS -> __init__`, `INSTANTIATES` |
| `via_instance` (`o.Inner()`) | `CALLS -> __init__`, `INSTANTIATES` | unchanged |

The typed-local receiver in the same fixture was already correct, which is what identified the `self.` path as the one at fault rather than the constructor redirect.

## Why no test covered this

Any test exercising `self.Inner()` aborted in `create_and_run_updater` before it could assert anything, because the conftest audit rejects the undocumented edge. That is why #1641 had to omit the shape from its fixture and assert on the resolver predicate instead; this PR is what lets that shape be tested at all.

## Tests

`codebase_rag/tests/test_self_nested_class_construction.py`.

The first assertion is over **every** `CALLS` edge in the fixture rather than the one pair, because the schema forbids the shape outright — a narrower assertion would miss a second construction elsewhere.

Two mutations, each with an execution count proving the line ran:

| mutation | dies |
|---|---|
| class filter removed (6 execs) | all three, but via the audit aborting the run — so it proves less than the count suggests |
| **self-dispatch dropped wholesale** (6 execs) | **only the control** |

The second is why the control exists. Deleting the fan-out satisfies "no CALLS targets a Class" perfectly while silently breaking dead-code reachability for every polymorphic method — a fix that passes the defect test and causes a worse bug. `Base.draw` calling `self.render()` must still reach both `Base.render` and `Derived.render`.

## Suite

10419 passed; the 24 failures and 4 errors are the known #1639 environment set, and the sorted FAILED set diffs empty against that baseline. Rebased onto `1dc99ccd` after the #1690 type-check hotfix.

Pushed after a single local review round, per the current review-gate rule.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected code relationship analysis for nested class construction.
  - Nested class calls are no longer incorrectly reported as method calls to a class.
  - Construction relationships continue to include instantiation and initializer calls.
  - Valid polymorphic method dispatch remains supported.

- **Tests**
  - Added coverage for nested class construction and polymorphic dispatch scenarios.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -731,7 +731,13 @@ class CallResolver:
                 self.function_registry.is_abstract(override_qn)
             ):
                 targets.add((self.function_registry[override_qn], override_qn))
-        return targets
+        # `self.Inner()` where `Inner` is a NESTED CLASS is a construction, not a
+        # virtual method call: the name resolves to `Outer.Inner`, which is a
+        # Class node, and CALLS may only target a Function/Method/Enum/Type. The
+        # construction already reaches the graph as INSTANTIATES plus a CALLS to
+        # `__init__`, so a class here is a duplicate the schema rejects outright
+        # (issue #1650).
+        return {target for target in targets if target[0] != cs.NodeLabel.CLASS.value}
 
     def js_member_twin_targets(self, callee_qn: str) -> set[tuple[str, str]]:
         # `View.prototype.lookup = function lookup(...)` registers TWO nodes for
--- a/codebase_rag/tests/test_self_nested_class_construction.py
+++ b/codebase_rag/tests/test_self_nested_class_construction.py
@@ -1,0 +1,114 @@
+"""`self.Inner()` constructs a nested class; it is not a virtual method call (#1650).
+
+`self.M()` fans out to every concrete override of `M`, so a method reached only
+polymorphically is not reported dead. That fan-out resolved `self.Inner()` to
+the nested CLASS `Outer.Inner` and emitted a `CALLS` edge to it, alongside the
+correct `CALLS -> __init__` and `INSTANTIATES`.
+
+`RELATIONSHIP_SCHEMAS` allows `CALLS` to target a Function, Method, Enum or Type
+-- never a Class; construction is what `INSTANTIATES` records. So the extra edge
+was both a duplicate and undocumented, and the suite's own `_audit_recorded_graph`
+rejected it, which is why no existing test could exercise this shape: any test
+that did aborted before it could assert anything.
+
+The controls matter here. The fan-out is load-bearing for real overrides, so a
+fix phrased as "drop the self-dispatch edges" would satisfy the assertion below
+and silently break dead-code reachability for every polymorphic method.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+from codebase_rag.constants import RelationshipType
+from codebase_rag.tests.conftest import create_and_run_updater, get_relationships
+
+
+def _edges(mock_ingestor: MagicMock, rel: RelationshipType) -> set[tuple[str, str]]:
+    return {
+        (call[0][0][2], call[0][2][2])
+        for call in get_relationships(mock_ingestor, rel.value)
+    }
+
+
+def _labelled_calls(mock_ingestor: MagicMock) -> set[tuple[str, str, str]]:
+    """(source qn, TARGET LABEL, target qn) for every CALLS edge."""
+    return {
+        (call[0][0][2], call[0][2][0], call[0][2][2])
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
Fixed in `6ee1a318`. `_edges` is now `_source_target_pairs`, which says what it returns and reads as clearly distinct from `_labelled_calls` — the latter carries the target LABEL precisely because the schema question here is about the label, not the pair.

vitali87:
@greptileai review

Issue #1449, opened 2026-08-26
test_live_dotnet_test_run_captures_dispatch_and_async fails instead of skipping when dotnet-trace cannot run


vitali87:
A machine with the .NET 8 SDK and a `dotnet-trace` built against .NET 10 fails the suite rather than skipping, with a message that has nothing to do with the assertion under test.

## Observed

```
FAILED codebase_rag/tests/test_dynamic_trace_speedscope.py::test_live_dotnet_test_run_captures_dispatch_and_async
subprocess.CalledProcessError: Command [dotnet-trace collect ...] returned non-zero exit status 131.
```

with the tool reporting on stderr:

```
You must install .NET to run this application.
App: ~/.dotnet/tools/dotnet-trace
Architecture: arm64
...apphost_version=10.0.11
```

The `dotnet build` step succeeds. The failure is `dotnet-trace` itself refusing to launch, because the global tool was installed against a .NET 10 runtime that is not present.

## Cause

The skip guard tests for the tool's existence as a file, never that it can execute (`codebase_rag/tests/test_dynamic_trace_speedscope.py:290`):

```python
def _dotnet_trace() -> str | None:
    """dotnet-trace on PATH, or in the default global-tools directory."""
    found = shutil.which("dotnet-trace")
    if found:
        return found
    candidate = Path.home() / ".dotnet" / "tools" / "dotnet-trace"
    return str(candidate) if candidate.exists() else None
```

`shutil.which` and `Path.exists()` both answer "is there a file here", which is not the question the skipif needs answered. An apphost whose runtime is missing is present, executable-bitted, and non-functional.

This is an asymmetry rather than a deliberate policy: the companion `_dotnet_with_sdk()` *does* probe by running the SDK, and the test body already handles two whole families of environmental failure by skipping (`_NET_NETWORK_ERRORS` for an unreachable NuGet, `_NET_SDK_ERRORS` for a too-old SDK). The trace tool is the one dependency checked by presence alone, so the same class of environment problem that skips politely at every other step raises here.

## Suggested fix

Probe by invoking, the way the SDK is probed — run `dotnet-trace --version` once at module scope and treat a non-zero exit or `OSError` as absent. That makes the guard answer "can this tool run" rather than "is this file here".

A narrower alternative is to add the missing-runtime signature to the skip markers, but probing is better: it covers every reason the tool cannot launch, not the one string currently observed.

## Impact

[...]

PR #1697, opened 2026-09-03
fix: probe formatters by invoking them, not by finding them on PATH (#1639)

vitali87:
Part of #1639 (does not close it — the node and dotnet guards follow in a
second PR).

## A user-facing bug the issue did not mention

Chasing #1639's rustfmt guard with a real failing shim turned up a product bug,
not just a test-infrastructure one.

`formatter_check` decided "is the formatter installed" with `shutil.which`. A
rustup shim for an uninstalled component **is** a real file on PATH, so that
check passes, the tool is invoked, it exits non-zero, and the function returns
`formatted=False`. The user is then told:

```
m.rs applied, but rustfmt would reformat it
```

That is a claim about their code that no formatter ever made. The function's
own docstring promised `None` when the tool is not installed, so the code
contradicted its documented contract.

Measured with a shim that mimics rustup exactly (exit 1, `error: 'rustfmt' is
not installed for the toolchain`):

| | `formatted` | message |
|---|---|---|
| before | `False` | `m.rs applied, but rustfmt would reformat it` |
| after | `None` | `m.rs: 2 edit(s) applied` |

## The fix

`_tool_runs(executable)`, a cached `--version` probe, gating the formatter
verdict. Non-zero exit means "not usable", which is what `which` was standing
in for. Same shape as `_runnable_dotnet_trace` and `_dotnet_with_sdk` in
`test_dynamic_trace_speedscope.py` (issue #1449), which solved this class for
`dotnet-trace` and whose reasoning is written out there.

The three `shutil.which` guards in `test_patcher.py` get the same treatment via
`_runnable_formatter`, so the test half of #1639's rustfmt item is done too.

## Verification, and its limits

**Verified with a real shim.** Under an identical rustup-style shim the old
guard returns `True` (hard failure, which is #1639's reported symptom) and the
new one returns `False` (correct skip). `test_patcher.py` passes both with the
shim on PATH and without it: 20 passed, 1 skipped in each.

**One honest caveat on that green.** The `1 skipped` in *both* runs is `gofmt`,
which is genuinely absent on this machine — not rustfmt. The rustfmt test
passes under the shim by taking its `else` branch and asserting
`formatted is None`, which is exactly the production behaviour this PR fixes. I
checked that with `-rs` rather than inferring it from the counts, since
identical skip totals would otherwise have looked like the shim did nothing.

**Not covered here.** The node/ESM and dotnet/`net10.0` halves of #1639 are not
in this PR. This machine runs Node 22 and .NET SDK 10, both working, so those
two failure modes cannot be reproduced here at all; they follow in a second PR
with the constructed cases labelled as constructed.

### DIFF
--- a/codebase_rag/constants/lang_tooling.py
+++ b/codebase_rag/constants/lang_tooling.py
@@ -158,6 +158,9 @@ PATCH_GOFMT_LIST_FLAG = "-l"
 # for an uninstalled component is a real file on PATH that exits non-zero, and
 # that is indistinguishable from "your code is misformatted" (issue #1639).
 PATCH_VERSION_FLAG = "--version"
+# gofmt has no --version and exits 2 on it; -h exits 0. The liveness probe must
+# use a flag the tool accepts, or a working formatter reads as a missing one.
+PATCH_GOFMT_PROBE_FLAG = "-h"
 PATCH_RUSTFMT = "rustfmt"
 PATCH_RUSTFMT_CHECK_FLAG = "--check"
 PATCH_RUSTFMT_EDITION_FLAG = "--edition"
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

vitali87:
@greptileai review

vitali87:
@coderabbitai review

Issue #1702, opened 2026-09-03
Property-access pass never recognises a call target: tree-sitter node compared by identity


vitali87:
## Defect

In the property-access pass in `codebase_rag/parsers/call_processor.py` the guard

```python
is_call_target = (
    parent is not None
    and parent.type == call_type
    and parent.child_by_field_name(func_field) is node
)
```

is always `False`. py-tree-sitter builds a fresh Python wrapper for every lookup, so `child_by_field_name(...)` never returns the same object as a node taken from `.children`; identity fails while `==` (node id) succeeds:

```
>>> f = call.child_by_field_name("function"); f is attr, f == attr, f.id == attr.id
(False, True, True)
```

So an attribute that is the function of a call (`obj.value()`) is never excluded, and when its leaf name is in `prop_names` it is resolved as a property access as well as a call, emitting an extra edge the guard was written to prevent.

## Fix

Compare with `==`, and add a test that drives the excluded shape (a property whose name is also called as a method on the same receiver) and asserts no property-access edge is emitted for the call site. Mutating `==` back to `is` must turn that test red.

The same pattern was found and fixed on #1543 in `trace/dispatch_site.py` (the dict-key dispatch shape never matched). A grep of `main` for `child_by_field_name(...) is <name>` finds only this remaining site.

PR #1706, opened 2026-09-03
fix: compare tree-sitter nodes by value when skipping the import base (#1703)

### LINKED ISSUE #1703: endpoint_prefixes compares tree-sitter nodes by identity: the import base leaks as an imported symbol
## Defect

`imported_module_qns` in `codebase_rag/parsers/endpoint_prefixes.py` (~line 198) skips the `from X import ...` base with

```python
for child in node.named_children:
    if child is base_node:
        continue
```

where `base_node = node.child_by_field_name(cs.FIELD_MODULE_NAME)`. This is the same defect as #1702: py-tree-sitter builds a fresh wrapper per lookup, so `child_by_field_name` never returns the same object as an element of `named_children`, and the `continue` never fires.

## Effect

The base is processed as though it were one of the imported symbols. It is normally masked by the `cs.SEPARATOR_DOT not in name` filter, but a **dotless** module name passes that filter and the module maps to itself:

```python
imported_module_qns("proj.m", parse(b"from routers import alpha\n").root_node,
                    {"proj.m", "routers", "routers.routers", "routers.alpha"})
# observed: ['routers.alpha', 'routers.routers']
# expected: ['routers.alpha']
```

`routers.routers` is not a real import; it is the base leaking through.

## Fix

`if child == base_node`. `Node.__eq__` compares tree context as well as node id, so it is strictly stronger than `.id ==` and cannot make two distinct nodes match.

Add a test with a dotless `from <module> import <name>` and assert the base does not appear among the results — mutating `==` back to `is` must turn it red.

## Context

Found by the local review round on #1702, which fixed the same pattern in `call_processor.py`. That issue stated a grep of `main` found no other instance; this is one, so the claim should not be relied on. Other `is` comparisons on tree-sitter nodes in `call_processor.py` (lines 606, 2330, 2625, 6592) were checked and are safe — they compare nodes drawn from cached `children`/`named_children` lists, where identity does hold. `class_ingest/mixin.py:90` and `call_processor.py:188` walk `.parent`, which is not cached, so they are latently the same hazard; no divergence could be constructed for either, but they would be worth a comment.

vitali87:
Closes #1703.

## What

`_from_import_targets` in `codebase_rag/parsers/endpoint_prefixes.py` skips the `from <module> import ...` base by testing each named child against it:

```python
base_node = node.child_by_field_name(cs.FIELD_MODULE_NAME)
for child in node.named_children:
    if child is base_node:      # never True
        continue
```

py-tree-sitter builds a fresh Python wrapper on every lookup, so the node returned by `child_by_field_name` is never the same object as the one in `named_children`. The skip never happened.

## Effect

A **dotted** base is saved by the `SEPARATOR_DOT not in name` filter below the skip, so the leak only shows on a dotless module name — which then maps to itself as though it had been imported:

| source | before | after |
|---|---|---|
| `from routers import alpha` | `{'routers': 'routers.routers', 'alpha': 'routers.alpha'}` | `{'alpha': 'routers.alpha'}` |
| `from routers import alpha, beta` | `{'routers': ..., 'alpha': ..., 'beta': ...}` | `{'alpha': ..., 'beta': ...}` |
| `from routers import alpha as a` | `{'routers': 'routers.routers', 'a': 'routers.alpha'}` | `{'a': 'routers.alpha'}` |
| `from pkg.sub import alpha` | `{'alpha': 'pkg.sub.alpha'}` | unchanged |
| `from . import alpha` | `{'alpha': 'proj.alpha'}` | unchanged |
| `from routers import *` | `{}` | unchanged |

`==` compares tree context as well as node id, so it is strictly stronger than `.id ==` and cannot make two distinct nodes match.

## Verification

`codebase_rag/tests/test_from_import_base_skip.py` is parametrised over four shapes, keeping a **dotted base as a control** — it must keep passing under mutation, because the dot filter covers that case independently of the skip.

| variant | result |
|---|---|
| fix applied | 4 passed |
| `is` restored | 3 failed, 1 passed |
| skip deleted outright | 3 failed, 1 passed |

The three failures in each mutant are the dotless cases by name; the control passes. That split is deliberate evidence: a fixture or collection error would have taken all four down uniformly, which is indistinguishable from real discrimination if you only check the exit code.

Also run: 168 tests across `test_route_call_endpoints`, `test_endpoint_prefix_resolution`, `test_endpoint_extraction` and the new file — all pass. `ruff check` / `ruff format --check` clean.

## Context

Same defect as #1702 (fixed in `call_processor.py` by #1704). That issue stated a grep of `main` found no other instance; this is one. The remaining `is` comparisons on nodes in `call_processor.py` (606, 2330, 2625, 6592) are safe — both sides come from cached `children`/`named_children` lists, where identity holds. `class_ingest/mixin.py:90` and `call_processor.py:188` walk `.parent`, which is not cached, so they are latently the same hazard; no divergence could be constructed for either.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected import analysis so the source module is no longer incorrectly reported as an imported symbol.
  * Ensured this behavior works consistently for dotted and dotless modules, including multiple and aliased imports.

* **Tests**
  * Added coverage for import statements with single, multiple, and aliased names.

### DIFF
--- a/codebase_rag/parsers/endpoint_prefixes.py
+++ b/codebase_rag/parsers/endpoint_prefixes.py
@@ -195,7 +195,12 @@ def _from_import_targets(module_qn: str, node: Node) -> dict[str, str]:
     base = _from_import_base(module_qn, base)
     out: dict[str, str] = {}
     for child in node.named_children:
-        if child is base_node:
+        # `==` not `is`: py-tree-sitter builds a fresh wrapper per lookup, so
+        # the node from child_by_field_name is never the same object as the
+        # one in named_children. With identity this never skipped the base,
+        # and a dotless module name passed the dot filter below and mapped
+        # to itself as though it were an imported symbol.
+        if child == base_node:
             continue
         if child.type == cs.TS_PY_DOTTED_NAME:
             name = _decode(child)
--- a/codebase_rag/tests/test_from_import_base_skip.py
+++ b/codebase_rag/tests/test_from_import_base_skip.py
@@ -1,0 +1,44 @@
+# `from <module> import <name>` must not report the module itself as one of
+# the imported names. The base is skipped by comparing it against each
+# named child, and that comparison used `is`: py-tree-sitter builds a fresh
+# wrapper per lookup, so the node from child_by_field_name is never the
+# same object as the one in named_children and the skip never happened.
+#
+# A dotted base (`pkg.sub`) hid the bug -- the dot filter below the skip
+# dropped it anyway -- so it only shows on a dotless module name.
+from __future__ import annotations
+
+import pytest
+
+from codebase_rag.parsers.endpoint_prefixes import _from_import_targets
+
+tree_sitter = pytest.importorskip("tree_sitter")
+tree_sitter_python = pytest.importorskip("tree_sitter_python")
+
+
+def _import_node(source: str):
+    parser = tree_sitter.Parser(tree_sitter.Language(tree_sitter_python.language()))
+    return parser.parse(source.encode("utf-8")).root_node.children[0]
+
+
+@pytest.mark.parametrize(
+    ("source", "expected"),
+    [
+        ("from routers import alpha\n", {"alpha": "routers.alpha"}),
+        (
+            "from routers import alpha, beta\n",
+            {"alpha": "routers.alpha", "beta": "routers.beta"},
+        ),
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@coderabbitai review

Issue #1660, opened 2026-09-02
Renaming the Go file that holds a receiver type drops the type's method CALLS on a reused updater


vitali87:
## Observed

Fixture: \`pkg/types.go\` (\`type T struct{}\`), \`pkg/methods.go\` (\`func (t T) M() int { return helper() }\`), \`pkg/util.go\` (\`func helper() int\`). Index with \`run(force=True)\` on a reused \`GraphUpdater\`, rename \`types.go\` to \`kinds.go\`, bump its mtime, \`run(force=False)\` on the same updater.

The result has no CALLS edge from \`T.M\` at all. A clean index of the renamed tree has \`proj.pkg.types.T.M -> proj.pkg.util.helper\` (under whatever module the type now resolves to).

Identical on \`origin/main\` and on the #1576 branch.

## Expected

A reused updater's result after the rename equals a clean index of the renamed tree.

## Related

- Found during the round-3 local review of #1576; pre-existing, out of that PR's scope.

Issue #1017, opened 2026-08-01
fix: two Rust modules can share one qualified name so import maps and scope walks serve the wrong module


vitali87:
## Summary

Two distinct Rust modules can map to one qualified name, and every consumer keyed by module qn (import maps, callers' scope walks, watch cleanup sets) then serves the wrong module. The #1007 branch detects the narrowest instance (a function-local `mod X { }` beside a bodyless `mod X;` in the same declaring module) and drops that mapping as the lesser error, but the underlying scheme cannot represent the collision, so one side always resolves through the other's namespace.

## Colliding shapes, all legal Rust (rustc verified during the #1007 reviews)

1. Function-local mod versus submodule file:

```rust
pub mod run;                       // src/foo/run.rs

pub fn f() -> u32 {
    mod run {
        use crate::alpha::helper;
        pub fn go2() -> u32 { helper() }
    }
    run::go2()
}
```

`go2` registers as `foo.run.go2`, inside `src/foo/run.rs`'s qn namespace. Its bare call resolves through run.rs's import map: `foo.run.go2 CALLS beta.helper` where cargo gives `alpha::helper`.

2. cfg-gated twins at file level:

```rust
#[cfg(feature = "ext")]
pub mod run;                        // src/foo/run.rs

#[cfg(not(feature = "ext"))]
pub mod run {
    use crate::alpha::helper;
    pub fn go2() -> u32 { helper() }
}
```

Both modules are indexed unconditionally; their import maps share the key `foo.run`, and whichever file parses later resets or overwrites the shared map, so one side's calls mis-bind depending on parse order.

3. Two impls of one type each containing a method-local `mod inner`: both key `foo.S.inner` and merge.

## Analysis

Function-registry collisions are handled by `register_unique_qn` (`natural@<start_line>`), and function-scope imports resolve through the span-keyed `function_locations` map, but MODULE qns have no equivalent deduplication: readers address modules by derived qn strings, so two modules with one qn are indistinguishable at every consumer. A fix needs a deduplication scheme for module qns (and span-aware resolution for callers registered inside the duplicated module), which touches the registry, module ingestion, import mapping, and the scope walk together.

Surfaced across rounds twelve and thirteen of the #1007 adversarial reviews; shape 1's map-merge and shape 2's order dependence reproduce on the branch, and main misresolves the same fixtures through raw-path trie fallbacks, so the ambiguity predates the branch.

vitali87:
[...]

PR #1710, opened 2026-09-03
fix: sweep definitions keyed outside the re-parsed file's own module (#1659)

### LINKED ISSUE #1659: Re-parsing a Go file whose receiver type lives in a sibling file registers a duplicate Method node
## Observed

Fixture: \`pkg/types.go\` (\`type T struct{}\`), \`pkg/methods.go\` (\`func (t T) M() int { return helper() }\`), \`pkg/util.go\` (\`func helper() int\`). Index with \`run(force=True)\` on a reused \`GraphUpdater\`, then re-parse \`methods.go\` on the same updater (edit it, or rename \`util.go\` so it joins as a dependent) with \`run(force=False)\`.

The second run registers a second Method node \`proj.pkg.types.T.M@3\` beside the first run's \`proj.pkg.types.T.M\`. The first run's method is keyed under the types module, so \`remove_file_from_state(methods.go)\` does not own it and does not remove it before the re-parse.

Identical on \`origin/main\` and on the #1576 branch.

## Expected

The snapshot after the re-parse equals a clean index: one Method node for \`T.M\`.

## Related

- Found during the round-3 local review of #1576; pre-existing, out of that PR's scope. The Go rename test added to \`test_incremental_deleted_state.py\` on #1576 compares CALLS edges only for this reason.

vitali87:
Closes #1659.

A definition's qualified name does not have to sit under the qn of the file that registered it, and `remove_file_from_state` assumed it did.

A Go method is keyed by its **receiver type's** module. With `type T struct{}` in `pkg/types.go` and `func (t T) M() int` in `pkg/methods.go`, the method registers as `proj.pkg.types.T.M`. The cleanup for `methods.go` sweeps the qn prefixes that file recorded — `proj.pkg.methods` — and nothing under that prefix matches, so the first run's registry entry survived. The re-parse then found the qn taken and registered a second Method beside it as `proj.pkg.types.T.M@3`.

## The fix

The ownership record this needs already existed and was already being computed, just only in the negative direction. `function_locations` is keyed `(registering module qn, line, col)`, so for this fixture the first run records:

```
('proj.pkg.methods', 3, 11) -> proj.pkg.types.T.M
```

`foreign_qns` was built from those entries to protect qns registered by modules this event does not touch (the #1017/#1025 inline-mod shape). The same loop now also collects the mirror set — qns whose registering module **is** touched — and a qn is swept when it matches a recorded prefix *or* is owned by this file, in both cases still subject to the existing `foreign_qns` veto. A qn registered by two modules stays protected, exactly as before.

Nothing changes for a file that was never parsed: it has no span records, so the owned set is empty and the prefix sweep is unchanged.

## Verification

The reproduction is the issue's second trigger — renaming `util.go` so `methods.go` joins the run as a dependent:

```
before:  {'proj.pkg.types.T.M', 'proj.pkg.types.T.M@3'}
after:   {'proj.pkg.types.T.M'}
```

and the snapshot now equals a clean index of the renamed tree.

The test asserts on Method **nodes**, not on CALLS edges. `test_a_reused_updater_resolves_a_go_call_after_a_same_package_rename` already covers this exact rename and compares calls only — and calls alone are satisfied by a graph carrying both the stale and the fresh method, which is why the duplicate survived alongside a passing test.

One note on the reproduction, recorded in the test because it cost me a green run: touching `methods.go` and re-running does **not** reproduce. The hash cache sees identical content and skips the file, so nothing is re-parsed and no duplicate can form. That test passed against unfixed code and was a statement about a run that never happened, not about the bug.

## Scope note

`#1660` (renaming the file that HOLDS the receiver type drops the method's CALLS) is a separate defect and is not fixed here. I checked rather than assumed: with this fix applied, that scenario still yields no CALLS at all where a clean index has `proj.pkg.kinds.T.M -> proj.pkg.util.helper`. It gets its own branch.

The full unit suite was run on this commit: 10453 passed, 23 failed, 4 errors. Every failure is the pre-existing #1639 oracle-toolchain issue on this machine (Node 18 / .NET SDK 8), which is exactly the issue's own baseline minus the rustfmt case #1697 already fixed. None are in the graph, incremental, watcher or registry paths. 175 targeted tests across every incremental/watch/registry/reingest file pass.

The configured local reviewer agent is not available in this session, so the pre-push local round could not be run; the PR's own Greptile review is the first review of this change.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2313,15 +2313,26 @@ class GraphUpdater:
         # records key by the REGISTERING file's module qn, so a qn recorded
         # under a module this event does not touch survives (issue #1025).
         touched_modules = recorded_qns
-        foreign_qns = {
-            location.qualified_name
-            for (
-                span_module,
-                _line,
-                _col,
-            ), location in self.factory.definition_processor.function_locations.items()
-            if span_module not in touched_modules
-        }
+        foreign_qns = set()
+        # The mirror of foreign_qns, and the reason the prefix sweep is not
+        # sufficient on its own: a definition's qn does not have to sit under
+        # the qn of the file that registered it. A Go method is keyed by its
+        # RECEIVER TYPE's module, so `func (t T) M()` in pkg/methods.go
+        # registers `proj.pkg.types.T.M` when `T` is declared in pkg/types.go.
+        # No prefix of methods.go matches that qn, so the entry survived this
+        # cleanup and the re-parse registered a second Method beside it as
+        # `...T.M@3` (issue #1659). The span records already name the
+        # registering module, which is exactly the ownership this needs.
+        owned_qns = set()
+        for (
+            span_module,
+            _line,
+            _col,
+        ), location in self.factory.definition_processor.function_locations.items():
+            if span_module in touched_modules:
+                owned_qns.add(location.qualified_name)
+            else:
+                foreign_qns.add(location.qualified_name)
         # Frontend registrations (libclang C/C++) have no span records; their
         # ownership is tracked per rel path at registration time.
         touched_rel = relative_path.as_posix()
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

PR #1715, opened 2026-09-04
fix: rebuild the C++ interface set instead of accumulating it (#1692)

### LINKED ISSUE #1692: cpp_module_interfaces accumulates across runs, but cannot simply be rebuilt on rehydration
## Defect

`definition_processor.cpp_module_interfaces` is filled by `_rehydrate_registry_from_graph` (the `MODULE_INTERFACE` branch of the module-rows loop) and by parsing, and nothing ever clears it. On a reused updater it only grows, the same shape as #1677 and #1687.

## Effect

`resolve_deferred_cpp_module_impls` skips an implementation whose interface is not in the set, because "emitting the edge anyway would mint a phantom the database drops". A stale entry defeats that guard: an interface deleted from the graph by another writer stays in the set, and an `IMPLEMENTS` edge is emitted to a `ModuleInterface` node that no longer exists.

## Why the obvious fix is wrong

Clearing the set before the refill (as #1688 does for `_rehydrated_module_qns`) **loses real edges**. Measured on a buffering ingestor, which is what production uses — `GraphService.ensure_node_batch` appends to `node_buffer` and only flushes at `batch_size`:

- add `iface.cppm` (`export module M;`) and `impl.cpp` (`module M;`) together, then run incrementally
- at rehydration time `proj.M`'s node write is still unflushed, so it is absent from the rows just fetched
- with the clear: `before=['proj.M'] after=[]`, and the run emits **no** `IMPLEMENTS` edge
- without the clear: the edge is emitted correctly

A test using an ingestor that writes through immediately (`_StatefulIngestor`) does **not** catch this — the edge is emitted in both cases. The buffering is essential to reproducing it.

## Why "preserve what this run parsed" is also wrong

Snapshotting the set before the clear and restoring it afterwards keeps the buffered interface, but the set persists across runs on a reused updater, so the snapshot also carries the previous run's stale entries. Verified: a planted `proj.GHOST` survives rehydration under that approach, which is exactly the staleness the fix was meant to remove.

## Fix

Distinguish interfaces registered by the current run's parsing from ones carried over. Options worth weighing:

- clear the set at the start of a run (where `_parsed_files` is cleared) rather than inside rehydration, so parsing repopulates it naturally and rehydration only adds graph-derived entries;
- track per-run registrations alongside `_frontend_owned_qns`-style ownership, and rebuild only the carried-over portion;
- flush pending node writes before the rehydration query, so a this-run interface is in the rows (heavier, and changes flush semantics for everything else).

A regression test must use an ingestor that buffers writes until `flush_all`, or it will pass with the bug present.

Found while implementing #1677: the clear was extended to this set, the GitHub Greptile review flagged the buffered-interface case, and the measurement above confirmed it as a real regression, so the change was dropped from that PR and recorded here.

vitali87:
Closes #1692.

## What

`_rehydrate_registry_from_graph` fills `definition_processor.cpp_module_interfaces` from the module rows and nothing ever cleared it, so on a reused updater (the watcher, the MCP server) an interface another writer deleted stays for the life of the process.

A stale entry defeats the membership check in `resolve_deferred_cpp_module_impls`, which exists precisely to stop an `IMPLEMENTS` edge being minted to a `ModuleInterface` the graph no longer holds — "a phantom the database drops", in its own docstring.

## Why the two obvious fixes are wrong

**A bare `clear()` is a regression, not a fix.** Production batches node writes, so an interface parsed in the current run is still in the ingestor's buffer when rehydration reads and is absent from the rows it fetches. Clearing outright drops it and loses the `IMPLEMENTS` edge for an interface and implementation added together. That is what an earlier attempt shipped and withdrew, and this branch carries the test that proves it.

**Snapshot-and-restore fails too.** The set survives across runs on a reused updater, so restoring "whatever was there" carries the previous run's stale entries straight back in.

## The fix

Three parts, each pinned by its own test:

1. **Capture** — the parse records what it registered into a companion set, `cpp_interfaces_parsed_this_run`.
2. **Per-run reset** — `run()` clears that set where it clears `_parsed_files`, so a previous run's entries are never mistaken for unflushed writes.
3. **Rebuild with restore** — rehydration clears the interface set, refills the graph-derived part from its rows, then puts this run's parsed entries back.

Stale entries go; unflushed ones stay.

## Verification

| variant | result |
|---|---|
| baseline | **17 passed** |
| bare `clear()` (the withdrawn regression) | 1 failed — `…parsed_this_run_survives_rehydration` |
| rebuild removed | 2 failed — `…graph_lost_is_forgotten`, `…not_revived_on_a_later_run` |
| per-run reset removed | 1 failed — `…not_revived_on_a_later_run` |

The last row was a real gap found in review: before that test, deleting the reset left all 16 tests green while reintroducing this very bug. `…graph_lost_is_forgotten` cannot catch it, because its interface is parsed by a *first* updater, so the second updater's companion set is empty and the reset is never what saves it. The new test parses with the same updater that runs again.

Tests use a `_BufferingIngestor` that holds node writes until `flush_all`, because a write-through fake hides every bug that depends on batching — the fixture lesson from #1688.

Full suite on this branch: **10513 passed, 40 skipped, 1 xfailed, 0 failed.** C++ tests confirmed to run rather than skip (the interpreter with all 15 grammars; the Python-only venv skips them silently).

## Scope

`cpp_interfaces_parsed_this_run` grows unbounded on an updater that only ever calls `reingest()`, since `run()` is the sole reset point. No correctness effect — `reingest` rehydrates at most once (`_hydrate_for_reingest` sets `_reingest_hydrated` on every path) so the set is never consulted there — but it is a slow leak in a long-lived watcher or MCP updater. Measured across three successive `reingest()` calls: rehydration count stays at 1 while the set grows. Noted rather than fixed here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved incremental processing of C++20 module interfaces so newly parsed interfaces remain available during the current run.
  - Prevented stale interface records from being incorrectly restored after graph updates or subsequent runs.
  - Correctly recognizes interfaces removed by another update, keeping implementation relationships in sync with the latest graph state.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -1163,6 +1163,10 @@ class GraphUpdater:
         # Reset per-run parse tracking so a reused updater does not reprocess
         # a previous run's files in Pass 3.
         self._parsed_files.clear()
+        # Per-run for the same reason: the set records what THIS run parsed,
+        # so a reused updater must not treat a previous run's interfaces as
+        # unflushed writes that rehydration has to preserve.
+        self.factory.definition_processor.cpp_interfaces_parsed_this_run.clear()
         # Per-run for the same reason: set on the in-sync early return below
         # and previously cleared only in `__init__`, so a reused instance kept
         # reporting a previous run's skip. `cli.py` reads it to decide what to
@@ -1840,6 +1844,20 @@ class GraphUpdater:
         # verification as a live target. Cleared only now that the rows are
         # in hand, so a failed query above leaves the previous set intact.
         self._rehydrated_module_qns.clear()
+        # Same rebuild for the interface set the loop below fills from these
+        # same rows. A stale entry defeats the membership check in
+        # resolve_deferred_cpp_module_impls, which is what stops an IMPLEMENTS
+        # edge being minted to a ModuleInterface the graph no longer holds.
+        #
+        # Interfaces THIS run parsed are put back: their node writes may still
+        # be in the ingestor's buffer and so absent from the rows just
+        # fetched, and dropping one loses the IMPLEMENTS edge for an interface
+        # and implementation added together. That is not hypothetical -- an
+        # earlier attempt at this cleared the set outright and did exactly
+        # that.
+        dp = self.factory.definition_processor
+        dp.cpp_module_interfaces.clear()
+        dp.cpp_module_interfaces |= dp.cpp_interfaces_parsed_this_run
         for row in module_rows:
             qn = row.get(cs.KEY_QUALIFIED_NAME)
             label = row.get(cs.KEY_LABEL)
--- a/codebase_rag/parsers/class_ingest/mixin.py
+++ b/codebase_rag/parsers/class_ingest/mixin.py
@@ -190,6 +190,7 @@ class ClassIngestMixin:
     rust_impl_method_traits: dict[str, str]
     rust_inherent_impl_methods: set[str]
     cpp_module_interfaces: set[str]
+    cpp_interfaces_parsed_this_run: set[str]
     _deferred_cpp_module_impls: list[tuple[str, str]]
     declared_module_qns: set[str]
     rust_function_modules: dict[str, str]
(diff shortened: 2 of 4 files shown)


vitali87:
@greptileai review

Issue #1716, opened 2026-09-04
The eval double answers [] for any query it does not emulate, so tests pass against lookups that never ran


vitali87:
## Observed

`evals/cgr_graph._StatefulIngestor` emulates "only the exact queries cgr emits, matched by identity". Everything else falls through to:

```python
case _:
    return []
```

An empty list is a valid answer for most of these queries, so an unemulated query is indistinguishable from a genuinely empty result. The double fails **open**: a test can be green because the lookup returned nothing, not because the code did the right thing.

That is not hypothetical. `CYPHER_UNRESOLVED_IMPORTER_PATHS` (#1682) was unemulated, so the waiting-importer lookup answered `[]` for every scenario, and that issue's tests could only ever cover the query layer — its own module docstring records the limitation. #1714 hit the same wall and both cases have now been added, which is what makes an end-to-end assertion mean anything there.

## Measured

Instrumenting `case _` to record what reaches it, over `test_incremental_deleted_state.py`, `test_reingest.py`, `test_incremental_eval.py`, `test_reingest_new_file_importers.py` and `test_incremental_deleted_dependents.py` — **136 passed** while six distinct queries were silently answered `[]`:

| constant | shape | note |
|---|---|---|
| `CYPHER_ALL_CSHARP_TYPE_LOCATIONS` | `MATCH (n) WHERE (n:Class OR n:Interface OR n:Enum) ... ENDS WITH '.cs'` | feeds C# partial-group joining |
| `CYPHER_ALL_GO_TYPE_LOCATIONS` | `MATCH (n) WHERE (n:Class OR ... OR n:Union) ...` | col-keyed location rehydration |
| `CYPHER_ALL_FUNCTION_LOCATIONS` | `MATCH (m:Module)-[:DEFINES]->(f:Function) ...` | location rehydration |
| `CYPHER_ALL_METHOD_LOCATIONS` | `MATCH (m:Module)-[:DEFINES]->(c)-[:DEFINES_METHOD]->(f:Method) ...` | location rehydration |
| `CYPHER_COUNT_PROJECT_MODULES` | `... RETURN count(m) AS count` | see below |
| `CYPHER_QUERY_EMBEDDINGS` | `MATCH (m:Module)-[:DEFINES]->(n) WHERE (n:Function OR n:Method) ...` | semantic embeddings, not graph structure |

[...]

PR #1724, opened 2026-09-04
fix: give the deletion ownership set the same path-derived fallback as the sweep (#1719)

### LINKED ISSUE #1719: A fresh updater's scoped re-ingest rehydrates locations before deleting the re-parsed subtree, so definitions land as DUP_QN variants
## Observed

On a **fresh** `GraphUpdater` (the shape the MCP tool builds per project), a scoped `reingest` of a file registers its definitions under a `DUP_QN` variant instead of the clean qualified name, so the result diverges from a clean index:

```
extra node:   ('Function', 'sib.util.helper@1')
missing node: ('Function', 'sib.util.helper')
extra edge:   ('Module', 'sib.util', 'DEFINES', 'Function', 'sib.util.helper@1')
missing edge: ('Module', 'sib.util', 'DEFINES', 'Function', 'sib.util.helper')
```

## Cause

`reingest` hydrates before it deletes:

- `graph_updater.py:3952` — `self._hydrate_for_reingest()`, which calls `_rehydrate_function_locations()` (`:3627`)
- `graph_updater.py:3996` — `self._reingest_delete(reparse, gone, hashes)`, which removes each re-parsed file's state and subtree

`CYPHER_ALL_FUNCTION_LOCATIONS` is `MATCH (m:Module)-[:DEFINES]->(f:Function) WHERE m.qualified_name STARTS WITH $project_prefix`. It does not exclude the files about to be re-parsed, and at the point it runs their nodes are still in the graph. So the rehydrated index carries a location for `sib.util.helper`, the re-parse of that same file registers `helper` again, `register_unique_qn` finds the qn taken, and the definition lands as `sib.util.helper@1`.

The guard at `:3617` (`if self._reingest_hydrated or self._parsed_files`) means a REUSED updater skips hydration and is unaffected. Only a fresh one — no prior `run()` in the process — takes this path.

## Why it has not been caught

`codebase_rag/tests/test_reingest.py::test_reingest_gives_same_stem_siblings_their_clean_index_qns[add_loser-fresh]` asserts exactly this equality and passes today, because the eval double answered `[]` for `CYPHER_ALL_FUNCTION_LOCATIONS` and rehydration silently no-opped. Emulating that query in #1716 makes the assertion real and the test fails. It is marked `xfail(strict=True)` there, pointing here.

## Suggested direction

Either exclude the paths about to be re-parsed from the location rehydration, or hydrate after `_reingest_delete` rather than before. The second is the smaller change if nothing between the two points depends on the rehydrated state; that wants checking rather than assuming, since `_hydrate_for_reingest` also calls `identify_structure` and `_rehydrate_registry_from_graph`, and the delete step reads `module_qn_to_file_path`.

A fix wants a test on the fresh-updater path comparing a scoped re-ingest against a clean index — the same shape as the existing parametrisation, which can then have its `xfail` removed.

## Provenance

Surfaced by #1716, which makes the eval double emulate the queries it previously answered `[]` for. Faithfulness of that emulation was checked before filing: the production query has no exclusion for the re-parsed file, and the hydrate-before-delete ordering is in the line numbers above, so a live Memgraph returns the same row.

vitali87:
Closes #1719.

A **fresh** `GraphUpdater` (the shape the MCP tool builds per project) doing a scoped `reingest` registered a re-parsed file's definitions under a `DUP_QN` variant instead of the clean name:

```
extra node:   ('Function', 'sib.util.helper@1')
missing node: ('Function', 'sib.util.helper')
```

## Cause

Not the hydrate/delete ordering the issue first guessed. `remove_file_from_state` computes two sets from the same question — which modules does this file own?

```python
recorded_qns      = {qn for qn, path in module_qn_to_file_path.items() if path == file_path}
module_qn_prefixes = recorded_qns or {path_derived_qn}     # sweep: HAS a fallback
touched_modules    = recorded_qns                          # ownership: had NONE
```

`module_qn_to_file_path` is filled only when a file is parsed (`DefinitionProcessor.process_file`), so on a fresh updater it is **empty**. The sweep survives that through its path-derived fallback; the ownership set did not. An empty `touched_modules` makes *every* rehydrated span "foreign", and `foreign_qns` then protects the entire graph from a sweep that consequently removes nothing.

The re-parse then finds its own qualified name still held by the rehydrated registry and disambiguates to `@1`.

## The fix

One line: ownership uses the same set as the sweep, fallback included.

```python
touched_modules = module_qn_prefixes
```

The two were answering the same question from different data. Where the sweep says "these are the prefixes this file owns", ownership must agree, or they disagree about which modules the event touches — which is precisely what happened.

## Verification

The regression test already existed, marked `xfail(strict=True)` by #1716 when the eval double started answering `CYPHER_ALL_FUNCTION_LOCATIONS` and the assertion became real. **That marker is removed here** — which is what `strict=True` was for: once fixed, the test passes and strict turns a lingering marker into a failure saying "remove me".

- `test_reingest_gives_same_stem_siblings_their_clean_index_qns` — all 8 parametrisations pass, `warm` and `fresh` alike.
- Mutation: reverting to `recorded_qns` reproduces exactly the original failure on exactly the original case, `[add_loser-fresh]`, and nothing else.
- 236 tests pass across the reingest, incremental, watcher-ownership, cross-language-collision and eval-double files.

The `#1025` ownership guard this touches still holds: `test_watch_registry_ownership.py` and the inline-mod cases pass unchanged. That guard exists so a sweep by prefix alone cannot deregister another file's definitions, and widening `touched_modules` only ever narrows `foreign_qns` for the file being removed — a file that is about to be re-parsed and re-register whatever it still defines.

## Provenance

Surfaced by #1716, which made the eval double answer queries it had been silently returning `[]` for. The test had been passing only because rehydration no-opped, so the equality it asserts was never exercised on this path. An honest double showing a red test is what made this findable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed scoped re-ingestion so re-parsed files no longer receive unnecessary duplicate-qualified-name variants.
  - Ensured existing code relationships are correctly removed and rebuilt during fresh re-ingestion.

- **Tests**
  - Re-enabled coverage for re-ingestion scenarios that previously expected the defect.
  - Updated test descriptions to reflect the corrected behavior.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2570,7 +2570,18 @@ class GraphUpdater:
         # they are never re-registered because it is not re-parsed. The span
         # records key by the REGISTERING file's module qn, so a qn recorded
         # under a module this event does not touch survives (issue #1025).
-        touched_modules = recorded_qns
+        # Uses the same set the prefix sweep uses, INCLUDING its path-derived
+        # fallback. `recorded_qns` is filled only when a file is parsed
+        # (`DefinitionProcessor.process_file`), so on a FRESH updater doing a
+        # scoped re-ingest it is empty -- and an empty `touched_modules` makes
+        # every rehydrated span foreign, which protects the whole graph from a
+        # sweep that then removes nothing. The re-parse re-registers a qn the
+        # rehydrated registry still holds and lands on a `@1` DUP_QN variant
+        # instead of the clean name (issue #1719). The prefix sweep already
+        # falls back to the path-derived qn for exactly this case; ownership
+        # has to fall back with it, or the two disagree about which modules
+        # this event touches.
+        touched_modules = module_qn_prefixes
         foreign_qns = set()
         # The mirror of foreign_qns, and the reason the prefix sweep is not
         # sufficient on its own: a definition's qn does not have to sit under
--- a/codebase_rag/tests/test_reingest.py
+++ b/codebase_rag/tests/test_reingest.py
@@ -304,7 +304,6 @@ def test_reingest_gives_same_stem_siblings_their_clean_index_qns(
     changed: list[str],
     deleted: list[str],
     fresh_updater: bool,
-    request: pytest.FixtureRequest,
 ) -> None:
     # The first same-stem sibling in walk order owns the bare module qn
     # (issue #1569). Adding the sibling that wins that order, or deleting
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

PR #1726, opened 2026-09-04
fix: distrust a hash-cache stamp from the future so edits are not skipped (#1665)

### LINKED ISSUE #1665: Hash-cache mtime in the future makes _is_already_in_sync skip every file, so edits are never indexed
## Summary

A `.cgr-hash-cache.json` whose mtime is in the **future** makes
`_is_already_in_sync` skip every file forever, so edits never reach the graph.
No single-file run is involved.

## Cause

`_is_already_in_sync` (`codebase_rag/graph_updater.py:2417`) fast-paths a file
when `stat.st_mtime <= cache_mtime` and the cached hash matches. If
`cache_mtime` is ahead of the wall clock, the first condition holds for every
file on disk regardless of how recently it was edited.

## How the stamp gets into the future

- a backwards system-clock adjustment (NTP correction, timezone-misconfigured VM)
- VM suspend/resume
- `cp -p` / `rsync -t` restoring a cache from a host whose clock ran ahead
- a restored backup or a copied container image

## Reproduction

1. Index a project normally.
2. Stamp the cache into the future:
   `os.utime(".cgr-hash-cache.json", (t, t))` with `t = time.time() + 3600`
3. Edit any source file.
4. Run a project sync.

Observed: the edit never reaches the graph. Expected: the edit is indexed.

## Note on why this surfaces now

Before #1619, a single-file run restamped the cache with `time.time()` on every
run, which accidentally healed a future stamp as a side effect. That restamping
was itself a data-loss bug (it hid sibling edits) and #1619 removed it, so this
pre-existing hole is no longer masked in that one path. The hole is independent
of #1619 and reproduces with no single-file run at all.

## Suggested fix

Either clamp on read:

```python
cache_mtime = min(cache_mtime, time.time())
```

or refuse the fast path when the stamp is in the future, and log it — a future
stamp indicates a clock or provenance problem worth surfacing rather than
silently correcting.

Found during review of #1664.

vitali87:
Closes #1665.

## The defect

`_is_already_in_sync` and the incremental hashing loop both fast-path a file
when `stat.st_mtime <= cache_mtime`, where `cache_mtime` is the mtime of
`.cgr-hash-cache.json`. A cache stamped **ahead of the wall clock** satisfies
that for every file on disk however recently edited, so nothing is re-indexed.

It does not self-correct: the stamp stays ahead, so every later run skips too.
Reproduced before fixing — index, `os.utime(cache, (time.time()+3600,)*2)`,
edit a file, re-run: runs 2 and 3 both index nothing.

The causes are ordinary rather than exotic: an NTP step backwards, VM
suspend/resume, `cp -p` or `rsync -t` from a host whose clock ran ahead, a
restored backup or a copied container image.

## The fix

`_trustworthy_cache_mtime` returns the stamp, or `0.0` when it is in the
future, and both readers use it.

**`0.0` rather than clamping to `time.time()`.** Clamping was my first attempt
and it does not work: "now" is never older than an edit made *now*, so a file
written in the same instant still satisfies `mtime <= now` and is skipped. `0.0`
disables the watermark and defers to the hash comparison. Cost is one re-hash
pass, after which the run rewrites the cache with a sane stamp — verified to
self-correct in exactly one pass.

## Three sites, three failure modes

The read guard alone was not enough, and the two extra findings both came from
review rather than from me.

| Site | Failure if unguarded |
|---|---|
| `_is_already_in_sync` (2929) | the fast path reports "in sync"; the edit is never indexed |
| the hashing loop (3097) | the shortcut `continue`s **before** the hash comparison, persists the stale hash, and re-stamps sanely — so the lie becomes permanent |
| the stamp **writer** (3540) | `cache_mtime or None` — `0.0` is falsy, so this took the `None` branch the comment beside it warns against, skipping `os.utime` and losing a sibling edit forever |

That third one is the one worth reading twice: **the read guard made the write
path worse.** A single-file run carries the previous stamp forward for the
siblings it never looked at; with `None` it instead keeps its own write time,
which is *later* than a sibling edited before the run. Reproduced: after one
single-file run over a poisoned cache, every subsequent project run reports
`in_sync=True` and the sibling's edit is gone. Fixed with `or 0.0` — the epoch
vouches for nothing, which is the honest claim; `None` vouches for now.

## A comment I shipped that was wrong

I originally guarded both readers, could not make a test discriminate the
second, investigated, and concluded in a comment that it had no failing case
because "the hash is the authority". **That was false**, and the reviewer
disproved it by construction: make the sync check decline for an *unrelated*
addition, and the poisoned stamp reaches the loop, whose shortcut skips the
edited file *before any hash comparison*. Measured under that mutation: cached
hash `77736f65…` vs actual `c8e27920…`, still diverged three runs later.

The comment is gone and there is a test in its place.

## Discrimination

| Mutation | Result |
|---|---|
| Stub the helper to return the raw stat | 2 red |
| Clamp to `time.time()` instead of `0.0` | red |
| Revert only the sync-check reader | red (that test only) |
| Revert only the hashing-loop reader | red (that test only) |
| Revert the stamp writer to `or None` | red (that test only) |

Each mutation fails its own test and nothing else. 131 passed across
`test_graph_updater_incremental.py`, `test_reingest.py` and
`test_hash_cache_guard.py`.

The first version of the main assertion was `_is_already_in_sync() is False`,
which is satisfied by *any* reason for declining — including a rescan that
finds nothing wrong. It pins "did not take the shortcut", not "did not lose the
edit". It now asserts the edited symbol reaches the graph, which fails against
the unfixed helper.

## Known gap, not fixed here

`_reingest_update_hashes` (4027, 4050) reads the raw stamp and writes it
straight back, so the watcher/MCP path can re-poison a cache. No edit is lost
today — reingest works from an explicit file list, and both batch readers are
now guarded — so this is robustness rather than a live defect, and it is a
different call path from the three above. Worth its own issue if you would
rather it were separate; happy to fold it in if you would rather it were not.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved change detection when cache timestamps are unreadable or incorrectly set in the future.
  * Ensured edited files are reprocessed and indexed instead of being hidden by stale cache metadata.
  * Preserved detection of edits to files not processed during single-file updates.

* **Tests**
  * Added regression coverage for future-dated cache timestamps, rehashing, indexing, and incremental update behavior.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -321,6 +321,36 @@ def _publish_hash_cache(
         return False
 
 
+def _trustworthy_cache_mtime(cache_path: Path) -> float:
+    """The cache mtime to compare files against, or 0.0 if it cannot be trusted.
+
+    Both readers use this instant as a watermark: "every file at or below it is
+    already accounted for". A stamp AHEAD of the wall clock makes that true of
+    EVERY file on disk however recently edited, so nothing is re-hashed, the
+    edit never reaches the graph, and no later run corrects it because the
+    stamp stays ahead (issue #1665). The causes are ordinary -- an NTP step
+    backwards, VM suspend and resume, `cp -p` or `rsync -t` from a host whose
+    clock ran ahead, a restored backup or a copied container image.
+
+    Returning 0.0 rather than clamping to `time.time()`: clamping is not
+    enough, because "now" is never older than an edit made now, so a file
+    written in the same instant still satisfies `mtime <= now` and is skipped
+    for the wrong reason. 0.0 disables the shortcut entirely and defers to the
+    HASH comparison, which is the authority the watermark is only an
+    optimisation over. The cost is one re-hash pass on a tree whose cache is
+    misstamped; the alternative is losing edits silently and permanently.
+
+    Shared by both readers on purpose. They answer the same question, and a
+    guard applied to one of two predicates is the shape that let the indexer
+    and the watcher disagree in issue #1636.
+    """
+    try:
+        stamped = cache_path.stat().st_mtime
+    except OSError:
+        return 0.0
+    return 0.0 if stamped > time.time() else stamped
+
+
 def _load_parser_fingerprint(stamp_path: Path) -> str | None:
     try:
         return stamp_path.read_text(encoding="utf-8").strip() or None
@@ -2907,7 +2937,7 @@ class GraphUpdater:
         # incidentally, by being an indexed and hashed file itself.
         if not self._exclusions_match_last_run():
             return False
-        cache_mtime = cache_path.stat().st_mtime
+        cache_mtime = _trustworthy_cache_mtime(cache_path)
         dir_mtimes_path = self.repo_path / cs.DIR_MTIMES_FILENAME
         old_hashes = (
             {} if self._cache_discarded_in_memory else _load_hash_cache(cache_path)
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1730, opened 2026-09-05
fix: preserve nested-class construction for optional receivers

### LINKED ISSUE #1680: An optional-union parameter annotation loses nested-class construction: o: Outer | None drops Inner()
A receiver annotated with an optional union does not resolve its nested-class construction, so `o.Inner()` produces no `INSTANTIATES` edge.

Raised by Greptile on #1678 and verified against both trees — it is **pre-existing on `main`**, not introduced by that branch, which is why it was filed rather than fixed there.

## Measurement

Fixture: `class Outer` with a nested `Inner`, four callers constructing `Inner` through differently-typed receivers.

| caller | receiver | edge recorded |
|---|---|---|
| `plain` | `o = Outer()` | yes |
| `annotated` | `def f(o: Outer)` | yes |
| `optional` | `def f(o: "Outer \| None")` | **no** |
| `optional_hint` | `o: Outer \| None = Outer()` | yes |

Identical on `origin/main` and on the #1678 branch. Only the optional-union **parameter annotation** is affected; an annotated local whose assignment gives a concrete inferred type is fine.

## Mechanism

`_resolve_class_qn_from_type` normalises through `_strip_optional` before resolving, so the upstream member-resolution path handles `Outer | None`. Consumers that call `_resolve_class_name` with the raw annotation do not, and a union string resolves to nothing.

## Suggested direction

Strip the optional before the class-name lookup wherever a declared type is turned into a class qn, rather than only in `_resolve_class_qn_from_type`. That makes the normalisation a property of the lookup instead of something each caller must remember, which is the shape of the current bug.

## Environment

```
cgr        0.0.842
repo       vitali87/code-graph-rag at main
verified   same result with and without the #1641 guard applied
```

atirna:
## Summary

- Normalize optional Python receiver annotations before deciding whether a nested class is constructed.
- Cover `o: Outer | None` constructing `Outer.Inner`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Fixes #1680

## Test Plan

- [x] Unit tests pass (`uv run pytest -q codebase_rag/tests/test_self_nested_class_construction.py codebase_rag/tests/test_call_resolver.py`)
- [x] New tests added
- [ ] Manual testing (describe below)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Nested-class construction now resolves correctly when the receiver type is optional, such as `X | None`.

* **Tests**
  * Added coverage confirming that nested-class construction from an optional receiver is represented correctly in the graph.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -491,7 +491,7 @@ class CallResolver:
         var_type = local_var_types.get(head)
         if not var_type:
             return False
-        typed_qn = self._resolve_class_name(var_type, module_qn)
+        typed_qn = self._resolve_class_name(self._strip_optional(var_type), module_qn)
         return bool(
             typed_qn
             and self.function_registry.get(typed_qn) == cs.NodeLabel.CLASS
--- a/codebase_rag/tests/test_self_nested_class_construction.py
+++ b/codebase_rag/tests/test_self_nested_class_construction.py
@@ -59,7 +59,7 @@ class Outer:
     def make(self):
         return self.Inner(1)
 
-
+    def make_optional(o: Outer | None): return o.Inner(1)
 class Base:
     def render(self):
         return "base"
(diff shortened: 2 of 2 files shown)


atirna:
thx for the review. I restored the independent `Outer.make -> Outer.Inner` `INSTANTIATES` assertion and kept the optional-receiver assertion; the focused resolver tests are green.

vitali87:
@greptileai review

Issue #1654, opened 2026-09-02
A .c file's #include emits no IMPORTS edge while a .h including the same header does


vitali87:
## What happens

Two files include a non-sibling header the same way; only one of them gets an IMPORTS edge.

- `bar.h` with `#include "foo.h"` produces `(Module proj.bar) -[IMPORTS]-> (Module proj.foo)`.
- `bar.c` with the same `#include "foo.h"` produces no IMPORTS edge at all.

Measured on a clean index of a two-file fixture per variant, main at ee6458f2 (reviewer probe while verifying #1573):

```
c-includes-nonsibling-header []
h-includes-nonsibling-header [('Module','proj.bar','IMPORTS','Module','proj.foo')]
```

## Why it matters

A `.c` translation unit's includes are invisible to every consumer of IMPORTS edges (dependents queries, importer lists, the incremental re-parse set). In #1573's test fixtures a `.c` includer had no edge into the header pair at all, so a dependent could only be tied in through a CALLS edge to an inline function.

## Scope

Diagnose why the C include handling differs between `.c` and `.h` sources (the same include syntax, same language spec) and make a `.c` file's `#include` emit the same IMPORTS edge a `.h` file's does. Add a test with both file kinds including the same header and assert the edge set is identical.

PR #1757, opened 2026-09-06
fix: match the std token when prefixing a system include so stdio.h lands under std. like signal.h

### LINKED ISSUE #1744: C/C++ system includes starting with "std" are misnamed: <stdio.h> becomes ExternalModule stdio.h while <signal.h> becomes std.signal.h
## Observed

`ImportProcessor._parse_cpp_include` (`codebase_rag/parsers/import_processor.py`, the `is_system_include` branch) decides whether a system include already carries the std prefix with

```python
include_path if include_path.startswith(cs.CPP_STD_PREFIX) else f"{cs.IMPORT_STD_PREFIX}{include_path}"
```

`CPP_STD_PREFIX` is `"std"`, so this is a substring match on the header name. `<stdio.h>`, `<stdlib.h>`, `<stdint.h>`, `<stdbool.h>`, `<stddef.h>` and `<stdarg.h>` are taken as already prefixed and become ExternalModule `stdio.h`, `stdlib.h`, ..., while `<signal.h>`, `<string.h>` or `<vector>` in the same file become `std.signal.h`, `std.string.h`, `std.vector`. Observed both shapes in one indexing run while reviewing #1654.

Identical on the C++ path today; #1654 exposes it to C, where these six are the most common includes.

## Expected

Match the token, not the substring: `include_path == "std"` or `include_path.startswith("std.")` (or `std::`), so every system header lands under the same `std.` prefix.

## Related

Same defect class as [[match-tokens-not-substrings]] in #1485/#105; found during the local review of the #1654 fix.

vitali87:
Closes #1744.

`ImportProcessor._parse_cpp_include` decided whether a system include already carried the std prefix with a `startswith("std")` test on the header name, which is a substring match. `<stdio.h>`, `<stdlib.h>`, `<stdint.h>`, `<stdbool.h>`, `<stddef.h>` and `<stdarg.h>` were taken as already prefixed and became ExternalModule `stdio.h` and so on, while `<signal.h>` or `<vector>` in the same file became `std.signal.h` and `std.vector`. Since #1654 the same branch serves C, where that family is the most common set of includes.

The test now matches the token: the path is left alone only when it is exactly `std` or starts with `std.`; everything else takes the `std.` prefix. The regression indexes a file with `<stdio.h>`, `<stdlib.h>`, `<signal.h>` and `<vector>` from a `.c` and a `.cpp` unit and asserts every IMPORTS target is under `std.`; on `main` two of the four land outside it.

### DIFF
--- a/codebase_rag/parsers/import_processor.py
+++ b/codebase_rag/parsers/import_processor.py
@@ -4167,9 +4167,15 @@ class ImportProcessor:
                 local_name = header_name
 
             if is_system_include:
+                # The token, not the substring: `startswith("std")` took
+                # `<stdio.h>`, `<stdlib.h>`, `<stdint.h>` and the rest of that
+                # family as already prefixed, so they became ExternalModule
+                # `stdio.h` while `<signal.h>` in the same file became
+                # `std.signal.h` (issue #1744).
                 full_name = (
                     include_path
-                    if include_path.startswith(cs.CPP_STD_PREFIX)
+                    if include_path == cs.CPP_STD_PREFIX
+                    or include_path.startswith(cs.IMPORT_STD_PREFIX)
                     else f"{cs.IMPORT_STD_PREFIX}{include_path}"
                 )
             elif resolved := self._resolve_cpp_include_target(include_path, module_qn):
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

Issue #1669, opened 2026-09-02
main carries 411 open Sonar issues; refactor per function, tests first, not gate-related


vitali87:
## What is happening

The Sonar Zero Issues Gate fails on #1662 with 12 open issues, 11 of them in functions that PR does not touch:

```
[CRITICAL] shell_command.py:169  Cognitive Complexity 17 > 15 (python:S3776)   _git_escapes_project
[CRITICAL] shell_command.py:259  Cognitive Complexity 18 > 15
[CRITICAL] shell_command.py:384  Cognitive Complexity 31 > 15
[CRITICAL] shell_command.py:446  Cognitive Complexity 28 > 15
[CRITICAL] shell_command.py:529  Cognitive Complexity 38 > 15
[CRITICAL] shell_command.py:650  Cognitive Complexity 36 > 15
[CRITICAL] shell_command.py:699  Cognitive Complexity 18 > 15
[CRITICAL] shell_command.py:753  Cognitive Complexity 24 > 15
[CRITICAL] shell_command.py:779  Cognitive Complexity 73 > 15
[CRITICAL] shell_command.py:1038 Cognitive Complexity 25 > 15
[CRITICAL] shell_command.py:1117 Cognitive Complexity 18 > 15
[MAJOR]    shell_command.py:947  Either merge this branch with the identical one on line 940 (python:S1871)
```

## Why

SonarCloud's last successful analysis of `main` is `ee6458f2` (v0.0.844, 2026-09-02 20:14 UTC). The advisory merge `033b64be` then added 945 lines to `codebase_rag/tools/shell_command.py` across 45 commits that were developed in the GHSA private fork, where Actions never run, so none of it was Sonar-analysed. The SonarCloud workflow on `main` has failed on both commits since (`033b64be`, `57677f8f`) because its "Run tests with coverage" step fails on the `xargs -0pt cat` case that #1662 fixes.

SonarCloud reports on a PR every issue that is absent from the target branch's last analysis, so these 12 are charged to whichever PR next touches the file. The gate is doing its job: the advisory merge introduced them and they never passed through it.

## The circularity

`main` cannot be re-analysed until its tests pass in the Sonar workflow; its tests cannot pass until #1662 merges; #1662 cannot merge while the gate (a `needs` of the required `All Checks Pass` job) counts these 12. Every other check on #1662 is green, including both `windows-latest` jobs, Greptile 5/5 and CodeRabbit at minimal risk, both anchored to the head.

[...]

PR #1760, opened 2026-09-06
test: split every composite assertion into one assertion per operand (Sonar S9073)

vitali87:
Part of #1669: the composite-assertion findings (python:S9073), which are 72 of the 411 open Sonar issues on `main`.

Every `assert a and b` SonarCloud reported was split into one assertion per operand, keeping the original message on each, so a failure names the operand that failed instead of the whole conjunction. The rewrite was done from the AST at the reported positions and then formatted, so only the reported assertions changed and no `and` inside a nested expression was touched. Short-circuit order is preserved: where a later operand relied on an earlier one being true, the earlier assertion now fails first.

57 test files under `codebase_rag/tests/` change; no production code does. Every touched file was run: 431 passed, 4 skipped (grammar-dependent skips that skip on `main` too).

## Summary by CodeRabbit

- **Tests**
  - Improved test diagnostics by splitting combined assertions into independent checks.
  - Precision, recall, type, presence, and value validations now report failures more specifically.
  - Test coverage and expected behavior remain unchanged.

### DIFF
--- a/codebase_rag/tests/test_ast_grep_service.py
+++ b/codebase_rag/tests/test_ast_grep_service.py
@@ -47,7 +47,8 @@ def test_replace_dry_run_produces_diff_without_writing(tmp_path: Path) -> None:
     target.write_text("print(x)\n")
     svc = AstGrepService(str(tmp_path))
     changes = svc.replace("print($A)", "log($A)", dry_run=True)
-    assert changes and changes[0]["applied"] is False
+    assert changes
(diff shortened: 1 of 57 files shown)


vitali87:
@greptileai review

vitali87:
Rebased onto current main (ab611856); the previous review compared this test-only branch against a main that had since merged #1701, so the hash-cache finding it reported was the base moving, not this diff.

@greptileai review

PR #1761, opened 2026-09-06
test: one raising call per pytest.raises block (Sonar S5778)

vitali87:
Part of #1669: the exception-test findings (python:S5778), six of the open Sonar issues on `main`.

Each `pytest.raises` block now holds exactly one invocation that can raise. The four async decorator tests build the coroutine outside the block and only `asyncio.run` it inside; the service-lifecycle test creates the context manager outside and enters it inside; the retry test builds its points list outside and calls `_upsert_with_retry` inside. The exception in every case still comes from the call the test is about, and nothing can now raise before the block is entered.

Three test files change, no production code. The three files pass: 65 passed.

## Summary by CodeRabbit

- **Tests**
  - Improved exception-handling test setup for asynchronous operations and service lifecycle failures.
  - Clarified retry-exhaustion test data preparation.
  - Preserved existing exception behavior and assertions.

### DIFF
--- a/codebase_rag/tests/test_decorators.py
+++ b/codebase_rag/tests/test_decorators.py
@@ -140,8 +140,9 @@ class TestAsyncTimingDecorator:
             async def async_failing() -> None:
                 raise ValueError("async error")
 
+            failing = async_failing()
             with pytest.raises(ValueError, match="async error"):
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

PR #1762, opened 2026-09-06
refactor: collect dead-code roots through one predicate instead of eighteen identical branches (Sonar S1871)

vitali87:
Part of #1669: the identical-branch findings (python:S1871) in `codebase_rag/dead_code.py`, eighteen of the open Sonar issues on `main`.

The root-collection loop tested a symbol against eighteen `elif` rules that each did the same thing, add the symbol to `roots`. Those rules are alternatives, not a priority order, so they now live in one predicate, `_is_root`, which returns on the first rule that matches, in the original order and with the original arguments, and the loop adds the symbol when it returns True. The per-rule comments moved with their rules. The helper also computes `qn in method_qns` and the paren-stripped leaf once instead of at every rule.

No behaviour changes: the dead-code test files (107 tests) pass, and the per-language root rules they cover (Python dunders and enum hooks, Go and Rust entry points, JS well-known symbols, C and C++ entry and operator roots, Java serialisation, C# attributes, Dispose and finalizers, Nest and React, entry points and test symbols) all go through the new predicate.

## Summary by CodeRabbit

* **Refactor**
  * Centralized reachability-root checks to improve consistency and maintainability.
  * Preserved existing dead-code detection behavior while simplifying candidate evaluation.

### DIFF
--- a/codebase_rag/dead_code.py
+++ b/codebase_rag/dead_code.py
@@ -489,6 +489,95 @@ def _walk(
                 stack.append(nxt)
 
 
+def _is_root(
+    qn: str,
+    props: PropertyDict,
+    config: DeadCodeConfig,
+    method_qns: set[str],
+    protocol_stubs: set[str],
+    method_to_class: dict[str, str],
+    class_decorators_norm: dict[str, frozenset[str]],
+    nest_factory_classes: set[str],
+    react_component_classes: set[str],
+    rust_test_modules: set[str],
+    rust_test_spans: dict[str, list[tuple[int, int]]],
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
Sonar scored the new predicate itself at 25: nineteen `if ... return True` branches. In 287cce14 the rules are a tuple of thunks evaluated lazily and in order by `any`, which keeps the first-match semantics with no branch per rule; the comments stay with their rules. Dead-code tests pass (107).

@greptileai review

PR #1763, opened 2026-09-06
refactor: merge the three remaining pairs of identical branches (Sonar S1871)

vitali87:
Part of #1669: the last three identical-branch findings (python:S1871) outside `dead_code.py`, which PR #1762 handles.

Each pair of branches did the same thing under two conditions and is now one branch. In the JS/TS handler the declared name and the assignment-derived name are tried in the same order through one `or` of walrus assignments, so the second lookup still runs only when the first yields nothing. In the Python binding collector, assignment and for-statement nodes share the branch that collects their `left` targets. In the sed validator, a standalone `-e` or `--expression` and a bundled cluster ending in a bare `-e` share the branch that consumes the next token; the comment explaining the cluster case is kept with it, and the later `-eSCRIPT` branch is reached for the same inputs as before.

No behaviour changes. The shell-command, JavaScript and TypeScript test files pass: 1404 passed, 1 skipped.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal name-resolution and target-binding logic without changing behavior.
  * Consolidated equivalent shell expression handling paths while preserving existing command behavior.

### DIFF
--- a/codebase_rag/parsers/handlers/js_ts.py
+++ b/codebase_rag/parsers/handlers/js_ts.py
@@ -99,9 +99,10 @@ class JsTsHandler(BaseLanguageHandler):
 
         while current and current.type not in lang_config.module_node_types:
             if current.type in lang_config.function_node_types:
-                if name := self._extract_node_name(current):
-                    path_parts.append(name)
-                elif name := self.extract_function_name(current):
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

PR #1764, opened 2026-09-06
chore: pytest.fail over assert False, returning fixtures without teardown, no empty match arms (Sonar S9077, S9100, S108)

vitali87:
Part of #1669: three small rules, fifteen of the open Sonar issues on `main`.

`python:S9077`: the five `assert False, message` sites in `test_nested_function_qualified_names.py` become `pytest.fail(message)`, which is what they meant.

`python:S9100`: three fixtures that had no teardown (`isolated_cgr_home` in `conftest.py` and the `_temp_home` fixtures in `test_cgr_state_and_status.py` and `test_workspaces.py`) return their path instead of yielding it, and the autouse `reset_session` fixture in `test_status_bar_config.py` drops a bare `yield` that had nothing after it. A returning fixture still runs for every test that requests it, and the monkeypatch teardown is pytest's own, not the fixture's.

`python:S108`: five `case _: pass` arms are removed from the Go and Java type-inference and variable-analysis matchers, and the do-nothing `case cs.TS_STATIC: pass` arm from the Java import parser. A `match` without a wildcard falls through on an unmatched value, and the `static` keyword matches none of the remaining arms, so no branch changes.

The nested-function, state, status-bar, workspace, Java and Go test files pass: 573 passed, 15 skipped (grammar-dependent skips that skip on `main` too).

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal parsing and type-analysis control flow without changing behavior.
  * Streamlined test fixture setup and return handling.

* **Tests**
  * Improved failure reporting in nested function and export tests.
  * Removed unused test imports and clarified fixture type annotations.

### DIFF
--- a/codebase_rag/parsers/go/type_inference.py
+++ b/codebase_rag/parsers/go/type_inference.py
@@ -189,8 +189,6 @@ class GoTypeInferenceEngine:
                 self._collect_var_declaration(node, var_types)
             case cs.TS_GO_SHORT_VAR_DECLARATION:
                 self._collect_short_var_declaration(node, var_types)
-            case _:
-                pass
         for child in node.children:
             self._collect_body_declarations(child, var_types)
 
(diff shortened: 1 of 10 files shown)


vitali87:
@greptileai review

PR #1765, opened 2026-09-06
refactor: accumulate one sampled speedscope stack in its own helper (Sonar S3776)

vitali87:
Part of #1669: one cognitive-complexity finding (python:S3776), `_accumulate_sampled` in `codebase_rag/trace/speedscope.py`, at 16 against the limit of 15.

The per-stack inner loop, which walks one sampled stack root first and adds the sample's weight to the edge between each in-scope frame and its nearest in-scope ancestor, is now its own helper, `_accumulate_stack`. The outer loop keeps its shape: a non-list sample or a bad frame index anywhere still aborts the whole accumulation with False, leaving what was already accumulated, and the ancestor still resets for every stack.

No behaviour changes. The full unit suite was run against the change: 10711 passed, 40 skipped, 1 xfailed.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined sampled-profile stack processing while preserving existing behavior and output accuracy.
  * Maintained correct handling of weighted relationships in sampled profiles.

### DIFF
--- a/codebase_rag/trace/speedscope.py
+++ b/codebase_rag/trace/speedscope.py
@@ -82,17 +82,30 @@ def _accumulate_sampled(
         if not isinstance(stack, list):
             return False
         weight = _sample_weight(cast("list[object]", weights), position)
-        ancestor: str | None = None
-        for frame_index in stack:
-            if not _valid_index(frame_index, len(names)):
-                return False
-            current = names[frame_index]
-            if current is None:
-                continue
-            if ancestor is not None:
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
Sonar flagged the cast literal repeated a third time; 800c6bf7 types the helper's parameter as `Sequence[object]` instead, which the narrowed list satisfies without a cast (ty clean, tests pass).

@greptileai review

PR #1766, opened 2026-09-06
refactor: split resolve_class_name into its three lookup steps (Sonar S3776)

vitali87:
Part of #1669: the first of the cognitive-complexity findings (python:S3776), `resolve_class_name` in `codebase_rag/parsers/py/utils.py`, which sat at 16 against the limit of 15.

The lookup has three steps that were interleaved in one function: the import map (with the fall-through for a mapped name that is not a registered node, the #652 phantom-caller guard), the module itself and each enclosing package, and finally the registry's search by simple name with its preference for a candidate nested in the current module. Each step is now its own helper and `resolve_class_name` is the `or` chain of the three in the original order. Conditions, return values and the comments that explain the two non-obvious steps are unchanged; the only structural difference is that the import-map step returns None to fall through instead of continuing inline.

No behaviour changes. The resolver, receiver, nested-class, Python, C++ and Java inheritance test files pass: 695 passed, 1 skipped, 1 xfailed (the skip and xfail are the same on `main`).

### DIFF
--- a/codebase_rag/parsers/py/utils.py
+++ b/codebase_rag/parsers/py/utils.py
@@ -16,30 +16,61 @@ def resolve_class_name(
     function_registry: FunctionRegistryTrieProtocol,
     require_registered: bool = False,
 ) -> str | None:
-    if module_qn in import_processor.import_mapping:
-        import_map = import_processor.import_mapping[module_qn]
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

PR #1767, opened 2026-09-06
refactor: split the C++ field-declarator name reader and the Go selector walk (Sonar S3776)

vitali87:
Part of #1669: two cognitive-complexity findings (python:S3776), both at 16 against the limit of 15.

In `codebase_rag/parsers/cpp/utils.py`, `_extract_name_from_field_declaration` scanned the children once to see whether any `function_declarator` existed and then scanned them again to read a name. The pre-scan is gone, since the loop answers the same question, and reading one declarator's name (its `declarator` field when that is a field identifier, otherwise the first field-identifier child) is now `_function_declarator_field_name`. The first declarator that yields a name still wins, and a declarator that yields none still lets the loop move to the next.

In `codebase_rag/parsers/go/type_inference.py`, the walk that turns a selector chain such as `e.trees.get` into its identifier segments moves out of `callee_segments` into `_selector_segments`; `callee_segments` keeps the plain-identifier case and the guard that any empty segment makes the callee unresolved.

No behaviour changes. The C++ and Go test files pass: 522 passed, 1 xfailed (the xfail is the same on `main`).

The local review found that no test pinned the declarator-list order the C++ reader preserves, so `test_cpp_field_declarator_list.py` parses `class A { void (*fp)(int), g(); };` and the three-declarator form and asserts the first declarator that carries a name wins.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved C++ parsing for field declarations with multiple declarators, correctly identifying the first named declarator.
  * Improved Go selector expression handling for more reliable type inference.
* **Tests**
  * Added coverage for C++ field declarations containing multiple declarators.

### DIFF
--- a/codebase_rag/parsers/cpp/utils.py
+++ b/codebase_rag/parsers/cpp/utils.py
@@ -183,28 +183,27 @@ def _extract_name_from_declaration(func_node: Node) -> str | None:
 
 
 def _extract_name_from_field_declaration(func_node: Node) -> str | None:
-    has_function_declarator = any(
-        child.type == cs.CppNodeType.FUNCTION_DECLARATOR for child in func_node.children
-    )
-    if not has_function_declarator:
-        return None
-
     for child in func_node.children:
-        if child.type == cs.CppNodeType.FUNCTION_DECLARATOR:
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

Issue #1738, opened 2026-09-05
remove_file_from_state never prunes method_return_types, so a deleted file's C++/Rust/Dart free-function and Go method return types persist


vitali87:
## Observed

`TypeInferenceEngine.method_return_types` is filled in `codebase_rag/parsers/function_ingest.py` for C++, Rust and Dart free functions and for Go receiver methods, and `GraphUpdater.remove_file_from_state` never touches it (zero references in `graph_updater.py`). On a reused updater a deleted file's entries persist: after deleting `pkg/a.go`, `function_registry` no longer holds `proj.pkg.a.A.Clone` but `method_return_types["proj.pkg.a.A.Clone"] == "A"`.

Measured in the local review of the #1668 fix, which prunes the sibling `go_function_return_types` map for exactly this reason and leaves this one alone.

## Expected

`remove_file_from_state` drops the deleted file's `method_return_types` entries with the same ownership filter it applies to the function registry (module-qn prefix, span-record ownership for Go methods keyed by their receiver type's module, `foreign_qns` excluded), so a re-parse re-records them and a deletion forgets them.

Same defect class as #1668 for a different map.

Issue #1668, opened 2026-09-02
TypeInferenceEngine._go_free_fn_index keeps a deleted file's return-type entries; a same-name sibling gets the stale one


vitali87:
## Observed

\`codebase_rag/parsers/type_inference.py\` builds \`_go_free_fn_index\` as a size-sentinel index over \`go_function_return_types\`. \`GraphUpdater.remove_file_from_state\` never prunes \`go_function_return_types\`, so a deleted Go file's entries persist on a reused updater. The index is filled with \`setdefault\`, so when a surviving or new sibling defines a free function of the same name, the stale entry from the deleted file wins and the sibling's return type is never recorded.

Noted (not probed) during the round-4 local review of #1576; the map grows on each run so the index does rebuild, which is why it is not a regression of that PR.

## Expected

\`remove_file_from_state\` drops the deleted file's \`go_function_return_types\` entries along with its other registry state, and the index is rebuilt or cleared with the other per-run memos.

## Related

- #1576 (resets the resolver's other module-map-derived memos); #1659, #1660 for the other Go reuse defects.

vitali87:
Variant noted in the round-5 local review of #1576: the same size sentinel also misses a changed return type on an existing qn (same key, same map size), so an edited Go free function keeps its old return type in the index until the map's size happens to change.

PR #1768, opened 2026-09-06
fix: prune a deleted file's C# method return types with the same ownership filter as the other return-type maps

### LINKED ISSUE #1753: remove_file_from_state never prunes csharp_method_return_types, so a deleted C# file's method return types persist
## Observed

`TypeInferenceEngine.csharp_method_return_types` (a `dict[str, tuple[str, int]]` filled by the C# type inference) is never touched by `GraphUpdater.remove_file_from_state`. On a reused updater, after `run()` then deleting `A.cs` and calling `remove_file_from_state(A.cs)`, the registry row for `proj.A.N.K.Self` is gone but `csharp_method_return_types` still holds `{'proj.A.N.K.Self': ('K', 0)}`.

Measured in the local review of #1752, which prunes the sibling `method_return_types` map (#1738) and leaves this one alone.

## Expected

`remove_file_from_state` drops the deleted file's `csharp_method_return_types` entries with the same ownership filter it applies to the function registry and to the other two return-type maps (module-qn prefix, span-record ownership, `foreign_qns` excluded), so a re-parse re-records them and a deletion forgets them.

Same defect class as #1668 and #1738 for a third map.

vitali87:
Closes #1753.

`TypeInferenceEngine.csharp_method_return_types` records a `(type, arity)` pair per C# method qualified name and is read by the C# type-inference engine through the same dict. `remove_file_from_state` prunes the function registry, the Go free-function return-type map (#1668) and the general method return-type map (#1738), but never this one, so on a reused updater a deleted C# file's entries persisted: after indexing, deleting `A.cs` and removing its state, the registry row for `proj.A.N.K.Self` was gone while the map still held `('K', 0)`.

The sweep now runs the ownership filter from #1752 over the union of both maps' keys, and `drop_method_return_types` pops the stale names from both. Neither map has a derived index, so nothing else needs invalidating, and the union cannot change which general-map entries go because the filter is applied per key.

The new test indexes two C# files, deletes one, removes its state and asserts its method's entry is gone while the sibling's stays. It fails on `main` with the entry still present, exactly as the issue describes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale C# method return-type information remaining after a file is deleted or reprocessed.
  - Preserved return-type information for files that remain in the project.

- **Tests**
  - Added regression coverage verifying that deleted C# files are removed from return-type tracking without affecting sibling files.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2983,7 +2983,14 @@ class GraphUpdater:
         # writes `...A.Clone`, its span record says `...A.Clone@3`), and a
         # few writers (deferred Rust functions, duplicate C++ out-of-class
         # methods) key the map by the `@line` name itself (#1752 review).
-        method_return_types = self.factory.type_inference.method_return_types
+        # The C# map (`csharp_method_return_types`, keyed the same way with a
+        # (type, arity) value) is the same defect for a third map (issue
+        # #1753), so both are swept by one filter.
+        type_inference = self.factory.type_inference
+        method_return_types = (
+            type_inference.method_return_types.keys()
+            | type_inference.csharp_method_return_types.keys()
+        )
         owned_natural = {_natural_qn(qn) for qn in owned_qns}
         stale_method_qns = [
             qn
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
Sonar's composite-assertion finding is split in cea1393b; the test passes.

@greptileai review

PR #1771, opened 2026-09-06
refactor: split the C++ include parser and the macro body scan (Sonar S3776)

vitali87:
Part of #1669: two cognitive-complexity findings (python:S3776), both at 16 against the limit of 15.

In `codebase_rag/parsers/import_processor.py`, `_parse_cpp_include` did three things in one body: read the include's path and kind from the node, derive the local name, and decide the module qn. Those are now `_cpp_include_spec` (the last string or system-lib child wins, an empty path records nothing), `_cpp_include_local_name` (the stem for `.h`/`.hpp`, the bare segment for `<vector>`) and `_cpp_include_full_name` (the `std.` token rule from #1744 for a system header, the resolved repo file for a quoted include that matches one, the `std.` prefix for a third-party header). The recording and the debug line are unchanged.

In `codebase_rag/parsers/cpp_frontend/frontend.py`, the token scan that finds the identifiers a macro body references moves out of `_process_macro_definition` into `_macro_body_identifiers`, with `_macro_parameter_list` deciding whether the macro is function-like (the `(` must abut the name) and which names are its parameters; the fallback when no closing parenthesis is found is kept.

No behaviour changes. The include and C++ frontend test files pass with libclang present: 55 passed.

## Summary by CodeRabbit

- **Refactor**
  - Improved internal processing of C++ macros and include directives while preserving existing behavior.
  - Maintained handling for quoted, system, and header-file includes, including name resolution and collision disambiguation.
  - C++ include statements without a usable path are handled more cleanly.

### DIFF
--- a/codebase_rag/parsers/cpp_frontend/frontend.py
+++ b/codebase_rag/parsers/cpp_frontend/frontend.py
@@ -19,7 +19,7 @@ from . import constants as fc
 from .qn import CppQnResolver
 
 if TYPE_CHECKING:
-    from clang.cindex import Cursor
+    from clang.cindex import Cursor, Token
 
 _NodeKey = tuple[str, str]
 _EdgeKey = tuple[str, str, str, str, str]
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

PR #1778, opened 2026-09-07
fix: prune the four remaining C# side tables when a file is deleted (#1769)

### LINKED ISSUE #1769: remove_file_from_state never prunes four C# side tables: generic methods, class generic arity, local functions and extension methods
## Observed

After `run()` on a project with `A.cs`, then deleting the file and calling `remove_file_from_state(A.cs)`, four C# side tables on `TypeInferenceEngine` still hold the deleted file's entries on a reused updater:

- `csharp_generic_methods`: `{'proj.A.N.K.Gen'}`
- `csharp_class_generic_arity`: `{'proj.A.N.K': 1}`
- `csharp_local_functions`: `{'proj.A.N.K.Loc.inner': ...}`
- `csharp_extension_methods`: `{'Twice': [('proj.A.N.Ext.Twice(int)', ...)]}`

`remove_file_from_state` touches only `csharp_partial_groups` among the C# side tables. Measured in the local review of #1768 (probe over the four maps before and after the delete).

## Expected

Each table drops the deleted file's entries with the same ownership filter the registry sweep and the three return-type maps use (module-qn prefix, span-record ownership on the natural qn, `foreign_qns` excluded), so a re-parse re-records them and a deletion forgets them. `csharp_extension_methods` is keyed by method name with a list of owners, so its prune removes the deleted owners from each list rather than the key.

Same defect class as #1668, #1738 and #1753, for four more maps.

vitali87:
Closes #1769.

## The defect

`remove_file_from_state` reached the registry, both return-type maps and `csharp_partial_groups`, but never `csharp_generic_methods`, `csharp_class_generic_arity`, `csharp_local_functions` or `csharp_extension_methods`. A deleted file's entries survived on a reused updater and went on steering resolution at a definition that is gone.

Same defect class as #1668, #1738 and #1753: a deleted file's side tables are never pruned.

## The fix

Three of the four maps key by the definition's qn, so they take the same ownership filter the registry sweep and the return-type maps already use.

`csharp_class_generic_arity` keys by a **class** qn instead, and that needed its own answer. Classes register no function span, so `owned_natural` can never name one and `foreign_qns` is vacuously false for them — the existing filter cannot protect a class at all. Ownership comes from the module index instead, by **longest** matching module qn:

```python
class_qns = {
    qn
    for qn in type_inference.csharp_class_generic_arity
    if self._owning_module(qn, module_index) in module_qn_prefixes
}
```

Longest rather than any prefix, because module qns nest. `proj/A.cs` records `proj.A` and `proj/A/B.cs` records `proj.A.B`, so both are prefixes of `proj.A.B.N.Nested`, which only `B.cs` owns. Matching any prefix errs in **both** directions, and the two errors are each other's mirror:

| deleted | `proj.A.B.N.Nested` should be | any-prefix | survivor-exclusion | longest |
|---|---|---|---|---|
| `proj/A.cs` | kept (B.cs owns it) | swept | kept | kept |
| `proj/A/B.cs` | swept (B.cs owns it) | swept | **kept** | swept |

The middle column is why this took two commits: excluding everything under a surviving module fixes the first row and silently reinstates #1769 on the second, for every file whose parent directory shares a sibling's stem.

## Verification

- 4 tests, one per direction plus the two original shapes. Each mutation reddens exactly one, and they are different ones: reverting to survivor-exclusion reddens the nested-file test; reverting to prefix-only reddens the sibling-directory test.
- `drop_csharp_side_tables` mutates in place — the lazily built `CSharpTypeInferenceEngine` shares these objects by reference, so rebinding them would leave the engine on the old ones.
- On the never-parsed-file fallback (`module_qn_prefixes = {path_derived_qn}`) the derived qn is absent from the module index, so this filter prunes nothing. Deliberate and documented: under-pruning a stale entry costs a wrong resolution, over-pruning a live sibling's costs a lost one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale C# type-resolution results after deleting a file.
  - Removed deleted files’ generic method, class generic arity, local function, and extension method metadata from subsequent analysis.
  - Improved handling of similarly named files, directories, namespaces, and sibling modules so remaining code continues resolving correctly.
  - Preserved type-inference information for files that were not deleted.

- **Tests**
  - Added regression coverage for file deletion and C# type-resolution scenarios.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -3166,12 +3166,17 @@ class GraphUpdater:
         the filter the registry sweep and the return-type maps use: already
         swept, or owned by this file's span records on the natural qn, or
         under one of its module prefixes with `foreign_qns` excluded.
-        `csharp_class_generic_arity` keys by a CLASS qn instead, and classes
-        record no function span, so `owned_natural` can never name one; the
-        module-qn prefix is the only ownership available, which is sound
-        because a C# class qn always sits under the module qn of the file
-        declaring it (a partial type's other halves are separate qns under
-        their own modules, each pruned by its own deletion).
+        `csharp_class_generic_arity` keys by a CLASS qn instead. Classes
+        record no function span, so `owned_natural` can never name one and
+        `foreign_qns` is vacuous for them; ownership comes from the module
+        index instead. A class qn does sit under the module qn of its
+        declaring file, but it can equally sit under a SHORTER module qn
+        belonging to a different file: `proj/A.cs` records `proj.A` and
+        `proj/A/B.cs` records `proj.A.B`, so deleting `A.cs` on the prefix
+        alone would take `B.cs`'s `proj.A.B.N.Nested` with it. A qn under a
+        surviving module is therefore kept, whichever prefix also matches.
+        A partial type's other halves are separate qns under their own
+        modules and each is pruned by its own deletion.
         """
 
         def owned(qn: str) -> bool:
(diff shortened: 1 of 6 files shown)


vitali87:
@greptileai review

vitali87:
Sonar's one open issue is a **false positive**, and `f1475f53` adds the evidence rather than dismissing it in the UI.

**The finding.** `python:S6199` — "Remove this unnecessary `list()` call on an already iterable object" at `codebase_rag/parsers/type_inference.py:674`.

(Note the Quality Gate shows *passed* beside this: the gate and the open-issue count answer different questions, so the Zero Issues Gate is the one that fails.)

**Why it is wrong.** That loop deletes from the dict it iterates:

```python
for name in list(self.csharp_extension_methods):
    ...
    else:
        del self.csharp_extension_methods[name]
```

Applying Sonar's suggestion gives `RuntimeError: dictionary changed size during iteration`.

**Measured, by applying the suggestion and running the suite:**

| | result |
|---|---|
| `list()` present | 6 passed |
| `list()` removed (Sonar's fix) | 3 failed — `RuntimeError: dictionary changed size during iteration` |
| restored | 6 passed |

The three reds are `test_a_deleted_csharp_files_side_tables_are_forgotten`, `test_a_sibling_directory_sharing_the_deleted_files_stem_is_kept` and the new `test_dropping_the_last_owner_of_an_extension_name_removes_the_key`. The first two are full parser-driven runs, so the delete branch is reached through the real call path and not only through a hand-built fixture.

The new test also pins the other half of the branch: with two owners under one extension name and one removed, the name keeps its key and loses only the removed entry; a name whose last owner goes is dropped entirely.

A dismissal in SonarCloud would be invisible from the code — the next reader sees an "unnecessary" `list()` and no reason for it. A test that reddens when it is removed puts the reason where they will find it. Marking the issue as a false positive in SonarCloud to clear the gate.

vitali87:
@greptileai review

PR #1779, opened 2026-09-07
fix: a single-file run must not prune Modules it never walked (#1756)

### LINKED ISSUE #1756: A GraphUpdater constructed on a file in a subdirectory makes that subdirectory the repo root and prunes every Module outside it
## Observed

`GraphUpdater(repo_path=<existing file>)` sets `repo_path` to the file's parent. For a file in a subdirectory (`pkg/module_a.py`) that parent is `pkg/`, so with the same project name the run's orphan prune (`_prune_orphan_nodes`) treats every Module outside `pkg/` as missing from disk and deletes it, and the run reads and writes a different hash cache (`pkg/.cgr_hash_cache` rather than the project root's).

Measured in the local review of #1755: after a project run then a single-file run constructed on `pkg/module_a.py`, the graph held no Modules at all (`modules=[]`). Identical on `main`; the next project run repairs it.

## Expected

Either the single-file constructor locates the project root (the enclosing directory that owns the hash cache, or the nearest ancestor the caller names) and keys the target relative to it, or the single-file run confines its orphan prune to the target the way it already confines `deleted_keys`. It must not delete Modules it was not asked about.

Found during the local review of the #1737 fix.

vitali87:
Closes #1756.

## The defect

`GraphUpdater(repo_path=<a file>)` sets `repo_path` to the file's **parent**. For a target in a subdirectory (`pkg/module_a.py`) that parent is `pkg/`, and `_prune_orphan_nodes` then resolves every Module's relative path against `pkg/`. A module recorded as `root_module.py` is tested as `pkg/root_module.py`, does not exist, and is deleted.

Every Module outside the target's own directory goes. The measured result in the issue was a graph holding **no Modules at all**.

Two things combine to make it total rather than partial:

- `CYPHER_ALL_MODULE_PATHS_INTERNAL` returns `path` and `qualified_name` but **no `absolute_path`**, so the component-aware containment gate that protects File, Folder and Package rows never runs for Modules.
- The remaining test is `(self.repo_path / path).exists()`, which is exactly the one that misreads the root.

## The fix

The prune is a whole-project claim, and a single-file run has not walked the project. That principle is already settled in this file — `deleted_keys` and the exclusion stamp are both guarded on `self._single_file is None`, with the comment "a single-file run walks one file and cannot speak for the project's exclusion set". The prune is the same claim, so it takes the same guard.

Chosen over the issue's other option (locating the project root and re-keying the target) because that changes what `repo_path`, `project_name` and the hash-cache location mean for every single-file caller, to fix a pass that this kind of run should not be making in the first place.

## Verification

- The new test builds a project with modules both at the root and in `pkg/`, runs a full build, then a single-file run on `pkg/module_a.py`, and asserts no module is swept.
- The target is deliberately **not** at the repo root. Every existing test in `TestSingleFileRunScope` uses a root-level target, where the derived `repo_path` equals the project root and the defect cannot fire — which is why they all pass on the broken code.
- The second ingestor is a `_MockIngestor`, not a plain `MagicMock`: the prune is gated on `isinstance(self.ingestor, QueryProtocol)`, so a plain mock makes the defect invisible rather than absent.
- `_swept_module_keys` counts deletes **not followed by a re-add**, because `CYPHER_DELETE_MODULE` also serves delete-before-reingest on a file that still exists.

## Measured

With a matching project name, the unguarded single-file run on `pkg/module_a.py` issues:

```
DELETE_MODULE: ['root_module.py', 'pkg/module_a.py', 'pkg/module_b.py']
```

All three, including the file it was asked to index. With the guard: none. That is the issue's `modules=[]` exactly.

The test's fixture passes `project_name="nested"` explicitly, and asserts `repo_path == repo/"pkg"` as a guard. Without the explicit name the derived one is `pkg`, every row fails the `qn.startswith(project_prefix)` gate before the path test runs, and the prune skips them for an unrelated reason — the first version of this test passed on the broken code for exactly that reason.

## What the guard still does

The early return sits **below** the two path-independent sweeps, which continue to run:

```python
logger.info(ls.PRUNE_SKIPPED_SINGLE_FILE)
self.ingestor.execute_write(cs.CYPHER_DELETE_ORPHAN_EXTERNAL_MODULES)
prune_unanchored_resources(self.ingestor)
return
```

Both take no path and no project, and delete only nodes with zero inbound edges, so a partial walk cannot make them over-delete. They are also precisely what a single-file run orphans: it deletes and re-ingests its target's entities, so an import or endpoint removed from that one file strands an ExternalModule or Resource that nothing else collects — `deleted_keys` is guarded off for these runs too. Skipping them was a P1 in review.

## Known limits (not addressed here)

- A subdirectory single-file run also **misroots what it writes**: modules and classes are keyed `nested.module_a` rather than `nested.pkg.module_a`. It issues no deletes, so the correct nodes survive and a duplicate wrong-key set is added. This PR stops the deletion half of #1756; the duplication half is live and needs the root-locating remedy.
- The same run overwrites `Project.root_path` with the subdirectory.
- A single-file run whose target **is** at the repo root has a correct `repo_path`, so its prune was safe; the guard skips it anyway. Conservative, and recoverable later with a `repo_path == <project root>` condition.

Each is pre-existing and worth its own issue rather than widening this one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed single-file processing so modules outside the target file’s directory are no longer incorrectly removed.
  - Preserved cleanup of genuinely orphaned external modules and unanchored resources during single-file runs.
  - Improved project-aware synchronization across repeated runs.
  - Added clearer logging when path-based orphan cleanup is skipped for single-file processing.
  - Added clearer reporting for structural failures during re-ingestion.

- **Tests**
  - Added regression coverage for nested project paths and orphan cleanup behavior.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -4734,6 +4734,15 @@ class GraphUpdater:
             # the containment gate that spares out-of-repo File and Folder
             # rows never runs for them.
             logger.info(ls.PRUNE_SKIPPED_SINGLE_FILE)
+            # The two sweeps below still run. Unlike the path-keyed loop they
+            # take no path and no project: each deletes only nodes with zero
+            # inbound edges, so a partial walk cannot make them over-delete.
+            # And a single-file run DOES orphan them -- it deletes and
+            # re-ingests its target's entities, which is exactly what strands
+            # an ExternalModule whose import was removed or a Resource whose
+            # endpoint was (#1756 review).
+            self.ingestor.execute_write(cs.CYPHER_DELETE_ORPHAN_EXTERNAL_MODULES)
+            prune_unanchored_resources(self.ingestor)
             return
 
         logger.info(ls.PRUNE_START)
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -4716,6 +4716,25 @@ class GraphUpdater:
         """Remove graph nodes whose files/folders no longer exist on disk."""
         if not isinstance(self.ingestor, QueryProtocol):
             return
+        if self._single_file is not None:
+            # A single-file run walked one file and cannot say what the
+            # project still holds, exactly as it cannot speak for the
+            # project's file set (`deleted_keys`) or its exclusion set (the
+            # stamp in `run`). The prune is a whole-project claim: it deletes
+            # every node whose path is absent from disk, on the premise that
+            # a full walk established which paths those are.
+            #
+            # Left unguarded it did not merely over-reach, it emptied the
+            # graph (issue #1756). `repo_path` is the TARGET'S PARENT here,
+            # so `pkg/module_a.py` makes it `pkg/` and every module's
+            # relative path is resolved against the wrong root:
+            # `root_module.py` is tested as `pkg/root_module.py`, is absent,
+            # and is deleted. Modules are the worst case because
+            # CYPHER_ALL_MODULE_PATHS_INTERNAL returns no `absolute_path`, so
+            # the containment gate that spares out-of-repo File and Folder
+            # rows never runs for them.
+            logger.info(ls.PRUNE_SKIPPED_SINGLE_FILE)
+            return
 
         logger.info(ls.PRUNE_START)
         total_pruned = 0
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
Rebased onto current `main` after #1546 merged (`cd3ac19e`); head is now `803c0fa5`. The 5/5 above is anchored to `98f80f3e`, the pre-rebase head, so it is stale — the commit subject is unchanged because a rebase preserves subjects, which makes subject-matching an unreliable staleness check here.

Re-verified after the rebase: 41 passed across `test_single_file_repo_path.py` and `test_graph_updater_pruning.py`. No conflicts; the rebase was clean.

@greptileai review

PR #1785, opened 2026-09-07
fix: reingest prunes the seeded module-qn map, with its own re-parsed paths (#1712)

### LINKED ISSUE #1712: reingest never prunes the seeded module-qn map, so a retained MCP updater accumulates
## Defect

`_prune_stale_seeded_module_qns` (added by #1711 for #1687) is called only from `run()`. `reingest` does not go through `run()`: it seeds the map at `graph_updater.py:3824` and `:3827` and never prunes.

`mcp/tools.py:904` calls `reingest` on a retained `_live_updater`, so an MCP session that indexes once and then only makes scoped re-ingest calls never re-enters `run()` and never prunes.

## Effect

Identical to #1687, in a path that fix does not reach. A qn whose `Module` another writer deleted stays in `module_qn_to_file_path` with a **real path**, so `known_module_paths()` keeps offering it and it wins the Rust sub-scope ownership arbitration in `finalise_rust_mod_scope_uses` (`import_processor.py:3757`) against the file that legitimately owns the scope.

Measured on `fix/1687-seed-map-rebuild`: after deleting a `Module` from the graph as a second writer would, five further `reingest` calls leave the qn in the map with its real path, and `known_module_paths()` still returns it.

## Why it was not fixed in #1711

The prune cannot simply be called from `reingest`'s seeding, because `reingest` seeds **twice** with different path sets — the second call covers only the gone files, so their entries exist for the state drop to key on — and a prune keyed on that call's `eligible_paths` would delete everything the first call seeded. It needs its own call site with the full surviving-path set, or a different key.

## Fix

Give `reingest` a prune with the correct path set, after both seed calls and after the re-parse has populated `_parsed_files`, reusing `_prune_stale_seeded_module_qns`. It already fails closed on an unreadable or empty read, so the only new decision is where the call goes.

A regression test wants a retained updater driven through repeated `reingest` calls with a graph-side deletion in between, asserting the qn leaves `known_module_paths()`. Note it must use an ingestor that **buffers** node writes until `flush_all`, as production does: a write-through fake makes a just-parsed module visible to the prune's read and hides the case the exemption exists for.

Found by the local review round on #1711.

vitali87:
Closes #1712.

## The defect

`_prune_stale_seeded_module_qns` had exactly one call site — inside `run()`. `reingest` does not go through `run()`: it seeds the module-qn map twice and never prunes.

So an MCP session that indexes once and then only makes scoped re-ingest calls holds a retained `_live_updater` that never re-enters `run()`. A qn whose `Module` another writer deleted stays in `module_qn_to_file_path` **with a real path**, so `known_module_paths()` keeps offering it and it wins the Rust sub-scope ownership arbitration against the file that legitimately owns the scope.

## Why the obvious fix is wrong

The helper exempts files "this operation parsed", because a re-parsed file writes its own map entry and its `Module` node may still be unflushed when the prune reads. It takes that set from `self._parsed_files`.

That field means different things on the two paths:

- `run()` **clears** it (line 1544) before pruning → "parsed this run"
- `reingest` **never** clears it → "parsed since the process started"

And `reingest` not clearing it is load-bearing, not an oversight. `_hydrate_for_reingest` reads it as a **liveness proxy**:

```python
if self._reingest_hydrated or self._parsed_files:
    self._reingest_hydrated = True
    return   # "a reused updater already has live state"
```

Clearing it there would force a full rehydration read on every scoped call — the cost that check exists to avoid.

So calling the helper as-is from `reingest` exempts every file the process has ever parsed. The map never shrinks on exactly the retained updater this issue is about: a fix that looks correct, passes a first-call test, and is **inert from the second call onwards**.

## The fix

`_prune_stale_seeded_module_qns(exempt_paths: set[Path] | None = None)`, defaulting to `_parsed_files` so `run()` is untouched. `reingest` passes `set(reparse.values())` — the paths *it* re-parsed — after both seed calls and after the re-parse. Same meaning at both call sites, without disturbing the liveness proxy.

## Verification

The fixture **buffers** node writes until `flush_all`, as production does. A write-through fake makes a just-parsed module instantly readable, so the exemption never has to do anything and the case it exists for is invisible. Two instrument checks guard this:

- `test_the_fake_satisfies_the_protocol_the_prune_gates_on` — `QueryProtocol` is runtime-checkable and needs `fetch_all` **and** `execute_write`; `isinstance` inspects the class, so a method from `__getattr__` does not satisfy it. Without this the prune returns early and every other assertion fails for an unrelated reason (measured — it cost a full mutation matrix before it was found).
- `test_the_buffering_fake_actually_buffers` — a pending module must not be readable before `flush_all`, or the fake is write-through and the exemption assertions pass for free.

Mutation matrix, with the seam each targets:

| mutation | seam | red |
|---|---|---|
| remove the call | reingest never prunes | the two prune tests |
| pass no exemption set (`_parsed_files` default) | stale exemption | the two prune tests |
| pass `set()` | no exemption at all | the unflushed-survival test |

Green: 5 passed. Restored: 5 passed.

One prediction was wrong and the docstring records it: I expected only the repeated-calls test to catch the `_parsed_files` variant. It catches both, because the first test's `run()` parses both fixture files, so the stale exemption covers the victim there too. Both fail for the same reason, so the repeated-calls test is a second instance rather than an independent axis — kept because it pins the multi-call shape the issue is about, which the first test reaches only incidentally.

Regression sweep: 155 passed across reingest, incremental, watch and single-file; 258 passed over a `-k` sweep of everything naming reingest, seed, prune, module_qn or watch. `ty` reports 9 diagnostics here and 9 on clean `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reingestion now removes stale module entries when the corresponding modules are no longer available.
  * Re-parsed files retain their current module entries during reingestion.
  * Repeated reingestion now consistently cleans up outdated module references.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2324,7 +2324,9 @@ class GraphUpdater:
                 continue
             module_map.setdefault(qn, self.repo_path / path)
 
-    def _prune_stale_seeded_module_qns(self) -> None:
+    def _prune_stale_seeded_module_qns(
+        self, exempt_paths: set[Path] | None = None
+    ) -> None:
         """Drop map entries for modules the graph no longer holds.
 
         `_seed_module_qns_from_graph` only ever adds, so on a reused updater
@@ -2343,11 +2345,17 @@ class GraphUpdater:
         different path sets, the second covering only the gone files, so a
         prune running inside it would delete what the first call seeded.
 
-        `reingest` therefore never prunes, since it does not go through
-        `run()`: a retained updater that indexes once and then only makes
-        scoped re-ingest calls (the MCP server's `_live_updater`) still
-        accumulates. That is the same class of bug in a path this fix does
-        not reach, not something it closes.
+        `reingest` calls this too (issue #1712), passing the paths IT
+        re-parsed as `exempt_paths`. It cannot use the default: the exemption
+        means "parsed by this operation, so the Module node may be
+        unflushed", and `_parsed_files` only carries that meaning on `run()`,
+        which clears it per run. `reingest` never clears it -- deliberately,
+        because `_hydrate_for_reingest` reads it as a LIVENESS PROXY
+        (non-empty means "this updater already has live state, skip the
+        rehydration read"). So on a retained updater the default would exempt
+        every file parsed since the process started, and the map would never
+        shrink -- the accumulation #1712 is about, in a fix that looks
+        correct and does nothing from the second call onwards.
         """
         if not isinstance(self.ingestor, QueryProtocol):
             return
@@ -2373,7 +2381,11 @@ class GraphUpdater:
         if not in_graph:
             logger.warning(ls.SEED_PRUNE_NO_VERDICT)
             return
-        parsed_this_run = {path for path, _lang in self._parsed_files}
+        parsed_this_run = (
+            exempt_paths
+            if exempt_paths is not None
+            else {path for path, _lang in self._parsed_files}
+        )
         for qn in [qn for qn in module_map if qn not in in_graph]:
             # A file this run parsed writes its own entry and its Module node
             # may still be unflushed, so the read above cannot see it.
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

PR #1788, opened 2026-09-07
fix: pin UV_PYTHON_INSTALL_DIR so the built image's venv actually runs

Victor-johnsson:
## Summary

The builder stage lets `uv sync` install uv's managed Python interpreter to its default location (`$HOME/.local/share/uv/python`, i.e. `/root/.local/...` since the builder stage runs as root). `.venv` only *symlinks* to that interpreter rather than containing it, and the final stage only copies `/app/.venv` out of the builder -- not `/root/.local` -- so the copied venv's `python` symlink dangles and the built image doesn't run at all.

Pinning `UV_PYTHON_INSTALL_DIR` under `/app` before `uv sync` runs, and copying that directory into the final stage alongside `.venv`, fixes it.

## Test plan

- Built the patched Dockerfile locally with both the default BuildKit builder (`docker build`) and the classic non-BuildKit builder (`DOCKER_BUILDKIT=0 docker build`), to also confirm compatibility with ACR Tasks-style remote builds that don't use BuildKit.
- Confirmed `.venv/bin/python` resolves to a real interpreter and runs (`readlink -f /app/.venv/bin/python` -> `/app/.uv-python/cpython-3.12.13-linux-*/bin/python3.12`; `python --version` succeeds).
- Confirmed `code-graph-rag --help` works through the real entrypoint (`/entrypoint.sh`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Container builds now consistently use the managed Python interpreter.
  * Improved reliability when running the application under the non-root user.
  * Ensured the required Python environment remains available after the container switches users.
  * Preserved access to the pinned Python installation in the final container image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,15 @@ WORKDIR /app
 # .venv only symlinks to that interpreter, so pin the install dir under /app
 # and copy it into the final stage too, or the venv's python symlink dangles
 # once /root is left behind.
+#
+# uv only populates this directory when it actually downloads a managed
+# interpreter -- if the pinned Python version ever matches this image's
+# system Python, uv uses that instead and never creates the directory,
+# which would make the unconditional COPY below fail the build outright.
+# Pre-create it so COPY always has something to copy, whichever
+# interpreter uv picks.
 ENV UV_PYTHON_INSTALL_DIR=/app/.uv-python
+RUN mkdir -p "$UV_PYTHON_INSTALL_DIR"
 
 COPY pyproject.toml uv.lock ./
 RUN uv sync --frozen --no-dev --extra treesitter-full --no-install-project --no-binary-package pymgclient
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,13 +18,13 @@ WORKDIR /app
 # once /root is left behind.
 #
 # uv only populates this directory when it actually downloads a managed
-# interpreter -- if the pinned Python version ever matches this image's
-# system Python, uv uses that instead and never creates the directory,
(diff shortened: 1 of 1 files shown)


vitali87:
@Victor-johnsson thanks for this PR. Can you pleaase address the bots' comments?

Victor-johnsson:
Yep! I'll fix this tonight ☺️

Victor-johnsson:
Addressed - pushed a follow-up commit that pre-creates `UV_PYTHON_INSTALL_DIR` before `uv sync` runs, so the final-stage `COPY` always has a directory to copy regardless of whether uv ends up using a managed interpreter or the builder image's system Python (@greptile-apps's finding).

Verified locally with the classic (non-BuildKit) Docker engine: `.venv/bin/python` still resolves and runs, `code-graph-rag --help` works through the real entrypoint.

Victor-johnsson:
Switched to the more direct fix: \`UV_PYTHON_PREFERENCE=only-managed\` instead of pre-creating the directory. My first pass just kept the build from failing, but didn't actually address the underlying gap @greptile-apps flagged — if uv ever picked the system Python, \`.venv\`'s symlink would point somewhere the final stage doesn't copy, breaking the image silently rather than failing the build. \`only-managed\` rules that out entirely: uv always uses its own downloaded interpreter, never the system one.

Verified locally again (classic Docker engine): \`.venv/bin/python\` resolves to a real managed interpreter and runs.

vitali87:
@greptile-apps review

PR #1795, opened 2026-09-08
fix: prune the two class-keyed maps when a file leaves the state (#1772)

### LINKED ISSUE #1772: remove_file_from_state never prunes class_field_types or class_inheritance
## Observed

After `run()` on a two-file C# project, deleting `A.cs` and calling
`remove_file_from_state(A.cs)`, two class-keyed maps on `DefinitionProcessor`
still hold the deleted file's entries on a reused updater:

- `class_field_types`: `{'proj.A.N.K': {'Field': 'string'}}`
- `class_inheritance`: `{'proj.A.N.K': ['proj.A.N.Base']}`

Measured with a probe over both maps before and after the delete, on a clean
tree at the #1753 fix. The probe asserts the precondition (that `A.cs`
recorded field types at all) before reading the outcome, so this is not a
repro that silently established nothing.

Neither map is mentioned anywhere in `remove_file_from_state`.

## Why it matters

Both maps steer resolution at a class that no longer exists:

- `class_field_types` types a receiver reached through a field (`_w.M()`),
  including a field inherited from a base class in another file.
- `class_inheritance` is walked to reach base-class members and drives the
  OVERRIDES arbitration.

A stale entry is a wrong answer rather than a missing one, which is the same
argument that motivated the sweep in #1738 and #1752.

## Expected

Each map drops the deleted file's entries. Both are keyed by a CLASS qn, not
a definition qn, so the ownership available is the module-qn prefix, as it is
for `csharp_class_generic_arity` in #1769: a class qn sits under the module qn
of the file declaring it, and a partial type's other halves are separate qns
under their own modules, each pruned by its own deletion.

Note these two maps are NOT C#-specific -- every language's class ingest
writes to them -- so unlike #1769 the prune has to be language-agnostic and
wants a fixture per affected language rather than a C#-only one.

Same defect class as #1668, #1738, #1753 and #1769, for two more maps.
Found while fixing #1769.

vitali87:
`remove_file_from_state` never mentioned `class_field_types` or
`class_inheritance`, so a deleted file's entries survived on a reused updater
and went on steering resolution at a class that no longer exists — a wrong
answer rather than a missing one. `class_field_types` types a receiver reached
through a field; `class_inheritance` is walked to reach base-class members and
drives the OVERRIDES arbitration.

Fixes #1772. Same defect class as #1668, #1738, #1753 and #1769, for two more
maps. Unlike #1769's four, these are written by every language, so the prune is
language-agnostic.

## Ownership is recorded, not inferred

The issue suggested attributing a class by its module-qn prefix. That is wrong,
for the reason #1769's review established: a C# class qn embeds its namespace,
so `proj/Core.cs` declaring `namespace Util` yields `proj.Core.Util.Helper`,
which sits under the **sibling** module `proj.Core.Util` (the qn of
`proj/Core/Util.cs`). A prefix rule errs in both directions — deleting the
sibling drops a live class, deleting the real declarer keeps a dead one.

So this adds a cross-language `class_owner_module`, the counterpart to the
C#-only `csharp_class_owner_module`, and reads ownership from it.

## Three writers, three ownership paths

Both maps are written from more places than the issue lists, and two of them
bypass class ingest entirely:

| Writer | Owner source |
|---|---|
| Class ingest (every language) | `class_owner_module`, recorded at ingest |
| JS prototype inheritance (`js_ts/ingest.py`) | same map, recorded at that write site |
| Rehydration from the graph | `rehydrated_definition_paths` |

The rehydrated case is the one that matters most in practice: an incremental
run re-parses only the changed files and rehydrates the rest, so on such a run
most classes never pass through ingest. Attributing only the parsed ones would
have left the majority of the map unpruned — the very defect this closes.

## Validation

* 5 new tests, parametrized over C#, Go, Python and JS plus a rehydration case.
  All fail on `main` and pass here.
* The C# fixture is `proj/Core.cs` (`namespace Util`) beside `proj/Core/Util.cs`,
  so the namespace/sibling-module collision is exercised rather than described.
* Five mutations, each reddening exactly the predicted cases and nothing else:
  prefix rule instead of the owner record → C#; removing each map's `pop`
  independently → C#+Python and C#+Go respectively; removing the JS owner
  record → the JS case only; removing the rehydrated half → the rehydration
  case only.
* 694 passed, 5 skipped, 0 failed across the inheritance, incremental,
  deleted-file, side-table and js_ts modules, on a settled tree under the host
  suite lock. No full suite run locally; CI covers that.
* `ruff check` and `ruff format` clean.

Purely additive: 425 insertions, no deletions, no existing signature changed.

## Note for review

`csharp_class_owner_module` is now a strict subset of the new
`class_owner_module` — same key, same value, written only for generic C#
classes. Harmless duplication left alone deliberately, since consolidating it
would touch #1769's prune; it belongs in its own PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed files are now fully cleared from class inheritance and field-type information.
  - Prevents outdated class relationships and type details from appearing after source files are deleted.
  - Cleanup works consistently across C#, Go, JavaScript, and Python, including data restored from the graph.
  - Unrelated files and their class information remain unchanged.

- **Tests**
  - Added regression coverage for deleted-file cleanup across supported languages and restored graph data.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -3133,6 +3133,11 @@ class GraphUpdater:
             module_qn_prefixes=module_qn_prefixes,
         )
 
+        # The two cross-language class-keyed maps, same defect for two more
+        # (issue #1772). Separate from the C# sweep above because every
+        # language's class ingest writes these.
+        self._prune_class_keyed_maps(module_qn_prefixes, file_path)
+
         for simple_name, qn_set in self.simple_name_lookup.items():
             original_count = len(qn_set)
             new_qn_set = qn_set - qns_to_remove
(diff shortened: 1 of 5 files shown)


vitali87:
@greptileai review

Issue #1794, opened 2026-09-08
reingest keeps a stale CALLS edge when a file yields no definitions


vitali87:
Found by the `fuzz_incremental_update` harness added for #1731, then reduced to a deterministic reproducer.

## Symptom

When a file's content is emptied (or replaced with anything that parses to no definitions), `GraphUpdater.reingest` keeps the `CALLS` edge that the file's now-deleted function used to emit. A full index of the same tree emits no such edge, so the incremental and batch paths disagree.

The graph ends up asserting that `proj.pkg.app.run` calls `proj.pkg.util.helper` when `run` no longer exists in the source at all.

## Reproducer

Fixture:

```
pkg/__init__.py   (empty)
pkg/util.py       def helper():\n    return 1\n
pkg/app.py        from pkg.util import helper\n\n\ndef run():\n    return helper()\n
```

```python
store = _StatefulIngestor()
updater = GraphUpdater(ingestor=store, repo_path=root, parsers=P, queries=Q, project_name="proj")
updater.run(force=True)

(root / "pkg/app.py").write_text("")      # emptied
updater.reingest(["pkg/app.py"])
```

Result:

```
stale edge: ('Function', 'proj.pkg.app.run', 'CALLS', 'Function', 'proj.pkg.util.helper')
missing:    []
```

A clean `run(force=True)` over the same tree emits no `CALLS` edge.

## Scope

Verified on the `warm` shape (the watcher's: reuse the updater that built the graph). Three content shapes, same fixture:

| new content of `pkg/app.py` | stale `CALLS` edge left |
| --- | --- |
| `""` (emptied) | yes |
| unparseable bytes (no valid definition) | yes |
| `def run():\n    return 42\n` (valid, call removed) | no |

So the defect is specific to the file parsing to **no definitions**. When `run` still exists and merely stops calling `helper`, the edge is correctly retracted — which is why the existing `test_reingest.py` edits do not catch it: every one of them leaves at least one definition in the file.

Deleting the file outright is also handled correctly; it is only the "file still exists but is now empty" state that leaks.

## Why it matters

This is the watcher's normal mid-write state: an editor that truncates before writing leaves a zero-length file, and a save caught at that instant leaves the graph with edges out of a function that no longer exists. Every consumer of `CALLS` (dependents, call graphs, dead-code roots) then reads an edge whose source node is gone.

[...]

PR #1796, opened 2026-09-08
feat: run fuzzing continuously with ClusterFuzzLite (#1731)

### LINKED ISSUE #1731: Fuzzing is declared but never runs: add ClusterFuzzLite with harnesses for the parsers, incremental updater and shell classifier
## Current state

- `codebase_rag/tests/fuzz_test_parsers.py` is a 20-line atheris harness. Despite the name it does not touch the parsers: it feeds a random string of up to 64 characters to `get_language_spec` and `get_language_for_extension`, which are extension-to-spec dictionary lookups.
- `pyproject.toml` carries a `fuzz` extra pinning `atheris>=2.3.0`.
- None of the workflows under `.github/workflows/` invoke it, and pytest does not collect it (it is a `__main__` script, not a `test_*` module). It has never executed in CI.

## What needs fuzzing

The tool indexes arbitrary repositories. Every file in a checkout reaches the tree-sitter grammars and the per-language query and qualification code in `codebase_rag/parsers/` and `codebase_rag/language_spec.py`, and the incremental path additionally parses whatever a watcher sees mid-write. Malformed, truncated and adversarial inputs are the normal case, and none of that surface has been fuzzed.

## Proposal

Set up ClusterFuzzLite so fuzzing runs continuously rather than as a one-off script.

1. **Harnesses**, under `fuzz/`, one per target, all atheris:
   - `fuzz_parse_source.py`: pick a language from the supported set via `FuzzedDataProvider`, hand the remaining bytes to the tree-sitter parser for that language and run the query and node-extraction pass over the tree. This is the main target.
   - `fuzz_incremental_update.py`: seed a temp repo with a small project, apply fuzzed edits (truncate, splice, rename, delete) and run `GraphUpdater` on the result. Targets the deletion, dependent-reparse and hash-cache paths that #1567-#1570, #1584 and #1665 turned up.
   - `fuzz_shell_command.py`: feed fuzzed strings to the `EXECUTE_SHELL` allowlist and dangerous-pattern classifier in `shell_command.py`; assert the classifier never raises and never returns "safe" for an input containing a known-dangerous token.
   - Fold the existing `fuzz_test_parsers.py` into the first harness or rename it to what it does.
2. **Seed corpora** under `fuzz/corpus/<harness>/`, built from the existing per-language test fixtures so the fuzzer starts from valid programs rather than noise.
3. **ClusterFuzzLite integration**: `.clusterfuzzlite/Dockerfile`, `.clusterfuzzlite/build.sh`, `.clusterfuzzlite/project.yaml` with `language: python`; a workflow `cflite_pr.yml` running `google/clusterfuzzlite/actions/run_fuzzers@v1` for 600 seconds on pull requests with `sanitizer: address`; a scheduled `cflite_batch.yml` for longer nightly runs. Crashes surface as workflow failures with the reproducer attached.
4. **Local entry point**: `uv run --extra fuzz python fuzz/<harness>.py -max_total_time=60`, documented in CONTRIBUTING.
5. **Triage rule**: a crash found by fuzzing gets a regression test in `codebase_rag/tests/` before the fix merges, and the reproducer goes into the corpus.

## Out of scope

- Fuzzing the Memgraph query layer or MCP transport. Different harness shape; separate issue once this lands.
- Coverage targets. Get it running first, measure after.

## Acceptance

- The three harnesses exist and run for at least 60 seconds locally without a crash on `main`.
- `cflite_pr.yml` runs on every PR and is a required status check on `main`.

vitali87:
Closes #1731.

## What was there

The `fuzz` extra and `codebase_rag/tests/fuzz_test_parsers.py` had been in the tree without ever running. No workflow invoked the harness, pytest did not collect it (a `__main__` script, not a `test_*` module), and despite its name it fed random strings to two dictionary lookups rather than to any parser.

## What this adds

Three harnesses under `fuzz/`, each asserting a property rather than merely "does not crash":

| Harness | Drives | Asserts |
| --- | --- | --- |
| `fuzz_parse_source.py` | a grammar for a fuzzer-chosen language, then the queries and name extraction over the tree | the pass never raises |
| `fuzz_incremental_update.py` | `GraphUpdater.reingest` over a fuzzer-chosen edit plan | the graph equals a clean full index of the same tree |
| `fuzz_shell_command.py` | the `EXECUTE_SHELL` allowlist and dangerous-command classifier | never raises, never calls a never-allowlisted program safe, never lets a prefix launder a dangerous trailing segment |

The parse harness goes through `QueryCursor` + `sorted_captures`, the same route `graph_updater.py` and `call_processor.py` take, so a capture bug is reached the way production reaches it. The incremental harness fuzzes both updater shapes: `warm` (the watcher's) and `fresh` (the MCP tool's, which must rehydrate the registry from the graph first).

Seed corpora in `fuzz/corpus/`, 44 seeds built from valid programs in each of the 15 languages plus degenerate inputs. ClusterFuzzLite runs the targets: `cflite_pr.yml` for 600s per PR, `cflite_batch.yml` for an hour nightly.

## A bug it found

`fuzz_incremental_update` turned up #1794 while being written: a file that parses to **no definitions** keeps the `CALLS` edges its removed functions emitted. Three ways in, all confirmed and reduced to deterministic reproducers:

| new content of a changed file | stale edge |
| --- | --- |
| `""` | yes |
| unparseable bytes | yes |
| truncated mid-`def` | yes |
| valid file, call removed | no |

The truncation case is the watcher's normal mid-write state, so this is reachable in ordinary use. The graph ends up asserting that a function which no longer exists still calls something.

That shape is suppressed by a narrow predicate (it asks the parser for zero complete definitions, not a golden diff) so the harness runs clean on `main`; any other disagreement still fails. Deleting the helper when #1794 is fixed re-detects it. Control: neutering `reingest` to a no-op is still caught on 9/20 random inputs.

## Verification, and a caveat

**atheris cannot build on macOS** — Apple Clang ships no libFuzzer, and the install fails with `Failed to find libFuzzer`. So the harnesses have not been run under real atheris locally; their first true fuzzing run will be on Linux in CI. CONTRIBUTING documents this, and `fuzz/` is excluded from the `ty` hook for the same reason.

To keep that from meaning "unverified", `codebase_rag/tests/test_fuzz_harnesses.py` (13 tests) drives the harness bodies under pytest with a faithful port of atheris' `FuzzedDataProvider`, transcribed from `src/native/fuzzed_data_provider.cc`. Without it, a rename in the code under test would break every harness while the local suite stayed green, surfacing days later as a ClusterFuzzLite build failure that reads like an infrastructure problem.

Every assertion is mutation-checked: planted faults fire, unmutated controls stay silent.

## Notes for review

Three things the local review round caught, all fixed here, each worth a second look:

1. **The corpora were encoded against a guessed provider.** `ConsumeIntInRange` reads from the **back** of the buffer and `ConsumeUnicodeNoSurrogates` eats a leading `string_spec` byte. All three corpora were inert: 19 of 21 parse seeds went to the Rust grammar as corrupted garbage, every shell seed lost its first character, and all eight incremental seeds collapsed onto one plan. The root cause was the test double — it seeded `random.Random` with the input, so it agreed with *any* encoding. It is now a faithful port, and four tests assert what each seed decodes to (all confirmed red against the old encodings). Definition nodes reached across the parse corpus: 45, up from 24 incidental matches.

2. **`evals` is not in the wheel.** `fuzz_incremental_update` imports `evals.cgr_graph`, but package discovery includes only `codebase_rag*`, `codec*`, `cgr*`, so pyinstaller would have frozen a target that fails on first run — reading as a fuzzing crash rather than a build mistake. Fixed with `PYTHONPATH` + `--paths`; two tests guard it, one of which asserts the premise so the workaround gets removed rather than left as dead weight if `evals` is ever packaged.

3. **The whitespace hooks corrupt the corpus.** `trailing-whitespace` and `end-of-file-fixer` rewrote the binary seeds on the first commit attempt, stripping exactly the trailing selector byte. `pre-commit.ci` runs with `autofix_prs: true`, so it would have redone that server-side on every PR. Both hooks now exclude `fuzz/corpus/`, and `.gitattributes` marks it binary.

## Summary by CodeRabbit

- **New Features**
  - Added automated fuzz testing for source parsing, incremental updates, and shell-command safety.
  - Fuzzing now runs automatically on pull requests and nightly, with crash reports and reproducing inputs available for investigation.
  - Added seed corpora covering malformed source, complex edits, and potentially dangerous commands.
  - Added build-time smoke tests to verify fuzz targets start successfully.

- **Documentation**
  - Added contributor guidance for running fuzz tests, regenerating seed data, and handling discovered crashes.
  - Documented an additional refusal condition for renaming certain references without rewrite locations.

- **Tests**
  - Added coverage validating fuzz harness behavior across corpus seeds and key safety invariants.


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1801, opened 2026-09-08
fix: memoise C# chained-call resolution so fluent chains are linear (#1800)

### LINKED ISSUE #1800: C# fluent chains make call resolution exponential: 67M resolve calls for 228 call sites, indexing never finishes
Indexing a C# repo with ordinary fluent-builder chains does not finish. Call
resolution for a chained invocation re-resolves its entire receiver prefix at
every link, with no memoisation and no depth guard, so cost grows
exponentially in chain length.

Found while trying to reproduce #1786 and #1787 against a real
`microsoft/aspire` clone. It explains #1786's "stalls the whole run" symptom,
which is not caused by the Java RecursionError that issue names.

## The cycle

All in `codebase_rag/parsers/csharp/type_inference.py`:

```
resolve_csharp_method_call     (:276)
  -> _try_extension_call         (:728)
  -> _resolve_receiver_class_qn  (:1004)
  -> _receiver_type_name         (:751)
  -> _invocation_return_type_name(:796)
  -> resolve_csharp_method_call  (:803)   <- closes the loop
```

**Correction (M1 CGR-2, measured):** there are THREE recursion sites back into
`resolve_csharp_method_call`, not two. The trace above lists
`_invocation_return_type_name` (:803) and `_resolve_receiver_class_qn` (:1004);
`_receiver_type_arity` (:891) recurses as well. This matters for the
arithmetic: the observed ~9x growth per +2 links is 3^2, which fits three
branches and not the two originally reported (which would predict 4x).

Resolving `a.B().C().D()` resolves its receiver, which resolves ITS receiver,
and so on. Nothing caches the answer for a receiver node, so each link
re-walks the whole prefix. There is no `lru_cache`, `@cache` or memo dict
anywhere in the file, and no recursion-depth limit.

## This is a complexity bug, not a correctness bug

Worth stating explicitly, because it determines what a regression test must
assert. The pre-fix code resolves these chains CORRECTLY - it just does so
exponentially. A resolution-coverage test (does `a.B().C().D()` bind to the
right targets?) therefore PASSES on unfixed main and does NOT carry this
regression.

The only assertion that fails pre-fix is a complexity one: the number of
resolver entries for a chain of N links. Any guard for this issue must count
resolver calls, not check that edges are correct.

(Established by M1 CGR-2 while building the fix; the local reviewer caught the
conflation in the first version of the tests.)

## Measurement

cProfile over ONE file from aspire
(`src/Aspire.Hosting.Python/PythonAppResourceBuilderExtensions.cs`):

```
ncalls              tottime  cumtime  function
67124421/228        128.576  615.678  csharp/type_inference.py:276(resolve_csharp_method_call)
67124307/770         13.820  615.663  csharp/type_inference.py:728(_try_extension_call)
67124307/775         51.530  615.662  csharp/type_inference.py:1004(_resolve_receiver_class_qn)
67124307/775         35.653  615.661  csharp/type_inference.py:751(_receiver_type_name)
33562076/575         11.518  615.660  csharp/type_inference.py:796(_invocation_return_type_name)
```

67.1 million calls for 228 real call sites. The `67124421/228` form is
cProfile's recursive-call notation.

## Reproduction

Two ways.

**Minimal, no database, no external repo** (script attached below): vary only
the number of links in a chain of extension methods.

```
 links    seconds    ratio
     2       0.06        -
     4       0.06     1.1x
     6       0.06     1.0x
     8       0.09     1.5x
    10       0.36     3.8x
    12       2.68     7.5x
    14      23.85     8.9x
```

**Real repo:** a full `index_repository` of `microsoft/aspire` (5000 `.cs`
files) against `memgraph-mage:3.3`. Pass 3 produced no log output for 33+
minutes at 100% CPU; killed at 46 minutes. The per-resolution gaps before it
wedged were `0.3s, 0.8, 1.9, 4.4, 25, 29, 159, 212s`. The file above takes
254s on its own; truncating it puts the cliff between 770 lines (0.10s) and
1047 lines (261s).

The trigger is unremarkable C#:

```csharp
stage
    .Comment("Copy requirements.txt for dependency installation")
    .Copy("requirements.txt", "/app/requirements.txt")
    .EmptyLine()
    .Comment("Install dependencies using pip")
    .Run("""apt-get update && pip install --no-cache-dir -r requirements.txt""")
    .EmptyLine();
```

## Suggested fix

Memoise the receiver resolution for the duration of a file's call pass. A
cache keyed on the receiver is enough to make the walk linear: the same
receiver prefix is currently recomputed once per enclosing link.

**Do not key the memo on `id()` of the node.** This paragraph originally
suggested `(node id, module_qn, caller_qn)`; that is unsafe. CPython recycles
`id()` for freed objects, so a stale entry can answer for an unrelated node,
and this repo has already been bitten by exactly that failure - see the
comment in `CallResolver.reset_resolution_caches`
(`codebase_rag/parsers/call_resolver.py:388`), "wildcard entries keyed by dict
id, which recycles". Use a stable key such as
`(module_qn, caller_qn, start_byte, end_byte)`.

An `_IN_PROGRESS` sentinel in the memo doubles as the cycle guard and gives
the "degrade to unresolved rather than hang" behaviour, so no separate depth
cap is needed.

Whatever memo is added must be cleared by `reset_resolution_caches`: the watch
pass deletes every CALLS edge and recomputes it, and answers cached from the
previous pass must not survive into the recompute.

## Impact

Any C# codebase using fluent builders, which is idiomatic in .NET
(`IResourceBuilder`, `HostApplicationBuilder`, LINQ-style chains, Aspire's
`DockerfileBuilder`). The run neither fails nor warns; it consumes a core
until something external kills it, and no graph is written for the affected
project. The calls that do get resolved are resolved correctly; the defect is
cost, not accuracy.

## Environment

`memgraph-mage:3.3`, Python 3.12, macOS arm64, this repo at 0.0.887.

vitali87:
`resolve_csharp_method_call` types a chained invocation by resolving its
receiver, and nothing cached the answer, so an n-link fluent chain re-resolved
its whole prefix once per branch per link. A real `microsoft/aspire` file spent
253s on its own and a full index made no progress for 33+ minutes at 100% CPU.

Fixes #1800.

## Three recursive branches, not two

The issue names two sites; there are three, all descending into the same
receiver node:

| site | branch |
|---|---|
| `_invocation_return_type_name` (:803) | receiver type name |
| `_receiver_type_arity` (:891) | receiver generic arity |
| `_resolve_receiver_class_qn` (:1045) | receiver class qn |

That reconciles the issue's own numbers: the measured ~9x growth per two links
is 3², which two branches would not produce.

## The fix

The body becomes `_resolve_csharp_method_call`; the public name is a memoising
wrapper. Wrapping the entry point collapses all three branches at once, and any
future caller with it.

**Keyed on `(module_qn, caller_qn, start_byte, end_byte)`, not node id.** The
issue suggested node id; `id()` is recycled by CPython for freed objects, and
this repo already carries a scar from that — see `reset_resolution_caches`,
"wildcard entries keyed by dict id, which recycles". `local_var_types` is
deliberately absent from the key: for C# it is a pure function of the enclosing
method (`_overlay_span_binding` is Rust-gated), so one key cannot need two
answers.

**No separate depth cap.** An `_IN_PROGRESS` sentinel doubles as a cycle guard,
giving the "degrade to unresolved rather than hang" behaviour the issue asks
for on a malformed parse tree. A `_MISSING` sentinel keeps a cached `None`
distinct from a miss — `None` means "unresolved", which is a legitimate and
common answer, and caching it is most of the win on chains that do not resolve.

**Cleared in `reset_resolution_caches`**, via the private
`_csharp_type_inference` attribute so a project with no C# does not build an
engine just to clear an empty dict.

## Measured

Resolver entries, not seconds — deterministic, and the quantity the fix changes:

| links | calls before | calls after | secs before | secs after |
|---|---|---|---|---|
| 8 | 4,916 | 29 | 0.09 | 0.05 |
| 12 | 398,574 | 45 | 2.95 | 0.06 |
| 14 | 3,587,219 | 53 | 29.57 | 0.06 |

Growth is now +4 calls per +2 links. End to end, 200 files x 14 links indexes
in 1.08s; on `main` a single 14-link file took 29.57s, so that corpus was
roughly 100 minutes. The memo holds 2,000 entries (~74 KB) for 200 files and
clears to 0 on reset.

## Correctness

Speed alone proves nothing — a memo returning wrong answers is just as fast. A
harness captured every emitted relationship with and without the memo across
fluent, generic, inheritance and cast fixtures: **58 identical relationships**.
That harness was validated with a known-positive (poisoning the memo drops
exactly the 7 chained CALLS edges), because an equivalence check that cannot
fail is not evidence.

The memo key was audited the same way: 10 distinct keys, 0 locals/result
conflicts, and the audit demonstrably reports conflicts when the key is
weakened to collide.

## Tests

`test_chain_resolution_work_grows_linearly` is the regression test and the only
one that fails on pre-fix code — the old code resolved these chains correctly,
just exponentially. The other three are correctness guards on the memo itself
and fail when it is made to return wrong answers:

* `test_long_chain_still_resolves_every_link` — all 12 links still bind.
* `test_real_world_mixed_arity_chain_resolves` — the reported aspire shape
  (`Comment().Copy().EmptyLine().Run()`), whose zero-argument links exercise the
  arity branch directly rather than incidentally.
* `test_memo_does_not_leak_between_resolution_passes` — reddens if the cache
  clear is removed.

995 passed, 1 skipped across the C#, resolution and type-inference modules, on
a settled tree under the host suite lock. `ruff check` and `ruff format` clean.

## Not covered

A full `microsoft/aspire` index is the real acceptance test and is not run
here: it needs memgraph and a 5000-file clone, and cannot run in CI. Worth a
follow-up for a periodic large-repo indexing benchmark — this class of
regression is invisible to a unit suite by construction, which is how it
survived.

## Summary by CodeRabbit

- **Performance**
  - Improved C# fluent method-chain resolution for more consistent performance, especially on long chains.

- **Bug Fixes**
  - C# method chains now resolve reliably without missing links, including mixed-arity extension-method chains.
  - Resolution results refresh after files are re-parsed, preventing stale analysis across passes.
  - Circular or repeated resolution paths are handled safely without interrupting analysis.

### DIFF
--- a/codebase_rag/parsers/call_resolver.py
+++ b/codebase_rag/parsers/call_resolver.py
@@ -412,6 +412,13 @@ class CallResolver:
         # old path and never learn the new one, and every JEDI fact
         # targeting the renamed file would miss its declared location.
         self._py_rel_to_module.clear()
+        # The C# chained-receiver memo (issue #1800) is a resolution cache
+        # like the rest and expires with them: it is keyed by byte span, and a
+        # re-parsed file's spans name different code. Reached through the
+        # PRIVATE attribute so a project with no C# does not build an engine
+        # just to clear an empty dict.
+        if (csharp := self.type_inference._csharp_type_inference) is not None:
+            csharp._call_memo.clear()
 
     def resolve_function_call(
         self,
--- a/codebase_rag/parsers/csharp/type_inference.py
+++ b/codebase_rag/parsers/csharp/type_inference.py
@@ -379,20 +379,7 @@ class CSharpTypeInferenceEngine:
         # Polly's hide-object-members regions) and the call must emit nothing;
         # the trie fallback was self-looping it onto the caller's own override.
         if receiver.type == cs.TS_CSHARP_BASE_EXPRESSION:
-            if class_qn := self._containing_class_qn(caller_qn):
-                seen: set[str] = set()
-                for root in self._partial_roots(class_qn):
-                    for base_qn in self.class_inheritance.get(root, []):
-                        if hit := self._find_method_by_arity(
-                            base_qn, method_name, arg_count, seen
-                        ):
-                            return cs.NodeLabel.METHOD.value, hit
-                seen = set()
-                for root in self._partial_roots(class_qn):
-                    for base_qn in self.class_inheritance.get(root, []):
-                        if hit := self._find_method_by_name(base_qn, method_name, seen):
-                            return cs.NodeLabel.METHOD.value, hit
-            return CSHARP_EXTERNAL_TARGET
+            return self._resolve_base_call(method_name, arg_count, caller_qn)
 
         receiver_class_qn = self._resolve_receiver_class_qn(
             receiver, local_var_types or {}, module_qn, caller_qn
@@ -445,6 +432,39 @@ class CSharpTypeInferenceEngine:
         ):
             return CSHARP_EXTERNAL_TARGET
         return None
+
+    def _resolve_base_call(
+        self, method_name: str, arg_count: int, caller_qn: str | None
+    ) -> tuple[str, str] | None:
+        """Resolve `base.X()` against the BASE chain only.
+
+        Split out of `_resolve_csharp_method_call` to keep that function under
+        the cognitive-complexity limit; the two nested walks below are its
+        densest part and are self-contained.
+
+        A first-party base's member wins when one exists; otherwise the base is
+        external (`object.Equals` in Polly's hide-object-members regions) and
+        the call must emit nothing, because the trie fallback was self-looping
+        it onto the caller's own override. Arity-exact matches are tried across
+        every partial part first, then name-only, mirroring the order the
+        instance path uses.
+        """
+        class_qn = self._containing_class_qn(caller_qn)
+        if class_qn is None:
+            return CSHARP_EXTERNAL_TARGET
+        seen: set[str] = set()
+        for root in self._partial_roots(class_qn):
+            for base_qn in self.class_inheritance.get(root, []):
+                if hit := self._find_method_by_arity(
+                    base_qn, method_name, arg_count, seen
+                ):
+                    return cs.NodeLabel.METHOD.value, hit
+        seen = set()
+        for root in self._partial_roots(class_qn):
+            for base_qn in self.class_inheritance.get(root, []):
+                if hit := self._find_method_by_name(base_qn, method_name, seen):
+                    return cs.NodeLabel.METHOD.value, hit
+        return CSHARP_EXTERNAL_TARGET
 
     def _externally_targeted(
         self,
--- a/codebase_rag/parsers/csharp/type_inference.py
+++ b/codebase_rag/parsers/csharp/type_inference.py
@@ -357,13 +357,7 @@ class CSharpTypeInferenceEngine:
         if func is None:
             return None
         if func.type != cs.TS_CSHARP_MEMBER_ACCESS_EXPRESSION:
-            # A bare `Foo(...)`/`Foo<T>(...)` follows C# simple-name lookup:
-            # an in-scope local function first (it shadows same-name method
-            # overloads), then an arity-matched member of the enclosing
-            # type. A miss falls to the generic simple-name path.
-            if func.type in (cs.TS_CSHARP_IDENTIFIER, cs.TS_CSHARP_GENERIC_NAME):
-                return self._resolve_bare_call(func, call_node, module_qn, caller_qn)
-            return None
+            return self._resolve_non_member_call(func, call_node, module_qn, caller_qn)
         method_name = safe_decode_text(func.child_by_field_name(cs.FIELD_NAME))
         receiver = func.child_by_field_name(cs.TS_CSHARP_FIELD_EXPRESSION)
         if not method_name or receiver is None:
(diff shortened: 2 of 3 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Pushed 91353ad1: extracted the `base.X()` walk into `_resolve_base_call` to clear the Sonar Zero Issues Gate.

The S3776 finding (cognitive complexity 53) is pre-existing — the identical issue sits on `main` at line 276 for the same function. Renaming the body to `_resolve_csharp_method_call` made Sonar's PR analysis treat it as new code, so the zero-new-issues gate fired. Extracted the two four-deep nested base-chain walks, which were its densest self-contained part.

The extraction turned `if class_qn := ...` into an early `if class_qn is None` return. Those differ only if the value can be an empty string; `_containing_class_qn` returns `str | None` and never `""`, so they are equivalent. Verified rather than argued: inverting that guard's polarity reddens `test_base_receiver_binds_base_chain_never_own_override` and `test_base_call_binds_first_party_base_by_name_when_arity_differs`, so the existing suite does cover the moved path.

996 passed, 1 skipped locally across the C#, resolution and type-inference modules.

vitali87:
@greptileai review

vitali87:
Pushed e4f85a6b. The Sonar gate was still red at 27 after the first extraction, so three more branches came out of `_resolve_csharp_method_call`:

| helper | owns | complexity |
|---|---|---|
| `_resolve_base_call` | the `base.X()` chain walk | 13 |
| `_method_or_property_target` | the property-vs-method check, previously duplicated | 1 |
| `_resolve_non_member_call` | bare `Foo()` simple-name lookup | 1 |
| `_external_or_unresolved` | the final external-target fallback | 3 |

`_resolve_csharp_method_call` is now 14, under the 15 limit.

To avoid guessing across 12-minute CI cycles I scored the function locally and calibrated the scorer against three known SonarCloud readings on this same function — 53 on `main`, 53 at d1cbe062, 27 at 91353ad1 — which it reproduced exactly.

Each extraction is mutation-verified rather than assumed covered: dropping the property check reddens 2 tests, the bare-call lookup 5, the object-virtuals check 1, and inverting the base-call guard 2. 996 passed, 1 skipped locally.

Three functions in this file remain above the threshold (`_receiver_type_arity` 18, `_find_extension_method` 38, `_resolve_receiver_class_qn` 39). All are pre-existing on `main` and untouched here, so the new-code gate does not flag them; folding them into a #1800 fix would be unrelated churn.

Issue #1798, opened 2026-09-08
reingest leaves a stale Package node when __init__.py is deleted (mislabelled, or duplicated)


vitali87:
Found by the `fuzz_incremental_update` harness from #1731 on its first 60-second run, then reduced to a deterministic reproducer.

## Symptom

Deleting a package's `__init__.py` should demote that directory from `Package` to `Folder`. A full index does exactly that. `reingest` does not: the node keeps its `Package` label and its `CONTAINS_FILE` / `CONTAINS_MODULE` edges stay anchored to the `Package`, so the incremental graph disagrees with a clean index of the identical tree.

```
after full index:       [('Package', 'proj.pkg')]
after reingest(delete): [('Package', 'proj.pkg')]     <- unchanged
clean index of same:    [('Folder', 'pkg')]           <- demoted correctly
```

## Reproducer

```python
FIXTURE = {"pkg/__init__.py": "", "pkg/util.py": "def helper():\n    return 1\n"}

store = _StatefulIngestor()
updater = GraphUpdater(ingestor=store, repo_path=root, parsers=parsers,
                       queries=queries, project_name="proj")
updater.run(force=True)

(root / "pkg/__init__.py").unlink()
updater.reingest([], deleted=["pkg/__init__.py"])
# {(label, uid) for ... if label in ("Package", "Folder")}
#   -> {('Package', 'proj.pkg')}, but a clean index gives {('Folder', '.../pkg')}
```

## Full delta from the harness

Node identity differs, and every containment edge moves with it:

```
extra nodes:   ('Package', 'proj.pkg', ...)
missing nodes: ('Folder',  '<root>/pkg', ...)
extra edges:   ('Package', 'proj.pkg', 'CONTAINS_FILE',   'File',   '<root>/pkg/app.py')
               ('Package', 'proj.pkg', 'CONTAINS_MODULE', 'Module', 'proj.pkg.app')
               ... (one per file and module in the directory)
missing edges: ('Folder', '<root>/pkg', 'CONTAINS_FILE',   'File',   '<root>/pkg/app.py')
               ('Folder', '<root>/pkg', 'CONTAINS_MODULE', 'Module', 'proj.pkg.app')
               ... (the same set, re-anchored to the Folder)
```

## Why it matters

[...]

PR #1803, opened 2026-09-08
fix: fuzz the rm-path guard the shell harness was skipping (#1731)

vitali87:
Follow-up to #1796, which merged before these two commits were pushed.

## The gap

`fuzz_shell_command.py` on `main` never exercises `_is_dangerous_rm_path`. The harness passes `project_root` to `_validate_segment` on the assumption that this reaches both root-aware guards, but only `_git_escapes_project` is called from there. `_is_dangerous_rm_path` has a single call site, inside `ShellCommander.execute` **after** `_validate_segment`, so passing a root does not reach it.

Measured divergence on `main` today, harness verdict vs production:

| command | harness | production |
|---|---|---|
| `rm *` | safe | blocked |
| `rm /` | safe | blocked |
| `rm /tmp/zzz` | safe | blocked |
| `rm -r -- -x/../../outside/victim` | safe | blocked |

All one-directional: the harness is never stricter than production, only laxer. So the fuzzer cannot find a regression in this guard, which is the point of fuzzing it.

`rm -rf /` is NOT in that table, and that is why the gap survived an initial check: it is caught by `_check_segment_patterns` regardless, so its "dangerous" verdict says nothing about the rm guard. The reason string is the tell — `"rm with root path"` from the pattern matcher, not `"rm targeting ..."` from the guard.

## The fix

Call the guard where `execute` calls it: in the segment loop, on the split segment, after `_validate_segment`. A spy confirms it now runs (`guard_calls=1`) and returns its own reason strings. `ls -la` and `rm ./build` stay safe, so the widening did not simply make the classifier stricter.

## Coverage

Five parity cases, each decided by exactly one of the guards this touches:

- `echo $(whoami)`, `` echo `id` `` — the subshell ordering
- `rm /tmp/zzz`, `rm *`, `rm -r -- -x/../../outside/victim` — the rm guard

Verified by mutation that each half is genuinely pinned: deleting the subshell block reddens on `echo $(whoami)`; deleting the rm call reddens on the rm cases; the pre-existing parity cases stay green through both. Before this, either half could have been reverted with the suite staying green.

## Also here

A comment-only commit recording that both measured #1798 shapes produce containment rows on the missing side only, and why the extra-side clause was deliberately kept rather than narrowed. Zero non-comment lines.

## Verification

- `test_fuzz_harnesses.py`: 16 passed, on this branch rebased onto current `main`
- 15 seed corpus files and 400 random inputs through the widened path: 0 failures
- The frozen target previously did 119,393 runs in 61s in `clusterfuzzlite-run-fuzzers:v1`

## Summary by CodeRabbit

* **Tests**
  * Expanded shell-command safety coverage to include subshell syntax, dangerous `rm` paths, empty commands, and punctuation-only commands.
  * Improved fuzz testing to align more closely with production safety classification, including consistent project-root handling and explicit rejection of unsafe or incomplete commands.

* **Documentation**
  * Added explanatory notes clarifying containment behavior in incremental update fuzz testing.

### DIFF
--- a/codebase_rag/tests/test_fuzz_harnesses.py
+++ b/codebase_rag/tests/test_fuzz_harnesses.py
@@ -245,6 +245,18 @@ def test_shell_harness_agrees_with_the_classifier(shell_harness: ModuleType) -> 
         ("bash script.sh", True),
         ("python3 script.py", True),
         ("make install", True),
+        # Decided ONLY by the subshell guard, which runs before pipeline
+        # patterns. Nothing else in `_classify` refuses these, so they go red
+        # if that call is dropped or moved after segmentation.
+        ("echo $(whoami)", True),
+        ("echo `id`", True),
+        # Decided ONLY by `_is_dangerous_rm_path`. `rm -rf /` above is caught
+        # by `_check_segment_patterns` regardless, so it cannot stand in for
+        # these: each needs the guard the harness calls explicitly, because
+        # `_validate_segment` does not call it.
+        ("rm /tmp/zzz", True),
+        ("rm *", True),
+        ("rm -r -- -x/../../outside/victim", True),
     ):
         dangerous, _reason = shell_harness._classify(command)
         assert dangerous is expected, f"{command!r} classified {dangerous}"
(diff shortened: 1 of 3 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1815, opened 2026-09-09
fix: re-index a deleted path that is still present on disk

### LINKED ISSUE #1799: reingest erases a file's definitions when it is named deleted but exists on disk
Found by the `fuzz_incremental_update` harness from #1731, then reduced to a deterministic reproducer.

## Symptom

When a path is passed in `deleted` but the file exists on disk at the time `reingest` runs, the file's definitions are erased from the graph and every importer is downgraded to a phantom `ExternalModule`, even though a clean index of the same tree indexes the file normally.

```
changed only (normal)              funcs: [... 'proj.pkg.util.helper']   phantom_ext: []
changed AND deleted (the race)     funcs: ['proj.main.main', 'proj.pkg.app.run']  phantom_ext: ['util']
deleted only, file back on disk    funcs: ['proj.main.main', 'proj.pkg.app.run']  phantom_ext: ['util']
```

Full delta against a clean index:

```
extra nodes  : ('ExternalModule', 'util')      <- phantom, path=proj.pkg.util.helper
missing nodes: ('File',     '<root>/pkg/util.py')
               ('Function', 'proj.pkg.util.helper')
               ('Function', 'proj.pkg.util.added')
```

`pkg/util.py` is present on disk throughout.

## Reproducer

```python
store = _StatefulIngestor()
updater = GraphUpdater(ingestor=store, repo_path=root, parsers=parsers,
                       queries=queries, project_name="proj")
updater.run(force=True)

(root / "pkg/util.py").write_text("def helper():\n    return 99\n")   # exists
updater.reingest(["pkg/util.py"], deleted=["pkg/util.py"])            # also named deleted
# -> proj.pkg.util.helper is gone; pkg/app.py's import now points at an
#    ExternalModule named 'util'
```

Passing the path in `deleted` alone, with the file present, reproduces it identically, so the trigger is "named as deleted while existing", not the combination with `changed`.

## Why it matters

This is the atomic-save race `reingest`'s own contract says it handles. `GraphUpdater.reingest`'s docstring states that `deleted` "forces removal even if a same-named file reappeared (the watcher's atomic-save race)" — but the observed behaviour also removes the definitions of a file that is *currently present and parseable*, and does not re-add them, so the graph ends up denying that a real file's symbols exist.

An editor that writes via delete-then-rename produces exactly this event pair, and the watcher forwards both. The result is not a transient inconsistency that the next pass repairs: the file is unchanged afterwards, so a hash-cache-based incremental run has no reason to revisit it, and the symbols stay missing until a full re-index.

Downstream, `pkg/app.py`'s `from pkg.util import helper` resolves to a phantom `ExternalModule` named `util`, so the import looks like a third-party dependency rather than a first-party module.

## Suggested fix

Decide the removal on the state of the tree at the moment of the re-ingest rather than on the event list alone: a path named in `deleted` that exists and parses should be re-indexed, not retracted. If the forced-removal semantics are wanted for the genuine race, they need to be followed by a re-parse of the reappeared file before the graph is flushed.

Present on `main`; not introduced by #1796. Reproducer committed to `fuzz/corpus/fuzz_incremental_update/` per the #1731 triage rule.

vitali87:
Closes #1799.

An editor that saves via delete-then-rename makes the watcher forward a DELETE for a path that exists again by the time `reingest` runs. `_reingest_split` put every path named in `deleted` into `gone` unconditionally, so the file was retracted and never re-parsed:

```
extra nodes  : ('ExternalModule', 'util')      <- phantom
missing nodes: ('File',     '<root>/pkg/util.py')
               ('Function', 'proj.pkg.util.helper')
```

`pkg/util.py` is present on disk throughout. No later incremental run repairs it: the file is unchanged, so its hash still matches and the walk never revisits it. Downstream, `pkg/app.py`'s `from pkg.util import helper` resolves to a phantom `ExternalModule`, so a first-party module reads as a third-party dependency.

## The fix

A deleted path that is still a file routes to `present`, which re-parses it. The forced removal is kept, not skipped: the graph still holds the OLD file's symbols and they must go. `_reingest_delete` already clears every reparse key before the re-parse writes the current one, so a definition the reappeared file no longer has does not survive.

`is_file()` rather than `exists()`, so a directory sitting where the deleted file was is still treated as a deletion — `test_reingest_deletes_a_file_a_directory_has_replaced` pins that and is what rules `exists()` out.

| path named in `deleted` | before | after |
|---|---|---|
| present file | retracted | re-parsed |
| absent | retracted | retracted |
| directory in its place | retracted | retracted |

## Report semantics

A present deleted path now appears in `reparsed` rather than `removed`. Swept unscoped: the only readers are `mcp/tools.py:1520` and `structural_delta.py:850`, both pass-through, and `realtime_updater.py:328` ignores the report. The change makes `structural_delta` strictly more accurate — a file listed as removed contributes no symbols to `touched`, so "tests reaching this change" previously omitted tests for a file that still exists.

## The corpus seed did not reproduce the bug

The seed committed as `repro_1799_deleted_but_present.bin` decoded to two plain deletes, leaving both paths genuinely absent at the call, so the race was never exercised. The harness passed on it — indistinguishable from passing because the code is correct.

It is regenerated from `build_corpus.py` (delete then recreate `pkg/util.py`), and verified three ways: it reaches `reingest` with the path on disk, it fails against the unfixed code, and the test asserting all this goes red when the old seed is planted back.

## Tests

Four in `test_reingest.py`, all red on `main`:

| test | pins |
|---|---|
| `..._reindexes_a_deleted_path_that_is_present_on_disk` (×2: `deleted_only`, `changed_and_deleted`) | graph equals a clean index |
| `..._drops_definitions_the_reappeared_file_no_longer_has` | the delete still happens — rules out "skip the delete" |
| `..._reports_a_present_deleted_path_as_reparsed_not_removed` | report semantics |

Plus `test_the_1799_seed_really_reaches_a_deleted_path_that_exists`, which observes `reingest`'s arguments rather than the decoded plan, so a seed that stops reproducing cannot pass silently.

Each mutation reddens only what it should: reverting to `gone[key] = path` reddens exactly the three new reingest tests; `exists()` instead of `is_file()` reddens only the directory-replacement guard; never deleting reddens 9 including pre-existing tests.

## The suppression is removed, not kept

`_resurrected_file_residue` and its three unit tests are deleted, so the differential oracle re-detects a regression instead of filtering it away. To rule out the filter having masked a co-occurring defect, I enumerated every plan the harness can encode (44,135 for ≤3 edits), ran all 515 that reach deleted-but-present, and got zero oracle failures.

## Verification

- `test_reingest.py` + `test_fuzz_harnesses.py`: 93 passed.
- Reingest/watcher/incremental/pruning/single-file subset under the host lock, `-n 4`: 402 passed, 0 failed.
- `ty check`: 9 warnings, identical on unmodified `main` — zero added.
- ruff check/format clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where files recreated during an atomic save could be incorrectly treated as deleted.
  * Recreated files are now re-parsed, ensuring current definitions are restored and removed symbols do not persist.
  * Processing reports now identify these files as reparsed rather than removed.
  * Files that become unreadable during processing are no longer reported as successfully reparsed.

* **Tests**
  * Added regression coverage for delete-and-recreate update sequences.
  * Expanded incremental update fuzz testing to detect recurrence of this issue.


vitali87:
@greptileai review

vitali87:
Fixed in a7a0ca8e — the finding is real and I reproduced it before fixing it.

`_reingest_reparse` had `except OSError: continue`, so a file that became unreadable between the split and the read was skipped silently while the report still named it in `reparsed`. `_reingest_delete` had already removed its entities, so the caller was told the graph holds definitions it no longer has. `reparsed` is now derived from the files that actually re-parsed, and the skip is logged.

Worth recording: this is **pre-existing on `main`**, not introduced here. A path named in `paths` has always been classified by `is_file()` the same way, and `reingest(["pkg/util.py"])` with the same race reproduces it identically on unmodified `main` — reported `reparsed=('pkg/util.py',)` with the function gone. The deleted-but-present fix gives it a second entrance, so it is closed here for both routes rather than only the new one.

Regression test `test_reingest_does_not_report_an_unreadable_file_as_reparsed` wins the race deterministically by swapping the file for a directory inside `_reingest_split`. Mutating `reparsed` back to `tuple(sorted(present))` reddens exactly that one test and nothing else.

Verification: 452 passed across reingest/watcher/incremental/pruning/single-file/structural-delta/fuzz, `-n 4` under the host lock. `ty check` still 9, identical to `main`.

@greptileai review

Issue #1582, opened 2026-08-31
Six stacked epic PRs show green while their last CI run failed: stale runs drop off the head and only CodeQL remains


vitali87:
Six of the ten open epic PRs present a green check set while their branch's last CI run **failed**. The failure is not shown as pending or missing — it is replaced on the PR page by an unrelated green workflow, so the PR reads as passing.

## Measurement

All ten layers of the Agent IDE epic (base is a sibling branch in every case):

| PR | branch | checks shown | CI runs at current head | last CI conclusion | red visible on PR? |
|---|---|---|---|---|---|
| #1539 | `feat/type-facts` | 29 | 1 | failure | **yes** |
| #1542 | `feat/graph-query-tools` | 29 | 1 | failure | **yes** |
| #1543 | `feat/edge-confidence` | 29 | 1 | cancelled | — |
| #1545 | `feat/rename-op` | 29 | 1 | failure | **yes** |
| #1546 | `feat/structural-delta` | 7 | **0** | failure | no |
| #1547 | `feat/postcondition-contract` | 7 | **0** | failure | no |
| #1548 | `feat/change-signature` | 7 | **0** | failure | no |
| #1549 | `feat/move-op` | 7 | **0** | failure | no |
| #1550 | `feat/extract-inline` | 7 | **0** | failure | no |
| #1551 | `feat/context-slice` | 7 | **0** | failure | no |

The seven checks on the lower group are five CodeQL `Analyze` jobs, the CodeQL umbrella, and a `null` CodeRabbit. No unit tests, no type check, no lint. `gh pr view 1551 --json statusCheckRollup` reports **0 FAILURE** — zero failures out of a set containing no tests.

For contrast, #1571 (based on another `fix/` branch) shows 69 checks including `Unit Tests` on six platform/version combinations, and #1566 (based on `main`) shows 29. Seven is not a smaller sample of the same population; it is a different population.

## Why the two groups differ

**Corrected.** This issue originally asserted that `ci.yml` and `sonarcloud.yml` restricting `pull_request` to `branches: [main, master, develop]` meant a push to a stacked branch triggers no CI at all. **That mechanism is wrong**, and a reader can falsify it in one query. Recorded rather than quietly edited, because the wrong mechanism is the more dangerous half: it would send someone to change workflow triggers that are not the problem.

Three stacked PRs whose bases have never changed (no `base_ref_changed` events in their timelines, only force-pushes) now show full CI:

[...]

PR #1551, opened 2026-08-30
feat: context tool, a graph-ranked token-budgeted context slice (#1536)

vitali87:
Implements T14 of epic #1521 (closes #1536). Stacked on #1548 (which carries the graph query tools of #1542 and the type facts of #1539); retarget once the stack lands.

## What

- `codebase_rag/context_slice.py`: `context(fetch_all, project, target, budget_tokens, repo_root, search=)`. `target` is a qualified name, a bare name, `path:line`, or free text (embedding similarity through the semantic extra; the best match becomes the target and each candidate's similarity is kept as a tie-breaker).
- Pieces, each with `why_included`: the target's source (distance 0); each direct caller's call line from the per-site location (#1522), not the whole caller; each direct callee's signature; the headers of the types it `RETURNS`/`ACCEPTS` (#1527); the full source of each reaching test with its depth and the symbol it reaches through; sections of documents whose `LINKS_TO` points at the target's file, those mentioning the name or else the first section.
- Ranking: graph distance, then trace hotness (`dynamic_call_count` on the caller's edge), then similarity, then name. Tokens counted with the shared `count_tokens`; pieces taken in order while they fit, so the total never exceeds the budget. The target alone is trimmed line by line to fit (`truncated`); a budget too small for its first line names it in `omitted`. Everything that did not fit is listed in `omitted`.
- Cypher: `CYPHER_CONTEXT_HOTNESS`, `CYPHER_CONTEXT_TYPES`, `CYPHER_CONTEXT_DOC_SECTIONS`; the in-memory emulator answers these plus the resolve and callees queries.
- MCP tool `context` (`target`, `budget_tokens` default 4000, `project`) under the ingestor lock; `cgr context TARGET [--budget N]`; docs in `docs/architecture/context-slice.md`.

## Tests

`codebase_rag/tests/test_context_slice.py` (5): with a 4k budget a fixture function's slice holds its source, its three callers' call lines, its callee's signature, the type it returns, the reaching test and the doc section that mentions it, and nothing from an unrelated module; the token count never exceeds the budget across shrinking budgets (trimmed target first, then omitted); distance then trace hotness orders the pieces; targets by bare name, location and free text (with and without a searcher); the MCP tool. Unit suite: 8963 passed plus the lock-guard fix.

## Summary by CodeRabbit

* **New Features**
  * Added a `context` CLI command and MCP tool for generating focused code context around a target.
  * Supports qualified names, file locations, and free-text target resolution.
[...]

Issue #1581, opened 2026-08-31
CodeRabbit auto-review is skipped on every stacked PR, so 14 of 18 open PRs have never been reviewed


vitali87:
CodeRabbit auto-review is disabled on any PR whose base is not the default branch, so on a stacked chain **only the bottom PR is ever reviewed**. Measured today: 14 of 18 open PRs have never been reviewed by CodeRabbit, and each carries a skip notice that is easy to read as a clean result.

## What happens

Every open PR with `baseRefName != main` has exactly one CodeRabbit comment:

> **Review skipped** — Auto reviews are disabled on base/target branches other than the default branch. [...] To trigger a single review, invoke the `@coderabbitai review` command.
>
> Configuration used: **defaults**

Measured across the two live stacks:

| PR | base | CodeRabbit skip notices | real CodeRabbit reviews |
|---|---|---|---|
| #1551 | `feat/extract-inline` | 1 | 0 |
| #1550 | `feat/move-op` | 1 | 0 |
| #1545 | `feat/structural-delta` | 1 | 0 |
| #1539 | `feat/scoped-reingest` | 1 | 0 |
| #1571 | `fix/incremental-implements-edges` | 1 | 0 |
| #1572 | `fix/incremental-deleted-dependents` | 1 | 0 |
| #1573 | `fix/incremental-init-package` | 1 | 0 |
| #1576 | `fix/incremental-stem-siblings` | 1 | 0 |

Full population: 14 of 18 open PRs have a non-default base. The 4 based on `main` are reviewed normally — #1566 (base `main`) has 10 non-empty comments including three `@coderabbitai review` invocations and their replies.

## Why it is a merge hazard rather than a nuisance

The four incremental PRs currently sit at `mergeStateStatus=CLEAN` with 69/69 checks green and bot comments present. Those comments are the skip notice and SonarCloud's Quality Gate — a gate, not a review. So the combination that normally reads as "ready" (CLEAN + all checks green + bot comments present) is fully satisfied on a PR that no reviewer has read.

This defeats the check CONTRIBUTING/CLAUDE.md already asks for. "Query both `.comments[]` and `.reviews[]`, require the body to be non-empty *and* carry the verdict you are looking for" catches it only if the reader opens the body, because a **count** cannot see it. And the count is not merely misleading, it is unstable: non-empty comments range from 2 to 5 across these PRs while the true review count is identically zero.

[...]

PR #1819, opened 2026-09-09
ci: fail a PR that has no CI run at its head commit

vitali87:
Closes the gap in #1582's option 2.

## The hole

A required status check that is **absent** does not block a merge. GitHub cannot fail a check nobody reported, so "required" in practice means "must not be red", not "must be present".

`All Checks Pass` — the ruleset's only required check — is a job inside `ci.yml`. Whenever `ci.yml` does not run at a PR's head, that job does not exist there, the requirement is vacuously satisfied, and the PR reports mergeable having run no tests.

Measured on #1551 while writing this:

```
mergeStateStatus:              CLEAN
mergeable:                     MERGEABLE
checks present:                7
'All Checks Pass' among them:  False
```

Seven CodeQL/CodeRabbit checks, zero unit tests ever run on that branch, and GitHub says it is ready to merge.

## The fix

A workflow that exists to be **present where `ci.yml` is not**. No `branches` filter, so it reports on every pull request, and it fails unless a run of the `CI` workflow exists at the exact head SHA.

That is the base-agnostic formulation #1582 settled on after two branch-filter theories were proposed and falsified: it does not ask *why* a run fired, only whether one covers the commit being merged.

Three details the issue's own caveats forced:

- Queries `actions/runs?head_sha=…`, **not** `gh run list --branch`. After a stack rebase several branches share a head, and the branch query returns runs belonging to *other* PRs — a trap #1582 records explicitly.
- Matches the workflow `name` exactly and prints every run found, either way. A bare `| length == 0` cannot distinguish "no run at this head" from "the name did not match".
- Asserts **existence only**, not conclusion. Whether the run passed is `All Checks Pass`'s job; duplicating it here would disagree with that gate while a run is in flight.

## Validation

Both directions, against live data, before committing:

| PR | base | CI runs at head | verdict |
|---|---|---|---|
| #1818 | `main`, merged clean | 1 | pass |
| #1547 | `main`, in review | 1 | pass |
| #1551 | feature branch | 1 | pass |
| #1549 | head pushed minutes ago | **0** | **fail** |

The last is correct and is the point — those commits are verified only by a local suite run.

The known-positive matters here: at #1818's head six workflows reported (`CI`, `ClusterFuzzLite PR`, `OSV-Scanner`, `PR #1818`, `PR Split Score`, `SonarCloud`), so the name match is genuinely selecting rather than trivially succeeding.

## Deliberately not included

**Adding this to the ruleset's required checks.** That is the owner's call and wants to be a separate, deliberate step. Until then it reports without blocking — which is the right order, since a gate that started blocking the moment it merged would block every PR currently open, including the five-layer stack.

Also not included: any change to workflow triggers. #1582 withdrew that fix after measuring a `pull_request` CI run firing for a PR whose base was not in the filter list, and I independently reproduced that before writing this.

## Summary by CodeRabbit

* **Chores**
  * Pull requests now verify that a CI run belongs to the specific pull request being reviewed.
  * CI validation checks only the intended CI workflow, improving accuracy.
  * Validation now includes CI runs beyond the initial results limit.
  * Run details provide clearer information about the pull requests associated with each run.

### DIFF
--- a/.github/workflows/require-ci-at-head.yml
+++ b/.github/workflows/require-ci-at-head.yml
@@ -37,9 +37,18 @@ jobs:
         run: |
           set -euo pipefail
 
+          # A missing head SHA must NOT be a pass. `workflow_dispatch`
+          # carries no pull_request payload, so an earlier `exit 0` here
+          # let anyone manufacture a green `CI Ran At Head` by dispatching
+          # this workflow by hand -- a gate that reports success without
+          # asserting anything, which is the exact defect it exists to
+          # prevent.
           if [ -z "${HEAD_SHA:-}" ]; then
-            echo "No pull_request head SHA in the event payload; nothing to assert."
-            exit 0
+            echo "::error::No pull_request head SHA in the event payload."
+            echo "This check asserts a property of a pull request head and"
+            echo "cannot do so outside a pull_request event. Dispatching it"
+            echo "manually proves nothing and must not report success."
+            exit 1
           fi
 
           echo "Head SHA: $HEAD_SHA"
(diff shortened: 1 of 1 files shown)


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
## A deliberate design point, measured after opening this

A peer measured something that bears directly on what this gate counts, and I reproduced it.

At #1551's current head there is a **hand-dispatched CI run whose `All Checks Pass` succeeded** — and `gh pr checks 1551` does not list it. Seven checks shown, `All Checks Pass` not among them. So a `workflow_dispatch` run on a stacked PR produces a real result that the PR page never surfaces.

This gate **counts that run** and would report pass. I think that is right, but it is a choice worth stating rather than leaving for someone to find:

- The gate asks *"was this commit tested?"*, and a dispatched run genuinely tested it. Excluding `workflow_dispatch` would make the gate unsatisfiable on any PR whose base is not `main`, since that is currently the only way to get CI on those branches at all.
- The consequence is that the gate can pass where the PR page shows no visible evidence. The run exists in the Actions API; it is the checks *list* that omits it.

So this closes the "no test ever ran" hole, and does **not** close the separate "a result exists but is invisible on the PR page" problem. That second one is #1581's territory, not this PR's.

If the preference is for the gate to require a *visible* result, the change is to also require `event == "pull_request"`. I have not made it, because today that would fail every stacked PR including ones whose suite genuinely passed — trading a false pass for a false fail, on branches nobody can currently fix.

vitali87:
@greptileai review

Issue #1617, opened 2026-09-02
CodeRabbit docstring-coverage warning contradicts the PR template's no-docstrings rule


vitali87:
## Problem

CodeRabbit's pre-merge "Docstring Coverage" check warns whenever fewer than 80% of the functions touched by a diff carry docstrings. The repository's own pull request template (`.github/pull_request_template.md`, line 38) requires the opposite:

```
- [ ] No new comments or docstrings (code should be self-documenting)
```

So any PR that touches test functions, which by convention here have no docstrings, gets a warning it is not allowed to clear. #1614 (a prose-only commit that added no functions) was flagged at 30% because the comment edit sat inside an existing undocumented test; satisfying the bot would have meant adding docstrings to 28 pre-existing tests, violating the template and turning an AST-identical doc fix into a 28-function change.

The warning is currently treated as acknowledged-and-declined at merge time, which weakens the "every bot finding cleared before merge" rule by making one class of warning routinely ignorable.

## Proposed fix

Disable or lower the docstring coverage pre-merge check in `.coderabbit.yaml` so the bot's threshold matches the template. If a docstring standard is wanted for non-test code, scope the check to `codebase_rag/` excluding `codebase_rag/tests/` and update the template line to say so, so the two sources agree.

Observed on #1614 at 3172edb4 (CodeRabbit pre-merge checks 4 passed, 1 warning after the description was templated).

PR #1822, opened 2026-09-09
ci: auto-review stacked PRs whose base is a feat/ or fix/ branch

### LINKED ISSUE #1581: CodeRabbit auto-review is skipped on every stacked PR, so 14 of 18 open PRs have never been reviewed

vitali87:
Closes #1581.

CodeRabbit auto-review defaults to the default branch only, so on a stacked
chain every layer above the bottom is skipped and only the bottom PR is ever
read. The issue measured 14 of 18 open PRs never reviewed, each carrying a
skip notice that is easy to read as a clean result.

`reviews.auto_review.base_branches` accepts regex patterns and defaults to
`[]`, which is precisely why every stacked PR was skipped.

## One correction to the issue

#1581 states the repo has no `.coderabbit.yaml` and the fix is a pure
addition. It has one now, added by 9320c7cf for #1617 to turn off the
docstring pre-merge check. So this is an edit, and that setting had to be
preserved. Verified it still parses as the string `off` after the change --
the quotes are load-bearing, since unquoted `mode: off` yields boolean
`False`, which is schema-invalid.

## Why two anchored prefixes rather than `.*`

A base that is neither the default branch nor a recognised stack layer should
keep default behaviour rather than silently opting in.

The `^` is load-bearing. Unanchored, under search semantics, `feat/.*` and
`fix/.*` would also match `notfeat/x`, `release/feat/x`, `myfix/x` and
`hotfix/urgent`. Anchored, all four correctly do not match. The trailing `.*`
is redundant under search semantics but required under fullmatch, and the
vendor docs do not say which is used, so it is kept.

## Verification

- Validated against the live vendor schema with `Draft202012Validator` (the
  dialect the schema declares): 0 errors. Four controls confirm the validator
  can fail -- string-not-array, a root typo key, `mode: false`, a bad enum.
- Prefix coverage is 9 of 9: every PR ever opened against a non-`main` base
  (#1548-#1551, #1274, #1260, #1114, #1112, #1111) used a `feat/` or `fix/`
  base. The 64 `chore/docs/test/ci/refactor/perf` PRs were all based on
  `main`.
- Patterns tested against real bases: `feat/postcondition-contract`,
  `feat/move-op`, `fix/incremental-init-package` all match; `main` does not.

Note when re-deriving that coverage: PRs retargeted to `main` before merging
now REPORT `base=main`, so recent history understates which prefixes were
stack bases. #1571-#1576 read as `main` today and were `fix/...` when the
issue was written.

## Known gap, not introduced here

Nothing validates `.coderabbit.yaml` before merge. `check-yaml` checks syntax
only, and the vendor schema sets `additionalProperties: false` at the ROOT
only -- so a `base_branchez` typo under `auto_review` validates with 0 errors
and silently restores the bug this PR fixes. Measured, not assumed.

The repo has the right precedent in `scripts/validate_labels.py`, wired into
pre-commit and `ci.yml`. A follow-up doing the same for `.coderabbit.yaml`
would close it. Filing separately rather than widening this PR.

## Summary by CodeRabbit

* **Chores**
  * Configured automatic reviews for recognized feature and fix branch bases.
  * Existing behavior for other branch bases remains unchanged.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -10,3 +10,18 @@ reviews:
   pre_merge_checks:
     docstrings:
       mode: "off"
+  auto_review:
+    # Auto-review defaults to the default branch only, so on a stacked chain
+    # every layer above the bottom is skipped and only the bottom PR is ever
+    # read (issue #1581: 14 of 18 open PRs had never been reviewed). The skip
+    # is easy to misread as a clean result -- the same "Review skipped"
+    # headline covers "already reviewed the last commit", which means the head
+    # IS covered, so the notice cannot be judged without opening its body.
+    #
+    # These patterns are anchored and cover the two prefixes stacks are cut
+    # from here. Deliberately not ".*": a base that is neither the default
+    # branch nor a recognised stack layer should keep the default behaviour
+    # rather than silently opting in.
+    base_branches:
+      - "^feat/.*"
+      - "^fix/.*"


vitali87:
@greptileai review

Issue #1828, opened 2026-09-09
Gloss nodes survive delete_project: CYPHER_DELETE_PROJECT does not traverse ANNOTATES


vitali87:
Follow-up to #1808, found while registering the `Gloss` node label (commit on `feat/1808-annotation-node`). Filing rather than fixing, because the schema PR is deliberately registration-only with no behaviour.

## The leak

`CYPHER_DELETE_PROJECT` (`codebase_rag/cypher_queries.py:92-97`) reaches nodes through containment and definition only:

```
MATCH (p:Project {name: $project_name})
OPTIONAL MATCH (p)-[:CONTAINS_PACKAGE|CONTAINS_FOLDER|CONTAINS_FILE|CONTAINS_MODULE|CONTAINS_SECTION*]->(container)
OPTIONAL MATCH (container)-[:DEFINES|DEFINES_METHOD*]->(defined)
DETACH DELETE p, container, defined
```

A `Gloss` hangs off its subject via `ANNOTATES`, which is in neither list. Deleting a project therefore deletes the symbol a gloss is about and leaves the gloss behind, with its `ANNOTATES` edge detached. Re-indexing the same project then produces fresh symbol nodes that the orphaned gloss is not attached to.

## Why it is not simply "add ANNOTATES to the traversal"

A gloss has a deliberately unusual lifecycle, and the two requirements pull apart:

- it MUST survive `index_repository` / `update_repository` / reingest, because unlike every other node its truth lives only in the graph and is not rebuilt from source
- it MUST NOT survive deletion of the project it describes, or it is unreachable garbage

The repo already has both halves of this pattern:

- `:IncompleteRun` (`cypher_queries.py:56-59`) is deliberately kept unreachable from this traversal *precisely so it survives* the rebuild
- `prune_unanchored_resources` (`services/resource_cleanup.py:42`) is a separate cleanup pass invoked from `delete_project` (`services/graph_service.py:363`) for exactly this class of node: one that escapes the containment traversal and needs its own sweep

So the likely fix is the `prune_unanchored_resources` shape, not extending the traversal: a pass that deletes glosses whose subject no longer exists. That keeps reingest survival intact while closing the leak.

## Suggested test

Two cases, since one alone passes for the wrong reason:

1. index, write a gloss, reingest, assert the gloss is still present and still attached
2. index, write a gloss, `delete_project`, assert the gloss is gone

[...]

Issue #1808, opened 2026-09-08
Annotation node: durable, symbol-keyed notes an AI agent writes about code and retrieves later


vitali87:
## Summary

Add an `Annotation` node: durable, symbol-keyed notes that an AI agent writes about code and retrieves later. Stored in the graph, never written into source files.

This is the answer that came out of researching #1592. That issue asks whether parsed code comments should be nodes; the evidence says no (see the comment there). But the node shape it proposes is right for a different payload — agent-authored annotations.

## The criterion

Across every system surveyed, a node is justified by exactly one thing: **edges out of it**.

- Kythe's `doc` node carries `param.0 VarX` — edges to the entities its text mentions
- Roslyn promotes doc comments (and only doc comments) to real syntax nodes so `<see cref="Foo"/>` is bindable
- Systems that only read documentation back on the entity use a field: SCIP, JDT, LSP, ctags, SPDX, and Doxygen's own XML output

An agent note like "this mirrors `parse_header`; safe because `validate()` runs first" **is** edges to other symbols. That is the case a node exists for.

## Why not in the source file

In-band payloads are disqualified by documented formatter behaviour:

- Black's style doc: comments "may sometimes be moved because of formatting changes, **which can break tools that assign special meaning to them**"
- gofmt **relocates** `//go:generate` directives to the end of the doc comment; Go 1.19's gofmt reformatted ~3% of public doc comments, ~13% of those not preserving inferred structure
- Black has shipped bugs **deleting** comments (#2874 power operators, #3909 parenthesized strings, #3162 implicit concatenation)
- Prettier's comment attachment is a maintainer-acknowledged architectural weakness (umbrella issue #275); #17015 shows placement changing *what the comment applies to*
- Ruff special-cases pragma comments specifically "to prevent Ruff from moving pragma comments around, thereby modifying their meaning"

Structured comments survive only because each formatter individually exempts them. That is not achievable across 15 languages.

## Key: `(path, qualified_name)`

Measured on four repos, two languages, 1 to 887 authors — for each commit, parse before/after, build qualified names by AST nesting, count survivors:

| Repo | Language | Authors | `qualified_name` survives |
|---|---|---|---|
| croft | Rust | 6 | 99.9% |
| code-graph-rag | Python | 1 | 99.2% |
| graphiti | Python | 22 | 99.1% |
| llama_index | Python | 887 | 98.5% |

[...]

PR #1830, opened 2026-09-09
feat: add the Gloss node for agent-authored notes (#1808)

vitali87:
Part of #1808. Schema and wire format only: registers the node and its edges, and nothing yet writes or reads one. Anchoring, staleness, retrieval and dedup are deliberately not here.

## The name is `Gloss`, not `Annotation`

The issue specifies `Annotation`. This codebase already uses that word for Java annotations and decorators (`JavaAnnotationInfo`, `MethodModifiersAndAnnotations`, `extract_annotation_info`, `TS_ANNOTATION`, and the `decorators` property), so the label would have read ambiguously against existing code permanently. A protobuf field number is append-only wire format, so a later rename is the expensive kind of change — which made this the moment to settle it rather than after merge.

The owner chose `Gloss` from a shortlist verified unused repo-wide. It is the term of art for a note attached to a specific passage, which is exactly this node.

The edge names are unchanged: `Gloss -[:ANNOTATES]-> Function`, `Gloss -[:MENTIONS]-> Class`.

The issue body still says `Annotation`. That is known and not a discrepancy to "fix" in either direction.

## Why this is not a smaller diff

A `NodeLabel` is a repo-wide contract, not a local enum entry. Four independent guards refuse a label that is not registered everywhere, so a schema-only slice does not exist — I tried it first and all four went red:

| Guard | What it refuses |
|---|---|
| `constants/graph.py` import-time check | a label with no `_NODE_LABEL_UNIQUE_KEYS` entry |
| `constants/graph.py` import-time check | a `RelationshipType` belonging to no capture group |
| `test_graph_audit` documented-labels | a label absent from `NODE_SCHEMAS` |
| `test_protobuf_node_parity` ×2, `test_protobuf_relationship_parity` | a label with no proto payload, or no `LABEL_TO_ONEOF_FIELD` mapping |

`_NOT_IN_PROTO` is deliberately an empty `frozenset()`, so a label without a payload fails rather than being quietly exempt. The mapping matters most of the four: without it `ensure_node_batch` returns early with only a `logger.warning`, and the node vanishes from the index without failing anything.

## Wire format

Append-only, per the comments in `schema.proto`:

- `Node.payload` field **22** (previous max 21)
- `RelationshipType.ANNOTATES = 30`, `MENTIONS = 31` (previous max 29)

Bindings regenerated with the pinned `grpcio-tools==1.76.0` the schema comments require, for the gencode/runtime stamp reason documented there.

## Off by default

`CaptureGroup.GLOSSES` is deliberately absent from `DEFAULT_CAPTURE_GROUPS`. Glosses are written, not parsed; indexing a repository must not begin inventing them. Making it a capture group rather than a parallel mechanism means the label and its edges obey the same enable/disable contract as everything else.

## Targets

`_GLOSS_TARGET_LABELS` is exactly the set `graph_query.resolve()` can return, so every anchor an agent can name is one an `ANNOTATES` edge can point at. Asserted by a test rather than left as a comment, since it is the kind of claim that drifts silently.

## Verification

- Four previously-red guards now green; 167 tests across the parity, audit and coverage files pass.
- `ty`: **23 diagnostics on this branch, 23 on `main`** — same command, same environment, untracked `probe1483/` moved aside. (That directory alone contributes 600+ and will mislead any local `ty` run in this clone.)
- `generate_readme.py` produces no diff; docs were already consistent.
- Local review returned 5/5 with no findings, having independently mutation-tested every registration site and confirmed the regenerated bindings match `schema.proto`.
- Every test here was mutated to prove it discriminates: dropping a label from the target tuple reddens the drift test and nothing else; adding `GLOSSES` to the defaults reddens the off-by-default test and nothing else.

## Known, deliberately deferred

**#1828** — `CYPHER_DELETE_PROJECT` traverses only `CONTAINS_*` then `DEFINES|DEFINES_METHOD`, so a gloss reachable only via `ANNOTATES` survives project deletion. Not fixed here because it is behaviour and this PR is registration.

It is not a one-line traversal fix. A gloss has two requirements that pull apart: it **must** survive re-indexing, since unlike every other node its truth lives only in the graph; and it **must not** survive deletion of the project it describes. Extending the traversal satisfies the second and breaks the first. The repo already has both halves of the pattern — `:IncompleteRun` is deliberately kept unreachable from that traversal so it survives a rebuild, and `prune_unanchored_resources` is a separate sweep already invoked from `delete_project` for exactly this class of node. Both test cases are written into the issue.

## Summary by CodeRabbit

- **New Features**
  - Added support for durable, agent-authored glosses—notes attached to code definitions and symbols.
  - Glosses can include authorship, timestamps, commit context, note content, and target information.
  - Added anchor tracking to show whether a gloss is exact, moved, stale, ambiguous, or lost.
  - Glosses can annotate a specific code definition and reference additional symbols mentioned in the note.
  - Added persistence support for gloss data in graph and serialized representations.


vitali87:
@greptileai review

PR #1838, opened 2026-09-10
fix: walk every parsed file for calls, not only those with call or function captures (#1837)

### LINKED ISSUE #1837: A class-only capture entry makes the call walk skip a file with module-level references
Found by CodeRabbit while reviewing #1833, and verified to PREDATE it: the
same behaviour reproduces on `origin/main` with that PR's changes absent.

## The gap

`_process_function_calls` (`codebase_rag/graph_updater.py`) skips a file whose
cached captures hold no call and no function:

```python
if cached and not cached.get(cs.CAPTURE_CALL) and not cached.get(
    cs.CAPTURE_FUNCTION
):
    continue
```

The premise is "no call sites and no functions, so the call walk has nothing
to find". That is false for a MODULE-LEVEL reference, which lives under
neither capture — a dispatch dict holding an imported handler is a real
`CALLS` edge the walk still has to emit.

The skip only fires when the entry is truthy, so a file with a CLASS capture
and nothing else takes it.

## Reproducer

Two files, differing only by the presence of a class:

```python
# pkg/plain.py
from pkg.util import helper
TABLE2 = {"h": helper}

# pkg/disp.py
from pkg.util import helper
class Marker:
    pass
TABLE = {"h": helper}
```

A full index on `origin/main`:

```
edges: [('proj.pkg.plain', 'proj.pkg.util.helper')]
```

`plain.py` gets its edge; `disp.py` does not. Captures confirm the mechanism:

```
disp.py:  keys=['class']  calls=0 funcs=0 classes=1   <- truthy, so skipped
plain.py: keys=[]         calls=0 funcs=0 classes=0   <- falsy, so walked
```

Adding an unrelated class to a module silently drops its module-level call
edges. This is a full-index bug, not only an incremental one.

## Suggested fix

The skip needs a positive reason to believe there is nothing to emit, rather
than inferring it from two capture kinds. Either drop the class-only case from
the skip (cheapest, costs one redundant walk per class-only file), or record
whether the combined query matched any module-level reference and skip only
when it did not.

## Test

- a file with a class AND a module-level imported-handler reference must emit
  the edge (fails today);
- the control that a genuinely empty file still skips, so the optimisation is
  not simply deleted.

## Why filed rather than fixed in #1833

That PR fixes a different defect in the same function (a stale captures entry
after a re-parse) and already carries the narrower `cached and ...` guard. This
one is older, affects the batch path too, and its fix is a judgement about what
evidence licenses the skip — so it wants its own change rather than being
folded into a rollback-adjacent PR.

vitali87:
Closes #1837.

## The defect

`_process_function_calls` skipped a cached file whose captures held no call and no function:

```python
if not cached.get(cs.CAPTURE_CALL) and not cached.get(cs.CAPTURE_FUNCTION):
    continue
```

The premise — "nothing for the walk to find" — is false for a **module-level reference**. A dispatch table holding an imported handler, a bare `x = handler`, or a JSX element is a real `CALLS` edge that lives under no capture kind at all. `process_calls_in_file` runs those passes *ahead of* its own no-calls early return precisely so such a file is covered, and the outer skip discarded that work before it could run. A captured class is positive evidence too: the Python decorator pass takes class nodes as its targets.

The skip only fired on a **truthy** cache entry, which inverts the intent: a file with a class and nothing else was skipped, while a file with no captures at all (no cache entry) was walked. Adding an unrelated class to a module silently dropped its module-level call edges.

This is a full-index bug, not only an incremental one.

## Reproducer

Two modules differing only by the presence of a class:

```python
# pkg/plain.py
from pkg.util import helper
TABLE2 = {"h": helper}

# pkg/disp.py
from pkg.util import helper
class Marker: pass
TABLE = {"h": helper}
```

On `origin/main`, `plain.py` gets its edge and `disp.py` does not.

## Why the skip is removed rather than narrowed

Narrowing the guard to "records none of call/function/class" was the obvious fix and turns out to be **unreachable**. The populator stores a capture key only when the query matched it, and drops an entry with no keys at all, so every cache entry has at least one non-empty kind.

Measured, rather than reasoned about:

- **0** entries with all three kinds empty, and **0** stored keys holding an empty list, over 105 real files — independently reproduced across **1277 files in 11 languages**.
- The old guard did fire on real code: **5 class-only files in this repo** (`cli_help.py`, `constants/mcp.py`, `exceptions.py`, `parsers/flow_access/constants.py`, `stack/constants.py`).

Shipping a dead branch that reads as a live optimisation is worse than removing it, so the walk is now uniform. Uncached files were always walked anyway.

## Cost

Measured on a full index of `codebase_rag/parsers` (~1250 files), fresh tree per run, 6 alternating iterations:

| | median |
|---|---|
| with the old skip | 5.978s |
| without it (this PR) | 6.123s |

**+0.145s, +2.4%**, for correctness on silently-dropped `CALLS` edges. `process_calls_in_file` retains its own internal no-calls early return, which is what keeps the delta small.

## Tests

`codebase_rag/tests/test_class_only_capture_call_walk.py`:

- a class-only file emits its module-level dispatch-table edge;
- a class-only file emits its decorator call (the pass that reads captured class nodes);
- a control asserting every parsed file reaches the walk.

Verified to discriminate rather than merely pass:

- Both behaviour tests go **red** without the fix.
- Reintroducing the old skip turns all three red, the control naming exactly `['disp.py', 'marker.py']`.
- An earlier version of the control passed even with the skip deleted entirely — it asserted an unreachable state. It was rewritten to assert the property that actually holds, and re-validated against a known-positive.

Spy hygiene: the test patches `type(processor).process_calls_in_file` because `CallProcessor` uses `__slots__`. Checked that the class attribute is pristine before and after each test, that `patch.object` restores on the exception path, and that it holds under `-n 4`.

Regression: 1208 passed across 117 call/dispatch/graph-updater/incremental test files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code relationship analysis by ensuring all parsed files are processed consistently.
  * Module-level references, dispatch-table calls, bare handler assignments, JSX references, and decorator calls are now detected reliably.
  * Class-only files no longer prevent relevant call relationships from being captured.

* **Tests**
  * Added regression coverage for module-level and decorator calls in class-only files.
  * Verified that every parsed file reaches call processing.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -4449,12 +4449,25 @@ class GraphUpdater:
             root_node = self._ast_for(file_path)
             if root_node is None:
                 continue
-            if captures_cache is not None and file_path in captures_cache:
-                cached = captures_cache[file_path]
-                if not cached.get(cs.CAPTURE_CALL) and not cached.get(
-                    cs.CAPTURE_FUNCTION
-                ):
-                    continue
+            # Issue #1837: there is deliberately no capture-based skip here.
+            # The old one ("no calls and no functions, so nothing to walk")
+            # was unsound in both directions. A MODULE-LEVEL reference -- a
+            # dispatch table holding an imported handler, a bare
+            # `x = handler`, a JSX element -- is a real CALLS edge that lives
+            # under no capture kind at all, and process_calls_in_file runs
+            # those passes ahead of its own no-calls early return precisely
+            # so such a file is covered. A captured class is positive
+            # evidence of work too: the Python decorator pass takes class
+            # nodes as its targets.
+            #
+            # Only a file recording none of call/function/class could license
+            # a skip, and no such CACHE ENTRY exists: the populator stores a
+            # key only when the query matched it and drops an all-absent
+            # entry (definition_processor.py), so every entry has a non-empty
+            # kind. A narrowed guard would be unreachable rather than merely
+            # cheap -- measured 0 hits over 105 real files, while the old
+            # guard skipped this repo's own class-only parsers/constants.py.
+            # Uncached files were always walked, so the walk is now uniform.
             self.factory.call_processor.process_calls_in_file(
                 file_path,
                 root_node,
--- a/codebase_rag/tests/test_class_only_capture_call_walk.py
+++ b/codebase_rag/tests/test_class_only_capture_call_walk.py
@@ -1,0 +1,149 @@
+"""A class-only capture entry must not make the call walk skip a file
+(issue #1837).
+
+`_process_function_calls` skipped any cached file whose captures held no call
+and no function. The premise -- "nothing to walk" -- is false twice over. A
+MODULE-LEVEL reference (a dispatch table holding an imported handler, a bare
+`x = handler`, a JSX element) is a real CALLS edge living under no capture
+kind at all, and `process_calls_in_file` runs those passes ahead of its own
+no-calls early return precisely so such a file is covered. A class is
+positive evidence too: the Python decorator pass takes captured class nodes
+as its targets.
+
+A file with NO captures never reaches the cache, so it was walked; a file
+whose only capture was a class had a truthy entry and was skipped. Adding an
+unrelated class to a module silently dropped its module-level call edges.
+This is a full-index bug, not only an incremental one.
+"""
+
+from __future__ import annotations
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

Issue #1797, opened 2026-09-08
One invalid UTF-8 byte silently drops every definition in a file


vitali87:
Found by the `fuzz_parse_source` harness from #1731, on its first real 60-second run, then reduced to a deterministic reproducer.

## Symptom

A single byte that is not valid UTF-8 anywhere in a source file causes **every definition in that file** to be dropped from the graph, including definitions that are pure ASCII and nowhere near the bad byte.

```python
clean file   -> ['proj.a.Greeter.greet', 'proj.a.other']
one bad byte -> []
LOST: ['proj.a.Greeter.greet', 'proj.a.other']
```

The two files differ by two bytes:

```
function Greeter.greet(n) return 1 end      <- indexed fine
function Greeter\xff\xff.greet(n) return 1 end  <- whole file lost
function other() return 2 end               <- valid, lost anyway
```

## Cause

`_generic_get_name` (`codebase_rag/language_spec.py:73-83`) decodes node text strictly:

```python
return name_node.text.decode(cs.ENCODING_UTF8)
```

`node.text` is a slice of the raw file bytes, which are read with `read_bytes()` and never validated as UTF-8. On undecodable bytes this raises `UnicodeDecodeError`.

Verified reachable through the production path, not just in isolation: instrumenting `LANGUAGE_FQN_SPECS[LUA].get_name` during a normal `GraphUpdater.run()` shows it called once and raising once, via

```
graph_updater.py:4259  _process_single_file
definition_processor.py:536  process_file
function_ingest.py:402  _ingest_all_functions
function_ingest.py:571  _resolve_function_identity
function_ingest.py:592  _try_unified_fqn_resolution
```

The run does not crash because the caller at `graph_updater.py:4058` catches `Exception`, logs `INCREMENTAL_FILE_FAILED` and abandons the file. So the failure is silent and total for that file rather than fatal for the run — which is why it has not been noticed.

## Why it matters

The tool indexes arbitrary repositories. A stray Latin-1 byte in a comment, an accidental binary blob committed with a source extension, or a file in a legacy encoding is enough to make every symbol in that file invisible to callers, dependents, dead-code and test-reachability queries. Nothing in the graph indicates the file was skipped; it simply has no definitions.

Not specific to Lua. `_generic_get_name` is the shared extractor, and the same strict decode appears in the other `_*_get_name` helpers in that module.

## Suggested fix

[...]

PR #1841, opened 2026-09-10
docs: state both guarantees safe_decode_text provides, and test them (#1811)

### LINKED ISSUE #1811: safe_decode_text is not safe: the name asserts a guarantee it does not provide
Found while fixing #1797. Low severity on its own; filed because the NAME is what caused the miss, and it will keep causing them.

## The problem

`parsers/utils.py`:

```python
@lru_cache(maxsize=50000)
def _cached_decode_bytes(text_bytes: bytes) -> str:
    return text_bytes.decode(cs.ENCODING_UTF8)      # strict

def safe_decode_text(node) -> str | None:
    if node is None or (text_bytes := node.text) is None:
        return None
    if isinstance(text_bytes, bytes):
        return _cached_decode_bytes(text_bytes)
    return str(text_bytes)
```

The "safe" refers only to the **None handling** — a null node or null text yields `None` instead of raising `AttributeError`. The decode underneath was strict and raised `UnicodeDecodeError` on any undecodable byte.

## Why it is worth an issue

There are 553 call sites. The name reads as "this is the decode that handles bad input", which is exactly the assumption that makes people stop looking.

It cost real time on #1797: I probed the extractors and the call pass, found and fixed both, and concluded the fix was complete — while a bad byte in a **Python decorator** still dropped every definition in the file, through this function, in the repo's primary language. The reviewer caught it. I had not probed `safe_decode_text` because its name told me it was already handled.

`safe_decode_text` now routes through `language_spec.decode_node_text` (fixed as part of #1797), so the immediate defect is gone. The naming problem remains.

## Suggested change

Rename to say what it actually guarantees. `decode_text_or_none` describes the None handling without implying anything about encoding. Alternatively keep the name and make it genuinely total, which after #1797 it now is — in which case the fix is a docstring stating that both the None case and the undecodable-bytes case are handled, so the next reader does not have to derive it.

Either way the point is that a name asserting a guarantee should be checkable against the implementation. A grep for "safe" that lands on a strict decode is a trap.

## Related

* #1797 — the defect this name hid.
* #1810 — the truncation mode, which `errors="replace"` does not address.

vitali87:
Closes #1811.

## The problem

`safe_decode_text` has ~550 call sites and a name that reads as "this is the decode that handles bad input". Until #1797 that was only half true: "safe" covered the None handling, while the decode underneath was strict and raised `UnicodeDecodeError` on any undecodable byte.

The issue records what that cost: on #1797 the extractors and the call pass were probed and fixed, and the fix declared complete — while a bad byte in a **Python decorator** still dropped every definition in the file, through this function, in the repo's primary language. `safe_decode_text` was not probed because its name said the case was already handled.

## The fix, per the issue's second option

The issue offers a rename (`decode_text_or_none`) or keeping the name and documenting both guarantees. Taking the second, on the owner's call.

It is now the accurate one: since #1797 the decode routes through `language_spec.decode_node_text` (`errors="replace"`), so both the None case and the undecodable-bytes case really are handled. `decode_text_or_none` would arguably *understate* it — the encoding guarantee is real, and a name mentioning only the None handling implies nothing about it. Keeping the name also leaves ~550 call sites and every in-flight branch untouched.

The issue's actual point is that **a name asserting a guarantee should be checkable against the implementation**. A docstring alone does not achieve that — it can drift from the code silently, which is how the original trap was set. So this adds tests that hold the code to what the docstring claims.

## Verified before writing it down

Both guarantees were confirmed by execution rather than by reading, since a docstring asserting a guarantee the code does not provide is exactly the defect being fixed:

| input | result |
|---|---|
| `None` | `None` |
| node with `text=None` | `None` |
| `b"@dec\xffrator"` | `'@dec�rator'`, no raise |
| `b"\xff\xfe\x00invalid"` | decoded, no raise |
| valid UTF-8 | unchanged |

With a control confirming a strict decode *does* raise on the same input, so "no raise" is evidence of the fix rather than of a weak fixture.

## Tests

`codebase_rag/tests/test_safe_decode_text_guarantees.py`, 8 tests, including:

- a **calibration** test asserting the bad bytes would break a strict decode, so every "does not raise" below it means something;
- both guarantees, separately;
- that valid UTF-8 is undisturbed by the replacement decode;
- that the decode **routes through the shared helper** rather than growing its own `errors="replace"`. Several passes decode node text independently and each abandons the file on its own; fixing one and leaving the rest losing data on the same input is what #1797 was.

Verified to discriminate: restoring the pre-#1797 strict decode turns exactly three red — both guarantee tests and the mechanism test — while the None-handling tests and the calibration stay green.

Regression: 768 passed across the parser/utils/definition/extraction test files.

## Summary by CodeRabbit

- **Documentation**
  - Clarified text-decoding behavior for null values, invalid byte sequences, valid UTF-8, and non-byte inputs.
  - Documented replacement decoding and related failure-handling guarantees.

- **Tests**
  - Added coverage confirming malformed bytes are safely replaced rather than raising errors.
  - Verified null handling, valid UTF-8 preservation, non-byte conversion, and use of the shared decoding behavior.

### DIFF
--- a/codebase_rag/parsers/utils.py
+++ b/codebase_rag/parsers/utils.py
@@ -266,6 +266,29 @@ def node_site_properties(node: Node) -> PropertyDict:
 
 
 def safe_decode_text(node: ASTNode | TreeSitterNodeProtocol | None) -> str | None:
+    """A node's text as `str`, or `None` when there is no text to decode.
+
+    "safe" covers BOTH failure modes, and both guarantees are real:
+
+    * a null node, or a node whose `text` is null, yields `None` rather than
+      raising `AttributeError`;
+    * undecodable bytes are replaced, never raised. The decode routes through
+      `language_spec.decode_node_text` (`errors="replace"`), so a single bad
+      byte cannot raise `UnicodeDecodeError` here.
+
+    The second half is worth stating rather than leaving to be re-derived.
+    Until #1797 the decode underneath was strict, so "safe" named only the
+    None handling -- and the name read as "this is the decode that handles
+    bad input", which is exactly the assumption that stops people looking.
+    It cost real time on #1797: a bad byte in a Python decorator dropped
+    every definition in the file through this function, in the repo's primary
+    language, because the per-file handler in `graph_updater` catches
+    `UnicodeDecodeError` and abandons the file. With ~550 call sites, a name
+    asserting a guarantee needs to be checkable against the implementation
+    (issue #1811).
+
+    Returns the value as-is via `str()` when `node.text` is not `bytes`.
+    """
     if node is None or (text_bytes := node.text) is None:
         return None
     if isinstance(text_bytes, bytes):
--- a/codebase_rag/tests/test_safe_decode_text_guarantees.py
+++ b/codebase_rag/tests/test_safe_decode_text_guarantees.py
@@ -1,0 +1,109 @@
+"""`safe_decode_text` must provide both guarantees its name asserts (#1811).
+
+The name reads as "this is the decode that handles bad input". Until #1797
+that was half true: "safe" covered only the None handling, and the decode
+underneath was strict. A bad byte in a Python decorator raised
+`UnicodeDecodeError`, the per-file handler in `graph_updater` caught it and
+abandoned the file, and every definition in that file vanished from the graph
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

Issue #1524, opened 2026-08-29
[Epic: Agent IDE] T3: Single-file scoped re-ingest with an interactive latency budget


vitali87:
Part of epic #1521 — **T3**.

## Dependencies
- Depends on: nothing — foundation ticket
- Unblocks: #1525 (T4)

## Problem
`update_repository` runs `GraphUpdater.run()` over the whole tree behind a file-hash cache. That is fine per commit; an agent editing a file needs the graph to reflect the edit in hundreds of milliseconds, and the cross-file call-resolution pass currently has project scope.

## Scope
- `reingest(paths) ` API + MCP tool: re-parse only the given files, delete/recreate their subgraphs, and re-run call resolution scoped to (a) those files and (b) the files that import them (via `IMPORTS` edges), not the project.
- Reuse the watchdog-based realtime updater's event path so both share one implementation.
- Benchmark harness: measure p50/p95 for a one-file edit on a medium repo (e.g. django) and record it in `docs/`.

## Acceptance
- One-file re-ingest on a 100k-line repo completes in < 1s p95 on a laptop, including scoped call resolution.
- Graph after `reingest` equals graph after a full `update_repository` for the same edit (property-based test over a set of edits).

Issue #1028, opened 2026-08-02
fix: files created during watch mode get no CALLS edges


vitali87:
## Summary

A file created while the realtime watcher is running is parsed and its definitions register (Function/Method nodes appear, and its import maps fill), but the call-recomputation pass never visits it, so none of its own outgoing CALLS edges are emitted until the next full index.

## Repro

Index a Rust project, start the watcher, create `src/new.rs` containing a function that calls an existing first-party function, and dispatch the create event. The new function's node exists; the CALLS edge from it does not. A subsequent full re-index emits it.

## Expected

A created file's calls resolve in the same event cycle, exactly as a modified file's do.

Documented in `test_watch_create_of_owned_module_keeps_its_own_imports` (test_rust_crate_path_trait_linking.py), which pins the import map instead of edges because of this gap. Known limitation declared across the #1007 adversarial review rounds; splitting it out for tracking.

PR #1842, opened 2026-09-10
refactor: remove the callerless register_parsed_file and pin the invariant it implied (#1784)

### LINKED ISSUE #1784: register_parsed_file has no callers, and it is the only appender that deduplicates _parsed_files
## Observed

`GraphUpdater.register_parsed_file` (`graph_updater.py:4324`) has **no callers**. A repo-wide grep for the name across every `*.py` returns only its own definition; there is no dynamic dispatch (`getattr`, string literal), no protocol declaring it, and no reference in docs or stubs.

```python
def register_parsed_file(
    self, file_path: Path, language: cs.SupportedLanguage
) -> None:
    # Watch-mode events parse outside run(): the file must join the
    # call pass's iteration set or its outgoing CALLS edges are never
    # emitted (issue #1028).
    if all(existing != file_path for existing, _ in self._parsed_files):
        self._parsed_files.append((file_path, language))
```

## Why it is worth more than a tidy-up

Two things make this more than dead weight.

**1. It is the only deduplicating appender, and it is the one that never runs.** The live appender is `_process_single_file` at `graph_updater.py:4265`, which appends unconditionally. So `_parsed_files` can hold duplicate entries for a file parsed repeatedly — on a retained updater (the MCP `_live_updater`) that is every scoped re-ingest of the same file. Anyone reading `register_parsed_file` would reasonably conclude the list is deduplicated. It is not.

**2. Its stated purpose describes a real defect class.** The comment cites #1028: a file parsed outside `run()` must join the call pass's iteration set or its outgoing CALLS edges are never emitted. That is the same shape as #1712 (`reingest` does not go through `run()`, so a `run()`-only step never happens on that path). If the watch-mode caller was refactored away rather than the need disappearing, the guarantee this function was written to provide may now be unmet on some path — and nothing would show it, because the function still exists and looks like it is doing the job.

## Expected

Decide which it is, and make the code say so:

- if the need is gone, delete it — a deduplicating appender that never runs is a false signal about the list's invariants;
- if the need remains, find the path that lost its call and restore it, with a test that fails when the call is removed.

Worth checking `git log -S register_parsed_file` for the caller that disappeared before choosing.

Found while implementing #1712, which needed the exact semantics of `_parsed_files`. Credit to the peer session that spotted the appender split.

vitali87:
Closes #1784.

Takes the **delete** branch, after checking which one applies. The investigation changed the conclusion, so it is worth reporting rather than just the fix.

## Confirming the function is dead

Unscoped, whole repo, every file type — not a subtree grep:

```
$ git grep -n "register_parsed_file" origin/main
origin/main:codebase_rag/graph_updater.py:4401:    def register_parsed_file(
```

One hit, its own definition. No dynamic dispatch (`getattr`, string literal, protocol), nothing in docs or stubs.

## Where the caller went

`git log -S` finds it exactly:

- **72648235** added it for #1028, called from the watcher in `realtime_updater.py`.
- **3974efa5** (#1524) replaced that call with `reingest`:

```diff
-                    self.updater.register_parsed_file(path, language)
+            self.updater.reingest((path,))
```

So the need did not disappear — the watcher moved onto a path that reaches `_process_single_file`, the **unconditional** appender.

## The issue's two premises, checked separately

The issue asks whether the need is gone (delete) or a path lost its call (restore). Both premises were tested by execution:

**Premise 2 — the #1028 guarantee — still holds.** A file created after `run()` joins `_parsed_files` via `reingest` and its outgoing CALLS edge is emitted. Nothing is unmet, so there is no lost call to restore.

**Premise 1 — "`_parsed_files` can hold duplicates" — does not reproduce.** Three consecutive reingests of the same file on a retained updater leave exactly one entry. The reason is `remove_file_from_state`, which strips the entry before the re-parse ("a re-parse re-registers"); `reingest` gets there via `_reingest_delete`, and the incremental path calls it directly. Both production callers of `_process_single_file` are preceded by a removal.

The appender **can** still duplicate — calling it directly without the removal produces a second entry — so the removal is load-bearing, not incidental. It is just not a path anything takes.

That makes this the delete branch: the invariant is real, but it comes from remove-then-append, and a deduplicating appender that never runs is a false signal about where it lives.

## The change

- Delete `register_parsed_file`.
- Record the invariant at the removal in `remove_file_from_state`, which is where it actually comes from — including why a deduplicating appender is deliberately not the mechanism, so this does not get "helpfully" reintroduced.
- Add `codebase_rag/tests/test_parsed_files_invariants.py` (7 tests) pinning the invariant to that mechanism.

## Tests

Beyond the outcome, one test asserts the appender **does not** deduplicate — otherwise the suite could not show where the invariant comes from, and would pass just as well if the mechanism moved. Also covered: remove-then-reparse restores exactly one entry, and the #1028 guarantee survives the deletion.

Verified to discriminate: deleting the removal from `remove_file_from_state` turns two red (`test_repeated_reingest_does_not_duplicate_entries`, `test_removal_then_reparse_restores_exactly_one_entry`); the others stay green, as they should, since they do not depend on it.

Regression: 378 passed across the incremental/watch/reingest/graph-updater test files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed repeated file reingestion so files are refreshed without creating duplicate entries.
  - Corrected remove-and-reprocess behavior to keep parsed file results consistent.
  - Ensured files added after an initial run are included in subsequent processing and relationship tracking.
  - Improved consistency so each parsed file is represented exactly once after a complete run.

- **Tests**
  - Added coverage for file parsing, reingestion, removal, and incremental processing scenarios.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -2926,6 +2926,16 @@ class GraphUpdater:
         # The call pass iterates _parsed_files; a removed file must leave it
         # (a re-parse re-registers), or deleted files keep contributing and
         # created files never do (issue #1028).
+        #
+        # This removal is ALSO what keeps the list free of duplicates. The
+        # appender in `_process_single_file` appends unconditionally, so
+        # every re-parse route has to strip the old entry first -- the
+        # incremental path calls this method, and `reingest` gets there via
+        # `_reingest_delete`. A deduplicating appender is deliberately not
+        # the mechanism: one existed (`register_parsed_file`, added for
+        # #1028) and was left callerless when #1524 moved the watcher onto
+        # `reingest`, where it read as the guarantee's source while never
+        # running. Removed in #1784; the invariant lives here.
         self._parsed_files = [
             entry for entry in self._parsed_files if entry[0] != file_path
         ]
@@ -4397,15 +4407,6 @@ class GraphUpdater:
         root_node = parse_with_preproc_recovery(parser, file_bytes, language).root_node
         self.factory._func_class_captures_cache.pop(file_path, None)
         return (root_node, language)
-
-    def register_parsed_file(
-        self, file_path: Path, language: cs.SupportedLanguage
-    ) -> None:
-        # Watch-mode events parse outside run(): the file must join the
-        # call pass's iteration set or its outgoing CALLS edges are never
-        # emitted (issue #1028).
-        if all(existing != file_path for existing, _ in self._parsed_files):
-            self._parsed_files.append((file_path, language))
 
     def _process_function_calls(self, only: Collection[Path] | None = None) -> None:
         # `only` scopes the pass to a re-ingested subset (issue #1524); every
--- a/codebase_rag/tests/test_parsed_files_invariants.py
+++ b/codebase_rag/tests/test_parsed_files_invariants.py
@@ -25,7 +25,7 @@ from __future__ import annotations
 
 from collections import Counter
 from pathlib import Path
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


vitali87:
@greptileai review

vitali87:
Addressed the P2 — good catch, the `== 2` did mandate an implementation detail.

`test_the_appender_itself_does_not_deduplicate` no longer asserts that the bare appender produces a duplicate. It now asserts the property that actually matters: that `remove_file_from_state` **runs** on the re-parse path, observed with a spy, rather than inferring it from a count the appender could also produce.

Checked both directions rather than assuming the reformulation is equivalent:

- **Deleting the removal** from `remove_file_from_state` still reddens three tests, including this one — detection is unchanged.
- **Adding the defensive dedup** you wanted to keep possible (`if all(e != filepath ...)` in the appender) now passes all 7. The old assertion would have failed exactly that change, which was your point.

On the T-Rex environment failure: the repo venv is Python 3.12 and `tree_sitter` is not importable from 3.11, so the focused test could not run there. It runs clean locally — 7 passed, and 378 passed across the incremental/watch/reingest/graph-updater files.

@greptileai review

PR #1844, opened 2026-09-10
fix: keep a surviving stem-sharing file's definitions when its sibling is deleted (#1773)

### LINKED ISSUE #1773: A sibling directory sharing a deleted file's stem is swept from the registry with it
## Observed

A directory whose name equals another file's stem produces module qns that
NEST: `proj/A.cs` records `proj.A`, and `proj/A/B.cs` records `proj.A.B`.
`remove_file_from_state` sweeps `function_registry` by module-qn prefix, so
deleting `A.cs` matches `proj.A.` and takes `B.cs`'s definitions with it,
even though `B.cs` still exists and this event does not re-parse it.

Measured during the local review of the #1769 fix, with `proj/A.cs`
declaring `public class K<T>` and `proj/A/B.cs` declaring
`public class Nested<T>`:

```
deleting A.cs removes proj.A.B.N.Nested from function_registry
while leaving proj.A.B.N.Nested.Deep(U)
```

The method qn survives because `foreign_qns`, built from span records,
protects it; the class qn has no span record and so is unprotected.

## Why it matters

The surviving file is not re-parsed by this event, so nothing restores the
entries. Until something else re-parses `B.cs`, calls into `Nested` do not
resolve, and a later re-parse can land on a `@line` DUP_QN variant beside the
row that was never removed -- the #1719 shape.

## Scope

This is NOT C#-specific: the prefix sweep is language-agnostic, and any
language where a directory can share a sibling file's stem is affected
(`a.py` beside `a/b.py`, `a.rs` beside `a/b.rs`, `A.cs` beside `A/B.cs`).

## Expected

The sweep keeps any qn that also sits under a SURVIVING module qn, i.e. one
in `module_qn_to_file_path` that is not among this event's own prefixes and
is longer than the matched prefix. #1769 applies exactly that filter to
`csharp_class_generic_arity`, where the same shape appeared; the registry
sweep and anything else keyed by prefix want the same treatment.

Found during the local review of #1769.

vitali87:
Closes #1773.

## The defect

A directory can share a sibling file's stem, and then module qns nest: `proj/a.py` records `proj.a` while `proj/a/b.py` records `proj.a.b`. `remove_file_from_state` sweeps `function_registry` by module-qn prefix, so deleting `a.py` matched `proj.a.` and took `b.py`'s definitions with it — though `b.py` still exists and the deletion event does not re-parse it, so nothing restores them.

Calls into those definitions then stop resolving, and a later re-parse can land a `@line` DUP_QN variant beside the row that was never removed (the #1719 shape).

## Reproduced in Python, not just C#

The issue measured it on `A.cs` beside `A/B.cs` and noted the sweep is language-agnostic. Confirmed — `a.py` beside `a/b.py` reproduces it identically on `origin/main`:

```
registry before: 4 entries
  from a/b.py: ['proj.a.b.Nested', 'proj.a.b.Nested.deep', 'proj.a.b.sibling']

registry after deleting a.py: 2 entries
  from a/b.py: ['proj.a.b.Nested.deep', 'proj.a.b.sibling']

LOST from the SURVIVING file a/b.py: ['proj.a.b.Nested']
```

Including the asymmetry the issue predicted: the **class** is lost while the **method** survives, because `foreign_qns` is built from function span records and a class registers none.

## The fix

A qn under a **surviving** module qn — one still in `module_qn_to_file_path`, not among this event's own prefixes, and longer than the prefix that matched — is kept, whichever shorter prefix also matches. `owned_qns` still wins, since that is the deleted file's own span-record evidence.

### Why not the `csharp_class_generic_arity` approach

#1769 hit the same shape and did **not** end up using this rule: it added an explicit `csharp_class_owner_module` record, because a C# class qn embeds its namespace, so a class can sit under a sibling module it was never declared in and a prefix rule errs in both directions.

That reasoning does not carry here. A Python/Rust/Go module qn is derived from the path, so "under a longer surviving module" does identify the real owner. Noted in the code comment so the divergence is deliberate rather than looking like an oversight.

## Tests

`codebase_rag/tests/test_sibling_stem_registry_sweep.py`, 6 tests. Three are controls, and they carry the weight — keeping the survivor's entries must not be achieved by weakening the sweep:

- the deleted file's own definitions are **still removed**;
- a plain deletion with no stem collision still removes everything under the module;
- the reverse direction (deleting `a/b.py` leaves `a.py` alone) — never broken, which is why it belongs here: a fix that kept anything a longer module matched would start failing it.

Verified to discriminate: reverting the fix reddens exactly the three defect tests while all three controls stay green.

One test was wrong on the first run and is worth noting, since it is the failure mode this repo keeps hitting: the deeper-collision case filtered survivors with `"c" in qn`, and the temp directory name contained a "c", so it picked up a definition from the file being **deleted** and reported it as a lost survivor. Now anchored on the nested module qn instead of a substring.

Regression: 712 passed across 62 incremental/deletion/registry/watch/reingest/C# test files.

Note: `ty` reports one `possibly-missing-attribute` on `graph_updater.py:5206`. It is pre-existing — the identical single diagnostic comes back with these changes stashed — and unrelated to this diff.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed file deletion handling so definitions in similarly named nested modules are preserved.
  - Deleting a file now removes only its own registry entries without affecting sibling or nested modules that share its name stem.
  - Preserved definitions from surviving files, including restored definitions, while still removing definitions tied to the deleted file.
  - Existing cleanup behavior remains unchanged for modules without naming conflicts.

- **Tests**
  - Added coverage for nested module name conflicts, deeper paths, reverse deletion scenarios, C# namespace patterns, and restored definitions.

### DIFF
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -3043,13 +3043,60 @@ class GraphUpdater:
 
         qns_to_remove = set()
 
+        # A directory may share a sibling FILE's stem, and then the module qns
+        # nest: `proj/a.py` records `proj.a` while `proj/a/b.py` records
+        # `proj.a.b`. Deleting `a.py` matches the prefix `proj.a.` and would
+        # take `b.py`'s definitions with it, though `b.py` still exists and
+        # this event does not re-parse it, so nothing restores them (issue
+        # #1773). `foreign_qns` does not save them: it is built from function
+        # SPAN records, so it protects `proj.a.b.Nested.deep` while leaving
+        # the class `proj.a.b.Nested`, which records no span -- exactly the
+        # asymmetry measured on `A.cs` beside `A/B.cs`, and reproduced here on
+        # `a.py` beside `a/b.py`. The sweep is language-agnostic, so this is
+        # not a C#-only shape.
+        #
+        # A qn under a SURVIVING module qn is therefore kept, whichever
+        # shorter prefix also matches. #1769 hit the same shape in
+        # `csharp_class_generic_arity` but could not use this rule: a C# class
+        # qn embeds its namespace, so it can sit under a sibling module it was
+        # never declared in, and a prefix rule errs in BOTH directions there.
+        # A Python/Rust/Go module qn is derived from the path, so "under a
+        # longer surviving module" does identify the real owner here.
+        surviving_modules = {
+            module_qn
+            for module_qn in self.factory.definition_processor.module_qn_to_file_path
+            if module_qn not in module_qn_prefixes
+        }
+
+        def _under_a_surviving_module(qn: str, matched: str) -> bool:
+            # Only a module LONGER than the prefix that matched can be the
+            # real owner; a shorter or equal one is the deleted file itself.
+            return any(
+                len(module_qn) > len(matched)
+                and (qn.startswith(f"{module_qn}.") or qn == module_qn)
+                for module_qn in surviving_modules
+            )
+
         for qn in list(self.function_registry.keys()):
-            if (
-                any(
-                    qn.startswith(f"{prefix}.") or qn == prefix
+            matched_prefix = next(
+                (
+                    prefix
                     for prefix in module_qn_prefixes
-                )
-                or qn in owned_qns
+                    if qn.startswith(f"{prefix}.") or qn == prefix
+                ),
+                None,
+            )
+            if (
+                matched_prefix is not None
+                and _under_a_surviving_module(qn, matched_prefix)
+                and qn not in owned_qns
+            ):
+                # Belongs to a file that still exists; the prefix match is an
+                # accident of the shared stem. `owned_qns` still wins, since
+                # that is this file's own span-record evidence.
+                continue
+            if (
+                matched_prefix is not None or qn in owned_qns
             ) and qn not in foreign_qns:
                 qns_to_remove.add(qn)
                 del self.function_registry[qn]
--- a/codebase_rag/graph_updater.py
+++ b/codebase_rag/graph_updater.py
@@ -3055,49 +3055,45 @@ class GraphUpdater:
         # `a.py` beside `a/b.py`. The sweep is language-agnostic, so this is
         # not a C#-only shape.
         #
-        # A qn under a SURVIVING module qn is therefore kept, whichever
-        # shorter prefix also matches. #1769 hit the same shape in
-        # `csharp_class_generic_arity` but could not use this rule: a C# class
-        # qn embeds its namespace, so it can sit under a sibling module it was
-        # never declared in, and a prefix rule errs in BOTH directions there.
-        # A Python/Rust/Go module qn is derived from the path, so "under a
-        # longer surviving module" does identify the real owner here.
-        surviving_modules = {
-            module_qn
-            for module_qn in self.factory.definition_processor.module_qn_to_file_path
-            if module_qn not in module_qn_prefixes
-        }
+        # Ownership must come from a RECORD, not from the qn's shape. The
+        # obvious rule -- keep a qn that also sits under a longer surviving
+        # module -- is wrong for C#, whose class qn embeds its namespace:
+        # `proj/Core.cs` with `namespace Util` yields `proj.Core.Util.Helper`,
+        # sitting under the SIBLING module `proj.Core.Util` from
+        # `proj/Core/Util.cs`. That rule would keep `Helper` after deleting
+        # the file that declares it (measured; caught in review of #1844).
+        # #1769 hit the same trap and answered it the same way, with an
+        # explicit owner record.
+        #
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
P1 confirmed and fixed — thank you, this was a real regression and I had flagged the risk in the PR body while wrongly assuming `owned_qns` covered it.

Reproduced exactly as described. `proj/Core.cs` declaring `namespace Util` yields `proj.Core.Util.Helper`, which sits under the sibling module `proj.Core.Util` from `proj/Core/Util.cs`:

```
registry after deleting Core.cs (my first version):
    proj.Core.Util.Helper            <- STALE, from the deleted file
    proj.Core.Util.Other.Sibling
```

And it was a genuine regression, not pre-existing: on `origin/main` the same probe removes `Helper` correctly.

**Fix:** ownership now comes from a record rather than the qn's shape. `class_owner_module` is the cross-language owner map added by #1772 — precisely `csharp_class_owner_module`'s counterpart, written by every language's ingest. A qn is exempt only when its **recorded** declaring module is still mapped to a file and is not itself being deleted. A qn with no recorded owner falls through to the prefix rule, i.e. pre-existing behaviour.

This turned out to fix more than it repaired. `main` also drops `proj.Core.Util.Other.Sibling` — the surviving sibling's class — which is #1773 itself in C#. Both directions are now correct:

```
registry after deleting Core.cs (now):
    proj.Core.Util.Other.Sibling     <- surviving file's class kept
    proj.Core.Util.Other.Sibling.Go
                                     <- Helper correctly gone
```

Two C# tests added, one per direction, since a fix satisfying only one is the bug in the other. Verified they discriminate: restoring the prefix-inference rule reddens `test_a_deleted_csharp_class_goes_even_when_its_namespace_mimics_a_sibling` and nothing else.

Regression: 714 passed across 62 files. Head is now `631c39cd`.

@greptileai review

vitali87:
Second finding confirmed and fixed — and you were right despite your own harness being blocked, which is worth saying since the T-Rex log reports "no cleanup conclusion could be drawn".

I verified it directly. With `class_owner_module` empty (the rehydrated case) and `rehydrated_definition_paths` populated as a real rehydrate writes it, the surviving sibling's class was swept:

```
owner map (empty = rehydrated, no ingest): {}
registry after deleting a.py: []
=> proj.a.b.Nested removed though a/b.py survives
```

**It is pre-existing, not a regression from this PR** — clean `origin/main` loses the same class in the same shape. My change fixed the ingested case and left the rehydrated one exactly as it was. Fixing it here anyway, since shipping half of it would leave the defect alive on precisely the incremental runs the issue is about.

**Fix:** consult both ownership records. `class_owner_module` is written at ingest; `rehydrated_definition_paths` is written when a row is read back from the graph, and `_prune_class_keyed_maps` already treats it as ownership evidence. A qn owned by a file that is not the one being deleted is kept.

One note on my own probe, since it nearly produced a false negative: the first version registered the class in `function_registry` without writing `rehydrated_definition_paths`, so it was simulating "registered from nowhere" rather than "rehydrated" — and reported the fix as not working. A probe that does not reproduce the mechanism cannot test it.

Two tests, one per direction: a rehydrated class from the SURVIVING file is kept, and a rehydrated definition from the DELETED file still goes. The second matters — keeping rehydrated definitions must not become "keep everything rehydrated". Verified: removing the rehydrated half reddens the first and nothing else.

Head is now `9450075d`. Regression 716 passed across 62 files.

@greptileai review

vitali87:
Both CodeRabbit findings confirmed by execution and fixed.

**Major — require a live module for rehydrated definitions.** Correct. My guard asked only "is the recorded path different from the file being deleted", and a definition recorded against a file removed EARLIER also has a different path. Reproduced with a ghost definition whose module is absent from `module_qn_to_file_path`:

```
ghost module live?  False
ghost in registry after deleting a.py: True   <- stale entry preserved
```

Applied the suggested shape: the recorded path must also be in the live module map. Ghost now removed, live sibling still kept.

**Minor — use a class to exercise the deleted rehydrated branch.** Also correct, and it was a real assertion-satisfied-by-both. `proj.a.top` is a function, its span puts it in `owned_qns`, and the `qn not in owned_qns` guard short-circuits before the rehydrated branch runs.

Demonstrated rather than assumed, with one mutation (`return True` — preserve every rehydrated definition, including the deleted file's) held constant:

- control using a **function**: 1 passed &nbsp;&nbsp;← never reached the branch
- control using a **class**: 1 failed &nbsp;&nbsp;&nbsp;&nbsp;← catches it

So the old fixture would have gone green on exactly the bug it was written to catch. Now uses a `Doomed` class with a fixture guard that it registered.

One self-correction: my first pass at that test left an `assert ... , None or True` line, which asserts nothing. Removed rather than shipped.

Head is now `fcdd700b`. Regression 716 passed across 62 files.

@greptileai review

Issue #1820, opened 2026-09-09
A successful marker clear settles a failed wipe's unattributed flag


vitali87:
A `wipe_database` that fails part way leaves `_graph_incomplete=True` with `_incomplete_project=None`, correctly: a wipe spans every project and writes no marker, so the damage cannot be attributed to one project.

The next SUCCESSFUL `_require_marker_cleared` for any project then resets all three fields, because an unattributed flag is treated as owned by nobody and therefore settleable by any clear:

```python
if self._incomplete_project in (None, project_name):
    self._graph_incomplete = False
    self._incomplete_project = None
    self._flag_from_failed_clear = None
```

That branch is right for the single-project case it was written for — a run raises an unattributed flag and the same run clears it — but it cannot tell that case apart from a wipe's flag, which no single project's success should settle. Reads then proceed against the half-wiped graph.

## Not a regression

Present identically before #1547 (the owner is `None` either way, so the pre-scoping code reached the same reset). Found by the local reviewer while checking #1547, and deliberately left out of that PR: it is a distinct defect with a distinct fix, and #1547 had already grown well past its original subject.

#1547 does make it reachable from more states, since it widens the attribution back to `None` when a second project is damaged. Worth fixing on that basis rather than despite it.

## Suggested direction

The flag needs to distinguish "unattributed because nothing has claimed it yet" from "unattributed because the damage genuinely spans every project". A wipe's flag should be settleable only by a completed wipe or a full re-index, not by one project's marker clear. A third state on the attribution field, or a separate flag for wipe-shaped damage, would both express that; the field currently overloads one `None` for both meanings.

## Reproducing

```
wipe_database(confirm=True)  with clean_database raising
  -> (True, None, None)
_require_marker_cleared('any-project')  succeeding
  -> (False, None, None)
_incomplete_refusal('any-project', DEFINITION) -> None
```

Expected: still refusing, because the wipe's damage was never repaired.

vitali87:
Sharper framing of the same defect, from the #1547 review on 29d4b842:

> After damage in two projects, `_invalidate_graph_for` widens `_incomplete_project` to `None`, meaning graph reads and scoped reingests must refuse for every project. `_require_marker_cleared` then treats that same `None` owner as clearable by any one project and resets `_graph_incomplete`.

[...]

PR #1846, opened 2026-09-10
fix: a successful marker clear settles only the project that owns the flag (#1774)

### LINKED ISSUE #1774: A successful marker clear for one project discards another project's incomplete flag
## Observed

`MCPToolsRegistry._graph_incomplete` is a single registry-wide boolean, but `_require_marker_cleared` sets it from **one project's** clear result:

```python
cleared = self._persist_incomplete(project_name, False)
self._graph_incomplete = not cleared
```

So a *successful* marker clear for project B unconditionally discards a flag project A earned from a failure that wrote no marker. The next scoped reingest of A is then accepted over a graph that may be partial, which is exactly what the flag exists to prevent.

Measured on unmodified `main` (probe against the test suite's fake store):

```
after failed wipe:       flag= True   marker_store= {}
after B delete succeeds: flag= False
reingest of A: ACCEPTED over a partial graph
```

The failed wipe writes no marker for A (it dies before the graph is gone, and the marker for A was never created), so the in-process flag is the *only* record that A is partial. Deleting an unrelated project B erases it.

## Why the marker does not cover this

The durable marker is per project, so `_persisted_incomplete(A)` correctly returns False here: no marker for A was ever written. The flag and the marker carry different information, and only the flag knows about a failure that never got as far as writing one. Joining them (`flag or persisted`) is right, but the flag itself is registry-wide while every writer of it is project-scoped.

## Expected

The flag should be per project, or the clear should be conditional on `project_name` matching whatever raised it — the same shape as the `_flag_from_failed_clear` attribution added in #1705, which already tracks *which* project's failed clear set the flag but is not consulted here.

## Scope

Pre-existing and not introduced by #1705: it reproduces identically at that PR's parent commit, where `_require_marker_cleared` has the same unguarded assignment. Found while reviewing #1705; filing separately rather than widening that PR.

vitali87:
Closes #1774.

## The defect

`_graph_incomplete` is a single registry-wide boolean, but `_require_marker_cleared` set it from **one project's** clear result:

```python
cleared = self._persist_incomplete(project_name, False)
self._graph_incomplete = not cleared
```

So a *successful* clear for project B discarded a flag project A earned. Reproduced on unmodified `main`:

```
after a marker-less failure:  flag=True
after B's successful clear:   flag=False
=> DEFECT REPRODUCES: B's clear discarded A's flag
```

The durable marker cannot cover this. It is per project, so `_persisted_incomplete(A)` correctly returns False: a wipe that dies before the graph is gone writes no marker for A at all, leaving the in-process flag the **only** record that A is partial.

## The fix, and why not the simpler one

The issue offers "per project, or conditional on `project_name` matching whatever raised it". The obvious reading — treat an unattributed (`None`) flag as unsettleable — **breaks three working paths**, and I confirmed that by trying it first:

```
FAILED test_delete_project_drops_the_retained_updater
```

`delete_project`, `index_repository` and `update_repository` each raise the flag before their first write with `_flag_from_failed_clear = None` ("Not a stranded marker: this flag must not be healed by one") and then settle it via their own `_require_marker_cleared` on success. So `None` does not mean "marker-less failure"; it also means "an operation is in flight and will clear its own flag".

The distinguishing fact is *which project the damage belongs to*, and no existing field carries it. So this adds `_incomplete_owner`, set at every one of the six sites that raise the flag:

- the four project-scoped paths (delete, index, update, reingest-mutated, delta) record their own project;
- the **wipe** records `None`, because it spans every project, writes no marker, and no single project's clear may settle it. Its own success path clears the flag inline.

A clear then settles the flag only when this project owns it. Same shape as #1769's `csharp_class_owner_module`, and as `_abandon_before_writing`, which already restores `flag_before` rather than clearing unconditionally.

## Tests

- the defect: an unrelated project's successful clear cannot settle a failed wipe's flag, and the reingest refusal it protects still fires;
- **two controls**, which are what constrain the fix: a completed `update_repository` still settles the flag it raised, and a completed `delete_project` does too. A fix that simply refused every clear would satisfy the defect test while wedging all three paths — which is exactly the failure my first attempt produced.

The controls cover two different owner-setting sites deliberately: pinning only the update path would leave delete and index free to regress.

Verified to discriminate: restoring the unconditional settle reddens the defect test alone; both controls stay green.

Regression: the full `test_mcp_update_and_search.py` passes (93), including the existing `TestIncompleteMarkerInvariant` suite that pins invariants (a), (b) and (c); 600 passed across 22 MCP/reingest/marker/delta files.

Note: `test_mcp_startup_key_validation.py` has 2 failures. They are pre-existing — the identical two come back with these changes stashed — and unrelated to this diff.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery tracking for incomplete project operations.
  - Prevented successful operations on one project from incorrectly clearing an unresolved failure from another project.
  - Preserved safeguards that block scoped reingestion while a prior operation remains incomplete.
  - Ensured project-specific recovery actions can clear their own incomplete status without affecting other projects.
  - Improved consistency across repeated project updates, deletions, marker-clearing attempts, and failed recovery scenarios.

### DIFF
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -1251,15 +1251,25 @@ class MCPToolsRegistry:
         Also syncs the in-process flag to the durable state, so this process
         and the next agree. Returns None on success, or the failure message.
         """
+        flag_before = self._graph_incomplete
         cleared = self._persist_incomplete(project_name, False)
         if not cleared:
-            if not self._graph_incomplete:
-                self._incomplete_owner = project_name
             self._graph_incomplete = True
             # Attribute the flag to this project's stranded marker, so the
             # recovery in `_hydrate_reingest_updater` heals only the flag it
-            # explains.
-            self._flag_from_failed_clear = project_name
+            # explains -- but ONLY when this failure is what raised it.
+            #
+            # An already-set flag belongs to whatever failure set it, and
+            # claiming it here hands this project recovery authority over
+            # someone else's damage: A fails leaving the flag up, B's clear
+            # then fails and claims the attribution, and when B's marker later
+            # recovers, `recoverable_here` is satisfied and B clears the latch
+            # A owns (caught in review of #1846). Same rule as the ownership
+            # claim below and as `_abandon_before_writing`, which already
+            # guards its attribution with `if not flag_before`.
+            if not flag_before:
+                self._incomplete_owner = project_name
+                self._flag_from_failed_clear = project_name
             return cs.MCP_INCOMPLETE_MARKER_STUCK.format(project=project_name)
 
         # A successful clear settles only the damage THIS project owns.
--- a/codebase_rag/mcp/tools.py
+++ b/codebase_rag/mcp/tools.py
@@ -806,12 +806,19 @@ class MCPToolsRegistry:
         # those definitions over what is left. After a completed delete the
         # project is gone and the not-indexed guard takes over.
         self._live_updater = None
+        # Claim ownership ONLY when raising the flag from clean. An
+        # already-set flag records damage an earlier failure found, and
+        # claiming it here would let this operation's own successful clear
+        # settle someone else's -- which is the #1774 defect arriving from
+        # the other side: a failed WIPE leaves owner=None, and an unrelated
+        # delete/index/update would overwrite that with its own project and
+        # then clear it (caught in review of #1846). Read before the raise,
+        # since the raise is what makes it look owned.
+        if not self._graph_incomplete:
+            self._incomplete_owner = project_name
         self._graph_incomplete = True
         # Not a stranded marker: this flag must not be healed by one.
         self._flag_from_failed_clear = None
-        # This operation's own project owns the flag, so its own successful
-        # clear may settle it; an unrelated project's may not (issue #1774).
-        self._incomplete_owner = project_name
         # Invariant (a). Nothing has been touched yet, so refusing costs
         # nothing; proceeding would leave a half-removed graph a fresh
         # registry cannot tell from a completed delete.
@@ -893,12 +900,19 @@ class MCPToolsRegistry:
         # resolve against the retained updater's definitions or against
         # whatever partial graph a failure left behind.
         self._live_updater = None
+        # Claim ownership ONLY when raising the flag from clean. An
+        # already-set flag records damage an earlier failure found, and
+        # claiming it here would let this operation's own successful clear
+        # settle someone else's -- which is the #1774 defect arriving from
+        # the other side: a failed WIPE leaves owner=None, and an unrelated
+        # delete/index/update would overwrite that with its own project and
+        # then clear it (caught in review of #1846). Read before the raise,
+        # since the raise is what makes it look owned.
+        if not self._graph_incomplete:
+            self._incomplete_owner = project_name
         self._graph_incomplete = True
         # Not a stranded marker: this flag must not be healed by one.
         self._flag_from_failed_clear = None
-        # This operation's own project owns the flag, so its own successful
-        # clear may settle it; an unrelated project's may not (issue #1774).
-        self._incomplete_owner = project_name
         # Persisted BEFORE the delete, and on a node the delete cannot reach:
         # this path removes the Project and rebuilds it, so a marker stored on
         # the Project would be destroyed by the very operation whose failure it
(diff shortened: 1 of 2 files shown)


vitali87:
@greptileai review

vitali87:
P1 confirmed and fixed. This one is a good catch and the reason is instructive: my test could not have found it.

Reproduced exactly as reported, through the public path:

```
after failed wipe:  flag=True  owner=None
delete_project('B'): success=True
after delete of B:  flag=False owner=None
reingest of A: ACCEPTED over a partial graph
```

**Why my test missed it.** It drove `_require_marker_cleared('project-B')` directly, so it never saw `_delete_project_sync` **raise** the flag for its own project first — overwriting the wipe's `owner=None` with `B` — and then legitimately settle what it now appeared to own. Testing the helper instead of the path that reaches it is the same mistake I flagged in #1845's PR body about asserting on the wrong subject, arriving from the other direction.

**Fix:** an operation claims ownership only when it raises the flag **from clean**. An already-set flag records damage an earlier failure found and is not this operation's to claim. Applied at all six raise sites, and it is the same rule `_abandon_before_writing` already applies to `_flag_from_failed_clear` (`if not flag_before`) — I had the precedent in front of me and did not carry it across.

The new test goes through `wipe_database` → `delete_project` → `reingest`, all public. Verified it discriminates: restoring the unconditional claim reddens it and nothing else. The two existing controls still pass, so delete/index/update still settle the flags they legitimately raise.

Head is now `a7b05eb7`. Full MCP suite 94 passed; 601 passed across 22 files.

@greptileai review

vitali87:
Third P1 confirmed and fixed. Same root cause as the second, in the other field — I guarded `_incomplete_owner` and left `_flag_from_failed_clear` unguarded beside it, which is the more consequential of the two since it is what grants the heal.

Reproduced the exact chain:

```
1. after A's failure:      flag=True owner=None attrib=None
2. after B's FAILED clear: flag=True owner=None attrib='B'   <- B claims A's flag
3. after B's recovery:     flag=False                        <- latch cleared
```

Step 3 is `recoverable_here = _flag_from_failed_clear == project_name` in `_hydrate_reingest_updater` firing for B on a flag A owns. B's scoped reingest then proceeded over A's partial graph.

**Fix:** the attribution now follows the same rule as the ownership claim — set only when this failure raised the flag from clean. Both fields are assigned under one `if not flag_before`, so they cannot drift apart again, which is how this survived the previous round.

Worth noting the precedent was already in the file: `_abandon_before_writing` guards its own attribution with `if not flag_before` and explains why. I had read that method while writing the first fix and still did not carry the rule across to its sibling.

A probe note, since it nearly cost me the finding: my first run reported 'refused, so the flag was not cleared' because the not-indexed guard raised — but the heal runs BEFORE that guard, and the printed state showed `flag=False`. The refusal masked a defect that had already happened. Making B indexed let the path complete and the finding was unambiguous.

Verified it discriminates: restoring the unconditional attribution reddens the new test and nothing else. Existing #1705 recovery tests all still pass (94 in the file), which matters because those pin the attribution's intended behaviour.

Head is now `a7f57383`. 602 passed across 22 files.

@greptileai review

vitali87:
Both directions of this finding are now fixed. The first half (a failed clear for B claiming A's flag) landed in `a7f57383`; the **second** half was still live and is fixed in `88d64dac`.

The second direction, reproduced:

```
1. after A's failed clear: flag=True attrib='A'
2. after B's clean clear:  flag=True attrib=None   <- A's licence discarded
3. A's own reingest: REFUSED
```

`_flag_from_failed_clear` is a project's licence to heal **itself**: A's failed clear strands a recoverable `writing=false` marker and records that A may lift the flag once that marker recovers. Clearing the attribution unconditionally on any successful clear threw that away, and A's own later reingest could no longer recover — refused indefinitely although A's graph was never touched.

It fails **closed**, so it is much less serious than the direction that wrongly allows a reingest. But it wedges a project with nothing wrong with it, and both directions come from the same unscoped assignment, so fixing one and not the other would have left the bug half-present. That is the same two-field trap I hit earlier on this PR, which is why I went looking for the mirror rather than stopping at the first half.

Now scoped: `if self._flag_from_failed_clear == project_name`. After the fix A keeps its attribution through B's clear and heals itself correctly, while B still cannot claim A's flag.

Verified it discriminates: restoring the unscoped clear reddens the new test and nothing else. Existing suite 95 passed; 603 across 22 files.

@greptileai review

vitali87:
Added a test pinning **#1820**'s reproduction, which this change also fixes. A peer session was about to pick that issue up, so this records the coverage rather than leaving it incidental.

#1820 is a third face of the same overloaded `None`: a failed `wipe_database` leaves the flag up with no owner, and before this change the next successful clear for **any** project settled it, so reads proceeded against a half-wiped graph.

Ran its reproduction verbatim on both trees:

```
                     origin/main      this branch
failed wipe:         flag=True        flag=True
any-project clear:   flag=False       flag=True    <- the defect
reingest:            ACCEPTED         REFUSED
```

It falls out of the ownership rule rather than needing its own change — an operation claims ownership only when it raises the flag from clean, and a wipe deliberately claims none. But 'falls out' is worth an assertion, or a later refactor can take it away while the other tests stay green.

The mutation is the useful part: restoring the unconditional settle reddens **all three** wipe tests together, which is what confirms they are one rule rather than three patches.

Note for anyone reading #1820: the field it names, `_incomplete_project`, exists nowhere in the tree — it appears to have been written against a shape never committed. The live field is `_incomplete_owner`.

Head is now `78dd12c0`. 97 in the MCP file, 604 across 22.

@greptileai review

PR #1865, opened 2026-09-12
fix: count special token spellings as ordinary text

lllleolin-max:
## Summary

- Count literal spellings such as `<|endoftext|>` as ordinary source text. Tiktoken's default `encode()` rejects them, causing query-result truncation or context pruning to raise `ValueError` when retrieved code mentions one.
- Use `encode_ordinary()` and add regression tests for all five special-token spellings in the configured encoding and for truncating a query result containing one.

## Type of Change

- [x] Bug fix

## Related Issues

Found during source review; no associated issue.

## Test Plan

- [x] New tests added
- Targeted token tests: 15 passed with `uv run --frozen --no-sync python -m pytest codebase_rag/tests/test_token_utils.py --noconftest -o addopts= -q`, including a rerun with the locked Tiktoken 0.12.0. Before the fix: 6 failed, 9 passed.
- Targeted `ty check`, repository Ruff hooks, Bandit, and README generation passed. README generation produced no changes.
- Core dependencies were installed from `uv.lock` using wheels only, excluding `pymgclient`. Full unit/integration suites were not run; `--noconftest` avoids unrelated parser/database fixture imports. The full pre-commit type check with locked `ty` 0.0.13 still reports 45 diagnostics, including 34 unresolved imports for optional packages such as PyTorch, libclang, and ast-grep. Heavy semantic-search dependencies were not installed.

## Checklist

- [ ] All pre-commit checks pass (full-project type check limitation above)
- [x] No new hardcoded production strings
- [x] No new type ignores, casts, `Any`, or `object` type hints

AI assistance: OpenAI Codex helped investigate, implement, review, and test this change.

## Summary by CodeRabbit

- **New Features**
  - Added provider-independent token estimates for message histories.
  - Tool payloads and structured argument values are now included in token-cost estimates, improving accuracy for messages containing tool calls.
  - Special token spellings are treated as ordinary text when counting tokens.
- **Bug Fixes**
  - Improved result truncation for content containing special token spellings, ensuring token limits are applied correctly.

### DIFF
--- a/codebase_rag/tests/test_token_utils.py
+++ b/codebase_rag/tests/test_token_utils.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import pytest
+
 from codebase_rag.types_defs import ResultRow
 from codebase_rag.utils.token_utils import count_tokens, truncate_results_by_tokens
 
@@ -17,8 +19,32 @@ class TestCountTokens:
         long = count_tokens("hello world this is a longer string with more tokens")
         assert long > short
 
+    @pytest.mark.parametrize(
+        "text",
+        [
+            "<|endoftext|>",
+            "<|fim_prefix|>",
+            "<|fim_middle|>",
+            "<|fim_suffix|>",
+            "<|endofprompt|>",
+        ],
+    )
+    def test_special_token_spellings_are_ordinary_text(self, text: str) -> None:
+        assert count_tokens(text) > 1
+
 
 class TestTruncateResultsByTokens:
+    def test_special_token_spellings_in_query_results(self) -> None:
+        rows: list[ResultRow] = [
+            {"content": "A source comment mentioning <|endoftext|>"},
+            {"content": "another result"},
+        ]
+        results, tokens, truncated = truncate_results_by_tokens(rows, max_tokens=1)
+
+        assert results == rows[:1]
+        assert tokens > 1
+        assert truncated is True
+
     def test_empty_results(self) -> None:
         results, tokens, truncated = truncate_results_by_tokens([], max_tokens=1000)
         assert results == []
--- a/codebase_rag/utils/token_utils.py
+++ b/codebase_rag/utils/token_utils.py
@@ -17,7 +17,7 @@ def _get_encoding() -> tiktoken.Encoding:
 
 
 def count_tokens(text: str) -> int:
-    return len(_get_encoding().encode(text))
+    return len(_get_encoding().encode_ordinary(text))
 
 
 def truncate_results_by_tokens(


vitali87:
Reviewed and verified this change independently; it is correct and the tests are sound. It is blocked only on CI approval, which is a maintainer action.

**The bug reproduces.** Against the configured encoding (`cl100k_base`):

- `encode("<|endoftext|>")` raises `ValueError: Encountered text corresponding to disallowed special token '<|endoftext|>'`
- `encode_ordinary("<|endoftext|>")` returns 7 tokens

So `count_tokens` raises on any retrieved source that merely *mentions* one of these spellings. `count_tokens` feeds `context_pruning.py` (`_content_tokens`, and the `PRUNED_PLACEHOLDER` sizing), so the crash reaches context pruning on real code, not just the truncation path named in the description.

**The fix is complete.** `git grep` across the whole repository finds exactly one tiktoken `.encode(` call site — the one changed here. Every other `.encode(` is UTF-8 byte conversion and unrelated.

**The tests are sensitive to the fix**, which I checked rather than assumed. Reverting `encode_ordinary` back to `encode` reddens exactly six tests and no others:

```
test_special_token_spellings_are_ordinary_text[<|endoftext|>]
test_special_token_spellings_are_ordinary_text[<|fim_prefix|>]
test_special_token_spellings_are_ordinary_text[<|fim_middle|>]
test_special_token_spellings_are_ordinary_text[<|fim_suffix|>]
test_special_token_spellings_are_ordinary_text[<|endofprompt|>]
test_special_token_spellings_in_query_results
```

Restored: `test_token_utils.py` and `test_context_pruning.py` together are 32 passed.

**Why it cannot merge yet.** All six workflow runs at head `2872694` are `action_required` — held pending maintainer approval because this is a cross-repository PR:

| Workflow | Conclusion |
|---|---|
| CI | `action_required` |
| Require CI At Head | `action_required` |
| SonarCloud | `action_required` |
| OSV-Scanner | `action_required` |
| ClusterFuzzLite PR | `action_required` |
| PR Split Score | `action_required` |

Only Greptile and CodeRabbit have reported, and neither is a required context. Across recent `pull_request` CI runs, every same-repo run is `success`/`cancelled` and this is the only `action_required` one, so the fork gate is the cause rather than a workflow-trigger problem.

Consequence for the merge gate: `All Checks Pass` is absent at the head and none of the eight jobs it aggregates has reported, so there is nothing to verify against. A maintainer needs to approve the workflow runs; after that the normal gate applies. Nothing is needed from the contributor.

vitali87:
Correcting my comment above: I said a maintainer approving the workflow runs would let the normal gate apply. The first half is right, the second is not, and the difference matters to you as a contributor — so I would rather say so than let you wait for a green that cannot arrive.

Approving the runs is **necessary but not sufficient**. `Require CI At Head` is a required check that cannot pass on any fork PR, independently of whether CI runs and passes. It is filed as #1866.

The mechanism: the gate matches a CI run to this PR through the run's `pull_requests[]` association, but GitHub does not populate that array for runs originating in a forked repository. An absent association is then treated as proof of a *foreign* one, so the filter matches nothing — permanently, not while something is still starting up.

The controlled comparison that separates this from a fork-approval problem is PR #1821, another fork PR whose workflows **were** already approved:

| | fork | CI run | `Require CI At Head` |
|---|---|---|---|
| #1821 | yes | in_progress, `pull_requests: []` | failure |
| #1835 | no | success, `pull_requests: [1835]` | pass |

#1821 shows the gate still failing after approval and after CI ran, which rules out approval as the remaining blocker. The only structural difference left is the fork.

So for this PR the state is: six runs held at `action_required` pending maintainer approval, and `Require CI At Head` will still not pass once they are approved, until #1866 is fixed. Nothing here is caused by your change, and nothing about it is yours to fix.

The review of the change itself stands unchanged — the bug reproduces, the fix is complete and correctly scoped, and the six tests are provably sensitive to it. Credit for finding #1866 goes to a parallel session that pushed past my initial "it's a startup race" reading; I had it wrong first.

lllleolin-max:
Thanks for the thorough review and the clarification. I really appreciate
you taking the time to investigate and keep me informed!

Vitali Avagyan ***@***.***> 于2026年9月13日周日 09:58写道：

PR #1873, opened 2026-09-13
test: keep startup key validation tests hermetic to provider env vars

### LINKED ISSUE #1871: Two startup-key-validation tests fail without a local Ollama server
## Summary

`codebase_rag/tests/test_mcp_startup_key_validation.py` has two tests that fail on a machine without a local Ollama server running, because they reach the network instead of mocking the provider. They fail on a clean checkout of `main` with no local changes.

## Reproduction

On `main` at `f403aedf`, with no Ollama running:

```
uv run pytest codebase_rag/tests/test_mcp_startup_key_validation.py -p no:randomly -q
```

```
FAILED TestStartupKeyValidation::test_remote_provider_without_key_fails_before_services
FAILED TestStartupKeyValidation::test_cypher_role_is_validated_too
2 failed, 1 passed
```

The errors:

```
ValueError: Ollama server not responding at http://localhost:11434.
  codebase_rag/providers/base.py:218

LLMGenerationError: Failed to initialize CypherGenerator: Ollama server not responding at ...
  codebase_rag/services/llm.py:134

Failed: DID NOT RAISE <class 'ValueError'>
```

## Why this is a defect, not just a local environment gap

Both tests assert that a **missing API key** is rejected *before* any service is constructed — `test_remote_provider_without_key_fails_before_services` says so in its own name. The failure shows the opposite happening: construction proceeds far enough to open a connection to `localhost:11434`. So the test cannot distinguish "the key check fired" from "the key check did not fire but Ollama happened to be reachable".

That makes the pass conditional on an ambient service. On a machine with Ollama running they pass; without it they fail for a reason unrelated to what they assert. Either way the assertion is not measuring the key check in isolation.

The third test in the class passes, so this is specific to these two paths.

## Suggested direction

Mock the provider construction so the key validation is exercised without any network call, and assert on the raised error identifying the role (`ModelRole.ORCHESTRATOR` / `ModelRole.CYPHER`) rather than on whatever error surfaces first. A test for "fails before services" should assert the service constructor was never called, which also pins the ordering the name claims.

## Scope

Found while verifying an unrelated branch: these two were the only failures in an 818-test run, and they reproduce identically on `main`, so they are not caused by that work.

wangzhengzhuo05:
## Summary

`codebase_rag/tests/test_mcp_startup_key_validation.py` has two tests that assert a missing API key is rejected *before* any service is constructed. They were not hermetic: `patch.dict(os.environ, {...})` merges into the ambient environment, so an exported provider key made `ModelConfig.validate_api_key` return early at `config.py:140`. The key check never fired, `create_server()` proceeded to build its services, and the tests failed with either `DID NOT RAISE <class 'ValueError'>` or an unrelated `Ollama server not responding` error — exactly the failure reported in #1871.

Reproduced on a clean checkout of `main` at `f403aedf`:

    $ ANTHROPIC_API_KEY=sk-ant-... uv run pytest codebase_rag/tests/test_mcp_startup_key_validation.py -p no:randomly -q
    FAILED TestStartupKeyValidation::test_remote_provider_without_key_fails_before_services
    FAILED TestStartupKeyValidation::test_cypher_role_is_validated_too
    2 failed, 1 passed

## Why

The pass was conditional on the developer's shell rather than on the code under test. Both tests assert "fails before services", yet they could not distinguish "the key check fired" from "the key check did not fire because an ambient key short-circuited it".

## How

Test-only change, one file.

- Added `_PROVIDER_KEY_ENV_VARS` (reusing the existing `cs.ENV_*_API_KEY` constants) and a `_isolated_environment` context manager that enters `patch.dict(os.environ, ...)` and then removes every provider key, so `validate_api_key` cannot short-circuit on the ambient environment.
- All three tests now use `_isolated_environment(...)` instead of the bare `patch.dict`.
- The two failure tests also patch `srv.CypherGenerator` and assert both `MemgraphIngestor` and `CypherGenerator` were `assert_not_called()`, pinning the "before services" ordering the test names claim and keeping the tests off the network.

Production ordering is unchanged and already correct: `create_server()` validates both roles (`server.py:85-86`) before constructing `MemgraphIngestor` (`:93`) and `CypherGenerator` (`:104`).

## Tests

- `uv run pytest codebase_rag/tests/test_mcp_startup_key_validation.py -p no:randomly -q` -> 3 passed (clean env)
- `ANTHROPIC_API_KEY=... OPENAI_API_KEY=... GOOGLE_API_KEY=... uv run pytest ...` -> 3 passed (the condition from the issue)
- `uv run ruff check` + `uv run ruff format --check` on the file -> clean
- Neighbouring `test_mcp_server.py` + `test_config_validation.py` -> 32 passed

Mutation check: with the two `validate_api_key` calls at `server.py:85-86` removed, the two tests fail (`2 failed, 1 passed`); restoring them returns to `3 passed`. The new assertions guard the startup ordering rather than the environment.

Fixes #1871

## Summary by CodeRabbit

- **Tests**
  - Improved startup validation test coverage by isolating environment variables from ambient provider credentials.
  - Added checks confirming that invalid role configurations produce clear diagnostics.
  - Verified that related generation and ingestion components are not initialized when startup validation fails.


wangzhengzhuo05:
Withdrawing this PR — the fix for #1871 has already landed on `main` in `06a2e12c` ("test: isolate the startup key tests from the ambient provider keys") with the follow-up `75882cb6`, and that solution is broader than this one: it rebuilds the environment with `clear=True` and adds the guard test that keeps `_EXEMPTING_ENV_KEYS` honest when a new provider joins `validate_api_key`. That guard test is the part this PR did not have, so there is no reason to keep a conflicting duplicate around.

For the record, the local verification this PR carried (in case any of it is useful):

- Root cause as reproduced here: `patch.dict` merges into the ambient environment, so an exported `ANTHROPIC_API_KEY` made `ModelConfig.validate_api_key` short-circuit (`codebase_rag/config.py:140`), the missing-key branch was never reached, and the two startup tests failed with `DID NOT RAISE` / "Ollama not running" instead of asserting the intended behaviour.
- Mutation check: removing the `validate_api_key` calls in `codebase_rag/mcp/server.py` made exactly the two key-validation tests fail; restoring the calls returned them to green — i.e. the tests were guarding the startup key check once the environment was isolated.
- Clean-env and ambient-`ANTHROPIC_API_KEY`/`OPENAI_API_KEY`/`GOOGLE_API_KEY` runs both passed with the isolation in place.

One small thing from that work worth a glance while you are in this file: this PR also cleared `OPENAI_API_KEY` and `GOOGLE_API_KEY` alongside the three keys now listed in `_EXEMPTING_ENV_KEYS`. If those two are not consulted by the exemption map, the current list is correct and nothing needs doing — the guard test looks like it would say so.

Thanks for the quick turnaround on #1871, and apologies for the duplicate.

Issue #1125, opened 2026-08-08
MCP startup bypasses ModelConfig.validate_api_key for remote providers


vitali87:
Found by Greptile on #1095 (verified with an executed harness: https://app.greptile.com/trex/artifacts/f49a9bed-4ee9-4c5a-9a86-426c881c4d68).

When a remote provider (e.g. anthropic) is configured for the orchestrator or Cypher role with no role API key, MCP server startup never invokes `ModelConfig.validate_api_key()`. It constructs `CypherGenerator` directly, which fails later with a wrapped provider-specific `LLMGenerationError` instead of the documented role-aware missing-key diagnostic.

This predates #1095 — the parent-commit harness shows the same bypass — and became more visible once `server.json` stopped marking the keys required (#1062).

Fix: validate the active orchestrator and Cypher configurations before constructing MCP services, preserving the Ollama exemption (`test_local_providers_skip_validation`).

PR #1874, opened 2026-09-13
test: isolate the startup key tests from ambient provider keys (#1871)

### LINKED ISSUE #1871: Two startup-key-validation tests fail without a local Ollama server

vitali87:
Closes #1871.

Two tests in `test_mcp_startup_key_validation.py` passed or failed on ambient machine state rather than on the code.

## The defect, measured

On `origin/main`, the same two tests, same commit, differing only in the environment:

```
ANTHROPIC_API_KEY unset  ->  3 passed
ANTHROPIC_API_KEY set    ->  2 failed
```

That inversion is the bug. Both tests assert that a **missing API key** is rejected at startup, before any service is constructed — but:

1. `patch.dict(os.environ, ...)` **merges** into the real environment, and `ModelConfig.validate_api_key` returns early for a config whose provider env var is set. On a machine holding a real key, the missing-key branch was never reached and the test asserted nothing.
2. Neither test mocked `CypherGenerator`, so `create_server()` built a real one and opened a connection to a local Ollama endpoint. On a machine without Ollama that surfaced as `ValueError: Ollama server not responding`, which looks nothing like the assertion that actually failed.

So the tests could not distinguish "the key check fired" from "the key check did not fire but the provider happened to be reachable".

## The fix

* `_isolated_env()` builds an explicit replacement mapping — the real environment minus every key the validator treats as an exemption, plus `TARGET_REPO_PATH` — and passes it with `clear=True`. PATH and the loader's variables survive because they are carried over, not because the merge is left on.
* Both tests now mock `CypherGenerator` and assert it was **never constructed**, which is the claim `..._fails_before_services` already made in its name and did not check.
* A new guard test reads `validate_api_key`'s source and fails if a provider env key is added there and not to the helper's list, so this cannot silently reopen.

## Verification

* The repaired tests pass with all three provider keys set **and** unset, and under random ordering.
* They still catch the defect they exist for (#1125): replacing both `validate_api_key` calls in `mcp/server.py` with `pass` reddens exactly those two and nothing else.
* The drift guard genuinely reddens — adding a fourth provider key to `validate_api_key` and not to the list fails it with the new key named.
* `patch.dict` restores exactly: the environment before and after is identical, and a later test in the same session sees no leakage.
* Related suites: 734 passed, 0 failed.

The third test in the file is deliberately untouched — it uses a local provider, exempted before the environment is consulted, and behaves identically either way.

## Summary by CodeRabbit

* **Tests**
  * Improved startup validation test isolation from ambient provider API-key environment variables.
  * Added coverage to ensure provider key handling remains aligned with validation logic.
  * Added checks confirming remote services are not initialized when role-specific API-key validation fails.


vitali87:
@greptileai review

PR #1875, opened 2026-09-13
fix: re-ingest hydration derives the package map without emitting containers (#1872)

### LINKED ISSUE #1872: Re-ingest hydration derives the whole repo and gives changed directories two container identities
## Summary

`GraphUpdater._hydrate_for_reingest` calls `identify_structure()` **unscoped**, which walks the entire repository and emits a `Package` or `Folder` node for every directory as it is on disk *now*. Any directory whose package-ness changed since the last index gains a node of the new kind while keeping the old one, so one directory ends up with two contradictory container identities — even though the scoped re-ingest never named it.

## Mechanism

`codebase_rag/graph_updater.py`, in `_hydrate_for_reingest`:

```python
self.factory.structure_processor.identify_structure()
```

`identify_structure`'s own docstring already describes this failure and names the remedy:

> `only` restricts BOTH the walk and the emission to the given repo-relative directories. A scoped re-ingest uses it: the unrestricted walk emits a node for every directory that changed on disk since the last derivation, including ones the call never named, so an unrelated directory whose `__init__.py` had been removed elsewhere gained a Folder node while keeping its Package node.

The scoped re-derivation inside `reingest` does pass `only`. The hydration call does not, so it re-introduces exactly the problem `only` exists to prevent.

Hydration cannot simply pass `only`: it needs the whole `structural_elements` map so a re-parsed module's parent lookup finds its enclosing package. What it does **not** need is to *emit* nodes for directories outside the call's scope. The map population and the node emission are one operation and need separating.

## Reproduction

Two sibling packages. `other/` loses its `__init__.py` on disk; the re-ingest names only `pkg/`.

```
after index:        {'other': ['Package'],           'pkg': ['Package']}
after pkg reingest: {'other': ['Folder', 'Package'], 'pkg': ['Folder', 'Package']}
```

Both directories end up holding two identities on `main` at `f403aedf`.

Note the predicate matters: a `Package` is keyed on a dotted qualified name (`proj.other`) and a `Folder` on an absolute path, so comparing raw uids cannot see that both describe one directory. Compare the `absolute_path` property, which both labels carry — my first probe compared uids and reported "no defect" against output that plainly showed the duplicate.

## Scope

Found while reviewing PR #1835, which fixes the *named* directory's flip on the scoped path. On that branch `pkg` is correctly reconciled and only the unrelated `other` remains duplicated, so #1835 strictly reduces this; the residue is this separate, older defect in the hydration path. Greptile raised it against #1835 and it is worth its own issue rather than expanding that PR.

## Suggested direction

Give `identify_structure` a way to populate `structural_elements` without emitting, and have `_hydrate_for_reingest` use it. The emission then happens only through the scoped call inside `reingest`, which already restricts itself to the flipped directories and their children.

A test needs a directory-level predicate (group container nodes by `absolute_path` and assert no directory carries more than one label), because the existing `_containers` helper in `codebase_rag/tests/test_reingest_package_flip.py` reduces a Package uid and a Folder uid to strings that can never collide, so a duplicate reads as two unrelated entries.

vitali87:
Closes #1872.

`_hydrate_for_reingest` called `identify_structure()` **unscoped**. That walks the whole repository and emits a `Package` or `Folder` node for every directory as it is on disk *now*, so any directory whose package-ness changed since the last index gained a node of the new kind beside its surviving old one — two contradictory container identities, for a directory the scoped re-ingest never named.

## Reproduction

Two sibling packages. `other/` loses its `__init__.py` on disk; the re-ingest names only `pkg/`.

```
after index:        {'other': ['Package'],           'pkg': ['Package']}
after pkg reingest: {'other': ['Folder', 'Package'], 'pkg': ['Folder', 'Package']}
```

`identify_structure`'s own docstring already described this failure and named `only` as the remedy — but hydration cannot use `only`. It needs the **whole** map, because a re-parsed module's parent lookup can reach any directory. What it does not need is to emit anything.

## The change

* `identify_structure` gains `emit: bool = True`. `emit=False` populates `structural_elements` and writes nothing — verified by spying every ingestor call: zero nodes, zero edges.
* `_hydrate_for_reingest` passes `emit=False`.
* New `_uncontained_dirs` derives a container for the ancestor directories of this call's files that the graph holds **no** container for.

That third piece is not optional, and it is the part a naive fix misses. Hydration's emission was the only writer of the container node for a directory that is **new since the last index**: the scoped re-derivation covers directories whose package-ness *flipped*, and a brand-new directory never flipped. Suppressing emission alone left `CONTAINS_FILE` and `CONTAINS_MODULE` edges pointing at a node nothing had written — and in production the merge query MATCHes both endpoints, so the edge is not created at all and the new file is **silently unparented**.

Measured before the fix: 6 dangling containment edges from adding `newplain/n.py`, `newpkg/p.py` and `pkg/sub/q.py` through a fresh updater.

The `_recorded_container_kinds` guard on that walk is load-bearing: an unconditional ancestor scope re-derives an ancestor that changed independently and gives it a second identity, which is the #1835 rule. Removing the guard reddens `test_an_independently_changed_ancestor_keeps_one_identity`.

## Verification

Five new tests. Mutation results, each reddening exactly what it should and nothing else:

| reverted | reddens |
|---|---|
| `emit=False` → `emit=True` | the unnamed-directory duplicate-identity test |
| the uncontained-directory walk | both new-directory tests |
| the `_recorded_container_kinds` guard | the #1835 ancestor-identity test |
| map population, emission still off | both map-contract tests |

368 passed across the re-ingest, structure and incremental suites. `ty` clean apart from one warning identical on `main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed re-ingestion scenarios where directories changing between package and folder types could receive duplicate identities.
  - Ensured newly discovered top-level and nested directories are correctly represented during re-ingestion.
  - Preserved file and module containment relationships after directory structure changes.
  - Re-ingestion now stops safely when an existing directory container type cannot be determined.
  - Corrected stale container cleanup when ancestor directory types change.
  - Prevented partial re-ingestion results when container validation fails.


vitali87:
@greptileai review

vitali87:
Both findings confirmed and fixed at `8bc24cce`. The P1 is the same defect class one level up from the one this PR fixes, which is worth stating plainly.

**Preserve recorded parent identity (P1).** Reproduced exactly as described — `outer/` still stored as a Package, its `__init__.py` removed on disk, then `outer/new/mod.py` added:

```
before: CONTAINERS {'new': ['Folder'], 'outer': ['Package']}   DANGLING 1
        Folder outer CONTAINS_FOLDER -> new
```

A child's parent identity comes from `structural_elements`, which hydration fills from DISK, so the edge named a Folder identity the graph does not hold and the new subtree would be disconnected.

Fixed by splitting the walk's two cases. A directory with NO container only needs deriving; one whose recorded kind DIVERGED from disk needs deriving **and** pruning, so it joins `flipped_dirs` rather than getting a second parallel mechanism. My first attempt derived without pruning and produced `outer: ['Folder', 'Package']` — the duplicate identity this PR exists to remove, recreated by its own fix. Now:

```
after:  CONTAINERS {'new': ['Folder'], 'outer': ['Folder']}    DANGLING 0
clean:  CONTAINERS {'new': ['Folder'], 'outer': ['Folder']}
```

That also changed an existing test's expectation. `test_an_independently_changed_ancestor_keeps_one_identity` asserted the ancestor KEEPS its stale Package node, which was right when reconciling was not possible and is now too weak — leaving it stale is what disconnects the child. It compares against a clean rebuild instead, pinning presence and absence at once, and it still catches the original #1835 bug: deriving the diverged ancestor without pruning reddens exactly that test.

**Unreadable container query (Major).** Also right — `_recorded_container_kinds` returns an empty set for both "no container" and "the read failed", so a transient failure would have derived every ancestor. The walk now aborts with `ReingestAborted` in the read-only prologue, and a test asserts no nodes were written when it does.

One correction to the shape, found by running it: my first version treated a sink with NO query surface as unreadable too, which aborted every re-ingest against a plain sink and reddened five of main's own tests. That is a legitimate configuration, not a failure — only a read that RAISED is the unknown. There is a control test for that direction now.

826 passed across the re-ingest, structure, incremental and MCP suites.

vitali87:
@greptileai review

vitali87:
Both fixed at `ab940cc5`. The P1 is a flaw in my own guard and the framing is worth keeping.

**Duplicate container query.** Correct — I asked twice, so a failure between the two reads turned an existing container into "no container", the directory was treated as new, the disk-derived identity was emitted WITHOUT joining `flipped_dirs`, the stale node was never pruned, and both identities survived. The guard reopened the exact defect it was added to close.

Now ONE read with three outcomes: a set (possibly empty, meaning no container node) or `None` meaning the read raised. A sink with no query surface still returns an empty set rather than `None`, because that is a legitimate configuration rather than a failure — an earlier version conflated them and aborted every re-ingest against a plain sink, reddening five of main's own tests.

I also extracted the row parsing into one shared helper. Two readers of a single query is how the parsing drifts, and that parsing already carries two non-obvious rules (read every row, not the first; narrow to `list[str]` before the membership test) that would have drifted separately.

Pinned by `test_a_read_that_fails_on_the_SECOND_ask_still_aborts`, which fails only the second container query — exactly the window the two-read version left open — and asserts both that it aborts and that no directory ends with two identities. It carries a fixture guard that the query was asked at least twice, so it cannot pass by never reaching the window.

**Compare containment edges (P2).** Also right: comparing container identities alone passes while the nested directory's edge still targets the pruned ancestor identity or is absent, so the node can be correct and the graph still disconnected. The ancestor test now compares `_containment` against the clean index as well as `_containers`.

827 passed across the re-ingest, structure, incremental and MCP suites.

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
@greptileai review

Issue #1787, opened 2026-09-07
Bulk 'verified IMPORTS edges' write fails 100% right after emission (nodes may not exist)


Victor-johnsson:
After a large `index_repository` run logs `Emitted N verified IMPORTS edges`, the very next ~15-20 flush batches (1000 relationships each) all fail completely:

```
INFO | Flushed 1000 relationships (0 successful, 1000 failed).
```

repeated ~15-20 times in immediate succession, right before "Analysis complete" for that unit. This looks like the edges are being created before their target module nodes are committed to Memgraph — a write-ordering issue, not a data problem (the same-named nodes exist moments later once the whole file/module's nodes are flushed).

**Repro:** `microsoft/aspire` (public repo), full `index_repository` run. `Emitted 28823 verified IMPORTS edges` followed immediately by this failure burst was reproduced consistently.

**Impact:** likely tens of thousands of dropped IMPORTS relationships on a repo this size — silent data loss in the graph (the job doesn't fail, just logs warnings).

**Environment:** same as #1786 (code-graph-rag v1.28.1, MCP stdio transport, microsoft/aspire).

Victor-johnsson:
Traced this a bit further, in case it's a useful lead.

**Call path:** `graph_updater.py`'s `run()` calls `flush_deferred_import_edges()` (`graph_updater.py:1701`) deliberately late, after Pass 3 -- per an existing comment there, so C# namespace imports can resolve against modules recorded during call processing (issue #1347).

In `import_processor.py:1344`, `flush_deferred_import_edges()` verifies each deferred import's target against `known_module_paths` -- an in-memory dict built during Pass 2 -- and only calls `_emit_import_edge()` (line ~1461) once that verification succeeds. The docstring there states the assumption explicitly: *"a guess that resolves nowhere... emits no edge, because the phantom endpoint is silently dropped by the database anyway."*

`_emit_import_edge()` (`import_processor.py:1307`) doesn't write immediately -- it calls `MemgraphIngestor.ensure_relationship_batch()` (`services/graph_service.py:376`), which queues the edge into `self._rel_groups` for a later batched write.

[...]

PR #1876, opened 2026-09-13
fix: retain failed node batches for retry

lllleolin-max:
## Summary

- Keep a failed node label's buffered rows available when `flush_nodes()` re-raises, while removing successful groups so a retry does not replay their CREATE writes.
- Serialize each node flush across its snapshot, write, and buffer update while allowing new nodes to be appended. A second explicit or automatic flush waits and then reads the remaining batch.
- Add regression coverage for serial and parallel flushes, CREATE and MERGE, batch-triggered failures, relationship recovery, and overlapping flushes with both successful and failed new rows.

## Type of Change

- [x] Bug fix

## Related Issues

Related to #1787: this addresses the failed-node-flush mechanism confirmed in the maintainer's investigation. It does not claim to reproduce or resolve the original full `microsoft/aspire` indexing report; the existing nodes-before-relationships ordering remains correct.

## Test Plan

- [ ] Full unit suite passes
- [x] New tests added: 12 initial regressions fail against unchanged production code; eight overlapping-flush regressions fail against the initial PR implementation. All 20 pass with this fix.
- [ ] Integration tests pass

117 focused tests passed across `test_graph_service_node_retry.py`, `test_graph_service.py`, `test_graph_service_calls_failure_logging.py`, and `test_ingestor_backend_selection.py`, with the normal repository conftest loaded. The Windows environment lacks native `mgclient`, so a repository-external test bootstrap supplied an empty import module; all exercised connections and cursors were the tests' own mocks. The environment also lacks pytest-xdist, so the run used `-o addopts=`. No real Memgraph/Neo4j or full-repository indexing run was performed.

Targeted Ruff lint/format, `ty check codebase_rag/services/graph_service.py`, and `git diff --check` passed. The actual configured pre-commit run passed whitespace, EOF, repository-pinned Ruff lint/format, and Bandit. README generation passed in the latest hook run. The full-project `ty` hook remains blocked by 45 diagnostics outside the changed production file, including unavailable optional dependencies and Windows `resource` attributes; the complete pre-commit gate is therefore not claimed as passing.

## Checklist

- [ ] All pre-commit checks pass
- [x] No new hardcoded strings in production code
- [x] No new `# type: ignore`, `cast()`, `Any`, or `object` type hints

AI assistance was used to investigate, implement, and test this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed node writes are now retained for retry instead of being discarded.
  * Successful node writes are not replayed during subsequent retries.
  * Concurrent flushes are serialized to prevent duplicate writes or lost failures.
  * Nodes added while a flush is in progress remain buffered.
  * Relationship writes retry pending nodes before proceeding.
  * Flush failures continue to be reported while preserving retryable data.

* **Tests**
  * Expanded coverage for retry behavior, buffering, failure reporting, and overlapping flushes.


vitali87:
@greptile review

Issue #1454, opened 2026-08-26
subprocess text=True decodes with the locale encoding in ~120 places, corrupting non-ASCII on Windows


vitali87:
PR #1450 fixed this for the Node oracle runner. That was a point fix on one instance of a class with **120 occurrences**.

## The defect

`subprocess.run(..., text=True)` with no `encoding=` decodes the child's output using the **locale** encoding. That is UTF-8 on the Linux and macOS runners and **cp1252 on the Windows runners**. Every tool cgr shells out to — cargo, javac, go, dotnet, node, npm, scala-cli — writes UTF-8. So on Windows any non-ASCII identifier, path, or message comes back mangled.

The failure surfaces far from its cause. It is not a decode error; it is a lookup miss for a name that should have been present, or a path that silently does not match. In #1450 it appeared as `KeyError: 'Café'` from a Ruby oracle whose parsing was entirely correct.

## Scope

```
text=True without encoding=, in codebase_rag/ and evals/:  120
```

Five other language oracles carry the identical bug the Node ones had:

- `evals/oracles/rust_oracle.py`
- `evals/oracles/scala_oracle.py`
- `evals/oracles/csharp_oracle.py`
- `evals/oracles/go_oracle.py`
- `evals/oracles/java_oracle.py`

Plus production paths: `parsers/go_frontend/frontend.py`, `parsers/java_frontend/frontend.py`, `parsers/java_lombok.py`, `parsers/stdlib_extractor.py`, `tools/health_checker.py`, `main.py`, `parser_loader.py`.

## Why it has stayed hidden

Every fixture in the repo is ASCII. An all-ASCII fixture **cannot** distinguish a correct decode from a locale decode — both produce identical output — so the bug is invisible to the existing tests by construction, on every platform. It only appears when a non-ASCII fixture meets a non-UTF-8 locale, which is why adding one Ruby fixture with a Japanese comment exposed it immediately.

The oracles are the sharpest case: an oracle is *ground truth*. A mangled identifier there does not crash, it silently scores a correct implementation as a miss.

## Fix

Pass `encoding=cs.ENCODING_UTF8` at every call site, as `evals/oracles/_common.py` now does.

The durable part is a gate, not the sweep. Options, roughly in order of preference:

1. A ruff/flake8 rule or a custom AST check in CI that fails on `text=True` without `encoding=`. Makes the class unrepresentable.
[...]

PR #1877, opened 2026-09-13
fix(ci): recognize fork PR runs with unambiguous source history

### LINKED ISSUE #1866: Require CI At Head can never pass on a fork PR: pull_requests[] is always empty for fork runs
`Require CI At Head` fails whenever it outruns GitHub's PR association, which is a race it loses on any fresh push.

## What happens

`.github/workflows/require-ci-at-head.yml` asserts a `ci.yml` run exists at the PR head AND that the run names this PR:

```
ci_count=$(echo "$runs" | jq --argjson pr "$PR_NUMBER" \
  '[.[] | select(.path == ".github/workflows/ci.yml")
        | select(.prs | index($pr))] | length')
```

Both workflows fire on the same `pull_request` event, so they start together. GitHub populates a run's `pull_requests[]` array **asynchronously, after the run is created**. A run that exists at the head but has not yet been associated comes back with `pull_requests: []`, the `index($pr)` filter drops it, `ci_count` is 0, and the gate reports "No 'CI' workflow run exists at head" while that run is sitting in the queue.

## Measured on #1821

The gate failed at `2026-09-13T01:48:45Z` (run 34445551108). Its own debug output, from the same job:

```
Runs at this head:
  .github/workflows/sonarcloud.yml [SonarCloud] PRs=[]: completed/skipped (pull_request)
  .github/workflows/ci.yml [CI] PRs=[]: queued/pending (pull_request)
  .github/workflows/cflite_pr.yml [ClusterFuzzLite PR] PRs=[]: in_progress/pending (pull_request)
```

The `ci.yml` run is right there on line 2. The gate printed it and then reported it absent. Note `PRs=[]` on **every** run including the `completed` one, so this is not specific to queued runs: nothing at that head was associated yet.

## Why this matters more than a flake

- It is a **required** check, so the false negative blocks the merge.
- It fails in the direction that looks like a real finding. The error text says the PR "has not been verified by the test suite", and the suggested remedies ("trigger one with `gh workflow run ci.yml`", "an empty commit is NOT sufficient") send you chasing a missing run that exists.
- It is timing-dependent, so it passes on PRs whose CI registered minutes ago and fails on whatever was just pushed. #1835, #1833 and #1547 pass right now; #1821 fails. That pattern reads as "a problem with #1821" rather than "a problem with the gate".

## Cause

Introduced by a5aa9178 ("ci: require the run to belong to this PR, and paginate the query"), which added the `select(.prs | index($pr))` filter. The intent is sound and documented: after a stack rebase several branches share a head, so a run at this SHA may belong to another PR. The defect is that absence of an association is treated as proof of a foreign run, when at startup it means "not known yet".

Before that commit the gate matched on path alone and had no race.

## Suggested fix

The association is only needed to exclude a run belonging to a *different* PR. That is `prs` being non-empty and not containing this PR. An empty `prs` is unknown, not foreign:

```
select(.prs | length == 0 or index($pr))
```

That keeps a5aa9178's protection (a run naming only PR #1234 is still rejected for PR #5678) and drops the race. It does admit an unassociated run at a shared head, so if that is judged too weak the alternative is to poll until the association appears or a timeout expires, and fail only then — the run's `head_branch` is another discriminator worth considering, since two PRs sharing a head SHA still differ by branch.

Either way the current formulation cannot distinguish "no CI run" from "the API has not caught up", and it is the gate's only assertion.

lllleolin-max:
## Summary

- Fix the permanent rejection of fork PRs whose CI runs have `pull_requests: []`. Match the workflow path, exact head SHA, source repository, and branch, then require paginated all-state PR history to identify this PR as the only PR ever using that exact source branch. Its current head must match and its creation time must be no later than the `pull_request` CI run. A second PR on the branch is ambiguous even if closed, at a different SHA, targeting another base, or created later; such branches require a fresh branch for this fallback.
- Keep explicit associations authoritative: a nonempty list must include this PR, and same-repository runs still require that association. Missing or malformed identity, timestamps, pagination data, and API failures fail closed. Add only read access to pull-request metadata.
- Preserve this gate's existing existence-only contract. `All Checks Pass` continues to enforce successful CI; queued, failed, or approval-required runs are not reported as successful tests.

The advisory `scripts/check_pr_gated.py` is unchanged and still conservatively rejects fork runs without explicit PR associations. This fixes the required workflow's run lookup; it does not claim that every advisory check passes or that a PR is ready to merge.

## Type of Change

- [x] Bug fix
- [x] CI/CD or tooling

## Related Issues

Closes #1866.

## Test Plan

- [x] New regression tests execute the actual workflow shell and jq projection with GitHub-shaped API responses. They cover forks, same-repository PRs, shared SHAs across repositories and branches, explicit foreign PR associations, reused source branches across open/closed PRs, later PRs, creation-time bounds, missing/null/malformed identity, pagination, API failures after partial results, and unfinished/failed run states.
- [x] Focused tests: `uv run --frozen --no-sync pytest codebase_rag/tests/test_require_ci_at_head.py --noconftest -o addopts= -q` — 43 failed / 45 passed against this PR's first revision, then 88 passed with the final history check. Bash and jq were available locally. `--noconftest` isolates this workflow test from application fixtures; the reused local environment lacks pytest-xdist and cannot load the application's CLI fixture. An independent review also ran the initial 72 cases with normal conftest and an external empty `mgclient` import shim.
- [x] Read-only replay of the old and new shell against live GitHub data: fork #1865 changes from exit 1 to exit 0; same-repository #1870 remains exit 0; #1549, with no run at its head, remains exit 1. #1865's CI was `action_required`, and #1870's was `failure`: these observations verify run existence, not successful CI or merge readiness.
- [x] `uv run --frozen --no-sync ruff check` and `uv run --frozen --no-sync ruff format --check` pass across the repository (1,246 Python files already formatted).
- [x] Actionlint 1.7.12 passes both workflows; its optional ShellCheck integration was disabled because ShellCheck is unavailable locally. `git diff --check` passes.
- [ ] Full unit/integration suite and hosted fork-PR workflow execution have not been run locally.

## Checklist

- [x] PR title follows Conventional Commits.
- [x] No new runtime dependencies; the `ci.yml` success gate is unchanged.
- [x] No `Any`, `object`, casts, or type-ignore annotations added.
- [ ] All pre-commit checks pass. Whitespace, EOF, YAML, Ruff lint and format checks pass. The whole-repository type hook reports 45 diagnostics in unchanged code, including missing optional dependencies in this local environment. The README hook regenerates dependency descriptions from local package metadata; its unrelated output was restored and is not included.

AI-assisted implementation and validation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CI validation for pull requests from forked repositories.
  - CI checks now verify the commit, branch, source repository, and event timing belong to the current pull request.
  - Added safeguards against incorrectly reusing workflow runs from other pull requests that share a source branch.
  - Incomplete, malformed, or unavailable pull-request history now causes validation to fail safely with clearer guidance.


vitali87:
@greptileai review

vitali87 (approved):
Thank you. LGTM

codebase_rag/tests/test_require_ci_at_head.py
```diff
@@ -0,0 +1,529 @@
...
+    head_repo: str = FORK,
+    env: dict[str, str] | None = None,
+    pr_pages: list[list[SourcePullRequest]] | None = None,
+) -> subprocess.CompletedProcess[str]:
+    bash = shutil.which("bash")
+    if bash is None or shutil.which("jq") is None:
+        pytest.skip("the Ubuntu workflow requires bash and jq")
+    workflow = yaml.safe_load(WORKFLOW.read_text(encoding="utf-8"))
+    step = workflow["jobs"]["require-ci-at-head"]["steps"][0]
+    return subprocess.run(
```
vitali87, line 126: The only CI failure is the repo's encoding guard (issue #1454): this call needs `encoding=cs.ENCODING_UTF8` next to `text=True` (import `constants as cs`). Everything else is green.

vitali87 (changes requested):

vitali87 (approved):

Issue #1604, opened 2026-09-01
evals graph emulator omits the project filter production applies to CALLS callees, masking the bug class it exists to catch


vitali87:
The `_StatefulIngestor` emulator in `evals/cgr_graph.py` answers `CYPHER_GRAPH_CALLEES` without the project filter that production applies, so the emulator returns rows production never would. A test mirror that is more permissive than the code it mirrors does not merely fail to catch this class of bug, it **masks** it.

## The divergence

Production (`codebase_rag/cypher_queries.py`, `CYPHER_GRAPH_CALLEES`):

```cypher
MATCH (caller)-[r:CALLS]->(callee)
WHERE caller.qualified_name = $qn AND callee.qualified_name STARTS WITH $project_prefix
```

Emulator (`evals/cgr_graph.py`, the `CYPHER_GRAPH_CALLEES` branch):

```python
for edge in self._out.get(source, ()):
    callee = (edge[3], edge[4])
    if (
        edge[2] == cs.RelationshipType.CALLS.value
        and callee in self.nodes
    ):
        rows.extend(self._graph_edge_rows(edge, self._GRAPH_SITE_KEYS, callee))
```

`callee in self.nodes` is the only guard. There is no `$project_prefix` test, so every outgoing `CALLS` target is returned.

## Consequence

A project function that calls an external dependency yields a callee row in an emulator-backed context slice and no row in production. Any assertion written against the emulator therefore passes on behaviour production does not have, and a regression that wrongly admits out-of-project callees would go undetected because the emulator already admits them.

## Why this shape matters

The emulator exists to make graph-backed behaviour testable without a live database. Its value depends entirely on agreeing with production. Where it is *stricter* a test fails loudly and someone investigates; where it is *more permissive* the suite stays green and stops meaning anything for that path. This is the second half of the pair to fix, not an incidental cleanup: fixing only the production query would leave the mirror still masking.

## Suggested fix

Filter `edge[4]` by the project prefix before adding the row, matching the `STARTS WITH $project_prefix` semantics of the production query, and add a case with an out-of-project callee asserting it is absent from both paths.

Found by CodeRabbit on #1551 and verified at source against `origin/feat/context-slice`.

vitali87:
**Not actionable on `main` — the code this describes does not exist there yet.**

`CYPHER_GRAPH_CALLEES` has no occurrence in `evals/cgr_graph.py` on `origin/main`:

```
[...]

PR #1878, opened 2026-09-13
fix(evals): scope incremental rehydration to the requested project

lllleolin-max:
## Summary

- Apply the production project-prefix filters to the eval ingestor's definition, module, and inheritance rehydration queries. With two indexed projects, the double currently restores the other project's symbols into a fresh incremental updater even though the production queries exclude them.
- Add regressions that index two Python projects with matching module and symbol names, then add a caller through a fresh `GraphUpdater.run()`. They verify the recovered registry, modules, and ordered bases, while preserving the legitimate call into the unchanged local module. An asymmetric inheritance control keeps an external base when its child belongs to the requested project.

This fixes the eval double's fidelity; production query behavior is unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Related to the emulator-parity concern in #1604. This addresses three different queries already present on `main`, rather than the branch-only CALLS case discussed there.

## Test Plan

- [x] New tests added

The four new tests fail against the original `evals/cgr_graph.py` and pass with the fix. The focused run covering `test_eval_rehydration_project_scope.py`, `test_eval_double_query_coverage.py`, `test_incremental_eval.py`, and `test_eval_capture_merge.py` passed **33 tests**, with **2 skipped** for unavailable C# and Go grammars.

Local Windows validation used `uv run --frozen --no-sync` and `-o addopts=`. An external bootstrap supplied an empty `mgclient` import module so the normal test conftest could load; graph operations ran through the actual eval ingestor, with no database driver exercised. The same focused selection also passed without conftest or the bootstrap. A full suite and live database validation were not run.

Ruff lint/format, the changed emulator's `ty check`, whitespace/newline hooks, Bandit, and `git diff --check` passed. Full-project `ty` produced the same 45 diagnostics before and after this change, including missing optional dependencies and existing warnings. README generation could not complete because a PyPI response ended with `IncompleteRead`.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project-scoped graph queries so definitions, inheritance relationships, and modules return only results belonging to the requested project.
  - Incremental updates now remain limited to the selected project while preserving data from other projects.
  - Project-specific registries, modules, and inheritance state remain isolated during incremental updates.

- **Tests**
  - Added coverage for project boundaries, parsing, graph relationships, registries, modules, inheritance state, and incremental update behavior.


vitali87:
@greptile review

PR #1879, opened 2026-09-13
fix(trace): preserve UNC hosts in CPU profile file URLs

lllleolin-max:
## Summary

- Preserve the server authority when converting V8 CPU profile file URLs to UNC paths. `file://nas/share/project/main.js` currently becomes `/share/project/main.js`, so it fails the project-prefix check and its calls disappear. The Windows conversion regression produces **0 records before the fix and the expected 4 afterward**.
- Compare project paths with native case rules: Windows UNC host, share, and directory casing cannot discard valid frames; POSIX paths remain case-sensitive. Only comparison operands are normalized, preserving the casing of emitted generated and source-mapped paths.
- Keep case-insensitive `localhost` authorities local, preserve percent-decoded spaces and Unicode, and retain drive-letter and hostless file-URL behavior. Non-file URLs remain excluded. This matches the UNC conversion described in [Node's file URL documentation](https://nodejs.org/api/url.html#urlfileurltopathurl-options).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation
- [ ] CI/CD or tooling
- [ ] Dependencies

## Related Issues

Found while checking Windows CPU profile import paths; no existing duplicate fix found.

## Test Plan

- [x] New tests added

`test_dynamic_trace_cpuprofile.py`, `test_dynamic_trace_sourcemap.py`, and `test_dynamic_trace_records.py`: **72 passed, 2 skipped** on Windows. The native POSIX case-sensitivity control skips on Windows; the live TypeScript test skips because Node and a local TypeScript compiler were unavailable. The original converter fails the UNC URL/conversion regressions. The published initial fix still fails four new cases: host/share/directory casing each drops all 4 expected edges, and a parsed source map to a mixed-case UNC path drops the expected edge.

The Windows regression writes a synthetic `.cpuprofile`, runs `convert_cpuprofile`, and reads the resulting trace through `read_trace_file`. It verifies the complete edge set for same-case and mixed-case UNC paths, original output casing, line numbers, sample counts, runtime-frame traversal, and vendored-code exclusion. A real local Source Map v3 fixture verifies remapped UNC output paths and `node_modules` exclusion; separate controls reject other servers, shares, and sibling directories. It does not mount a share or claim SMB integration coverage. The URL conversion regressions run on every platform; the UNC repository tests require Windows.

Local checks used `uv run --frozen --no-sync`, with an external empty `mgclient` import module solely to let the normal test conftest load; the converter does not use a database. Ruff lint/format, targeted `ty`, whitespace/newline hooks, README generation, Bandit, and `git diff --check` passed. Full-project `ty` still reports the same 45 diagnostics as the unmodified upstream baseline, including unavailable optional dependencies and existing warnings. A full repository suite was not run.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CPU profile path handling for local files, network shares, and UNC-style paths.
  - Made project-root and dependency-directory matching reliable across platform-specific path casing.
  - Prevented profiles and trace frames from outside the project—including unrelated network paths—from appearing in converted records.
  - Preserved valid source-map paths under case-varied UNC roots while excluding dependency paths outside the project scope.
  - Retained local caller-to-callee trace relationships when remote HTTP frames are excluded.


vitali87:
@greptile review

PR #1885, opened 2026-09-13
ci: gate automatic release tags on successful source CI

### LINKED ISSUE #1663: Gate the release in version-bump.yml on green CI for the same SHA
## Problem

`version-bump.yml` and `ci.yml` both trigger on `push` to `main` and run independently. When the GHSA-wvxg-744g-6pcg fix merged (`033b64be`), CI went red on ubuntu, windows and the base install (run 33685599965: `Unit Tests (windows-latest, py3.12/3.13)`, `Unit Tests (ubuntu-latest, py3.12/3.13)`, `Unit Tests (base install)` all failed). Version Bump did not wait for it: it bumped to v0.0.845 and, because the commit message carried a GHSA id, cut the release and published to PyPI on a tree whose CI had never passed.

The failures turned out to be test-only (see #1662), but the release path had no way to know that. A red CI on the same SHA still produces a PyPI release.

## Why CI never ran before the merge

The fix was developed in the GHSA temporary private fork. GitHub does not run Actions on those forks at all (`/actions/runs` and `/actions/workflows` both 404 by design), so the first CI run on the change was on the merge commit, after merge. Advisory-fork merges are therefore exactly the commits most likely to reach `main` untested, and they are also the ones `version-bump.yml` releases immediately.

## Proposal

Make the release step in `version-bump.yml` conditional on `ci.yml` having succeeded for the same commit SHA. Options, in rough order of preference:

1. Trigger the bump/release from `workflow_run` on `ci.yml` (`types: [completed]`, `conclusion == 'success'`, branch `main`) instead of from `push`, so it cannot start before CI finishes.
2. Keep the `push` trigger but poll the commit's check runs / `ci.yml` workflow run for the SHA and fail (or skip the release step) unless it concluded `success`.
3. Fold the bump/release into `ci.yml` as a final job that `needs` the test matrix.

Whichever form, the version bump commit itself should not be the thing that gets gated (it is a chore commit), only the tag + PyPI publish.

## Not in scope here

- #1662 fixes the tests that were red on that SHA.
- This is a release-workflow change, so it is the owner's call; filing rather than implementing.

lllleolin-max:
## Summary

- Preserve the automatic version bump, then require successful `ci.yml` coverage of its original main SHA before pushing a release tag. A failed, missing, or unfinished run cannot trigger the tag's PyPI/binary/Docker publishing workflows or the subsequent GitHub release step. Interim tags retain their existing silent token push.
- Verify the active workflow identity, both source repositories, main branch, push event, exact SHA, and bump parent. Select the latest run/attempt across all pages; API errors, malformed responses, and incomplete pagination fail closed. Manual version-bump dispatch is restricted to main.
- Keep the existing release cadence and security-release decision. The gate polls for at most 90 attempts at 30-second intervals, bounded by a 50-minute step timeout. The job budget rises from 40 to 100 minutes to accommodate this wait alongside the existing highlights request, which can take 25 minutes.

This implements option 2 in #1663. The CI evidence covers the triggering source commit, not the generated chore commit: version metadata and generated NEWS/README remain part of the existing untested bump. The bump is already pushed if the gate refuses a release. Existing maintainer-controlled tag pushes, release creation, and manual PyPI dispatch are outside this automatic version-bump gate.

## Type of Change

- [x] Bug fix
- [x] CI/CD or tooling

## Related Issues

Fixes #1663.

## Test Plan

- [x] New regression tests added: 57 passed. They execute the actual workflow Bash with stubbed GitHub/Git commands, preserving step order and demonstrating that the bump remains while release tagging is blocked.
- [x] Verified red on the exact upstream workflow at `96e017e39b803a6f58b1db3da11d0a55656a42e9`: 53 failed, 4 passed.
- [x] Read-only GitHub API snapshots replayed through the local workflow shell: main CI run `34741163588` (`failure`) blocks the tag; source CI run `34741154226` (`success`) permits the simulated tag push.
- [x] Whole-repository Ruff lint and format checks, `git diff --check`, YAML validation, and actionlint 1.7.12 passed. ShellCheck was unavailable locally.
- [ ] Full unit/integration suite and hosted workflow execution: not run. No tag, release, or package was published during validation.

Local regression command: `uv run --frozen --no-sync pytest codebase_rag/tests/test_release_ci_gate.py --noconftest -o addopts= -q` with Git Bash and jq on PATH. The reused local environment requires `--noconftest` to avoid an unavailable CLI fixture dependency and `-o addopts=` because pytest-xdist is absent. Production CI keeps its normal test configuration.

## Checklist

- [x] PR title follows Conventional Commits.
- [x] No `# type: ignore`, `cast()`, `Any`, or `object` type hints.
- [ ] All pre-commit checks pass: whitespace, EOF, YAML, Ruff lint and format passed; the type hook reports 45 diagnostics in unchanged code, including absent optional dependencies. README generation passed on one bounded retry after an external `IncompleteRead`. No unrelated generated file changes remain.

AI assistance was used for implementation, regression tests, and review.

## Summary by CodeRabbit

* **Chores**
  * Release automation now runs on the main branch and records highlight-generation and news-update results in the job summary.
  * Release preparation fails clearly when no usable highlights are generated or an update step fails.
  * Releases validate complete, successful CI evidence and wait for eligible runs before creating tags or publishing releases.
  * Malformed or incomplete CI responses now block releases with explicit errors.

* **Tests**
  * Added comprehensive coverage for successful, delayed, failed, malformed, and incomplete CI responses to ensure releases fail safely.


vitali87 (changes requested):
clear bots' comments

vitali87:
@greptileai review

PR #1886, opened 2026-09-13
fix: a declared rename outranks the empty-container uniqueness check

### LINKED ISSUE #1836: A nested empty container is never declared, so a valid container rename rolls back
Found by Greptile on #1547 and deferred there: it is a pre-existing limit of
the empty-container matcher, not a regression that PR introduced.

## The gap

`structural_delta._empty_container_renames` will only pair a fingerprintless
container with a candidate when the operation DECLARED that pair:

```python
if len(matches) == 1 and len(peers) == 1:
    # The only admissible evidence: the operation says it did this.
    if (qn, matches[0]) not in declared:
        continue
```

That conservatism is right — an empty container has no descendants and no
fingerprint, so nothing in the two snapshots can distinguish "renamed" from
"removed, and a different one added".

The problem is what ends up in `declared`. `rename` builds it from
`report.hierarchy` (`editing/rename.py`), and `_hierarchy` is the symbol plus
every method it OVERRIDES or is overridden by, transitively. A nested
container is not in that chain. So renaming `OldContainer` to `NewContainer`
when it contains an empty nested class declares only the outer pair, the inner
one is never admissible, and the delta measures it as one removal plus one
addition. The contract then fails and rolls back a valid rename.

## Why it is a design question, not a one-line fix

Two directions, and they trade off:

1. **Widen what is declared.** Include nested containers under the renamed
   symbol in `pairs`. Their new qn is derivable the same way the outer one is
   (`parent + "." + new_name` with the prefix rewritten). This keeps
   "declared" as the only admissible evidence.
2. **Widen what counts as evidence.** Allow a one-to-one inference for an
   otherwise-matching fingerprintless container. This is what the reviewer
   suggested, and it is the weaker guarantee — it accepts a pairing nothing
   demonstrated, which is exactly what the current comment refuses.

(1) looks right to me: the operation genuinely does know it renamed the nested
container, so the fix is to say so rather than to relax the standard of proof.
But it is a judgement about what a declaration means, so it wants deciding
rather than assuming.

## Test to write

Two cases, since one alone passes for the wrong reason:

- a container rename with an empty NESTED container must satisfy its
  postcondition (fails today);
- an unrelated empty container appearing and another disappearing in the same
  edit must still NOT be paired (the control that keeps the conservatism).

## Not urgent

`move` is not wired to an operation and this needs a nested empty container
specifically, so it is a narrow shape. Should be closed before anything relies
on empty-container renames.

vitali87:
**Does not close #1836.** I originally thought it did, the local review caught that it does not, and I verified the review is right. Filing it as what it actually is.

## What this fixes

`_pair_lone_containers` (Pass 4 of rename detection, for empty containers with no fingerprint and no descendants) required BOTH candidate sets to be unique — `len(matches) == 1 and len(peers) == 1` — and consulted the caller's `declared` pairs only afterwards.

Rename two empty classes in the same file and each sees two matches and two peers, so both pairings are refused **before** the declaration is read. The contract then sees two removals plus two additions where the operation named two renames, and since an unexpected rename is a contract failure it rolls back a correct edit.

The declaration is the disambiguation the uniqueness checks exist to substitute for, so it is now the only test and the peer scan is gone.

## This deliberately loosens the gate

Worth stating plainly, because my first version of the comment claimed the opposite. With one removed and two added containers, the old code refused on `len(matches) != 1` and this admits the declared one. That is sound because a declaration is still *required* — not because nothing widened. `matches` is filtered on path and label before `declared` is consulted, so an absent, cross-file or mismatched-label target is still refused.

## Why it is not #1836

#1836 is about what gets **declared**; this is about what is **admitted given** a declaration. The nested container in that issue is never declared at all: `rename.py` builds `pairs` from `report.hierarchy`, and `_hierarchy` (`rename.py:140`) walks only `overrides` edges, so no descendant of a renamed symbol is ever named.

Verified by driving the issue's own scenario, with `declared` holding only the outer pair as the real caller produces:

```
declared: [('m.OldContainer', 'm.NewContainer')]
paired  : [('m.OldContainer', 'm.NewContainer')]
LEFT OVER old: ['m.OldContainer.Inner']
LEFT OVER new: ['m.NewContainer.Inner']
```

Still a removal plus an addition, still rolls back.

I am not fixing that here. It needs a containment query `graph_query.py` does not have, and the issue states option (1) "is a judgement about what a declaration means, so it wants deciding rather than assuming" — so it is the owner's call, not mine to assume.

## Tests

| Test | Proves |
|---|---|
| Two sibling empty renames, both declared | Both reported. Red on main before the fix. |
| Only one of the two declared | The undeclared one **stays** a removal plus an addition. The known-positive: without it, a guard widened to admit the group could pass everything in it. |
| Nested container, outer declared (`strict` xfail) | The test #1836 asks for. Pins the real gap and will announce itself if the upstream fix lands. |

`42 passed, 1 xfailed`.

Mutation-checked: reinstating a uniqueness precondition ahead of the declaration reddens exactly the two new passing tests and no others.

## Why the existing coverage missed this

The three pre-existing declared-rename tests each rename a **single** empty class, so both uniqueness counts are 1 and the old guard never fired on any of them. They passed throughout and could not have caught this — insensitive to the condition by construction.

## Review notes I acted on

Two claims in my own first comment were wrong and are corrected in `8af21062`: the comment blamed the peer scan alone when `matches` uniqueness blocked independently (measured: both counts are 2), and "loosens nothing" was false as described above. This function's comment was corrected once before for an unverified assertion, so it seemed worth not repeating that.

## Summary by CodeRabbit

- **Bug Fixes**
  - Declared renames of empty containers are now recognized even when multiple matching additions and removals exist in the same file.
  - Unspecified pairings continue to be reported separately as removals and additions rather than being inferred as renames.
  - Improved reporting for multiple declared empty-container renames in a single file.


vitali87:
Re-checked this PR end to end after the fact, and two of its claims are wrong. Recording them here, with the measurements, and recommending this PR be **closed in favour of #1880**.

**1. The xfail is at the wrong layer and can never announce the real fix.** The PR body says the `strict=True` xfail "will announce itself the moment the upstream fix lands". I ran it against #1880 — which *is* the upstream fix, widening what `rename.py` declares — and it **still xfails**:

```
2 passed, 40 deselected, 1 xfailed    # my 3 tests, run against #1880's code
```

Reason: the test drives `observe()` with an explicit `declared_renames` holding only the outer pair, so `rename.py` is never on the path. It then asserts the delta should pair an **undeclared** nested container — which is option (2) from the issue, the one it rejects. A correct option-(1) fix leaves that undeclared pair as a removal plus an addition, exactly as #1880 does. So the marker would sit xfailed forever under the right fix, and XPASS only under the wrong one. The opposite of its stated purpose.

**2. #1880 already covers what this PR fixes.** Both of my sibling tests pass against #1880's code, by execution, not by reimplementation as I first checked:

| Test | Against #1880 |
|---|---|
| Two declared sibling empty renames in one file → both reported | pass |
| Only one of the two declared → the other stays a removal plus an addition | pass |

#1880's own new file: `12 passed`, including `siblings`, `deeply-nested`, refusal cases, and declaration-uniqueness cases. Its approach also keeps a uniqueness check with real meaning — uniqueness *among declared candidates* — where this PR deleted the check outright.

**3. The two PRs conflict.** `git merge-tree --write-tree` between the branches: `CONFLICT (content): codebase_rag/structural_delta.py`. Both rewrite `_pair_lone_containers`; only one can land as written.

**What this PR would still be for.** Nothing #1880 does not do. Its one honest contribution — the sibling case — is in #1880's `siblings` parametrization. Keeping this open would only stand in #1880's way.

The caveat that applies to #1880 rather than to this PR: it is a cross-repository PR, so it sits behind the fork gate (`Require CI At Head` cannot pass on a fork until #1866 is fixed, and its workflow runs need maintainer approval). That is a decision for the owner, and it is not a reason to prefer this PR — it is a reason to fix #1866.

I am not closing this myself only because the owner asked for a re-check before I acted; the recommendation is to close it.

vitali87:
Superseded by #1945 — this PR should be closed.

#1880 merged first and fixed #1836 by the same route this PR took: moving the `declared` test to the front of the `matches` filter in `_pair_lone_containers` and replacing the path/label `peers` scan with a declared-based one. The production change here is now redundant, which is why this branch went `DIRTY` — the conflict is against `1c32bb8b`, #1880's `structural_delta.py` half.

Rather than rebase a duplicate fix, #1945 keeps only the two tests, rebuilt on current main. They earn their place: main has no coverage for the two-renames-in-one-file case, because its three existing declared-rename tests each rename a single empty class, so the uniqueness precondition never fires on them.

Verified before dropping the production change — on the issue's own nested shape, the pre-#1880 guard reports `[]` even with the widened `declared`, while the post-#1880 guard reports both renames. So #1880's change is load-bearing and this one is genuinely redundant, not merely a different spelling.

Correcting the record on two things I said about this PR earlier: I described #1880 as touching only `editing/rename.py` — it changed `structural_delta.py` too, which is where the overlap is. And the note on this PR that it does not close #1836 was right at the time but is now moot: #1880 closed it, and #1945 adds the missing coverage.

I have left this branch untouched rather than force-pushing over it.

vitali87:
Closing as superseded. The same change reached `main` through two other PRs:

- #1880 (merged 2026-09-15) carries the `_pair_lone_containers` fix in `codebase_rag/structural_delta.py`.
- #1945 (merged 2026-09-16) carries the test coverage: two declared empty-class renames in one file, and the nested-container case.

This branch's tree is now byte-identical to `main`, so the PR shows 0 changed files and merging it would be a no-op:

```
$ git diff --stat origin/main...a0e7754115de189613ffbe66476fe1813f0d08a8
(empty)
```

The branch is not an ancestor of `main` (`git merge-base --is-ancestor` returns non-zero) — its three commits are unique, but `a0e7754`, which merged `origin/main` in, resolved to main's already-equivalent content and cancelled the diff.

Note the comparison here is of the resulting tree, not of intent: #1880 was authored independently, so its fix may be shaped differently while producing an identical file. Either way the behaviour this PR aimed at is on `main`.

The #1836 gap this PR's description was careful to distinguish itself from — a nested container never being *declared*, because `_hierarchy` (`rename.py:140`) walks only `overrides` edges — is unaffected by this closure and remains tracked on #1836.

PR #1895, opened 2026-09-13
fix: key the stale type-fact filter on the fact's file, not its module qn (#1892)

### LINKED ISSUE #1892: Scoped reingest drops an unchanged colliding module's RETURNS/ACCEPTS facts: the stale filter keys on module qn, not file
Found in passing on #1891 (Greptile flagged the same shape in the Parameter
mirror of this code; this is the pre-existing original).

## The mechanism

`GraphUpdater._reingest_delete` discards rehydrated type facts for the files
it is about to re-parse so the deferred pass does not emit both the old and the
new `RETURNS`/`ACCEPTS` edge (#1527):

```python
stale_modules = {
    base_module_qn(Path(key), self.project_name) for key in (*reparse, *gone)
} | {qn for qn, path in qn_to_path.items() if path in stale_paths}
pending = self.factory.definition_processor.pending_type_facts
pending[:] = [fact for fact in pending if fact.module_qn not in stale_modules]
```

The key is the **module qn**. Two distinct files can share one: `foo.py` and
`foo/__init__.py` both map to `proj.foo`. Re-ingesting `foo.py` therefore also
discards the queued facts of the **unchanged** `foo/__init__.py`.

## When it costs an edge

The discarded facts only matter when their edges were detached in the same
run — i.e. the reingest set also contains the file defining the annotated
type. `reingest([models.py, foo.py])` with `foo/__init__.py` holding
`def other(g: Gadget)`:

1. `models.py` re-parses → `Gadget` deleted and recreated → `foo/__init__.py`'s
   `ACCEPTS` into it is detached.
2. The prologue rehydrates `foo/__init__.py`'s facts (its file is not in the
   set), then the filter drops them because `proj.foo` is a stale module.
3. Nothing rebuilds the edge.

## Fix

`PendingTypeFact` has no file identity; add the owning relative path (the
definition rows already return `path`) and filter on `fact.path in {*reparse,
*gone}`. #1891 does exactly this for `PendingParameterType`, with a regression
(`test_scoped_reingest_keeps_a_colliding_modules_facts`) that can be adapted:
swap `OF_TYPE` for `ACCEPTS` and the Parameter node for the Function.

Not folded into #1891: the fact type is shared by every language's ingest and
the change is a separate contract.

vitali87:
## Summary

`GraphUpdater._reingest_delete` discards the rehydrated `RETURNS`/`ACCEPTS` facts of the files a scoped re-ingest is about to re-parse, so the deferred pass does not emit both the old and the new edge (#1527). It keyed that filter on the module qn, and two files can share one: `foo.py` and `foo/__init__.py` both derive `proj.foo`, so re-ingesting `foo.py` also discarded the facts of the unchanged `foo/__init__.py`.

- `PendingTypeFact` gains the owning file's relative `path`, filled by both parser sites (function and method, from the props that already carry it) and by the rehydration from the graph. The filter keys on `fact.path not in {*reparse, *gone}`, the same key the Parameter mirror (#1891) uses two lines below; a fact ingested without a file keeps the module-qn key rather than surviving every filter.

**What the measurement changed.** The issue says the discarded fact costs the unchanged file its `ACCEPTS` edge when the type's file is in the same re-ingest. Measured on the issue's scenario before fixing: the fact is discarded exactly as described, but the edge is not lost on `main`, because `RETURNS`/`ACCEPTS` are in the inbound-edge capture since #1527 and `_reingest_resolve` restores them after the re-parse. With the restore switched off the edge does go missing under the old filter and stays under the new one. So this is the second, independent path back to the edge and consistency with #1891, not a fix for a lost edge; the issue carries the table. An edge-level regression with the restore active passes on both filters and would prove nothing, so the one here disables the restore.

## Type of Change

- [x] Bug fix

## Related Issues

Closes #1892.

## Test Plan

- [x] `codebase_rag/tests/test_reingest_colliding_module_type_facts.py`: the unchanged colliding file's fact survives the filter and its edge is rebuilt from the fact with the restore disabled (function and method owners); the re-parsed sibling's stale fact is dropped; a fact without a file falls back to the module key; every producer (both parser sites, captured at the emitter since the queue empties on emit, and the graph rehydration) records the owning path.
- [x] Five mutations, each reddening only its tests: filter keyed on module qn again, no fallback for a pathless fact, requeue drops the path, function site drops the path, method site drops the path.
- [x] Related files green (71 passed): parameter nodes end-to-end, type facts, reingest hydration / empty-file calls / package flip, incremental implements edge.
- [x] The pre-commit `ty` hook is red at this machine's baseline (13 diagnostics identical to `main`, none in a touched file) and was skipped for the commit; every other hook ran.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scoped re-ingestion when a module file and package initializer share the same module name.
  * Preserved type facts, including return and accepted-type information, for unchanged files during re-ingestion.
  * Improved handling of type facts without an associated file path by falling back to module-based matching.


vitali87:
@greptileai review

vitali87:
@greptileai review

Issue #1908, opened 2026-09-13
change_signature: support definitions in languages other than Python


vitali87:
`change_signature` (#1533) reads the definition's parameter list and every call site's argument list with the Python grammar, and refuses a definition in any other language:

```
Cannot change the signature of myproj.lib.tool.helper: lib/tool.js is not Python, and only Python definitions are supported for now
```

The graph side is language-neutral already: `CALLS` edges carry `line`/`col`/`end_line`/`end_col`/`arg_count`/`kwarg_names` for every frontend, `hierarchy()` and `call_node_at()` in `codebase_rag/editing/sites.py` are shared with `rename`, and the postcondition contract measures the delta the same way. What is Python-specific is the reading and rendering of headers and argument lists:

- `_param_spec` walks tree-sitter-python's `parameters` children (`identifier`, `typed_parameter`, `default_parameter`, `typed_default_parameter`) and refuses splats and separators;
- `_bind_arguments` walks an `argument_list` (`keyword_argument`, `list_splat`, `dictionary_splat`);
- new parameter text is validated with `ast.parse`, and `=literal` values are checked against the annotation with `ast.literal_eval`.

Extending it means one header reader and one argument binder per language, behind the same `ParamSpec`/binding types, with the language chosen from the definition's path. Languages without keyword arguments (JS/TS, Go, Rust, C, C++, Java, C#) simplify the binder to positionals only; languages with named arguments (Kotlin, Swift, Scala, Ruby keyword args, PHP 8) need the keyword path too.

Per the rule that the same syntax can mean different things across grammars, each language needs its own red test on a real fixture before the reader is ported.

PR #1912, opened 2026-09-13
feat: rebuild the change_signature edit operation on main (#1533)

### LINKED ISSUE #1533: [Epic: Agent IDE] T11: Edit algebra op 2: `change_signature(qn, new_params, arg_mapping)`

vitali87:
Closes #1533. Replaces #1548, which is closed once this merges: that branch sat 431 commits behind main with unresolved review threads, so the operation is rebuilt from scratch on today's `rename`, contract and transaction code rather than rebased.

## What it does

`change_signature(qn, new_params, mapping)` rewrites a Python function's or method's parameter list and every call site the graph knows, on the definition and on every override in both directions, in one edit transaction.

- `new_params` lists the new parameters in order as they would be written in a header (`n: int`, `limit: int = 10`); a bare old name carries that parameter over with its annotation and default.
- `mapping` says where each new parameter's value comes from at a call site: an old parameter's name, its 0-based index, or `=literal` to insert that text at every site. A new parameter absent from the mapping takes the old parameter of the same name.
- Sites are bound as Python would bind them and re-rendered in the new order. Values stay positional until one goes by keyword or a defaulted parameter is left out; from there every later value is spelled by keyword. Keyword values keep their original relative order under their new names, so renaming by index renames keyword callers too.
- An old parameter fed to a new one of another name is renamed, and the body follows: bare references become the new name (attributes and keyword-argument names that share the spelling are left alone, a comprehension that only reads it follows). The rename refuses when the old name is used inside a nested function, lambda or class, or re-bound by `global`, `nonlocal` or an import, and when the new name is already read in the body. Without this the first smoke test failed with `NameError` on the renamed definition.
- A recursive call, or a call nested in another call's arguments, is rewritten inside out: the inner rewrite is folded into the enclosing value before the enclosing site is rendered, so `helper(helper(1))` becomes `helper(helper(1, 1), 1)` as one edit per site rather than two that overlap.
- Spelling a kept parameter out again (`b: int = 0` for an existing `b`) re-annotates it; only the bare name carries the old spelling over. A `@staticmethod` has no receiver; a method whose first parameter is not `self` or `cls` refuses rather than remapping it as a value.
- A site the mapping cannot complete (missing required value, splat, surplus argument, undeclared keyword, generator argument) or that the graph resolved by guesswork is left as written and listed as `unmapped` with a reason; `allow_heuristic` rewrites the guessed ones.
- With a re-ingest the applied change is measured through the structural delta and undone when the contract fails: every site of the changed signature must read as mapped or be in the unmapped list, so a caller the graph never saw is caught after the fact.

Refusals, with nothing written: unknown definition, non-Python definition, `*args`/`**kwargs`/`/`/`*` in the header, overrides whose parameter names differ (names the member), duplicate new names, a required parameter after a defaulted one, a mapping naming an unknown new or old parameter or feeding two new parameters from one old one, and a literal that cannot be a value of the declared type (`='no'` for `int`; `=None` fits `int | None` and `Optional[int]`; a non-literal is not checked).

## Surface

- `codebase_rag/editing/signature.py`: the operation.
- `codebase_rag/editing/sites.py`: `AMBIGUOUS`, `hierarchy()` and `call_node_at()`, moved out of `rename.py` so both operations share one answer to "which call does this `(line, col)` name".
- `change_signature_expectation()` gains `heuristic_allowed`, the leave `rename_expectation` already carried; without it `allow_heuristic` rewrote the guessed site and the contract then undid the whole change.
- MCP tool `change_signature` and CLI `cgr change-signature QN PARAM... --map NEW=SOURCE`.
- `docs/architecture/change-signature.md`.

Python only. Other languages are #1908; the refusal message points there.

## Tests

`codebase_rag/tests/test_change_signature.py` runs every case against a real index of a fixture repo through the in-memory ingestor, applies the change to disk, and executes the rewritten fixture with a subprocess so the call sites are shown to agree with the new definition, not just to parse. Covers: add with literal, reorder, rename by index, keyword spelling after an omitted default, every unmapped reason, every refusal, a method hierarchy, dry run, undo history, a contract failure rolling back, and the MCP tool (schema, payload, refusal as a payload).

## Summary by CodeRabbit

- **New Features**
  - Added a `change-signature` operation for updating Python function and method parameters.
  - Rewrites supported definitions, overrides, bodies, and known call sites.
  - Supports parameter mappings, dry runs, heuristic resolution, unmapped-site reporting, validation, rollback, and edit tracking.
  - Available through both the CLI and MCP tools.

- **Documentation**
  - Added usage guidance covering parameter rules, constraints, call-site handling, and examples.


vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Superseded by a `pr-split` replacement stack on `main` (`--max-loc 500 --stack`):

| # | layer | diff |
|---|---|---|
| #2163 | contract expectation, shared call-site helpers (`sites.py`), constants | +286/-71 |
| #2161 | `signature_spec` + `signature_contract` | +448/-0 |
| #2162 | `signature_bind` | +491/-0 |
| #2166 | `signature` (the changer) | +479/-0 |
| #2164 | CLI + MCP tool + docs | +412/-1 |
| #2169 | test support + acceptance, transaction and MCP tests | +425/-0 |
| #2168 | mapping tests | +358/-0 |
| #2165 | definition and renamed-parameter tests | +365/-0 |
| #2167 | hierarchy, receiver and default-guard tests | +471/-0 |

Linear native stack #2163 → #2161 → #2162 → #2166 → #2164 → #2169 → #2168 → #2165 → #2167. The top has the same diff as the restructured branch against the merge-base and merges cleanly into `main`.


It's built from this PR rebased onto `main` (fix-pr-issues' `c0f23007`), plus one restructuring commit. pr-split can't split a single new file, and two of them were over the limit:

- `editing/signature.py` (1322 lines) → `signature_spec` (specs, mapping sources, result types), `signature_bind` (header reading, argument binding, rendering), `signature_contract` (the postcondition check: `SignatureChanger._enforce_contract` lifted out unchanged as `enforce_contract`) and `signature` (the changer). `cli.py` and `mcp/tools.py` import `parse_mapping`, `sites_for`, `unmapped_for` and `SignatureRefused` from `signature_spec`.
- `tests/test_change_signature.py` (1521 lines) → split by section into four test files over a shared `change_signature_support.py`.
- The call-site test helpers pick edges from `edge_props`. `main` now keys edge properties by site, so the old five-element key failed with KeyError on the rebased branch.

Local checks: on the full restructured head, the four signature test files plus edit-contract, rename, MCP parity, CLI smoke and readme tests pass (221 passed, 1 skipped). Layer 1 was also checked on its own (edit-contract, rename, MCP parity, CLI smoke: 139 passed, 1 skipped), and layers 2-4 add modules that nothing imports until layer 5.

This PR's description says it replaces #1548. The earlier split of #1548 (#2041-#2050, based on `feat/postcondition-contract`) is therefore likely superseded too; whether to close it is the owner's call.

Issue #1896, opened 2026-09-13
A tuple-unpacked local from an annotated return is never typed, so its member calls resolve by bare name


vitali87:
Found in passing by the local review of #1893's fix (tuple-unpacking is the
real shape of `codebase_rag/trace/sourcemap.py:225`, which that fix does not
reach).

## The mechanism

```python
def parsed(self) -> tuple[int, int, SourceMap] | None: ...

_ol, _oc, inner = parsed          # `parsed` typed from the return annotation
return (inner.base_dir / source).resolve()
```

`parsed` gets its type from the call's return annotation, but the unpacking
assignment (`pattern_list` / `tuple_pattern` on the left) is never walked, so
`inner` has no entry in the local type map, `inner.base_dir` has none either,
and the receiver falls to the bare-name trie — `resolve` matches every class
defining it (five in `trace/resolution.py`).

Reproduced minimally: the same receiver with `inner: SourceMap` annotated by
hand emits no false edge; tuple-unpacked, it emits
`app.from_sections -> engine.Engine.resolve`.

## Fix shape

In the variable analyzer's assignment walk, when the left is a `pattern_list`
/ `tuple_pattern` and the right's type is `tuple[A, B, C]` (annotation text,
Optional stripped), bind each target name to the corresponding element type;
`list[...]`/`Sequence[...]` element types are already recovered for loop
variables, so the `tuple[...]` split is the missing case. `#1893` pins the
shape as a strict xfail (`test_a_tuple_unpacked_receiver_is_still_matched_by_name`)
so it flips when this lands.

PR #1919, opened 2026-09-14
fix: unpack a tuple-annotated name into typed locals (#1900)

### LINKED ISSUE #1900: A typed parameter annotated tuple[...] unpacked into locals binds nothing
Found in passing by the local review of #1896's fix.

`#1896` types `_n, w = f()` from `f`'s `tuple[...]` return annotation, and
`p = f(); _n, w = p` by tracing `p` to its defining call. It does NOT cover a
right-hand side whose tuple type comes from an annotation on the name itself:

```python
def use(p: tuple[int, Widget]) -> int:
    _n, w = p
    return w.render()        # w untyped -> bare-name fallback -> Banner.render too
```

`p` IS typed (`local_var_types["p"] == "tuple[int, Widget]"`), but the
unpacking pass only follows an identifier back to a call. Extending
`_defining_call` / `_tuple_return_elements` to also accept a stored
`tuple[...]` type string for the identifier (split with the same
`_split_top_level`) closes it; the regression is the shape above with a decoy
class defining `render`, asserting only `Widget.render`.

vitali87:
Closes #1900

## What was wrong

`def use(p: tuple[int, Widget]): _n, w = p` bound nothing. #1896's unpacking pass traces the right-hand name back to a defining CALL and reads that callee's return annotation; here there is no call, the tuple type sits on the name itself, as the annotation text the typed-parameter pass stored in the local type map. The same gap covered an annotated local whose value the engine cannot read, `q: tuple[int, Banner] = opaque()`, and a bare declaration `q: tuple[int, Banner]`, which were not typed at all.

## The fix (`codebase_rag/parsers/py/ast_analyzer.py`)

* A new `_process_assignment_annotation` pass runs after both value passes: an annotated name the value left untyped takes its annotation text, quotes stripped. It is a last resort, not a pre-emption. Inference from the value keeps priority because it resolves further than the annotation's text does: `x: Base = Derived()` stays `Derived`, and `items: List[Widget] = load()` keeps the `list[Widget]` a loop can unwrap (the first cut of this branch stored the annotation first and regressed both; the local review caught it against the base tree). A declaration with no value has no right-hand side at all and is typed here alone.
* A new `_unpacked_elements` helper feeds the unpacking pass: the defining call's return annotation first and, when that yields nothing, the name's stored text through the same top-level splitter (`_tuple_elements_from_text`, `Optional[...]` / `| None` stripped). The stored text has no owner, so `Self` inside it keeps its text rather than resolving the module qn as if it were a method's.

A stored text that is not a tuple (`list[Widget]`, `Widget`), a union of two tuples, or a count mismatch binds nothing. Where a name is reassigned from a call before the unpacking, the nearest call wins, consistent with #1896's nearest-binding rule.

## Tests (`codebase_rag/tests/test_tuple_unpacked_locals.py`)

Seven tests reading the engine's local type map (a name left unbound falls to the bare-name fallback, which can pick the right class by coincidence, so an edge cannot tell binding from luck): a tuple-typed parameter; an annotated local from a tuple literal; one from an unreadable call; a bare declaration; value-over-annotation precedence with a quoted forward reference; a `List[Widget]` local whose loop variable must stay `Widget` (the regression); and `Self` inside a tuple on a name in a three-part module.

Mutations, one per rule, each restored from HEAD afterwards:

| removed | red |
|---|---|
| annotation overwrites the value | precedence, container loop |
| quote strip | precedence |
| declaration without a value skipped | declaration |
| stored text only when there is no call | unreadable call, declaration |
| module qn as `Self`'s owner | `Self` |
| the annotation pass | precedence, declaration, tuple literal, unreadable call |

## Builds on #1905

The splitter and the unpacking pass this extends landed with #1905 (issue #1896); this branch is one commit on top of that head.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type detection for annotated Python names, including optional, quoted, forward-reference, and `Self` annotations.
  * Restricted annotation-based inference to the correct scope and read position.
  * Preserved earlier inferred types while correctly handling later reassignments.
  * Tuple-unpacked variables are now typed correctly from annotated parameters or locals, including values without assignments.
  * Preserved explicit value types when they take precedence over annotations.
  * Improved loop-variable typing for annotated container values.

* **Tests**
  * Added coverage for scope handling, reassignment, tuple unpacking, annotations, and unsupported types.


vitali87:
@greptileai review

vitali87:
Both findings reproduced and fixed in 188198b4 (also rebased onto current `main`). The second one's cause is different from the one stated, and the difference matters, so the detail is below.

## 1. Stale annotation after an uninferable reassignment — fixed

Confirmed exactly as described. `p: tuple[int, Banner] = opaque(0); p = opaque(1); _n, b = p` bound `b` to `Banner` from an annotation describing a value the name no longer held.

Fixed with the clearing rule the unpacked-call path already obeys: a later binding of the same name in the caller's own scope drops the annotation (`_rebound_after`). A bare `q: T` declaration is exempt, because it has no value of its own and the assignment that follows is the value it was declared for — without that exemption `test_a_declared_but_unassigned_name_carries_its_annotation` goes red.

Tests: `test_a_reassignment_the_engine_cannot_read_drops_the_annotation`, plus `test_an_unreassigned_annotation_still_binds` as the control, so the fix cannot be achieved by never reading annotations at all.

## 2. Nested annotations leaking into the outer scope — real symptom, different cause

The symptom reproduces. The stated cause does not: it is not `_process_assignment_annotation` that records the nested name.

Instrumenting each pass on the exact repro shows the entry is written by **`_process_assignment_simple`**, before the annotation pass runs. And the leak reproduces with **no annotation anywhere**:

```python
def outer(v) -> int:
    def inner() -> int:
        v = Banner()          # plain assignment, no annotation
        return v.render()
    return v.render()
```

| tree | local type map for `outer` |
|---|---|
| `origin/main` | `{'v': 'Banner'}` |
| this branch | `{'v': 'Banner'}` |

So it is pre-existing on `main` and not a regression from this PR. Filed as #1922 with the reproduction.

What this PR does do is apply the scope filter to its own pass, so the annotation pass cannot add to the problem: `_scope_of(assignment) != caller.id` now skips it, matching `_process_assignment_unpacking`.

One consequence worth flagging for the test: asserting `"v" not in types` after a full build would **fail on `main` too**, so it would be measuring #1922 rather than this change. `test_the_annotation_pass_reads_only_the_analysed_scope` therefore drives the pass directly with an empty map and asserts it records nothing — a test that can only fail if this pass's scope rule is wrong.

## Verification

Each of the three new rules was mutated separately; every one reddens exactly one distinct test and nothing else:

| rule removed | red |
|---|---|
| the clearing rule | `test_a_reassignment_the_engine_cannot_read_drops_the_annotation` |
| the bare-declaration exemption | `test_a_declared_but_unassigned_name_carries_its_annotation` |
| the scope filter | `test_the_annotation_pass_reads_only_the_analysed_scope` |

64 tests pass in the focused file. The two SonarCloud composite assertions (python:S9073) that failed the gate are split; checked by AST rather than grep, since a line-anchored pattern cannot see a wrapped composite.

vitali87:
@greptileai review

vitali87:
Fixed in d0ac5c94, and the fix restructures the rule rather than patching it — the finding was correct that the previous round's clearing rule was wrong, though the mechanism is worth stating precisely because it changes where the test belongs.

## What was actually wrong

The clearing rule I added last round lived in the annotation pass and was **position-blind**: it dropped the map entry whenever the name was rebound anywhere later in the function, including after the unpack that consumed it.

```python
def use() -> int:
    p: tuple[int, Banner] = opaque(0)
    _n, b = p
    r = b.render()        # b IS Banner here
    p = opaque(1)         # ...but this cleared the entry retroactively
    return r
```

The local type map is flat per function, so it cannot express "Banner until line 5, unknown after". Clearing the entry to protect the later state untypes the earlier use, which is the regression reported.

## The fix

Supersession moved out of the pass that writes the map and into `_unpacked_elements`, which reads it — where `_defining_call` already tests bindings before the site it is resolving. The predicate is now nearest-binding: the binding closest before this unpack is what the name holds here, and the annotation applies unless that binding replaced it.

The annotation completes exactly two binding shapes, which is what the existing tests pin:
- the annotated statement itself, `q: tuple[int, Banner] = untyped()`, where the call's return says nothing and the annotation is the only thing that can type it;
- the assignment a bare `q: T` declaration was made for.

Any other nearest binding (`p = opaque()`, `p = supplied`, `p += x`, `for p in xs`, `with cm as p`) is a value the annotation never described.

## A second defect the same change closes

Moving the test to the read point also fixed the gap my local reviewer found independently: a **typed parameter** rebound in the body kept unpacking its stale annotation, while the annotated-local twin was correctly cleared. A parameter's type is seeded into the map before any assignment pass runs, so it has no annotated assignment node for a write-time rule to reach. At the read point both paths obey one predicate.

| shape | before | after |
|---|---|---|
| `def use(p: tuple[int, Banner])` then `p = opaque()` | `b` typed `Banner` | `b` unbound |
| `q: tuple[int, Banner] = opaque(0)` then `q = opaque(1)` | `b` unbound | `b` unbound |
| unpack, then rebind further down | `b` unbound (regression) | `b` typed `Banner` |

## Verification

66 tests pass. Five mutations, each reddening exactly the tests its rule protects:

| rule removed | red |
|---|---|
| the supersession test | rebound parameter; reassignment |
| nearest binding → earliest binding | reassignment |
| the annotated statement counts as supersession | ordering; unreassigned; tuple-typed local; untyped-call local |
| the bare-declaration exemption | declared-but-unassigned |
| the scope filter | annotation pass scope |

Two new tests: `test_a_rebound_typed_parameter_does_not_unpack_its_annotation` and `test_an_unpack_before_a_later_rebinding_still_reads_the_annotation`, the second being the regression this finding reported.

On the summary's note that the previous two findings were "resolved without explanation": they were answered in [this comment](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666409810) before being resolved — the first fixed, the second shown to reproduce identically on `origin/main` with no annotation involved and filed as #1922.

vitali87:
@greptileai review

vitali87:
Both fixed in 083c2803.

## The bare declaration covered too much

Correct and reproduced. `_declared_before` matched any earlier bare declaration, so after `q: tuple[int, Banner]; q = opaque(0); q = opaque(1)` the second assignment was still treated as the one the declaration was made for, and the stale tuple shape unpacked into a name that no longer held it.

A declaration binds no value, so it covers the **first** assignment after it and nothing later. Everything after that is an ordinary rebinding and supersedes the annotation like any other.

| shape | before | after |
|---|---|---|
| `q: T` then one `q = untyped()` | `b` typed `Banner` | `b` typed `Banner` (unchanged, and required) |
| `q: T` then `q = opaque(0)` then `q = opaque(1)` | `b` typed `Banner` | `b` unbound |

`test_a_declaration_covers_only_the_assignment_it_was_made_for` pins it. Mutating the rule back to "any assignment after the declaration" reddens exactly that test and nothing else.

## Helper naming

Renamed the two helpers this branch added, to match the neighbouring `_render_targets` / `_local_types` / `_functions` rather than sitting terser than the file's convention:

- `_walk` → `_nodes_under`
- `_engine` → `_type_inference_engine`, which also gained its missing `-> PythonTypeInferenceEngine` return type.

67 tests pass.

## On the resolved threads

The summary notes three threads resolved "without explanation". Each was answered on the PR before being resolved, and the reasoning is in those comments rather than in the thread resolution itself:

- stale tuple types after a reassignment — [fixed here](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666409810), with the clearing rule;
- nested annotations escaping scope — [same comment](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666409810): the symptom reproduces identically on `origin/main` with no annotation involved, so it is pre-existing and filed as #1922, where a partial fix and the remaining leak paths are now written up;
- retaining types until rebinding — [restructured](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666957950) so the position test happens at the unpack site rather than in the pass that writes the map, because that map is flat per function and cannot hold "Banner until line 5, unknown after".

vitali87:
@greptileai review

vitali87:
Fixed in 983b4243, and the finding was right about the symptom while pointing at the wrong rule, which is worth spelling out.

## What the repeated declaration actually broke

Reproduced:

```python
def use() -> int:
    q: tuple[int, Widget]
    q = opaque(0)
    q: tuple[int, Banner]
    q = two()
    _n, b = q
    return b.render()
```

The map held `q = tuple[int, Widget]` — the FIRST declaration, after a second one had superseded it. `b` came out as `Banner` only because `two()` resolves on its own, so the stale entry was not visible in that one edge.

The cause is not in `_declared_before`, which the finding names. It is the annotation pass: its `var_name in local_var_types` guard makes the first annotation of a name permanent, so the second declaration was skipped entirely. Among several annotations of one name the LAST is the one describing the value bound after it, so the pass now walks in document order and may replace its own earlier entry, tracked by a `written` set so an inferred type from the value passes still wins over both.

`_declared_before` did need the related change — nearest declaration before the binding rather than earliest — and that one has its own discriminating case, below.

## Verification, including one gap I am not going to paper over

69 tests pass. Mutations:

| rule removed | red |
|---|---|
| last annotation wins (restore the `in local_var_types` guard) | repeated-declaration test |
| nearest declaration (`max` → `min`) | second-declaration-covers-its-own-assignment test |
| document order (`sorted` → capture order) | **nothing** |

The third is inert and I am reporting it as inert rather than as a pass. The tree-sitter captures already arrive in document order for every fixture I could build, so sorting them is currently belt-and-braces: correct, cheap, and load-bearing only if that order ever changes. I could not construct an input where it matters.

The `max` → `min` mutation was also inert at first. The annotation-pass fix alone made the repeated-declaration test pass, so the nearest-declaration rule was unproven until I found the case that separates them: a second declaration followed by a call the engine cannot read, where only that declaration can type the unpack. That is `test_a_second_declaration_covers_its_own_assignment`, and it goes red under `min`.

## On the resolved threads

The summary again lists threads resolved "without explanation". Each was answered in a PR comment before being resolved — [round two](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666409810), [round three](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5666957950), [round four](https://github.com/vitali87/code-graph-rag/pull/1919#issuecomment-5667213392) — including the one finding declined with evidence, which reproduces identically on `origin/main` and is now PR #1928 against issue #1922.

vitali87:
@greptileai review

PR #1924, opened 2026-09-14
refactor: split the C# field-map and member-name helpers (#1669)

vitali87:
## Summary

Two functions in `codebase_rag/parsers/csharp/utils.py` carry SonarCloud `python:S3776` findings on `main` (`build_field_type_map` and `synthesize_method_name`). Both are split into named helpers with no behaviour change.

- `build_field_type_map` now reads a property through `_property_field` (name and type, or nothing when either is absent) and a field declaration through `_declared_fields` (every declarator of one declaration sharing its type). Last-wins assignment is unchanged, and so is insertion order.
- `synthesize_method_name` now delegates both operator forms to `_operator_name`, which picks the `operator` field for a binary or unary operator and the `type` field for a conversion operator, then applies one shared guard. Destructor prefixing and the reserved-keyword drop are untouched.

Public signatures are unchanged. The callers are `language_spec._csharp_get_name`, `parsers/utils.py` and `class_ingest/mixin.py`.

The file's third flagged function is not touched: #1669 tracks these as per-function follow-ups, and mixing them would make the behaviour-preservation argument harder to check.

## Type of Change

- [x] Refactor (no functional change)

## Related Issues

Part of #1669.

## Test Plan

- [x] New `codebase_rag/tests/test_csharp_utils_units.py`: 21 tests calling the helpers directly on parsed nodes, one per branch, written and run against the UNREFACTORED code first so they pin current behaviour rather than the new shape. Covers a class with no body, an empty body, a property, a single field, several declarators of one field, an initialised field, properties and fields together, a method (not a field), a repeated name, generic and nullable type annotation, a plain method, binary and conversion operators, a destructor, a constructor, a reserved-keyword name with its non-reserved control, a nameless node, and the signature pairing for a method and an operator.
- [x] 43 tests green (the 21 new plus the existing C# definition and property-read suites).
- [x] 14 mutations, each reddening exactly its own tests: a missing body returning content; property and field types left unannotated; a partial property pair kept; only the first declarator mapped; the declarator filter dropped; a missing variable declaration ignored; field and property declarations skipped; a conversion operator reading the operator field; the operator prefix dropped; operators falling through to the plain name; the destructor prefix dropped; a reserved keyword kept.
- [x] Three of those 14 are DOCUMENTED no-ops, recorded as such in the harness so their greenness is never read as coverage: the grammar guarantees a property carries both name and type, a field declaration carries exactly one variable declaration, and only `variable_declarator` children carry a `name` field, so those guards cannot be observed failing. The other eleven all redden.
- [x] The local reviewer ran the old and new code differentially over 47,001 nodes from the repo's own C# fixtures plus 223 hand-built edge-case sources, with zero differences in values and insertion order, and validated the harness with three planted defects first.
- [x] Confirmed that `tree_sitter.Node` defines neither `__bool__` nor `__len__`, so replacing `node and node.text` with `node is not None and node.text` cannot change behaviour for a childless node.
- [x] The pre-commit `ty` hook is red at this machine's baseline (identical on `main`, nothing in a touched file) and was skipped for the commit; every other hook ran. `ty check` on the changed module passes.

## Note

The C# tree-sitter grammar lives in an optional dependency group. A worktree synced without it makes these tests skip rather than fail, which reads as a pass locally; it must be installed to exercise them.

## Summary by CodeRabbit

* **Refactor**
  * Improved consistency when analyzing C# properties, fields, operators, constructors, destructors, and methods.
  * Preserved existing handling for multiple field declarations, initializers, generic and nullable types, and reserved keywords.

* **Tests**
  * Expanded coverage for C# member mapping, name generation, and method signature extraction across a broader range of syntax patterns.


vitali87:
@greptileai review

Issue #1911, opened 2026-09-13
evals: _StatefulIngestor keeps one CALLS site per caller/callee pair


vitali87:
`evals/cgr_graph._StatefulIngestor` keys a relationship on its endpoints only:

```python
edge = (str(from_label), from_val, str(rel_type), str(to_label), to_val)
self.edges.add(edge)
...
self.edge_props[edge] = dict(properties)
```

The production store keys `CALLS`, `REFERENCES` and `INSTANTIATES` on `(line, col)` as well (`constants/graph.py`, the relationship key table), so a caller that reaches the same callee twice holds two sites there and one here. The last site written wins and the earlier one vanishes from every graph read the double answers, including `CYPHER_GRAPH_CALLERS`.

Consequences seen while building the `change_signature` operation (#1533):

- `def run(): return helper(helper(1))` yields one caller row, so a test of nested call sites has to wrap `fetch_all` and duplicate the row by hand.
- `def run(): return K.helper(2) + K().helper(3)` yields one site; the test had to split the two calls across two callers.
- Any eval that compares an incrementally mutated graph against a clean re-index cannot see a lost or gained second site between the same two nodes.

Proposal: key site-bearing relationships on the same tuple the production store uses, `(from, rel, to, line, col)` for the three site kinds and `(from, rel, to, line, col, alias)` for `IMPORTS`, and leave the endpoint-only key for the structural relationships. `_delta_sites`, `_graph_rows` and the subtree deletes already iterate `self._in` / `self._out`, so they need no change beyond the key type.

vitali87:
Fixed on main by #1915 (merge 82ba5f7e, commit 04584bc0 'key site-bearing relationships the way the store does'). The PR did not say 'Closes', so this stayed open.

PR #1926, opened 2026-09-14
feat: cgr check --isolated measures an edit without keeping the re-ingest (#1718)

### LINKED ISSUE #1718: cgr check: measure a base ref without mutating the shared graph
## Context

`cgr check --base <ref>` (PR #1546) re-ingests the files that differ between the base and the working tree and reports the structural delta. The re-ingest brings the shared project graph up to the working tree, so:

- a second run on an unchanged tree reports nothing, because the delta was already applied;
- a prior non-failing check therefore makes a later `--fail-on-found` check pass for the same edits;
- the command is only meaningful when the graph reflects the base before the run.

This is documented in `codebase_rag/structural_check.py` and is the intended contract for the first version.

## Proposal

Add an isolated mode: run the comparison against a scratch copy of the affected subgraph (or a temporary project name seeded from the base-state rows) and discard it afterwards, so `cgr check` never mutates the shared graph and is idempotent. The delta computation (`structural_delta.observe`) is already independent of which project it reads; the work is the scratch-graph plumbing and the restore.

Raised by Greptile on #1546 as a heavy-lift finding; split out so the first version can land with its documented contract.

vitali87:
Closes #1718

## What was wrong

`cgr check` measures the working tree by re-ingesting the files that differ from `--base`, and that re-ingest is also what brings the shared graph up to the tree. So a second run on the same edit reports nothing, and a check that ran once makes a later `--fail-on-found` pass for the same edits. The first version documented this as its contract; the issue asks for a mode that never mutates the shared graph and is idempotent.

## The fix

`cgr check --isolated` (`run_check(..., isolated=True)`) runs exactly the check it runs today, then puts back what the re-ingest replaced.

* **Scope is the updater's, not a guess from the diff.** `GraphUpdater.reingest` now publishes `reingest_scope` (changed, deleted, dependents, same-stem survivors) right before its `before_write` hook, while nothing has been written. The guard captures inside that hook.
* **Capture** (`codebase_rag/check_isolation.py`): the module subtrees at those paths, walked with the same relations `CYPHER_DELETE_MODULE` walks so the capture equals the delete; the File nodes at those paths and the Folder/Package nodes of every ancestor directory (a package indicator appearing or vanishing flips the container above the file); every relationship touching any of it, either direction, with both ends addressed by the key the batch writer uses; and the properties of far ends the check can prune or re-grade (ExternalModule, Resource, Gloss).
* **Restore** is the inverse of the re-ingest's writes: findings at the scope's paths the capture did not see are deleted, the subtrees are deleted again, File and container nodes the graph did not have before are deleted, orphaned ExternalModules and unanchored Resources are pruned, and the captured nodes and edges are re-emitted through the batch API and flushed. The hash cache is put back byte for byte with its timestamps. A re-ingest that fails after its first write is restored the same way before the error propagates; one that aborts in its prologue captured nothing and restores nothing.

Rejected shapes: a scratch project clone (cost proportional to the graph, qualified names embedded in properties to rewrite, and `delete-project` reaches the File nodes the two projects share) and a reverse re-ingest from the base contents (reads only from disk, re-resolves callers, so the result is a fresh graph rather than the one that was there).

## Tests

`codebase_rag/tests/test_check_isolation.py` on the eval store, whose module delete and capture queries are modelled by value: after an edit of every kind the re-ingest handles differently (a rename whose callers dangle, a deleted module, a new module in a new directory with a new import, a flipped package indicator) the store's nodes, edges and edge properties are identical to before; the isolated delta equals the applied delta; two isolated runs report the same delta while two applied runs do not; the hash cache is byte- and mtime-identical; a flipped container and a new directory are put back and dropped; a CodeSmell and an ExternalModule the re-ingest creates beside the subtree do not outlive the restore (findings enabled explicitly, with the mid-run known positive); a failure in the re-parse still restores; the updater names its scope before the hook. The applied mode keeps its controls.

`codebase_rag/tests/integration/test_check_isolation_e2e.py` replays the edit on a real Memgraph so the Cypher itself is exercised (the variable-length walk on both kinds of scope node, `startNode` for the direction, `properties()` of both ends, the label-union delete), with a gloss whose anchor the re-parse re-grades to STALE and the restore must return to EXACT.

Mutations, each restored from HEAD afterwards; every one reddens exactly the tests its rule protects:

| rule removed | red |
|---|---|
| the restore itself | every state-equality test |
| re-emitting captured edges | state equality, flipped container, findings, failure path |
| the hash-cache restore | hash cache |
| publishing `reingest_scope` | every state test, plus the scope contract |
| deleting path nodes the capture did not see | state equality, flipped container, new directory |
| restoring only on success | the failure path |
| deleting uncaptured findings | findings |
| pruning orphaned ExternalModules | state equality, findings |
| the edge direction | every state test |
| the site properties in the dedup key | the per-site test |
| ancestor directories in the capture | state equality, flipped container |
| discarding the store's queued writes | the buffered-failure test |
| the IO refusal | the refusal test |
| the finding cleanup's project scope | the cross-project test |

The per-site mutation was inert at first: the fixture could not hold two parallel edges, because the eval store collapses them (issue #1911, fixed by #1915). Replaced with a test that drives the production query shape directly. The project-scope mutation is applied to the store's modelling of the query rather than the Cypher text, since that store dispatches on query identity and an edited string silently matches nothing; the real Cypher is covered by the integration test.

## Limits, and what review changed

Three findings from the local review round, all fixed:

* **A failed re-ingest's queued writes.** The batch API buffers until `flush_all`, so a run that raised after its first write left its partial parse queued, and the restore's own flush committed it right after the deletes meant to remove it. The restore now discards the store's pending writes before its first delete. The existing failure test could not see this: the eval store writes batches through immediately and its `flush_all` is a no-op, so the test passed for the wrong reason. It is now driven through a double with production's buffering semantics, with a known-positive that the double really does commit on flush.
* **The finding cleanup was not project-scoped.** `$paths` are repo-relative and two projects in a shared graph can hold the same one, exactly the reason the module delete beside it carries a project test. Added, with tests at both tiers.
* **IO resource links cannot be restored.** The capture walks out from the re-parsed modules and never reaches a `Resource`, while the endpoint pass deletes every network `RESOLVES_TO` edge graph-wide and rebuilds from the edited tree; a Resource-to-Resource `FLOWS_TO` chain touches no scoped node at all. Rather than lose them silently, `--isolated` now refuses when the capture enables those relationships and says how to proceed. Widening the capture to the IO neighbourhood is the larger alternative and is not attempted here.

A documented merge limit was also widened: `ensure_node_batch` merges properties rather than replacing them, so a property the check ADDS to any surviving node is not removed. That covers the File, Folder and Package nodes at the scope's paths as well as the far-end nodes the docstring originally named. I could not construct an input where the re-parse actually adds a key to one, so this is stated as a limit rather than a demonstrated bug.

## Not covered here

The MCP write tools (`rename`, `reingest`) keep applying their edits; the guard is built so they could adopt it, but this PR wires only the CLI check.

## Summary by CodeRabbit

- **New Features**
  - Added `cgr check --isolated` to measure structural changes without retaining re-ingest changes.
  - Isolated checks restore the graph and hash cache, enabling consistent repeated runs.
  - Glosses can be re-anchored after symbol changes and identified as moved, ambiguous, or lost.
  - Added safeguards and guidance for unsupported I/O relationship capture.

- **Documentation**
  - Added help text and architecture documentation covering isolated checks, restoration, and rollback behavior.

- **Tests**
  - Added coverage for restoration, findings, failed re-ingests, repeated checks, and integration scenarios.


vitali87:
@greptileai review

vitali87:
All four fixed in cb5804c6 (rebased onto current `main`). Every one reproduced; the second needed a wider fix than the finding proposed, and the detail matters.

## 1 and 4. The validated capture was not the one the run used

Correct. `run_check` validated the caller's selection and then built `GraphUpdater` without it, so the updater fell back to the configured default, which can enable the very IO links `_refuse_unrestorable_capture` had just refused. `capture=capture` now reaches the updater.

`test_the_validated_capture_is_the_one_the_run_uses` records what the updater is actually constructed with. Removing the forward reddens it.

## 5. An unreadable cache was deleted

Correct, and the sharper way to put it is that absent and unreadable shared one representation. `_FileSnapshot` caught every `OSError` and set both fields to `None`, which `put_back` reads as "there was no cache" and unlinks. A permission error on a cache that exists therefore deleted it.

Now `FileNotFoundError` marks confirmed absence, any other `OSError` leaves the file alone, and only confirmed absence licenses the unlink. Two tests, including the control that a genuinely missing cache is still removed, so the distinction is not achieved by never unlinking.

## 3. Findings kept their rewritten properties

Correct. A finding is keyed on file, line, column and rule, so an edit that moves code without moving the match keeps the qualified name; the cleanup spares it as captured, and its properties were never captured because the far-props CASE listed only ExternalModule, Resource and Gloss. The finding labels join that list, and the eval store's modelling with them.

## 2. The global sweeps — fixed, and wider than reported

The finding is right that `restore()` ran two repo-wide sweeps that could collect a pre-existing orphan anywhere in the graph. Those are gone, replaced by a set computed before the first delete: the shared nodes reachable from the scope now, minus the ones the capture saw. That set is exactly this check's own work.

Removing my sweeps was necessary but not sufficient, which the suggested fix would have missed. `reingest` runs `CYPHER_DELETE_ORPHAN_EXTERNAL_MODULES` **itself**, at `graph_updater.py:5442`, before the restore ever runs. So an unrelated orphan was still destroyed with my sweeps removed. The guard now also captures orphaned shared nodes up front and re-emits them, so the restore can undo a sweep it does not control.

`test_an_unrelated_orphan_survives_an_isolated_check` covers it, and it was red until the capture was added, not merely until the sweeps were removed.

## Verification

23 tests pass. Five mutations, each reddening exactly the test its rule protects:

| rule removed | red |
|---|---|
| forwarding the capture | validated-capture test |
| unreadable treated as absent | unreadable-cache test |
| re-emitting captured orphans | unrelated-orphan test |
| finding labels in the far-props CASE | none — see below |
| finding labels in the store's modelling | finding-properties test |

The fourth is an honest gap rather than a pass: the eval store dispatches on query IDENTITY, so editing the Cypher text makes it match no case and silently do nothing. Only the store's modelling of that query is reachable from the unit tier, which is what the fifth row mutates. The production Cypher is covered by the integration test, which needs a Memgraph this host does not have.

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Both points addressed in 5029e794, but they need different answers: one is already fixed and the other was fair.

## Finding properties — already fixed, before this review ran

The P1 says restoration keeps findings by qualified name without restoring their captured properties. That was true and was fixed in an earlier commit on this branch. The finding labels are in the far-props capture now:

```
CASE WHEN x:ExternalModule OR x:Resource
     OR x:Gloss OR x:CodeSmell
     OR x:SecurityIssue OR x:Pattern
     THEN properties(x) END AS far_props
```

with `_CHECK_FAR_PROP_LABELS` in the eval store extended to match, and `test_a_findings_properties_come_back_when_its_name_survives` pinning it: a finding whose qualified name survives the re-parse has its snippet rewritten mid-run, and the restore must put the captured value back. Mutating the labels out of the store's modelling reddens that test.

So the thread is stale rather than the code being wrong. I have not resolved it by hand; the re-review on this head should clear it.

## `_isolated` naming — fair, renamed

Renamed to `_measure_then_restore`, which says what it does: run the measurement, then put the graph and the hash cache back. The old name described the flag rather than the behaviour.

## On the three threads resolved without explanation

Each was answered on the PR before being resolved, and the reasoning is in those comments:

- global orphan deletion and shared-node deletion — [this comment](https://github.com/vitali87/code-graph-rag/pull/1926#issuecomment-5667501226), which also records that removing my own sweeps was necessary but NOT sufficient, because `reingest` runs a repo-wide orphan sweep itself at `graph_updater.py:5442`, so the guard now captures pre-existing orphans and re-emits them;
- the unreadable cache — same comment: absent and unreadable shared one representation, so a permission error deleted a cache that merely could not be read.

25 tests pass.

vitali87:
@greptileai review

vitali87:
Seven threads were open; six are already fixed on this branch and one was not. Verified each against the code at d473d91b rather than from recollection, since several were filed before the commits that address them.

## Newly fixed: a dangling subtree edge crashed the capture

The only genuinely outstanding one. The eval store accepts an edge without requiring its endpoints, so an edge can point at a node it never recorded, and `_check_scope` read that child directly:

```
KeyError: ('Function', 'p.app.ghost')
```

Reproduced on a store holding one Module and a `DEFINES` edge to a never-created child. Fixed with the `child in self.nodes` guard, and `test_a_dangling_subtree_edge_does_not_crash_the_capture` pins it — removing the guard reddens exactly that test.

A real graph cannot hold a dangling edge, so this is a property of the double. It still matters, because a capture that raises takes the isolated check down with it, and the double is what the unit tier runs.

## Already fixed, with the evidence

| thread | state |
|---|---|
| unreadable cache treated as absent | fixed: `FileNotFoundError` marks absence, other `OSError` leaves the file alone |
| validated capture not forwarded | fixed: `capture=capture` reaches `GraphUpdater` |
| finding properties not restored | fixed: finding labels are in the far-props CASE and in `_CHECK_FAR_PROP_LABELS` |
| production and emulator traversals diverge | fixed: `HAS_FIELD` added to the capture walk, with two tests pinning the three walks together |
| shared node classified as new from scope membership | fixed: the created-set is computed BEFORE the first delete, and pre-existing orphans are captured and re-emitted |

The `HAS_FIELD` one was a real catch and worth calling out: the module delete walks it, the double walks it, and my capture query did not, so the re-ingest deleted Field nodes the capture never saw and the restore could not put them back. Silent data loss in the mode that promises none.

## Not fixed, deliberately: the property-merge limit

`ensure_node_batch` merges rather than replaces, so a property the check ADDS to a surviving node is not removed. That is documented in the module docstring, including that it covers the File/Folder/Package nodes at the scope's paths as well as far-end nodes.

I could not construct an input where a re-parse actually adds a key to one of those nodes — `structure_processor` emits a fixed four-key row — so it is stated as a limit rather than fixed on speculation. Restoring exact properties means deleting and recreating every captured node, which trades a demonstrated non-problem for a larger blast radius.

26 tests pass.

vitali87:
@greptileai review

vitali87:
Split this oversized PR into smaller reviewable replacement PRs and registered the linear portions with native GitHub stacks.

Replacement DAG:

- #2090 — feat: add isolated check graph capture
  - #2094 — feat: query isolated check scopes
    - #2091 — feat: emulate isolated check capture
    - #2100 — feat: run structural checks isolated (also depends on #2089)
      - #2092 — feat: expose cgr check isolated flag
        - #2099 — docs: document isolated structural check
      - #2093 — test: add isolated check fixtures (also depends on #2091)
        - #2096 — test: cover isolated graph restoration
        - #2095 — test: cover isolated restore cleanup
        - #2098 — test: cover isolated refusals and edge sites
          - #2097 — test: cover isolated check integration (also depends on #2096 and #2095)
  - #2089 — feat: publish reingest scope

Native stacks registered:

- #2101: #2090 → #2094 → #2091
- #2102: #2100 → #2092 → #2099
- #2103: #2093 → #2098 → #2097

Validation on the replacement head:

```text
uv run ruff check codebase_rag/check_isolation.py codebase_rag/structural_check.py codebase_rag/tests/check_isolation_helpers.py codebase_rag/tests/test_check_isolation_graph_restore.py codebase_rag/tests/test_check_isolation_report_disk.py codebase_rag/tests/test_check_isolation_restore_cleanup.py codebase_rag/tests/test_check_isolation_refusals_sites.py codebase_rag/tests/test_check_isolation_updater_scope.py codebase_rag/tests/integration/test_check_isolation_e2e.py evals/cgr_graph.py
# All checks passed!

uv run python -m compileall -q codebase_rag/check_isolation.py codebase_rag/structural_check.py codebase_rag/tests/check_isolation_helpers.py codebase_rag/tests/test_check_isolation_*.py codebase_rag/tests/integration/test_check_isolation_e2e.py evals/cgr_graph.py
# passed

uv run pytest -q codebase_rag/tests/test_check_isolation_graph_restore.py codebase_rag/tests/test_check_isolation_report_disk.py codebase_rag/tests/test_check_isolation_restore_cleanup.py codebase_rag/tests/test_check_isolation_refusals_sites.py codebase_rag/tests/test_check_isolation_updater_scope.py
# 25 passed, 1 skipped, 2 warnings
```

While doing this split I hit another `pr-split` issue and filed it as vitali87/pr-split#186: non-stacked `execute` ignored the DAG `depends_on` bases, so I repaired the branches/PR bases manually before linking the native stacks.

vitali87:
Superseded by the split replacement DAG listed above.

PR #1927, opened 2026-09-14
fix: a local binding of an imported module's name shadows the module (#1907)

### LINKED ISSUE #1907: Python call resolver: a local binding does not shadow an imported module name
## What happens

A function-local binding of a name that is also an imported module does not stop the Python call resolver from linking `name.attr()` to the module's function. The CALLS edge points at code the call cannot reach.

```python
from . import helpers          # helpers.make_pair() -> tuple[int, Banner]

def use(supplied) -> int:
    helpers = supplied         # or: def use(helpers), for helpers in xs, case [helpers], ...
    w = helpers.make_pair()
    return w.render()
```

On `main` (dee19db1) the graph holds `use -> proj.helpers.make_pair` and, through the bare-name fallback, `use -> proj.engine.Banner.render`. Python never reads the module here: `helpers` is local for the whole body.

## Where

`codebase_rag/parsers/call_resolver.py`: the attribute-call path consults `import_mapping[module_qn]` for the receiver name without asking whether the caller's own scope binds it. The type-inference side of the same question was fixed for tuple unpacking in #1905 (`_locally_bound_names` in `codebase_rag/parsers/py/ast_analyzer.py`, which already enumerates every binding form: parameters, assignment / augmented / `for` / `with ... as` targets, walrus, `except ... as`, body-level import, nested def or class name, `case` capture, an enclosing def's locals); the resolver can reuse it.

## Expected

No CALLS edge to `proj.helpers.make_pair` when the receiver name is bound in the caller's scope (a body-level `from . import helpers` re-binding it to the module is the one binding that keeps the edge).

Found while answering Greptile's review of #1905, whose T-Rex artifact shows the empty local type map (the binding IS recognised there) next to the resolver's edges.

vitali87:
Closes #1907

## Defect

A function-local binding of a name that is also an imported module did not stop the Python call resolver from linking `name.attr()` to the module's function:

```python
from . import helpers          # helpers.make_pair() -> tuple[int, Banner]

def use(supplied) -> int:
    helpers = supplied
    w = helpers.make_pair()
    return w.render()
```

`main` holds `use -> proj.helpers.make_pair`. Python never reads the module here: `helpers` is local for the whole body, before and after the binding runs. The same holds for a parameter, a `for`/`with ... as` target, a walrus, a `case` capture, and an enclosing def's local. The two-part resolver path consulted the module's import map for the receiver name without asking whether the caller's own scope binds it, and the resolver's `(call_name, module_qn)` cache then shared that answer between callers that shadow the name and callers that do not.

## Fix

- `codebase_rag/parsers/py/ast_analyzer.py`: `shadowed_import_names(caller, module_qn)` on the Python analyzer, the import-map keys `_locally_bound_names` (from #1905) reports as bound in the caller. A body-level import the map reflects is a re-import, not a shadow, and stays out.
- `codebase_rag/parsers/call_processor.py`: for a Python caller, record that set on the resolver under the caller's qn before its calls resolve.
- `codebase_rag/parsers/call_resolver.py`: `_receiver_is_untyped_shadow` runs ahead of the cache and every probe. A dotted call whose head is a shadowed import name and has no entry in the local type map is a call on an unknown value: no import-probe answer, no cache read or write, no bare-name trie guess. A typed shadow (`helpers: Widget = supplied`) still resolves through its type as any local does.

## Tests

`codebase_rag/tests/test_local_binding_shadows_import.py`

- `test_a_local_binding_shadows_the_imported_module[<shape>]`: eight binding forms, each red on the base (`assignment`, `assignment_after_call`, `parameter`, `for_target`, `with_alias`, `walrus`, `case_capture`, `enclosing_def_local`).
- `test_the_unshadowed_call_keeps_its_edge`: control, the same body without the binding keeps the edge.
- `test_a_body_level_reimport_is_not_a_shadow`: `from . import helpers` inside the body keeps the edge.
- `test_siblings_do_not_share_an_answer_through_the_cache[shadowed_first|use_first]`: both callers have an empty type map, so both are cache-eligible; whichever resolves first, `use` keeps its edge and `shadowed` gets none.
- `test_a_typed_shadow_resolves_through_its_type`: `helpers = Banner()` shadows the module but is typed, so `helpers.render()` is `Banner.render`.

## Mutations

Each applied, run, and restored as separate commands, anchors asserted unique.

- A: the resolver's early refusal never returns (`return None` -> `pass`): exactly the eight shadow shapes red (8 failed, 5 passed); the controls, re-import, cache-order and typed-shadow tests stay green.
- B: the shadow check moved to after the cache read: exactly `test_siblings_do_not_share_an_answer_through_the_cache[use_first]` red (`use` cached the hit and `shadowed` inherited it); `shadowed_first` and every other test green.
- C: the typed exemption dropped (`return head in shadowed`): exactly `test_a_typed_shadow_resolves_through_its_type` red (the typed local became an unknown value and lost its edge); every other test green.

Base for the runs: the tests are red on `main` + #1905 for all eight shapes (8 failed, 3 passed before the fix).

## Pre-existing, out of scope

- An annotation-only local (`helpers: Banner = supplied`) is not typed by the local type map, so as a shadow it is an unknown value and gets no edge; on `main` the same call got a bare-name trie guess. The map's annotated-assignment gap is separate from this issue.

## Depends on

#1905 (`_locally_bound_names`).

## Local review round (78b232f1 -> 1d98c447)

greptile-local, 4/5, one confirmed P2 false suppression: the optional-import idiom

```python
try:
    from . import helpers
except ImportError:
    helpers = None
return helpers.make_pair()[0]
```

lost a call edge that `main` resolves. `_locally_bound_names` unions bindings over the whole body, so the handler's `helpers = None` put back a name the try body's import had already exempted via `_reimports`. Reproduced on the branch (no edge) and on `main` (edge present) before changing anything, so it was a regression this change introduced.

Fix: `_guards_an_optional_import` exempts a binding in an `except` handler when the same try statement's body re-imports that name. A handler guarding some other import still shadows.

Two further tests, each mutation-pinned:

- `test_an_optional_import_fallback_is_not_a_shadow` — reddens alone when the predicate is dropped from the filter.
- `test_a_handler_guarding_another_import_still_shadows` — the discriminating control; reddens when the predicate is made unconditional, which would otherwise exempt every handler binding.

Shapes probed directly, all correct: `try/except/else/finally`, a binding in `else:` or `finally:` (still shadows, neither is the import's fallback path), nested try in both directions, an `except` inside a nested `def` (does not leak, the walk stops at the function boundary), and `import os as helpers` under a different module (still shadows, `_reimports` requires the map to agree on the path).

Re-review of 1d98c447: 5/5, no findings. 268 passed across `test_local_binding_shadows_import.py`, `test_tuple_unpacked_locals.py`, `test_call_resolver.py` and `test_call_processor.py`.

Pre-existing and unchanged: an import nested deeper in the try body (`try: / if True: / from . import helpers`) is not seen, since `_reimports` is applied to the try body's direct children only. Same limitation `_reimports` already had on `main`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Python call resolution when local variables, parameters, loop targets, aliases, or other bindings shadow imported names.
  - Correctly handles `global` and `nonlocal` declarations, re-imports, optional imports, and typed shadowed names.
  - Prevents incorrect links to imported functions when a local binding takes precedence.
  - Improved resolution consistency across nested scopes and independent callers.


vitali87:
@greptileai review

PR #1933, opened 2026-09-14
refactor: extract the NEWS.md entry parser (#1669)

vitali87:
## Summary

`format_latest_news` in `codebase_rag/readme_sections.py` carries a SonarCloud `python:S3776` finding on `main`. The bullet-collection loop moves into a new `_news_entries(content)` helper returning the entries and the marker count, leaving `format_latest_news` to read the file, apply the count and join. Behaviour is unchanged.

One simplification is worth calling out because it is the only place the refactor could have changed behaviour. The old loop ended with two branches:

```python
elif current and line.strip():
    current.append(line)
elif current:
    bullets.append("\n".join(current)); current = []
```

A blank line with no entry open fell through both and did nothing. The new loop ends with an unconditional `else: close()`, and `close()` is a no-op when nothing is open, so the two are equivalent. That equivalence is the main thing verified below rather than asserted.

## Type of Change

- [x] Refactor (no functional change)

## Related Issues

Part of #1669.

## Test Plan

- [x] 7 new tests in `codebase_rag/tests/test_generate_readme.py`, written and run against the UNREFACTORED code first so they pin current behaviour: only the first marker counts; a blank line before the marker still closes the entry; an empty file; a file with no bullets; an indented marker; `limit=0` rendering nothing; a blank line between bullets not stopping collection.
- [x] 21 tests green in that file (7 new plus 14 pre-existing).
- [x] 10 mutations, each reddening exactly its own tests: the marker compared unstripped; a later marker overwriting the count; the marker not closing the open entry; a bullet not closing the previous one; a blank line not closing; continuation lines not collected; the trailing entry dropped; the marker count tested against `None` rather than falsily; the limit ignored; the wrong exception caught on an unreadable file.
- [x] The local reviewer ran the old and new loop differentially over 1,952,904 cases (every line sequence up to length 4 over a 14-shape alphabet, plus 40,000 random longer ones, crossed with LF/CRLF/trailing-newline forms and limits 0, 1, 2, 3, 5, 100) with zero divergence, and separately enumerated 98 state-transition pairs covering the collapsed branch in both directions. The harness was validated with a planted defect first (272,640 diffs), so the zero is meaningful.

### Two honest notes on the tests

- Three of the seven new tests (empty file, no bullets, blank line between bullets) kill no mutation I could construct: they document behaviour at the edges rather than discriminate. They are kept because they are cheap and pin real boundaries, but they should not be counted as coverage of the refactor.
- The reviewer suggested strengthening one of them. The strengthening turned out to be worthless, because the assertion I was about to add passes under both the working and the broken code while an existing assertion in the same test already discriminates. It was reverted in favour of a comment recording why the existing assertion is load-bearing, rather than adding a line that looks like rigour and is not.

- [x] The pre-commit `ty` hook is red at this machine's baseline (identical on `main`, nothing in a touched file) and was skipped for the commits; every other hook ran, including `generate README sections`. `ty check` on the changed module passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved latest news formatting for blank lines, wrapped bullets, indented release markers, and multiple release markers.
  * Correctly handles empty or bullet-less news content and zero-entry limits.

* **Tests**
  * Added coverage for latest news formatting edge cases and release marker handling.


vitali87:
@greptileai review

Issue #1903, opened 2026-09-13
A function nested in a method is ingested twice: once by the module-function pass with no class context, once by the method pass


vitali87:
Found in passing by the local review of #1901's fix.

## The mechanism

A `def inner()` nested inside a method is reached by two passes in `call_processor.py`: `_process_calls_in_functions` (module-level function walk, `class_context=None`) and `_process_methods_in_class` (class walk, `class_context` set). Both resolve `inner`'s calls and both emit CALLS edges for the same call sites.

## When it shows

Only when the two passes disagree. With #1901, the method pass types `w = self.parse()` from the enclosing method's `self` and binds `w.render()` to `Widget.render`; the module-function pass has no class context, leaves `w` untyped, and still emits the bare-name fallback edge to an unrelated `Banner.render`. So a nested function's call site carries both the correct edge and the false one. Before #1901 the two passes agreed (both untyped) and the duplication was invisible.

Reproduction: `class Parser: def parse(self) -> Widget: ...; def run(self): def inner(): w = self.parse(); return w.render(); return inner()` with a decoy `class Banner: def render(self)`. `test_a_closure_inside_a_method_sees_the_methods_self` in `test_self_receiver_typing.py` asserts the correct edge with `in` rather than `==` because of this and points here.

## Fix shape

Either skip functions whose enclosing scope is a method in `_process_calls_in_functions` (the class pass owns them), or run the module pass with the enclosing method's class context. The first keeps one owner per definition.

Issue #1901, opened 2026-09-13
self.method() receivers are never typed: a two-part call has no self entry in the local type map


vitali87:
Found in passing by the local review of #1896's fix; a shared limitation of
the single-name and unpacking assignment paths, not introduced by either.

```python
class Parser:
    def parse(self) -> tuple[int, SourceMap]: ...
    def run(self) -> str:
        _n, inner = self.parse()      # binds nothing
        x = self.parse()              # x untyped too
```

`_resolve_method_qualified_name("self.parse()", module_qn, local_var_types)`
needs a receiver type for `self`, and the local type map never carries a
`self` entry — the class context is known to the caller but not written into
the map. So `self.method()` on the right-hand side of an assignment, single
name or unpacked, is untyped, and a later call on the result falls to the
bare-name fallback.

Fix shape: seed `local_var_types["self"]` (and `cls`) with the enclosing class
qualified name when the callable is a method (the class context is already
threaded to `resolve_function_call` as `class_context`), or special-case a
`self.`-prefixed receiver in `_resolve_method_qualified_name` to use the
enclosing class. Regression: the shape above with a decoy class defining the
called method, asserting only the enclosing class's method.

PR #1937, opened 2026-09-14
test: assert the closure's receiver edge is the only one (#1903)

vitali87:
## Summary

`test_a_closure_inside_a_method_sees_the_methods_self` asserted membership rather than equality, because a function nested in a method was ingested twice: once by the class pass with the enclosing class as context, and once by the module-function pass without it. The second pass emitted a bare-name `render` edge to the colliding decoy class, so the test could only assert that the correct edge was present among others.

#1903 made the class pass the single owner of such a function, so that decoy edge is gone. The assertion is now `==`, and the comment records why rather than leaving a caveat that no longer applies.

Membership is the weaker form precisely here: it passes whether or not the decoy edge comes back, so without this change a regression of #1903 would leave this test green.

## Type of Change

- [x] Test improvement (no production change)

## Related Issues

Follows up #1903 (fixed in #1906). Part of the #1901 work in #1904, where the weakened assertion was introduced and flagged for exactly this follow-up.

## Test Plan

- [x] Verified the premise before changing the assertion, rather than assuming the merged fix did what its title says: driving the closure fixture against `main` at `ae519452` yields exactly `{proj.widget.Widget.render}` for the `.inner` caller, where it previously also carried the decoy.
- [x] Confirmed the tightening discriminates. Against polluted output (the correct edge plus a decoy) the old `in` form passes and the new `==` form fails; against correct output the new form passes. Without that check both forms pass on correct output and the change would be untested.
- [x] `test_self_receiver_typing.py` and `test_nested_function_single_owner.py`: 16 passed, 2 skipped. The two skips are JavaScript cases in the latter file, skipped for a missing parser in this environment and unrelated to this change.
- [x] The changed test run on its own by node id: 1 passed, so the result is not an artifact of it being skipped or deselected.
- [x] The pre-commit `ty` hook is red at this machine's baseline (identical on `main`, nothing in a touched file) and was skipped for the commit; every other hook ran.

## Summary by CodeRabbit

- **Tests**
  - Improved validation of nested functions defined inside class methods.
  - Tests now confirm that ownership information is reported precisely, without duplicate or overly broad entries.
  - This strengthens regression coverage for code analysis involving method closures and class context.


vitali87:
@greptileai review

PR #1946, opened 2026-09-15
refactor: split the source-root discovery walk into per-signal helpers (#1669)

vitali87:
Part of #1669.

`discover_python_source_roots` carried cognitive complexity 31 (Sonar `python:S3776`, threshold 15): one `os.walk` body doing directory filtering, three independent discovery signals and the recording/dedup logic inline.

The function's own comment already named the structure ("three signals"), so the split follows it:

| helper | signal |
| --- | --- |
| `_nested_package_root` | a package whose parent is not a package, below the repo root |
| `_src_dir_roots` | what a `src` directory exposes (PEP 420 children, single modules) |
| `_pyproject_roots` | setuptools `package-dir` remaps declared in that directory |
| `_walkable` | the ignore/dot-prefix directory filter |
| `_record_root` | identity-remap skip, top-level keying, pair dedup, debug log |

The walk now collects from the three signals and records each pair. No behaviour change: the public signature, the return shape and the ordering are unchanged.

Complexity after, whole file: every function at 6 or below, `discover_python_source_roots` at 3 (measured with `cognitive_complexity`, same metric family as the rule).

## Verification

17 new unit tests in `test_python_source_roots_units.py`, one per branch of the walk, **written and run against the unrefactored code first** — all 51 tests in the two source-root files passed before the refactor and pass unchanged after it. Each asserts the whole mapping rather than membership, so a spurious extra root fails the test; `x in roots` would not see it.

A 14-mutation harness pins that those tests can actually fail: each mutation names the tests it must redden, and the run is compared against that prediction rather than against "something went red". One mutation is a **documented no-op**, verified rather than assumed: dropping the `src`-child `__init__.py` filter makes that branch produce a pair identical to one the package signal already records, and `_record_root` deduplicates it. Driving `_src_dir_roots` both ways gives `[]` with the filter and `[('proj', 'src.proj')]` without — equal to the package signal's own pair. The filter is redundant with dedup rather than load-bearing; it is kept because it matches the original and avoids a pointless second record, and it is marked in the harness so its greenness is never read as coverage.

Beyond the unit tests, a **differential check** compares the original walk (reconstructed from `origin/main`) against the refactored one on the same trees: 10 hand-picked edge cases the unit tests do not reach (`src` that is itself a package, nested `src` under a package, `src/src`, a dotted-name file, a pyproject inside an ignored directory, an identity remap) and 300 randomly generated repository shapes. **Zero differences.** The harness was validated with a known-negative — a deliberately broken "original" that skips the pyproject signal diverges on 20 of the same 300 shapes — so the zero reflects equivalence rather than a dead comparison.

A local review round found two things the sweeps above could not, both now fixed:

- A comment claimed the `src`-child `__init__.py` filter was an inert no-op. **It is not.** When `src` is itself a package, `_nested_package_root` skips `src/child` (its parent is a package), so nothing else records the pair and there is no duplicate for `_record_root` to absorb — dropping the filter invents a root. Reproduced directly (`{}` shipped vs `{'proj': [('proj', 'src.proj')]}` without it) before accepting the finding. The earlier "no-op" conclusion came from driving `_src_dir_roots` in isolation, which proves what the helper does and not what the composed walk does. The comment now states the real reason, and `test_a_src_child_package_under_a_package_src_is_not_a_root` pins it.
- Nothing pinned candidate **ordering**, which decides resolution: `resolve_via_source_roots` returns the first disk-confirmed match. Two tests now pin it, using parent names deliberately out of walk order (`zz`/`aa`/`mm`) so an unsorted walk is observable, plus one pinning that the package signal is recorded before the pyproject signal.

Each of the three new tests was verified to redden when the line it pins is mutated away, and to redden alone.

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal discovery of Python source roots while preserving existing behavior, including package detection, `src` layouts, project configuration mappings, filtering, and duplicate handling.

- **Tests**
  - Expanded coverage for source-root discovery across nested packages, namespace packages, module files, ignored directories, project remappings, duplicate candidates, nested configurations, and ordering rules.


vitali87:
@greptileai review

PR #1947, opened 2026-09-15
fix(evals): walk CONTAINS_SECTION in the double's module subtree (#1938)

### LINKED ISSUE #1938: evals double omits CONTAINS_SECTION from the module subtree walk, so Sections survive a re-parse there but not in production
## What happens

`evals/cgr_graph._MODULE_SUBTREE_RELS` says in its own comment that it is "what `CYPHER_DELETE_MODULE` walks", but omits `CONTAINS_SECTION`:

```python
# What CYPHER_DELETE_MODULE walks: the definitions plus what hangs off them
# by ownership.
_MODULE_SUBTREE_RELS = _DEFINES_RELS | {
    cs.RelationshipType.HAS_PARAMETER.value,
    cs.RelationshipType.HAS_FIELD.value,
}
```

The production query walks five relations:

```
OPTIONAL MATCH (m)-[:DEFINES|DEFINES_METHOD|CONTAINS_SECTION|HAS_PARAMETER|HAS_FIELD*0..]->(c)
```

Measured on `origin/main`:

| | relations |
|---|---|
| `CYPHER_DELETE_MODULE` | CONTAINS_SECTION, DEFINES, DEFINES_METHOD, HAS_FIELD, HAS_PARAMETER |
| `_MODULE_SUBTREE_RELS` | DEFINES, DEFINES_METHOD, HAS_FIELD, HAS_PARAMETER |
| in the delete only | **CONTAINS_SECTION** |

## Why it matters

`CONTAINS_SECTION` is how a document's heading structure hangs off its Module, and it is in the delete query deliberately: the query's own comment says that without it "a re-indexed document keeps every Section from its previous parse (issue #1426)".

So in the unit tier, a re-parse of a document does **not** remove its Section nodes, while in production it does. Any test that asserts on Section survival across a re-ingest is testing the opposite of production behaviour, and a defect in that area cannot be reproduced against the double at all.

This is the same class as #1911 (the double's edge key ignoring `MERGE_KEY_PROPS_BY_REL`): the emulator is the oracle for the incremental and structural-delta tests, so where it diverges from production the tests are blind rather than wrong-looking.

## Expected

`_MODULE_SUBTREE_RELS` includes `CONTAINS_SECTION`, or the comment stops claiming it mirrors the delete and says which relations it deliberately omits and why.

A test asserting the two lists agree would stop the next divergence. They have drifted at least twice: `HAS_FIELD` was added to both, and `CONTAINS_SECTION` to only one.

## Found by

Noticed while adding a capture query for #1718 that must walk exactly what the delete walks. CodeRabbit flagged the same class of drift on that PR for `HAS_FIELD`, which my capture query was missing; checking the third implementation of the same walk turned this up. Not fixed there because it is pre-existing and its blast radius is the document tests rather than that PR.

vitali87:
Closes #1938.

`evals/cgr_graph._MODULE_SUBTREE_RELS` says in its own comment that it is "what `CYPHER_DELETE_MODULE` walks", and omitted `CONTAINS_SECTION`.

| | relations |
|---|---|
| `CYPHER_DELETE_MODULE` | CONTAINS_SECTION, DEFINES, DEFINES_METHOD, HAS_FIELD, HAS_PARAMETER |
| `_MODULE_SUBTREE_RELS` | DEFINES, DEFINES_METHOD, HAS_FIELD, HAS_PARAMETER |

`CONTAINS_SECTION` is how a document's headings hang off its Module, and it is in the delete query deliberately: without it "a re-indexed document keeps every Section from its previous parse (issue #1426)".

So a re-parsed document kept its Sections in the double while production deleted them. A test asserting on Section survival across a re-ingest pinned the **opposite** of production behaviour, which is worse than missing coverage: the green actively certifies the wrong thing.

## Fixing the class, not the instance

The list is a hand-maintained copy of one that lives in the query, so restating the missing name would drift again on the next relation to join the walk. That is the shape of an inventory check that fails open: a guard driven by an explicit list cannot catch the omission it exists to prevent.

So `TestTheModuleSubtreeWalkMirrorsTheDeleteQuery` **derives** the expected set by extracting the walk from `CYPHER_DELETE_MODULE` itself and asserting exact equality. A restated list would be the same hand-maintained copy one layer down.

Guards on the instrument itself, because a check that cannot fail in the direction you care about is not a check:

- `test_the_extractor_reads_a_real_walk` pins two relations by name and a length floor, so a regex matching a **shorter** walk than the real one cannot make the comparison vacuous while staying green.
- The extractor asserts its match is not `None`, so a future edit that the regex cannot read fails loudly rather than yielding an empty set. Probed: a named variable, a space after the colon, a lowercase relation name, an alias rename and a newline reflow all fail loudly; `*0..`→`*1..` still reads correctly.
- `test_a_nested_subsection_goes_with_its_parent` drives the variable-length hop. Sections nest, so a single-level fixture passes even if the walk stops after one hop — verified by mutation that the nested case reddens while the single-level one stays green.

## Merge-order note

This test compares the walk against the double's mirror of it, so **any two PRs that each add a relation to that walk are now order-dependent**: the second to merge must merge `main` in first, or the parity test fails on `main`. #1931 adds `DEFINES_CONSTANT` to the same walk and is affected.

That is the test converting a silent divergence into a loud failure, which is the point — but it is an obligation that did not exist before, so it is worth stating.

## Verification

Red before the fix, with the predictions written down first: the parity test and both Section tests failed, the control passed. Failure reasons checked, not just the colour — `only in the query ['CONTAINS_SECTION'], only in the double []`, and the Section genuinely outliving its Module.

Drift is caught in **both** directions: mutating the query side instead (adding a relation to the walk, leaving the double alone) reddens the parity test naming the new relation.

17 passed in the file; 207 passed across the reingest/incremental/eval-double dependents. `_MODULE_SUBTREE_RELS` has exactly one use site, and no test outside the new file pairs Sections with this double, so nothing silently changes meaning.

## Summary by CodeRabbit

- **Bug Fixes**
  - Module deletion now also removes owned document sections, including nested sections.
  - Reprocessing a document correctly clears its previously stored sections.
- **Tests**
  - Added coverage to verify module-subtree cleanup matches the deletion behavior and handles direct and nested sections.


vitali87:
@greptileai review

vitali87:
**Merge-order resolution, pre-verified.**

#1931 and #1947 both add a relation to `CYPHER_DELETE_MODULE`'s walk and to the eval double's mirror of it, so they conflict in `evals/cgr_graph.py`. I merged them in a throwaway worktree to find out exactly how, rather than leaving the second merger to discover it.

The conflict is a single hunk and the resolution is **keep both lines**:

```
<<<<<<< HEAD
    cs.RelationshipType.DEFINES_CONSTANT.value,
=======
    cs.RelationshipType.CONTAINS_SECTION.value,
>>>>>>>
```

becomes

```python
    cs.RelationshipType.DEFINES_CONSTANT.value,
    cs.RelationshipType.CONTAINS_SECTION.value,
```

Composed state verified: walk and double both carry all six relations (CONTAINS_SECTION, DEFINES, DEFINES_CONSTANT, DEFINES_METHOD, HAS_FIELD, HAS_PARAMETER) and the parity assertion holds.

There is no ordering preference — either PR may merge first — but the second **must** merge `main` in and resolve this hunk, or the parity test fails on `main`. That failure is the test working as designed: it converts a silent divergence between the double and production into a loud one. Resolve by hand, not by script.

PR #1949, opened 2026-09-15
docs: state the f-string brace convention in cypher_queries

vitali87:
Documents an undocumented convention that 12 lines in `codebase_rag/cypher_queries.py` already depend on. Docstring only — no query text changes.

## Why

The module has **37 f-string query literals** and **12 lines relying on doubled braces** (`{{qualified_name: $qn}}`), with no comment anywhere saying so. #1930 added one more when `CYPHER_FIND_BY_QUALIFIED_NAME` became an f-string to interpolate `_SNIPPET_LABELS`, which is what prompted this.

## Every claim executed, not reasoned about

The first draft got three claims wrong, and a local review falsified them by running them. A docstring whose only value is accuracy is worse than nothing when it is wrong, so each statement here was executed:

| claim | verified |
|---|---|
| `f"...{name: $x}"` → `NameError` at import | yes |
| `f"...{id: $x}"` (builtin key) → `TypeError` | yes |
| `f"RETURN {id}"` → `'RETURN <built-in function id>'` (quiet) | yes |
| `f"-[:CALLS*{2}]->"` → `'-[:CALLS*2]->'` (quiet) | yes |
| stray brace in a `.format()` template → `KeyError` at call time | yes |

What the first draft had backwards, and why it matters:

- **Map syntax is never the quiet case.** The colon parses as a format spec, so a missed doubling raises either way. I had described it as the usual quiet failure.
- **The quiet case is a colon-free brace resolving as a name** — usually a builtin, which is in scope whatever this module defines. Cypher's own quantifier is the shape that would actually bite here.
- **Non-f-string literals are not uniformly safe.** Three are `.format()` templates — `CYPHER_AUDIT_MISSING_REQUIRED`, `CYPHER_AUDIT_IS_NULL` (formatted in `graph_audit.py`) and `CYPHER_MEMORY_LIMIT_SUFFIX` (`graph_dialects.py`) — where a stray brace raises `KeyError` at **call** time, strictly quieter than the import-time failure being warned about. The draft said adding a brace to a non-f-string was "always safe", which would have walked someone into exactly that.
- `_SNIPPET_LABELS` and `_GLOSS` are defined **in this file** from the `constants.graph` enums, not in `constants.graph`. The draft's pointer sent readers to the wrong module.

## Verification

- All **96** rendered module-level string constants dumped from base and branch and diffed: **byte-identical**. No behaviour change.
- Module imports cleanly; `CYPHER_FIND_BY_QUALIFIED_NAME` still renders its full 12-label set with no unrendered `{{`/`}}`.
- `ruff check` and `ruff format --check` clean repo-wide (1280 files).

No tests run: there is no rendered-output delta to test, and the constant diff is the stronger evidence.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation explaining brace-escaping conventions for Cypher query definitions.
  * Clarified how literal braces are handled across formatted and non-formatted query strings.


vitali87:
Fixed the P2 in `dd3af780`.

You were right that `CYPHER_MEMORY_LIMIT_SUFFIX` is not in this file — it lives in `constants/graph.py:1041` and is formatted at `graph_dialects.py:153`. The docstring said "the templates here are" and listed it, so a reader scanning this module for it would not find it. It now names the two that are actually here (`CYPHER_AUDIT_MISSING_REQUIRED`, `CYPHER_AUDIT_IS_NULL`, both formatted in `graph_audit.py`) and mentions the third parenthetically as belonging to `constants.graph`, since the rule applies there too and it is worth knowing about.

One thing the fix turned up, worth recording because it is the same trap the docstring is about. Checking my list of templates with a grep for `<name>.format(` also matches `CYPHER_SET_PROPS_RETURN_COUNT` — a false hit. That constant is concatenated at `cypher_queries.py:466` and `:487` and never formatted; the grep matched its own `RETURN count(` text. Two templates in this file, not three. I verified each by reading its call sites rather than trusting the match.

All 96 rendered string constants still diff byte-identical against base, so this remains docs-only.

Note the head has moved to `dd3af780`, which invalidates the review anchored to `b8f2c9c5`. That is the usual sequence — a finding, a fix, and the fix discards the review that found it — so this needs a fresh review at the new head before it can be gated.

vitali87:
@greptile review

PR #1952, opened 2026-09-15
fix(scripts): a closed PR's cleared run association is not a missing one (#1944)

### LINKED ISSUE #1944: check_pr_gated.py cannot distinguish a closed PR's cleared run associations from a missing one
On a MERGED or closed PR, `scripts/check_pr_gated.py` reports:

```
PR #1940: NOT verifiably gated -- 1 reason(s)
  - the CI run at c7dac6d1 does not resolve to #1940; owners: none (could not be determined). Branches sharing a head SHA after a rebase report each other's runs
```

The PR was fine. It merged cleanly at `eb9b32c4`, with 31 SUCCESS + 1 SKIPPED contexts and a 5/5 review anchored to the exact head.

## Mechanism

The ownership block (around line 648) fetches each run at the head and unions `detail.pull_requests[].number`. GitHub **clears `pull_requests` once the PR is closed or merged**, so `owners` comes back empty and the `pr not in owners` branch fires.

Measured on 2026-09-15, same repo, same workflows, same author, all same-repository (non-fork):

| PR | state | run created | `pull_requests` |
|----|-------|-------------|-----------------|
| #1940 | MERGED | 2026-09-14T20:51:32Z | `[]` |
| #1869 | MERGED | 2026-09-13T09:43:09Z | `[]` |
| #1939 | OPEN | 2026-09-14T20:53:35Z | `[1939]` |
| #1930 | OPEN | 2026-09-14T17:19:26Z | `[1930]` |

#1939's run was created two minutes *after* #1940's and resolves; #1869's is two days old and does not. Age is not the variable, and neither is fork status — all four are same-repo. The open/closed split is the only one that holds. GraphQL still reports the association correctly (`associatedPullRequests` returns `{number: 1940, state: MERGED}`), so the information exists; only the REST run detail drops it.

## Why it matters

The empty-`owners` case is deliberately fail-closed, and the code comment says so — it cites Greptile on #1625 and the fail-open shape the checker exists to catch. That reasoning is right and should not be reverted. The defect is that **one branch now serves two populations**: a genuinely unverifiable run, and a perfectly ordinary merged PR.

The consequence is not a bad merge — it is the opposite, and it is corrosive in a subtler way. The reason text asserts a specific cause ("Branches sharing a head SHA after a rebase report each other's runs") that is simply untrue here: exactly one branch holds this SHA. An operator who checks the claim finds it false, and a gate that cries wolf on merged PRs trains people to override it on open ones. That is the same failure mode #1936 and #1941 are about from the other direction.

This is adjacent to #1877 but not fixed by it: that PR fixed the **required workflow's** run lookup for fork PRs and states explicitly that `check_pr_gated.py` "is unchanged and still conservatively rejects fork runs without explicit PR associations". The script keeps the gap.

## Suggested fix

Read PR state first and handle the closed case as its own arm, rather than widening the empty-`owners` branch:

- If the PR is MERGED or CLOSED, say so plainly — gating a merged PR is moot, and the script should report that instead of a rebase-collision theory.
- While the PR is OPEN and `owners` is empty, fall back to GraphQL `associatedPullRequests` on the head OID before concluding anything. Keep fail-closed if that is also empty.
- Either way, stop asserting the shared-SHA cause unconditionally. Assert it only when a second branch actually holds the head (`git ls-remote` / the runs' own `head_branch` set disagree), and otherwise report what was observed: "no run at this head carries a PR association".

A regression test wants all three shapes: open with associations, open without, and merged.

Found while gating #1940 (issue #1932), which then turned out to have already merged.

vitali87:
Closes #1944.

`check_pr_gated.py` reported correctly gated MERGED PRs as ungated:

```
PR #1930: NOT verifiably gated -- 1 reason(s)
  - the CI run at 9b10a9f2 does not resolve to #1930; owners: none (could not be determined)
```

Reproduced on **#1930, which this session had merged an hour earlier after verifying its gate by hand** — so the tool contradicted a merge it had itself certified.

## Mechanism

GitHub **clears** a workflow run's `pull_requests` once its PR closes or merges. Measured, same repo and workflow, differing only in PR state:

| PR | state | run's `pull_requests` |
|----|-------|------------------------|
| #1930 | MERGED | `[]` |
| #1947 | OPEN | one entry |

## The fix, and the fail-open inside it

The first commit excused an empty owner set when the PR is MERGED or CLOSED.

That was **not sufficient, and a peer session caught it**: `_gh_stdout_or_empty` returns `""` for a failed `gh` call *and* an empty body alike, so an **unreadable** run and a **cleared** one reach the owner loop identically. Excusing on PR state alone covered both — reopening exactly the fail-open the empty-is-unverified rule closed (Greptile on #1625).

So the second commit tracks whether every run's detail actually parsed. Only a run that was **read** and reported no owner is excused:

- a failed detail fetch keeps failing closed on a merged PR exactly as on an open one;
- a populated owner set naming a different PR still fails, whatever the state — that is the rebase collision the message exists for;
- an open PR with no owner still fails closed.

Verified against live data: run `34874043546` returns a 35-key object with `pull_requests` length 0, so the genuine cleared case is a **truthy dict** and is excused, while `""`/`{}` is not.

The excused case emits a **caveat** rather than passing silently, so the untested assumption is visible in the output.

## Verification

Every claim is pinned by a test that was observed to fail:

| mutation | reddens |
|---|---|
| `all_details_read` dropped from the condition | fetch-failed test, alone |
| state gate dropped | open-PR-no-owner test, alone |
| populated non-matching owner set excused | open-PR-other-owner test, alone |
| `if not detail: continue` guard removed | fetch-failed test, alone |
| caveat wording changed | CLOSED test only — the MERGED test pins different tokens |
| `CLOSED` dropped from the state tuple | CLOSED test, alone |

Live behaviour after the change: #1930 gated-with-caveat, #1947 still strict with CI-in-flight as its only reason.

## A vacuous assertion, fixed

The pre-existing ownership test could **never fail**:

```python
reasons = check_pr_gated.check("1826")          # binds the (reasons, caveats) TUPLE
assert not any("does not resolve to" in r for r in reasons)
```

Iterating the tuple yields two **lists**, so `in` is list membership, not substring. Demonstrated both directions: the expression is `True` whether the reason is present or absent. It had been green without testing anything.

Unpacked and given a failure message; its original intent is unchanged. Audited every `check()` call site — 8 on `main`, exactly one vacuous; a second that looks similar deliberately captures the tuple to assert its shape and is correct. A control-validated AST scan (proven to flag the known positive first) returns **0 hits across 1298 files**, so this is not systemic.

132 passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pull request checks now remain blocked when any CI run detail cannot be read or omits valid pull request ownership data, including for merged or closed pull requests.
  - Closed or merged pull requests are cleared only after all CI run details have been successfully verified.
  - Improved handling of CI run ownership checks prevents incomplete or malformed details from being treated as valid.
- **Tests**
  - Added regression coverage for unreadable, incomplete, and malformed CI run details across open, merged, and closed pull requests.


vitali87:
@greptileai review

vitali87:
@greptileai review

PR #1953, opened 2026-09-15
refactor: split the ast-grep rule validation into two guards (#1669)

vitali87:
Part of #1669.

`_parse_rule` carried cognitive complexity 21 (Sonar `python:S3776`, threshold 15): a type dispatch, a mutual-exclusion check, three near-identical "this modifier applies to that selector only" guards, and the construction, all inline.

Two helpers come out of it:

| helper | responsibility |
| --- | --- |
| `_require_exactly_one_selector` | a rule selects by `pattern` **or** `kind`, never both and never neither |
| `_require_selector_applies` | the three modifier guards, as one table-driven loop |

`_parse_rule` itself becomes a flat dispatch: string shorthand, reject non-mappings, validate, build. Complexity after: `_parse_rule` 6, `_require_selector_applies` 4, `_require_exactly_one_selector` 1. No behaviour change.

## Verification

The three guards were three sequential `if`s and are now one loop, so **the loop's order is what preserves which error a multi-violation rule reports**. That precedence was unpinned, so it is pinned now. It was also checked against the original directly: the `origin/main` function was extracted and run standalone against four multi-violation rules, and the reported error is identical in every case.

`test_ast_grep_tier_languages.py` already covers the main validation branches end to end through `load_pattern_configs`, which is the right level for them. The new file does not duplicate that. It calls `_parse_rule` directly to pin only what the YAML path cannot express:

- the truthiness edges — an empty string is a *present* pattern, not an absent one, and `pattern: ""` beside a real `kind` is a valid kind rule
- non-string field values being stringified (YAML will not produce an `int` key value)
- the guard precedence described above
- `_parse_rules` itself: empty sections, non-list sections, and per-item order

Each test asserts the whole `_Rule` rather than one field, so a refactor that drops or invents a field fails here.

A 13-mutation harness confirms the tests discriminate: every mutation names the tests it must redden and is compared against that prediction. Two mutations that reddened ~50 language tests apiece were dropped rather than kept — a mutation that blasts most of the suite demonstrates that the suite runs, not that any test pins the branch.

A local review round verified equivalence far more thoroughly than I had: an exhaustive cross-product of 5 fields x 13 values (371,293 mapping inputs) plus 11 non-mapping inputs, **zero mismatches** against the base implementation in raised-vs-returned, message text and every `_Rule` field. That covers the guard-order question and the `name_head` precompute-vs-in-loop question outright, and it confirmed the 12 tests are not duplicates: they kill 6 of 7 mutations that `test_ast_grep_tier_languages.py` kills none of. Coverage analysis confirmed the pruned tests removed no unique line or arc.

It also found a real defect in the test file, since fixed. The docstring claimed whole-object assertions catch any dropped field, but **no test expected a truthy `has_child` or `name_head`**, so those two slots sat at their defaults and equality could not see them being dropped. Verified by deleting `has_child=...` from the constructor: all 12 tests still passed. Two positive-value tests now close it, and each was confirmed to redden alone when its field is dropped. The docstring now states the condition the claim depends on, rather than asserting the conclusion.

One mutation survives and is an **equivalent mutant, not a gap**: moving `name_head` to last in the guard tuple is unobservable, because `name_head` requires `pattern` while `has_child`/`name_child` require `kind`, and the exactly-one-selector guard means only one can be absent. No input distinguishes them.

## Summary by CodeRabbit

- **Refactor**
  - Internal rule validation was reorganized without changing validation rules or resulting behavior.
  - Error messages were slightly reworded.

- **Tests**
  - Added comprehensive coverage for rule parsing, validation, modifier applicability, value handling, unknown keys, empty inputs, and parsing order.


vitali87:
@greptileai review

PR #1956, opened 2026-09-15
refactor: split the cypher response cleaning into per-format helpers (#1669)

vitali87:
Splits `_clean_cypher_response` (cognitive complexity 19, over Sonar's
threshold of 15) into three helpers, one per responsibility.

## What moved

| Helper | Responsibility | Complexity |
|---|---|---|
| `_fenced_block` | extract the query inside a ``` fence, or `None` | 2 |
| `_strip_bold_spans` | drop `**...**` labels and a following `:` | 6 |
| `_strip_markdown` | bold spans, then backticks, then a leading `cypher` | 1 |
| `_clean_cypher_response` | choose a branch, terminate the query | 4 |

No function in the file is over the threshold now.

Four markdown-decoration constants move to `constants/security.py` beside the
existing `CYPHER_*` values. `cli.MARKDOWN_FENCE` already holds the same
string, but the guard table should not depend on the CLI's constants, so this
keeps a cypher-scoped copy.

## Behaviour is unchanged, and that is tested three ways

**Characterisation tests first.** 20 tests in
`test_clean_cypher_response_branches.py`, written and passing against the
UNREFACTORED function, so they pin shipped behaviour rather than describe the
new structure. They assert the WHOLE returned string; the eight existing tests
in `test_llm_service_unit.py` mostly assert one property (`endswith(";")`,
`"```" not in result`), which a broken version that mangles the query and
still terminates it would also satisfy.

Three pinned outputs are quirks rather than design, and are pinned as-is so a
later decision to change one has to edit a test that says so:

* an unterminated fence keeps its backticks (`` ```MATCH... `` -> `` ```MATCH...; ``)
* an unterminated bold span is left in place
* empty input becomes a bare `;`

**A differential sweep.** The original implementation reconstructed and
compared against the refactor over the hand-picked edges plus 20,000 random
strings built from the tokens that drive the branches: 0 differences. A
sabotaged implementation differs on 36 of the hand cases, so the harness can
see a difference when one exists.

**A mutation harness.** 15 mutations, each with its expected red tests written
down before running.

## One real defect the characterisation tests caught

Extracting `_fenced_block` to return `str | None` made the natural caller
`_fenced_block(query) or query`. That is a behaviour change: `"``````"` splits
to `['', '', '']`, so the block is `""` -- falsy -- and the `or` fell back to
the raw text, returning `` ```````; `` where the original returns `;`. An
empty fence is a successful extraction, not a failed one. Fixed with
`query if block is None else block` and a comment saying why.

Part of #1669.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Cypher responses wrapped in Markdown formatting, including fenced blocks, bold labels, inline backticks, and leading language prefixes.
  * Preserved incomplete fences and unterminated formatting without incorrectly altering query text.
  * Ensured query termination, whitespace trimming, and multiline formatting behave consistently.

* **Tests**
  * Added coverage for formatting variations, empty responses, incomplete markup, and query termination behavior.


vitali87:
@greptileai review

PR #1963, opened 2026-09-16
fix(go): bump x/mod to v0.40.0 for sumdb advisories

vitali87:
Closes the remaining half of code scanning alert #78.

## Why this is needed

GO-2026-6179 and GO-2026-6180 each name **two** affected modules:

| Module | Fixed in | Before | After |
| --- | --- | --- | --- |
| `toolchain` | 1.26.6 | fixed in #1962 (`go-version: ">=1.26.6"`) | unchanged |
| `golang.org/x/mod` | 0.40.0 | `v0.22.0` | `v0.40.0` |

Pinning the CI toolchain closed the exploitable half, since `cmd/go` performs
sumdb verification. Scorecard's OSV check matches declared manifest versions,
so it kept reporting both advisories against the `x/mod v0.22.0` pin. A
successful Scorecard run on `2a3c6a61` (which already contained the toolchain
fix) still scored 8 with both IDs.

## What cascades

`x/mod v0.40.0` declares `go 1.25.0` and requires `x/tools v0.49.0`, so the
bump is not isolated:

- `golang.org/x/tools` v0.28.0 => v0.49.0
- `golang.org/x/sync` v0.10.0 => v0.22.0
- the module's own `go` directive 1.23 => 1.25.0

Pinning the directive back to 1.23 with the new dependencies fails outright
(`go: updates to go.mod needed`), so there is no variant of this fix that
keeps 1.23.

## Behaviour change worth knowing

The `go` directive raises the minimum toolchain needed to **build** the
bundled `gotypes` helper. The runtime build runs under `GOTOOLCHAIN=local`,
so a user on Go 1.23/1.24 no longer builds the helper and drops to the
tree-sitter backbone.

That path degrades cleanly rather than breaking indexing. Driven with an old
toolchain on PATH:

```
WARNING  Go frontend tool failed to build; using tree-sitter.
stderr: go: go.mod requires go >= 1.25.0 (running go 1.23.4; GOTOOLCHAIN=local)
```

It returns empty facts, raises nothing, and passes the toolchain's own message
through verbatim. CI resolves `>=1.26.6`, which satisfies the new floor.

## Verification

- `govulncheck ./...`: 2 vulnerabilities => **No vulnerabilities found**
- `go build`, `go vet ./...`: exit 0
- `go mod verify`: all modules verified; `go mod tidy` is a no-op
- Go-related tests: 455 passed, 0 failed (32 errors are the known no-Docker `@memgraph` baseline)
- Built binary run against a sample module: call sites, externals and IMPLEMENTS pairs all correct, so the facts contract holds rather than merely compiling

## Summary by CodeRabbit

* **Chores**
  * Updated the Go toolchain requirement to version 1.25.0.
  * Refreshed supporting Go tooling dependencies to newer versions.


vitali87:
@greptileai review

Issue #1964, opened 2026-09-18
Java varargs param_types records the modifier, not the element type: `final String... xs` -> 'final...'


vitali87:
## Symptom

`extract_type_facts` (`codebase_rag/parsers/type_facts.py`, `_extract_java_type_facts`) records a Java varargs parameter's type as its **modifier or annotation** whenever one is present, so `param_types` on the Function/Method node is wrong and the deferred `ACCEPTS` edge for that parameter never resolves.

Verified on `main` (3cd90fe8):

```java
class C { void m(final String... xs, @NonNull java.util.List<String>... ys) {} }
```

```
param_types on main: ['final...', '@NonNull...']
```

Expected `['String...', 'java.util.List<String>...']`.

## Mechanism

```python
if child.type == cs.TS_SPREAD_PARAMETER:
    # `String... names`: the type is the first named child.
    inner = next(iter(child.named_children), None)
    types.append(f"{_text(inner)}{cs.LANG_ELLIPSIS}")
```

tree-sitter-java puts a `modifiers` node first inside `spread_parameter` when the parameter carries `final` or an annotation:

```
spread_parameter
  modifiers
    final
  type_identifier  'String'
  ...
  variable_declarator
    identifier [name]  'xs'
```

so "the first named child" is `modifiers`, not the type. A plain `String... xs` is unaffected, which is why the existing tests are green.

## Fix

Take the first named child that is neither `modifiers` nor `variable_declarator`:

```python
inner = next(
    (c for c in child.named_children if c.type not in (cs.TS_MODIFIERS, cs.TS_VARIABLE_DECLARATOR)),
    None,
)
```

The same shape was found and fixed for the Parameter-node enumerator in the #1804 follow-up PR (`java_declared_parameters`), by a local review of that branch; this is the pre-existing twin in `type_facts.py`, filed separately rather than widened into that PR.

A regression test needs a `final` or annotated varargs, since the unmodified form cannot go red.

Issue #1804, opened 2026-09-08
Add a Parameter node label: 12 languages already extract parameters, the data is discarded


vitali87:
## Summary

Add a `Parameter` node label. Twelve languages already have complete, edge-case-hardened parameter extractors in the repo; they are gated off by default and their output is discarded. This issue persists what is already computed.

## Why this one first

It is the cheapest of the four granularity issues and needs no new per-language parsing. The extractors already handle: varargs, Go grouped declarations (`a, b int`), TypeScript `this` parameters occupying no slot, Rust `self_parameter` exclusion, C `variadic_parameter`, Dart `optional_formal_parameters` flattening, PHP `$`-prefixed names, Java `spread_parameter` varargs with the name in a nested declarator.

## What exists today

**Parameter names — 12 languages**, `*_positional_parameter_slots` in `codebase_rag/parsers/utils.py`:

| Language | utils.py |
|---|---|
| Go | 807 |
| JS/TS | 871 |
| C++ | 895 |
| Java | 914 |
| C# | 951 |
| Lua | 1000 |
| Scala | 1024 |
| Rust | 1053 |
| C | 1074 |
| PHP | 1104 |
| Dart | 1129 |
| Python | `flow_access/processor.py:142` |

Dispatched by `_lean_parameter_slots`, `codebase_rag/parsers/flow_access/processor.py:171-204`. Each returns `tuple[list[str | None], int | None]` — one entry per formal slot, `None` where the slot binds no simple name, plus the variadic index.

**Parameter types — 9 languages**, already persisted as the parallel `param_types: list[string]?` property on `Function`/`Method` (`types_defs.py:895`, `:899`), extracted in `codebase_rag/parsers/type_facts.py:192-203`. Per `type_facts.py:15-18` the array is parallel to declared parameters in source order, `""` for unannotated.

**Type resolution** already works: `TypeReferenceResolver` (`type_facts.py:264-357`) resolves an annotation string to a project type QN via imports, module scope, then unique-suffix match.

## Blockers

1. **Gated off by default.** `flow_access/processor.py:633` enables the slot extractors only when `FLOWS_TO` is enabled; `CaptureGroup.IO` is not in `DEFAULT_CAPTURE_GROUPS` (`constants/graph.py:366-373`).
2. **No positions retained.** Every extractor returns bare `str`/`str | None`. No `start_point` is read anywhere. This is the main new work: ~1 line per extractor.
[...]

PR #1965, opened 2026-09-18
feat: declared parameters for the eleven non-Python languages (#1804)

vitali87:
Part of #1804. Closes the "other eleven enumerators" item from the last comment there.

## What

`declared_parameters` now covers Go, JavaScript, TypeScript/TSX, C, C++, Java, C#, Lua, Scala, Rust, PHP and Dart, each as its own `*_declared_parameters` case in `parsers/parameter_nodes.py` mirroring the Python one. `emit_declared_parameters` and the deferred `OF_TYPE` pass are unchanged: every language now feeds them through the same `DeclaredParameter` shape, so Parameter nodes, `HAS_PARAMETER` and `OF_TYPE` reach the graph for all twelve languages behind the existing opt-in `parameters` capture group.

## The three rules the Python review made explicit, applied per language

- **Declaration order, not the taint slot table.** Each enumerator walks the declaration itself. Scala reads every curried list (`def f(a)(b)(implicit c)`), where the taint table reads only the first; Python's keyword-only parameters were the same shape.
- **A variadic is one flagged slot.** Go `...T`, JS/TS `...rest`, Java `T...`, C# `params T[]` (the grammar inlines it as a bare `array_type` + `identifier`), Scala `T*`, PHP `...$x`, C++ `Args&&... args`. `type_name` follows the owner's `param_types` spelling where one exists (`...Opt`, `String...`, `params int[]`).
- **The receiver is excluded by structure, never by name.** Rust `&self` and `self: Box<Self>`, TypeScript `this: T`, Java `C this` take no position; Go's receiver is a separate field and Lua's `:` receiver is implicit. A C# extension method's `this C self` IS caller-supplied (`c.E(x)` or `E(c, x)`) and keeps its slot. Only Python needs `has_receiver` from the call site.

## Index policy for slots that bind no name

An unnamed C/C++/Go parameter, a JS/TS destructuring pattern, a Rust `_` or tuple pattern, a Lua `...`: the slot keeps its declaration position and yields no node, so the parameters after it still agree with the source and with `param_types`. Compacting would have made `HAS_PARAMETER {index}` disagree with the declaration, which is the defect the Python round found in the taint table.

## Two conventions worth knowing

- PHP names drop the `$` sigil, as `Field` nodes already do for promoted properties, so `private int $x` names the same thing from both labels. The taint table keeps the sigil because it matches in-body uses; a node does not.
- C/C++ `type_name` is the declared type specifier without the declarator's pointer or reference marks (`const char *b` → `char`), the same reading `extract_type_facts` gives a return type. `OF_TYPE` resolves by the names inside the annotation, so this loses nothing for resolution.

## Tests

- `test_parameter_nodes_languages.py`: 35 cases on tree-sitter nodes, every one a shape where the taint slot table answers differently (receiver, variadic, default, unnamed slot, curried list, destructuring, promoted property, initialising formal, sibling default value in Dart).
- `test_parameter_nodes_languages_end_to_end.py`: one twelve-language fixture through `GraphUpdater` + the eval store, asserting each callable's rows, the full `OF_TYPE` set (every language resolves its `Widget`), that every source file contributes, and that no Parameter is orphaned. This is what proves JS/TS's own ingest module and the deferred C++ out-of-class method reach the emitter.
- Mutation run: sixteen single-axis mutations of the enumerators; each reddened exactly the test named for it. The Lua `...` position is unobservable (the grammar requires it last) and the test says so in its docstring.

Grammar shapes were read from a parse dump of each language's grammar before any reader was written, per the "same syntax, different semantics" rule: Dart's default value is a *sibling* of the parameter, C#'s `params` is not wrapped in a `parameter`, C++'s C-style `...` is an anonymous token while C's is a named `variadic_parameter`, TS wraps the real pattern in `required_parameter`.

## Local review round

One `greptile-local` round on the first commit found three P1s and three P2s, all fixed in the second commit with a red test each: Rust closure parameters walked named children only, so a bare pattern or `_` took no position; Java `final String... xs` read `modifiers` as the element type; Dart annotations (`@Deprecated('x') int b`) leaked into `type_name`; Dart `super.x`, old-style function-typed parameters (`void cb(int i)`) and the `:` default separator were not read. The same varargs mistake exists in `extract_type_facts` on `main`; filed as #1964 rather than widened here.

## Summary by CodeRabbit

* **New Features**
  * Added parameter declaration support across 13 programming languages, including variadic, default, receiver, destructured, unnamed, and typed parameters.
  * Parameter indexes, names, types, defaults, and variadic status are now captured more consistently across supported language syntaxes.
  * Parameter information is now available throughout indexing and graph relationships for supported languages.

* **Tests**
  * Added comprehensive language-specific and end-to-end coverage to validate parameter extraction and type relationships.


vitali87:
@greptileai review

vitali87:
@greptileai review

Issue #2017, opened 2026-09-19
C#: every `split(DUP_QN_MARKER, 1)[0]` reader mistakes a verbatim identifier (`@event`) for a duplicate marker


vitali87:
## Symptom

A C# verbatim identifier opens with `@` (`class @event`, `class @lock`), the same character as the duplicate-qn marker (`Box@12`). Every reader that derives a natural name with `qn.split(cs.DUP_QN_MARKER, 1)[0]` turns `@event` into the empty string: `_natural_qn` in `graph_updater.py` and the other marker readers (constructor lookups, the C++ destructor name, the #2014 partial key before its review fix).

Measured on the #2014 branch by its local review: two verbatim-named partial types in one directory merged into one group and cross-typed each other's fields; the same readers on `main` have the same blind spot for any verbatim-named type.

## Mechanism

The marker is `@` followed by a line number; the readers cut at the first `@` without checking that a number follows.

## Expected

One shared helper strips only a numeric marker suffix (`head, sep, tail = qn.rpartition("@"); head if sep and tail[:1].isdigit() else qn`, as #2014 now does for the partial key), and every marker reader goes through it, with a verbatim-named fixture in the test.

## Provenance

Found by the local review of #2014, which fixed the partial-key reader only.

Issue #2016, opened 2026-09-19
C#: reingest never prunes the syntactic partial index, so a re-parsed file appends its part qns again


vitali87:
## Symptom

Re-parsing a file that holds a partial part on a reused updater (`GraphUpdater.reingest`) appends the part's qns to its `_csharp_partial_index` group again: two same-file parts become `['Bench', 'Bench@5', 'Bench', 'Bench@5']`, and a cross-file partial becomes `['A.App.Bench', 'B.App.Bench', 'A.App.Bench']`.

Measured on the #2014 branch (ff036b14) by its local review; the index code is unchanged from `main`.

## Mechanism

`class_ingest/mixin.py` appends every partial part to `self._csharp_partial_index[key]` at ingest and nothing removes a module's entries before that module is re-ingested. Resolution tolerates the duplicates today (visited sets, `min(group)`), but `_join_csharp_partials` in `graph_updater.py` does `old.remove(qn)`, which removes one occurrence only, so a joined group can keep a stale duplicate.

## Expected

Re-ingesting a module drops that module's part qns from every group before adding them back, so a group holds each part once.

## Provenance

Found by the local review of #2014 while checking incremental re-index of a two-part file.

PR #2018, opened 2026-09-19
fix(csharp): same-file partial parts share one partial group

### LINKED ISSUE #2014: C#: two partial declarations of one class in the same file are never joined into a partial group
## Symptom

```csharp
// src/App.cs
namespace App
{
    public partial class Bench
    {
        public void Run() { config.Each(base.Handle); }
    }

    public partial class Bench : BaseHandler { }
}
```

Measured on `main` (3cd90fe8) with the eval store: the second part registers as `proj.src.App.App.Bench@16`, and `csharp_partial_groups` holds two singleton groups, `{Bench: [Bench], Bench@16: [Bench@16]}`, with the base list under `Bench@16` only. Any lookup that spans the parts through `_partial_roots` (members, bases, `base.X` from #2011) sees one part. Splitting the same two parts across two files (`App.cs` and `BenchBase.cs`) joins them as expected.

## Mechanism

The syntactic partial key in `class_ingest/mixin.py` is the declaring directory plus `class_qn[len(module_qn) + 1:]`. A second same-name part in one file gets the duplicate marker (`Bench@16`), so its key suffix is `App.Bench@16`, not `App.Bench`, and the two parts never share a group. The Roslyn join replaces the syntactic groups only when facts are available.

## Expected

Same-file parts of a partial type share one group, the same as cross-file parts: strip `DUP_QN_MARKER` from the key's last segment (the marker is a registration artefact, not part of the declared name), so both parts key to `proj.src.App.Bench`.

## Provenance

Found in passing while answering a review thread on #2011; the fixture there uses two files to sidestep it. Related: a dotted file stem (`Bench.Base.cs`) also defeats the key on `main` because the directory is taken by a single `rsplit`; #1999 fixes that one.

vitali87:
Closes #2014.

Two partial declarations of one C# class in the same file register as `Bench` and `Bench@24` (the duplicate marker), and the syntactic partial key was the declaring directory plus the class qn's suffix after the module qn, so the second part keyed to `...App.Bench@24` and never shared a group with the first. Every lookup that spans the parts through `_partial_roots` saw one part: a field declared on the second part left a receiver in the first part untyped, and the bare-name fallback bound whichever same-named method sat nearest. Parts split across files joined as expected.

The key now strips the duplicate marker from the declared-name suffix, so both parts key to `proj.src.App.Bench`. The marker is a registration artefact, not part of the declared name, and the group holds each part's own qn (with its marker) as before.

Test (`test_csharp_same_file_partial_parts.py`, real parsers over the eval store): a receiver typed by a field declared on the other same-file part binds `Widget.Go`, not the decoy's `Go`; red on `main` (the decoy bound), red again with the marker kept in the key.

After the local review: a verbatim identifier (`@event`, `@lock`) opens with the marker's character, and cutting at it emptied the declared name so every verbatim-named partial type in a directory merged into one group; the key now strips only a numeric suffix, with a test (red with the cut at the first `@` reinstated). The same blind spot in the other marker readers is #2017, and the partial index's growth on reingest is #2016.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved C# partial-type handling so numeric registration suffixes are removed correctly.
  * Preserved verbatim identifiers such as `@event` when resolving partial types.
  * Improved member and method resolution across partial class declarations in the same file.
  * Prevented distinct verbatim-named types from being incorrectly merged.


vitali87:
@greptileai review

PR #2019, opened 2026-09-19
fix(dart): a named constructor call records INSTANTIATES on its class

### LINKED ISSUE #2012: Dart: a named constructor call (`Box.of(1)`) emits CALLS to the constructor but no INSTANTIATES edge
## Symptom

```dart
class Box<T> {
  Box(T value);
  Box.of(T value);
}

void run() {
  final a = Box(1);        // INSTANTIATES Box, CALLS Box.Box
  final b = Box.of(1);     // CALLS Box.of, no INSTANTIATES
  final c = new Box.of(1); // same, once #2010 lands
  final d = Box<int>.of(1);// same
}
```

Measured on `main` (3cd90fe8) and on the #2010 branch with the eval store.

## Mechanism

A named constructor registers as a method `Box.of` and the call `Box.of(1)` resolves to it directly, so the class branch of the call processor (which emits INSTANTIATES and redirects CALLS to every constructor) never runs: it runs only when the callee resolves to the class itself. The comment there records the named form as "invoked by its own name and resolves as an ordinary method", which is right for the CALLS edge but leaves the construction unrecorded, so a class only ever built through named constructors (factory-style `Widget.fromJson`, `Duration.zero`-like statics aside) has no INSTANTIATES edge at all.

## Expected

A call whose target is a constructor of a registered Dart class (a method whose qn is `<class>.<name>` and whose declaration is a `constructor_signature`) emits INSTANTIATES to that class beside the CALLS edge, for the bare, `new`, and type-argument forms alike.

## Provenance

Found in passing while probing #2010 (`Box.of(1)` and `new Box.of(1)` in the same fixture).

vitali87:
Closes #2012.

A Dart named constructor (`Box.of`, `const Box.fixed`, `factory Box.build`) registers as a method under its own name, so `Box.of(1)` resolves straight to that method and the call processor's class branch, which records INSTANTIATES and redirects CALLS to the constructors, never runs for it. The call kept its CALLS edge and the class got no INSTANTIATES, so a class built only through named constructors had no construction recorded at all.

The class-ingest pass now records every Dart constructor signature's qn (default, named, const and factory) in a constructor set that reaches the call processor through the type-inference tables, and a resolved Dart callee in that set that is not the class itself takes INSTANTIATES on its owning class (every class-typed variant) beside the CALLS edge the method path already emits. A static method that merely returns the class is not in the set and stays a plain call.

Test (`test_dart_named_constructor_instantiates.py`, real parsers over the eval store): the named, const and factory forms take INSTANTIATES `Box` beside CALLS to their constructor; the static factory takes CALLS only. Mutations, each restored: constructor set never filled (three named forms red), emission off (same three red), every Dart method treated as a constructor (the static-factory control red). Related Dart test files: 71 passed.

After the local review: the constructor set is per-file state, so it is dropped in the same sweep that forgets a re-parsed or deleted file's return types (a named constructor rewritten as a static factory no longer leaves a stale qn stamping INSTANTIATES; test on a reused updater, red without the prune), and twin classes take the `overload` stamp the class branch gives a two-candidate construction (test on the edge property, red without it).

## Summary by CodeRabbit

* **New Features**
  * Dart named-constructor calls now record class instantiation relationships alongside constructor calls.
  * Instantiation relationships accurately distinguish overloaded class definitions.

* **Bug Fixes**
  * Prevented stale instantiation relationships when code changes a named constructor into a static factory method.

* **Tests**
  * Added coverage for named, overloaded, and factory-constructor scenarios.


vitali87:
@greptileai review

vitali87:
The integration job failed pulling `testcontainers/ryuk:0.8.1` (Docker Hub returned a 500 on the manifest request), unrelated to the change; re-triggered with an empty commit.

vitali87:
@greptileai review

PR #2032, opened 2026-09-19
fix(dart): bind member reads on construction receivers (new, const, generic) (#2015)

### LINKED ISSUE #2015: Dart: a member read on a construction receiver (`Box(1).width`, `new Box(1).width`) emits no REFERENCES edge
## Symptom

```dart
class Box<T> {
  Box(T v);
  int width = 1;
  String describe() => 'b';
}

int bareRead() { return Box(1).width; }       // INSTANTIATES Box, CALLS Box.Box, no REFERENCES Box.width
int newRead() { return new Box(1).width; }    // same (after #2013)
int genericRead() { return Box<int>(1).width; } // same (after #2013)
void chained() { Box<int>(1).describe(); }    // CALLS Box.describe: the CALL chain types its receiver
```

Measured on `main` (3cd90fe8) for the bare form and on the #2013 branch (fff0efcf) for all three, with the eval store: no form records the `width` read.

## Mechanism

`dart_member_read_name` (codebase_rag/parsers/dart/utils.py) walks the receiver chain with `_walk_chain` WITHOUT `allow_calls`, so a call-result receiver breaks the chain and the read has no static name. The call pass walks the same chain with `allow_calls=True` and types a `Box(...)` hop from the constructed class, which is why `Box<int>(1).describe()` binds while `Box<int>(1).width` does not.

## Expected

A member read on a construction receiver names the member on the constructed class (`Box.width`), for the bare, `new`, `const` and type-argument forms alike, the way the call pass already types the same receiver. A read on a plain call-result receiver (`factory().width`) can follow through the callee's recorded return type.

## Provenance

Measured while answering a review thread on #2013; the thread asked for the type-argument form only, but the bare form on `main` has the same gap, so the fix belongs to the read pass rather than to the generic shape.

vitali87:
Closes #2015.

## Problem

A Dart member read on a construction receiver emitted no REFERENCES edge unless the receiver was the bare `X(1)` form. The call pass already types the same receiver, which is why `Box<int>(1).describe()` bound while `Box<int>(1).height` did not.

Three shapes were unhandled by the shared chain walker, for different reasons:

| Form | Parse | Why the chain broke |
|---|---|---|
| `new X(1).m` | `new_expression` | a base `_chain_part` rejected outright |
| `const X(1).m` | `const_object_expression` | same, and a *different* node type from `new` |
| `X<int>(1).m` | `relational_expression` | no construction node at all: the grammar reads `<`/`>` as comparisons and splits the call's parens into a sibling `parenthesized_expression`, where the walk bottomed out |

## Fix

All three now contribute the constructed class plus a `()` hop, so the receiver types through the same resolver path the bare form already uses. Since the generic walk reaches the call's parens rather than the construction, a parenthesized hop defers to its parent.

## The ambiguity, and why a shadow check is needed

`X<int>(1).x` and the chained comparison `a < b > (1).x` are token-for-token identical after parsing; no tree-shape check can separate them. Requiring `<` then `>` excludes the flipped order, but not the same order.

That residual ambiguity is **not** harmless. When the left operand also names a class, the resolver binds the fabricated receiver and emits a wrong edge:

```dart
dynamic cmp(dynamic Box, dynamic b) { return Box < b > (1).height; }  // wrongly bound Box.height
```

So a local or parameter of that name now suppresses the construction reading, reusing the shadow spans the bare-read path already computes: if the operand is a variable in scope, the site is a comparison, not a construction. Where no such variable exists and no class matches, the name resolves to nothing and no edge is emitted.

## A note on the issue's example

The issue reads a plain field (`Box(1).width`). Dart extraction does not record plain fields as nodes, so `width` has nothing to bind to and that example cannot produce an edge on any branch. The gap is real for members the read pass can bind, so it is reproduced here on a getter. The field-node gap is separate (the Parameter/Constant-node family, #1804/#1806).

## Tests

- `test_construction_receiver_getter_read_is_referenced` — bare, `new` and type-argument forms, with the already-working bare read and generic CALL as controls.
- `test_const_construction_receiver_read_is_referenced` — the `const` form.
- `test_real_comparison_is_not_read_as_a_construction` — both operator orders, using a local named after the class, the only shape in which a fabricated receiver resolves.

Every fix condition was mutated individually and each turns a test red on its own: the `new_expression` branch, the `const_object_expression` branch, the parenthesized deferral, the operator-order guard, and the shadow suppression.

Two of these tests earned their keep only after being rewritten: the first guard test passed either way (its fixture's own reads satisfied the assertion), and its replacement asserted on the operator order the guard rejects structurally, so it discriminated nothing. That second gap is what let a wrong-edge bug ship green in the first commit here; it was caught by the local review and is fixed in the second.

Dart suites on the pushed head: 134 passed, 5 skipped, 0 failed. A wider run before the last fixes: 760 passed, 0 failed, 71 errors all `@memgraph` (the documented Docker-less baseline on this host, identical on `main`).

## Follow-up spotted, not fixed here

`outer.Box<int>(1).height` (an import-prefixed generic construction) reassembles as `outer().height`, dropping the `.Box` hop. Same mis-parse family, distinct shape; I'll file it separately rather than widen this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Dart code analysis for getter references on constructed objects, including `new`, `const`, and generic constructions.
  - Prevented comparisons that resemble generic constructions from generating incorrect references or construction calls.
  - Preserved accurate method call detection for generic object constructions, including cases involving shadowed identifiers.

- **Tests**
  - Added coverage for constructed-object getter reads, constant constructions, generic constructions, shadowed identifiers, and comparison expressions.


vitali87:
@greptileai review

vitali87:
Fixed in 2df61f70.

The false `CALLS` edge is real and confirmed: `Box < b > (1).abs()` with a parameter named `Box` bound `Box.abs`.

One correction to the report, which makes this worse rather than better. It states both the parent and PR-head executions report `FALSE_BOX_ABS: True`. Measured in a clean worktree checked out at `origin/main`, the parent does **not** reproduce it:

```
origin/main:  FALSE_BOX_ABS: False   edges: []
this branch:  FALSE_BOX_ABS: True    edges: ['a.f--CALLS-->a.Box.abs']
```

So it is a regression introduced by this PR, not pre-existing.

**Fix.** `_get_call_target_name` has no access to scope, so it cannot apply the shadow check that makes the ambiguous reading safe. Instead of threading shadow spans through that path, the ambiguous relational reading is now opt-in (`allow_ambiguous`), and only the read path takes it.

**The call path loses nothing by declining.** Measured by disabling the new branch outright: `X<int>(1).m()` and `X(1).m()` both still bind through the pre-existing `argument_part` hop, so the generic call never needed this code. Only `new X(1).m()` depends on it, and that shape is unambiguous. `test_shadowed_comparison_emits_no_construction_call` asserts exactly that split and goes red if the gating is removed.

Dart suites: 135 passed, 0 failed.

@greptileai review

vitali87:
Fixed in 3949467c — the fixture now declares `class Box<T>`.

Correct catch: `Box<int>(1)` against a non-generic `Box` is `disallowed_type_instantiation_expression`, so the fixture was not valid Dart.

I checked that this does not weaken the test before applying it. The call-site parse is context-free here, so the declaration cannot change what is exercised: `Box<int>(1).abs()` parses to an `argument_part` selector either way, and the mutation that removes the ambiguity gating still turns `test_shadowed_comparison_emits_no_construction_call` red.

The two non-generic `Box` fixtures in `test_dart_getter_reads.py` are deliberately left alone: one uses `Box` as a comparison OPERAND (`Box < b > (1).height`, where it is a `dynamic` parameter, not a type) and the other as `const Box(1)`. Neither applies type arguments, so neither triggers the diagnostic.

@greptileai review

vitali87:
Ready to merge, blocked only on a human approval.

Gate verified against the exact head `3949467c`:

- `scripts/check_pr_gated.py 2032` -> **gated** ('All Checks Pass' present and satisfied)
- every check pass or skipping, none failed
- Greptile **5/5**, and its anchor SHA is `3949467c` itself, not a predecessor
- 0 unresolved review threads (the CodeRabbit thread auto-resolved when the line it sat on changed)

`gh pr merge` is refused by the classic protection layer (`reviewDecision=REVIEW_REQUIRED`), and `--auto` is unavailable because auto-merge is disabled for this repository. Not using `--admin` on my own PR. Over to you for the approval.

PR #2034, opened 2026-09-19
docs: correct the C# frontend default and document the web-search egress path (#2028)

### LINKED ISSUE #2028: C# backend description disagrees with the current Tree-sitter default, and its external-data-flow description omits the web-research/search path

vitali87:
Closes #2028.

Two claims in the security model had drifted from the code. Both verified by execution, not by reading.

## The C# default was wrong, in the unsafe-sounding direction

`security.md` and `graph-schema.md` both said the frontend defaults to `CSHARP_FRONTEND=auto`, resolving to Roslyn wherever `dotnet` is on PATH. `security.md` drew the conclusion:

> indexing a C# repository runs parts of that repository's build with your privileges, **and this is the default**

Commit 3990056f ("harden C# indexing") changed the shipped default to `treesitter` and the docs were never updated. So the security model has been *understating* the project's own safety, and telling readers to set a value that was already in force.

```
AppConfig.model_fields['CSHARP_FRONTEND'].default  ->  treesitter
```

The build-executing behaviour is real for `auto`/`hybrid`/`roslyn`, so it is kept and attributed to those opt-in modes rather than deleted. The C++ `hybrid` default in `language-frontends.md` is a genuinely different matrix and is untouched.

## A third egress path was undocumented, and it is the unconfigured one

Requirement 4 is that the paths by which code can leave the machine are documented. The boundary section listed two, "both chosen by configuration". The CLI agent's `research` tool is a third, and unlike the other two it needs no configuration at all: the search backend registers unconditionally with keyless DuckDuckGo (`main.py`, "The search backend registers unconditionally").

Documented with what is actually true of it:

- what crosses (the model-composed query) and what cannot (a query carrying a verbatim span of recorded repository content is **refused**, #1128) — including the limit that keeps that honest, since it is a cheap string match and paraphrased content still passes;
- the sub-agent isolation, so fetched web content never shares a context with repository reads and shell;
- that MCP exposes neither `research` nor `web_search`, deliberately.

## Requirement 4 itself

Adding a path that needs no configuration made the page self-contradictory: requirement 4 asserted every egress path is "controlled by explicit configuration", which the shipped design does not do. Amended to the property the code actually maintains — documented, and either configured **or gated** so repository content cannot cross unrecognised. Caught by the local review, not by me.

## Checked while here

The `DOTNET_CLI_TELEMETRY_OPTOUT=1` the C# frontend sets is now recorded, since it sits in the paragraph warning about running the analysed build. No other undocumented egress exists in a session: no telemetry, no update check, no image pull. The only other outbound call in the tree is a PyPI fetch whose sole non-test caller is `scripts/generate_readme.py`, i.e. build-time tooling.

## Follow-ups not taken here

`docs/getting-started/configuration.md` has a `CPP_FRONTEND` row but none for `CSHARP_FRONTEND`, `WEB_SEARCH_PROVIDER` or `SERPDIVE_API_KEY`, so the configuration reference never states the C# default or the search backend at all. Identical on `main`; I will file it rather than widen a docs fix into a different file.

## Summary by CodeRabbit

- **Documentation**
  - Identified `treesitter` as the shipped default for the C# frontend.
  - Clarified that `auto`, `hybrid`, and `roslyn` request Roslyn and fall back to tree-sitter when .NET is unavailable.
  - Documented logging levels for expected and explicit fallbacks.
  - Clarified that the .NET telemetry opt-out does not restrict build, restore, or source-generator network access.
  - Documented egress paths, exact egress-gate matching, redirect behavior, and opt-in package restoration.


vitali87:
@greptileai review

vitali87:
Re-triggering the review: the 5/5 above is anchored to `f3695101`, and the head is now `94711b4c` after an empty commit that re-triggered a stuck CI run. The diff is byte-identical between the two, but the anchor rule is anchor-to-head, so the artifact needs to name the commit that would merge.

For the record on why the empty commit was pushed: the `CI` run scheduled at 21:10 sat in `queued` for 31 minutes with none of its 15 jobs dispatched (`createdAt == updatedAt`), while every other workflow on the branch completed and only 2 runs were queued repo-wide. The superseded run has been cancelled.

@greptileai review

vitali87:
All five findings fixed in ae0a7fed. Every one verified against the code first, and every one was right — they all point the same way, that my wording claimed more than the implementation guarantees, which in a security document is the dangerous direction to be wrong in.

**`hybrid`/`roslyn` do not force Roslyn.** `resolve_frontend_mode` degrades every toolchain-backed mode to the fallback when the toolchain is absent, and the C# resolver's own comment says an explicit HYBRID/ROSLYN 'degrades to tree-sitter with a warning'. Corrected in both documents. Worth flagging: my local reviewer cleared this exact phrase as 'checked, not a defect' — it verified on a host WITH `dotnet` installed, where the distinction is invisible. A second reviewer on a different environment is what caught it.

**Telemetry claim narrowed.** `DOTNET_CLI_TELEMETRY_OPTOUT=1` disables .NET CLI telemetry and establishes nothing about MSBuild, restore or the analysed project's source generators. Reworded, with the absence of any network restriction on them stated.

**The egress gate is narrower than 'any verbatim span'.** `TAINT_SPAN_CHARS = 24`: a long recording matches on a 24-character window, a short one only as a complete token. A shorter quote passes just as a paraphrase does. Now stated with the window, since a reader deciding what is safe to put in a query needs the real rule, not a flattering summary of it.

**The TLS guarantee was not enforced.** Confirmed both halves: base URLs are accepted as given including `http://` (`OLLAMA_BASE_URL` ships as plain HTTP), and `web_search.py` sets `follow_redirects=True` with no HTTPS target requirement. Reworded to httpx's default certificate verification plus the dependency on configured endpoints.

**C# build-time egress added to the inventory.** The boundary paragraph now counts four paths — three query-time, one build-time — in that order, with `CSHARP_FRONTEND=treesitter` named as the control.

@greptileai review

vitali87:
Fixed in a3a645c1. Correct finding, and verified independently at `graph_updater.py:1099-1101`:

```python
if settings.CSHARP_FRONTEND == cs.CSharpFrontend.AUTO:
    logger.info(ls.CSHARP_FRONTEND_AUTO_FALLBACK)
else:
    logger.warning(ls.CSHARP_FRONTEND_UNAVAILABLE)
```

The code's own comment says the split is deliberate: 'AUTO promises hybrid only where the toolchain exists, so a missing dotnet is the expected fallback (info); an EXPLICIT hybrid/roslyn request that cannot run stays a warning.'

My previous commit said every Roslyn-backed mode degrades 'with a warning', which flattens that. Your point about the consequence is the part that made it worth fixing rather than rewording: someone watching for a warning to learn that automatic Roslyn analysis did not run would miss the `auto` case entirely. Both documents now state the fallback without attributing a level, then name the two levels separately.

This is the second correction in a row to the same sentence, and both times the error was mine adding a detail I had not checked -- first 'force it', now 'with a warning'. The underlying fact (every Roslyn-backed mode degrades without the toolchain) was right; the embellishments were not.

@greptileai review

vitali87:
Superseded by #2038; closing rather than rebasing.

While this was in review, `8806eec3` ('docs: refresh security threat model and trust boundaries') rewrote `docs/architecture/security.md` on `main` and independently fixed the headline half of #2028: the C# default is stated as `treesitter` in both documents, requirement 4 is reworded, and the research/search path is documented as a data path with the note that no search API key is required for it to be available. `graph-schema.md` on `main` now reads almost exactly what this PR proposed for it.

That left this branch conflicting in both files and defending a shrinking delta, so rebasing four commits onto a rewritten document was the wrong shape. #2038 instead carries the four details the rewrite does not state, each established by execution during THIS PR's review and re-verified against the code, written into the new structure's voice rather than restoring my superseded wording:

- `DOTNET_CLI_TELEMETRY_OPTOUT=1` covers CLI telemetry only;
- `auto` logs its downgrade at INFO while explicit modes log WARNING;
- the taint window is 24 characters, so a shorter verbatim quote passes;
- TLS is not enforced (`http://` base URLs accepted, `follow_redirects=True`).

Thanks to both bots — the five CodeRabbit findings and the Greptile log-level probe are what produced everything that survived into #2038. Closing this one with no work lost.

PR #2038, opened 2026-09-19
docs: four verified details the security rewrite left open (#2028)

vitali87:
Part of #2028. Replaces #2034, which is superseded.

## Why this is a new PR

While #2034 was in review, `8806eec3` ("docs: refresh security threat model and trust boundaries") rewrote `docs/architecture/security.md` on `main` — 167 lines — and independently fixed the headline half of #2028:

- the C# default is now stated as `CSHARP_FRONTEND=treesitter`, in both this document and `graph-schema.md`;
- requirement 4 is reworded away from "controlled by explicit configuration";
- the research/search path is documented as a data path, including that **no search API key is required for it to be available** and that the taint check is not comprehensive DLP.

That is most of what #2034 did, arrived at independently. Rebasing #2034 onto it would have meant resolving four commits against a rewritten document to defend a shrinking delta, so I closed it and rebuilt the remainder as one commit written into the new structure's own voice.

## What is left, and why it is worth keeping

Four details the rewrite does not state. Each was established by execution during #2034's review — two by CodeRabbit, one by a Greptile probe, one by me — and each is re-verified against the code here:

| Detail | Evidence |
|---|---|
| `DOTNET_CLI_TELEMETRY_OPTOUT=1` disables **CLI** telemetry only; restore, MSBuild and the analysed project's source generators are unrestricted | `_DOTNET_ENV`, `frontend.py` |
| A missing toolchain logs an explicit `hybrid`/`roslyn` at **WARNING** but an `auto` downgrade at **INFO** — so watching only for warnings misses the `auto` case | `graph_updater.py:1098-1101`, where the code comment says the split is deliberate |
| The taint check matches a long recording on a **24-character** window and a short one only as a complete token, so a shorter quotation passes | `TAINT_SPAN_CHARS = 24`, `taint.py` |
| TLS is **not enforced**: base URLs are accepted as given including `http://` (the shipped Ollama default), and the search client sets `follow_redirects=True` with no HTTPS target requirement | `config.py:211`, `web_search.py` |

The first two matter to anyone deciding whether an indexing run was safe; the last two matter to anyone deciding what is safe to put in a query. The rewrite already says paraphrased content passes the taint check — the length floor is the other half of that same limit.

## Scope

19 added lines in one file, no restored wording from #2034, and `graph-schema.md` untouched since `main` now says what that PR proposed for it.

## Summary by CodeRabbit

* **Documentation**
  * Expanded security guidance for repository parsing, including C# CLI telemetry settings, network behavior, and fallback to Tree-sitter when the toolchain is unavailable.
  * Clarified logging levels for explicit versus automatic frontend modes.
  * Detailed recording text-matching rules for values below and at least 24 characters.
  * Documented certificate verification, accepted HTTP or HTTPS base URLs, and redirect behavior for search requests.


vitali87:
@greptileai review

vitali87:
Fixed in a2d54cea. Both halves correct, and both confirmed by running the live implementation rather than reading it:

```
record('api key here')    -> taints('what is api key here for')  = True   # multi-word, 12 chars
record('SECRET_TOKEN_X')  -> taints('explain SECRET_TOKEN_X')    = True   # complete short quote BLOCKED
record('x'*80)            -> taints('xxxxxxxx')                  = False  # short quote from a LONG recording passes
```

So my wording had it backwards in both directions. 'Entirely one token' is not the rule — `record()` classifies on normalised length against `TAINT_SPAN_CHARS`, so a multi-word value under 24 characters is a short recording too. And 'a shorter quotation passes' is true only of quotes taken from LONG recordings; a short recording is matched as a complete value in either direction, so quoting one in full is blocked however short it is.

Reworded around the length split, which is the thing that actually determines the outcome, with the reason the split exists (a common short value must not refuse every later query containing it inside a longer word).

This is the point of the paragraph, so it is worth getting exactly right: a reader predicting when research egress is refused would have got both of these cases wrong from what I wrote. Good catch.

@greptileai review

Issue #1936, opened 2026-09-14
check_pr_gated.py accepts a review of an older commit: the artifact's anchor is never compared to headRefOid


vitali87:
## Symptom

`scripts/check_pr_gated.py` reports a PR as gated on the strength of a review artifact that reviewed an **older commit**. A stale review and a fresh one are indistinguishable to it.

## Measured

PR #1906 right now:

```
head:    7928c5ba8efe8af0f8d3f6b943171eb016deee72
anchor:  9aeb04f825538b27f0ff25fd31b56bbf08679000   (Greptile, updated 2026-09-13T21:25Z)
```

The head is two pushes past the reviewed commit, and those two commits are the substantive ones: a duplicate-edge fix and a dead-code removal. The script's only remaining objection is `'All Checks Pass' has not concluded`. When that check concludes, it will report the PR gated.

## Mechanism

The script does verify a review artifact positively: `REVIEW_VERDICT_MARKERS` requires a verdict phrase, and `TRUSTED_REVIEWERS` requires the author be a known bot rather than the PR author (added for CWE-345 after a spoofing finding on #1625). Both good.

What it never does is compare the artifact's commit anchor to `headRefOid`. It fetches `headRefOid` (line 626) for the CI-run checks, but no review path uses it. So a review that names `Last reviewed commit: <sha>` satisfies the gate whatever that sha is.

This is the exact trap `CLAUDE.md` warns about under "Verifying a review actually happened":

> Anchor the artifact to the head SHA. Bots name the commit they reviewed ("Last reviewed commit: `<sha>`"); if that is not the head you are about to merge, the review is stale and does not count, however good it looks. A stale review and a missing review have the same consequence: do not merge yet. **There is no CI job that enforces this any more** (the Greptile gate was dropped because the bot re-scores in place and the gate lagged behind), **so the check is yours to make.**

The doc is right that no CI job enforces it, and says the check is the agent's to make by hand. But the same doc also says of this script: "Its output is the state; a check list, a `mergeStateStatus`, and your recollection are not." An agent that follows that instruction literally will merge on a stale review, because the tool it was told to trust does not model staleness.

## Fix shape

Extract the 40-hex sha from the review body the same way the doc's own snippet does, and report a reason when it is not `headRefOid`:

```
- review artifact anchors to 9aeb04f8, not the head 7928c5ba, so it reviewed
  neither of the 2 commits since
```

[...]

PR #2086, opened 2026-09-20
feat: EnumVariant nodes under HAS_VARIANT for Rust, Java, C, C++, C# and PHP enums (#1807)

### LINKED ISSUE #1807: Add an EnumVariant node label (greenfield: no variant extraction exists today)
## Summary

Add an `EnumVariant` node label. This is the only one of the four granularity items with **zero existing extraction** — enums are classified as nodes, but their variant bodies are never walked.

## Why it matters, and why it is ironic

Enum variants are among the most consistently documented symbols in real code: in 1,500 sampled crate files from `~/.cargo/registry`, 373 enum-variant `///` doc comments appeared in an 800-file subsample, and 32% of all Rust `///` blocks precede a field/variant rather than a named definition. So this is where doc comments most want to attach — and it is the most expensive to build.

## What exists today

Enum *classification* works. `determine_node_type` (`codebase_rag/parsers/class_ingest/node_type.py:26-33`) matches `TS_ENUM_DECLARATION | TS_ENUM_SPECIFIER | TS_ENUM_CLASS_SPECIFIER | TS_RS_ENUM_ITEM`, and `NodeLabel.ENUM` (`constants/graph.py:197`) / `NodeType.ENUM` (`types_defs.py:108`) exist.

But every tree-sitter query captures the **declaration node only, as `@class`**, and never descends:

| Language | language_spec.py |
|---|---|
| Rust | 427-428 — `(enum_item name: (type_identifier) @name) @class` |
| Java | 507-508 |
| C | 537 — `(enum_specifier) @class` |
| C++ | 564, 568 |
| PHP | 605-606 |
| C# | 659 |

Greps for `enum_variant`, `enum_member`, `enum_constant`, `enumerator`, `enum_body` across `codebase_rag/` return **no extraction code**. `constants/ast_rust.py:20` defines `TS_RS_ENUM_ITEM` but there is no `TS_RS_ENUM_VARIANT`. Every `variant` hit in the codebase is the unrelated duplicate-QN machinery in `function_registry.py:89-98`.

The only variant-adjacent code is `parsers/rs/utils.py:398-401`, which treats an enum-discriminant expression block as a scoping boundary — it walks *past* variants, not into them.

## Work required — pure greenfield

- ~15 new extractors, one per language
- ~15 new grammar-node constants
- New query fragments in `language_spec.py` for every language
- New dispatch (no existing hook)
- Positions, docstrings, discriminant values

Rust variants carry payloads (tuple and struct variants), so the shape is not uniform even within one language.

## Proposed shape

```
EnumVariant {qualified_name, name, path, absolute_path,
             start_line?, start_col?, value?, docstring?, index?}

Enum -[:HAS_VARIANT {index}]-> EnumVariant
```

## Recommendation

**Do this last of the four.** It has the highest per-language cost and no scaffolding to reuse. The `Parameter` and `Field` issues should establish the node-emission and position-capture patterns first.

Consider scoping an initial version to the languages where enums are most load-bearing and best documented — Rust, TypeScript, Java, C# — rather than all 15 at once.

vitali87:
Closes #1807.

Adds an `EnumVariant` node label with `Enum -[:HAS_VARIANT {index}]-> EnumVariant`, opt-in through the `enum_variants` capture group, for the six languages where enums are most load-bearing: Rust, Java, C, C++, C# and PHP. The issue recommends scoping an initial version rather than doing all 15 at once, and that is what this is.

Each variant carries its qualified name, name, path, absolute path, the position of its NAME, its declaration index, the discriminant as written where it has one, and its doc comment.

## Provenance

The feature commit was not written in this session. It sat on an unpushed local branch in this clone, complete with tests, which is why `gh pr list` showed #1807 as unclaimed — that command cannot see branches that were never pushed. I confirmed with the other session on this machine that the branch was not theirs, rebased it onto current `main` (clean cherry-pick), and reviewed it as adopted code rather than assuming it was sound.

Two of the three commits here are the result of that review.

## What the review changed

**The opt-in test did not test the opt-in.** `test_the_default_index_emits_no_variant_and_no_edge` passed with the `rel_enabled` early return in `emit_declared_variants` deleted entirely — twice, on a quiet host. There are two independent gates: the in-function check, and `services/filtering.py`, which drops the nodes and edges regardless. The default index is empty either way, so the assertion could not distinguish working code from broken. Added a test that asserts the gate directly, with the variant count as a known-positive control so an empty result cannot be read as a parse that found nothing.

**The eval double drifted from the delete query.** `CYPHER_DELETE_MODULE` gained `HAS_VARIANT`; `_MODULE_SUBTREE_RELS` in `evals/cgr_graph.py`, a hand-maintained copy of the same walk, did not. The repo ships `TestTheModuleSubtreeWalkMirrorsTheDeleteQuery` to catch exactly this — it had already lost `CONTAINS_SECTION` once (#1938) — and it was **red on this branch**:

```
AssertionError: the double's module subtree drifted from CYPHER_DELETE_MODULE:
  only in the query ['HAS_VARIANT'], only in the double []
```

Past the red test, the consequence is that the double stops modelling production: an `_StatefulIngestor` test re-indexing an enum would see orphaned variants that production deletes, and could pin the opposite of real behaviour.

I missed this in my own verification because I chose "affected" test files by name resemblance — enum, field, capture, protobuf — and nothing in `test_eval_double_query_coverage.py` suggests enums. The correct selector is the changed symbol, not the theme.

**Three assertions that were not being made.** Each confirmed to redden on its own mutation:

| Added | Mutation that proves it | Result |
|---|---|---|
| a re-parse drops a stale variant with its owner | remove `HAS_VARIANT` from the double | red |
| the `HAS_VARIANT` edge carries its index | edge `index` -> constant `0` | all six languages red |
| a variant is positioned at its NAME | `name_node.start_point` -> `node.start_point` | red |

The re-index test mirrors `test_a_reparse_takes_stale_parameters_with_their_owner`; its absence is why the drift reached the branch through the feature's own test file. For the position test, PHP's `case Hearts = 'H';` puts the name at column 9 and the node at column 4 — measured, not assumed, so the five-column gap makes the assertion discriminating rather than accidentally correct.

## Verified

- 113 passed across the nine affected suites, `test_eval_double_query_coverage.py` among them, selected by symbol this time.
- Extraction checked against real grammar output for all six languages, including Rust `1 << 3` and tuple/struct variants, C `(1+2)` / `-1` / `'x'`, C++ scoped enums and `sizeof(int)`, C# attributes and trailing commas, PHP backed and pure enums. A Java `RED(1)` argument list correctly yields **no** value: it is a constructor call, not a discriminant.
- `CYPHER_DELETE_MODULE` direction and `*0..` placement confirmed correct and consistent with the existing `HAS_FIELD` entry.
- Protobuf modelling matches `Field` exactly, including the reserved field for `absolute_path` and the `_NOT_EXPORTED` entry; round-trip verified.
- `ty check` reports 13 diagnostics on this branch and **13 on `origin/main`**, none in the files this PR touches. Pre-existing, measured in a separate worktree rather than assumed.

## Not addressed

An anonymous `typedef enum { A, B } color_t;` in C produces no `Enum` node at all, on `main` too, so it gets no variants either. Out of scope here; worth its own issue if C typedef enums matter.

## Summary by CodeRabbit

- **New Features**
  - Added optional enum-variant capture for Rust, Java, C, C++, C#, and PHP.
  - Enum variants are now represented with their names, declaration order, source locations, documentation, and written values when available.
  - Added graph relationships connecting enum declarations to their variants.
  - Added serialization support for enum-variant data.

- **Bug Fixes**
  - Re-indexing and project removal now correctly clean up associated enum variants and relationships.


vitali87:
@greptileai review

vitali87:
Not my PR — flagging one thing and leaving the decision with whoever owns it.

`check_pr_gated.py 2086` reports **gated**. Two of the three review anchors are nevertheless stale:

| bot | anchor | at head `71ec232b`? |
|---|---|---|
| Greptile | `840c99b9` (4/5) | no |
| Copilot | `840c99b9` | no |
| CodeRabbit | names `71ec232b` twice, no red-flag phrases | yes |

Unresolved threads: 0.

This is issue #1936 reproducing on a live PR: the script verifies a review artifact positively (verdict phrase, trusted author) but never compares its anchor to `headRefOid`, so a stale review and a fresh one are indistinguishable to it. PR #1939 is open against that.

The practical exposure here is small: one commit since the review, `71ec232b docs(schema): declare the index property on HAS_VARIANT`, 7 insertions in `codebase_rag/types_defs.py`, and CodeRabbit did review it. So this is not "unreviewed code" so much as "two of three bots have not seen the last commit".

Re-triggering Greptile costs a comment and removes the ambiguity, which seems worth it given the merge gate asks for an anchored artifact. I have not done it, since re-triggering on someone else's PR changes their review state.

Flagging rather than acting because the CLAUDE.md is explicit that the script's output is the state — and this is the documented case where that is not sufficient on its own.

vitali87:
@greptileai review

vitali87:
@greptileai review

vitali87:
Adopted as orphaned (the session that created it is offline; no push in ~24h; confirmed with peer sessions) and rebased onto main with --force-with-lease, replacing the merge-main commits. Rebased rather than merged because a merge-main commit trips the ruleset's require_extra_approval_for_unattributed_changes.

- Conflicts: `ast_java.py`, `ast_php.py` and `ast_rust.py` constants. Both sides had added lines at the end of the file, and both are kept.
- Semantic conflict: main changed `evals/cgr_graph.py` to key `edge_props` by site (6-tuple), which made the edge-index assertion in `test_enum_variant_nodes.py` raise KeyError in 6 cases. Now reads via `props_for()` (a00323ce).
- Local: the PR's two test files plus the two receiver-typing test files pass, 40 passed / 0 skipped with all grammars installed. Setting the HAS_VARIANT edge index to a constant turns `test_every_variant_becomes_a_node_in_declaration_order` red in all 6 languages.

PR #2044, opened 2026-09-20
feat: expose change-signature via API, CLI, and MCP

vitali87:
Public editing API exports, facade/apply layer, CLI command, MCP tool, and generated descriptions.

## Files changed

- `codebase_rag/editing/__init__.py`
- `codebase_rag/editing/signature.py`
- `codebase_rag/cli.py`
- `codebase_rag/cli_help.py`
- `codebase_rag/mcp/tools.py`
- `codebase_rag/tools/tool_descriptions.py`

## Diff stats

**+345** additions, **-1** deletions (346 LOC)

## Dependencies

This PR depends on: `pr-5`

## Dependency graph

Merge in this order:

```
pr-1: feat: add change-signature types and constants
├── pr-2: feat: parse and render signature parameters
│   ├── pr-4: feat: rewrite signature definitions and call sites
│   │   └── pr-5: feat: plan change-signature graph edits
│   │       └── pr-6: feat: expose change-signature via API, CLI, and MCP  <-- this PR
│   │           ├── pr-7: test: cover change-signature units
│   │           │   ├── pr-8: test: cover change-signature acceptance flows
│   │           │   └── pr-9: test: cover change-signature regressions
│   │           └── pr-10: docs: document change-signature architecture
│   └── pr-5: feat: plan change-signature graph edits
│       └── pr-6: feat: expose change-signature via API, CLI, and MCP  <-- this PR
│           ├── pr-7: test: cover change-signature units
│           │   ├── pr-8: test: cover change-signature acceptance flows
│           │   └── pr-9: test: cover change-signature regressions
│           └── pr-10: docs: document change-signature architecture
├── pr-3: feat: map change-signature call arguments
│   ├── pr-4: feat: rewrite signature definitions and call sites
│   │   └── pr-5: feat: plan change-signature graph edits
│   │       └── pr-6: feat: expose change-signature via API, CLI, and MCP  <-- this PR
│   │           ├── pr-7: test: cover change-signature units
│   │           │   ├── pr-8: test: cover change-signature acceptance flows
│   │           │   └── pr-9: test: cover change-signature regressions
│   │           └── pr-10: docs: document change-signature architecture
│   └── pr-5: feat: plan change-signature graph edits
│       └── pr-6: feat: expose change-signature via API, CLI, and MCP  <-- this PR
│           ├── pr-7: test: cover change-signature units
│           │   ├── pr-8: test: cover change-signature acceptance flows
│           │   └── pr-9: test: cover change-signature regressions
[...]

PR #2108, opened 2026-09-21
fix: classify change_signature as a non-reader

vitali87:
`test_every_graph_reader_is_guarded` is red on #2044 across every platform:

```
these tools read the graph but never refuse on a partial one:
['change_signature']. Guard them, or add them to _NOT_GRAPH_READERS.
```

## The classification, not a guard

`_NOT_GRAPH_READERS` is deliberately fail-closed (#1547): a new tool counts
as a reader until someone classifies it, and this test fails until it is
either guarded or listed. The message offers both remedies and the test goes
green either way, so it names the unclassified tool without deciding which
side it belongs on.

`change_signature` is a graph-driven **edit**, like `RENAME` which already
sits in that set. It runs its own re-ingest behind the incomplete-run marker
and holds the result to the postcondition contract, so refusing up front on a
partial graph would block the operation that repairs it.

This is the owner's own resolution in `e5e9f1c1` ("fix(mcp): guard and
classify change_signature after the rebase onto main"), which lands on a
sibling layer of the 1549 stack. This branch carries neither that
classification nor its rationale, so the question reads as open here when it
has already been decided; the comment is reproduced verbatim to keep the two
layers consistent.

### Why not add it to `_READS_THE_GRAPH`

Tempting, because `change_signature` does route through `_graph_query`, whose
refusal is gated on `if tool in _READS_THE_GRAPH` — so the guard is indeed
not firing today. But making it fire is the wrong outcome: it would refuse
the edit precisely when the graph is partial, which is when the re-ingest
behind it is what restores the graph.

## Verification

`226 passed, 2 skipped` for `test_mcp_project_scoping.py`. Deleting the
classification line reddens `test_every_graph_reader_is_guarded` naming
`change_signature`, so the line is load-bearing rather than decorative.


vitali87:
@greptileai review

vitali87:
Superseded. The classification now sits directly on its base, pr-split/split-pr1548-rebased/pr-6 (#2044): bcc4c7df cherry-picks the owner's e5e9f1c1, with the same comment verbatim. 04d6eeed also fixes what this PR left red: `_reingest_for_contract` still called the deleted `_updater_for_reingest` (it now delegates to `_guarded_rename_reingest`), and three ty argument-type errors on `fetch_all` (fixed by typing it `QueryFn`).

PR #2109, opened 2026-09-21
fix: classify move, and stop kwonly params suppressing TOO_MANY

vitali87:
Two unrelated failures on #2074.

## 1. `move` read as an unguarded reader

```
these tools read the graph but never refuse on a partial one: ['move'].
```

`_NOT_GRAPH_READERS` is deliberately fail-closed: a tool counts as a reader
until classified. The `MOVE` enum member lands on this layer ahead of its MCP
handler, so nothing dispatches it and it reads nothing yet.

Classified as a non-reader beside `RENAME` and `CHANGE_SIGNATURE`, per
`e5e9f1c1`: these are graph-driven **edits** that run their own re-ingest
behind the incomplete-run marker and hold the result to the postcondition
contract, so refusing up front on a partial graph would block the operation
that repairs it.

The comment records what must happen next: when the handler lands it routes
through `_graph_query` and must move to `_READS_THE_GRAPH`, or the
dispatcher's `if tool in _READS_THE_GRAPH` silently skips its refusal.

## 2. Keyword-only params suppressed a real TOO_MANY verdict

`test_extra_positional_to_a_keyword_only_callee_is_too_many` is red here for
the same reason as on the 1550 stack. The suppressor counted `kwonlyargs`,
but keyword-only parameters absorb no surplus positional:

```
>>> def helper(a, *, b=1): pass
>>> helper(1, 2, b=3)
TypeError: helper() takes 1 positional argument but 2 positional arguments
           (and 1 keyword-only argument) were given
```

#2107 carries the full reasoning. The oracle on this branch restated the
buggy expression as well, so it is replaced with one that decides by
**calling** the function with one surplus positional.

## Verification

`283 passed, 2 skipped` across `test_mcp_project_scoping.py`,
`test_structural_delta.py` and `test_variadic_header_detection.py`.


vitali87:
@greptileai review

